PR #8, opened 2019-02-02
Fix stale publish rule in Makefile and gofmt some files

balleddog:
- I found a stale filename in Makefile after I ran `make publish` unsuccessfully
It seems to related to [this commit](https://github.com/scylladb/scylla-operator/commit/7c6d59e851e1ff1b1a2940016f0510e08044220d)
- Some go source files need to be`gofmt`ed


yanniszark:
Thanks for your contribution @balleddog! :smile: 
The changes look good to me!

PR #13, opened 2019-04-29
example/gke/gke.sh: make check_cluster_readiness zone agnostic

jwnx:
check_cluster_readiness was hardcoded for us-east1, which is a bug. This
patch changes it to use the GCP_ZONE parameter


yanniszark:
@jwnx this has been fixed in the tolerations pr. 
We should probably close this in favor of #12

gnumoreno:
@yanniszark there are some conflicting files here. Can you check that and resubmit?

PR #14, opened 2019-05-03
add support for monitoring 2.3

jwnx:
Removes helm charts and update docs to use `scylla-grafana-monitoring` instead of a
kubernetes job to load dashboards.


gnumoreno:
@jwnx there are some conflicting files here. Can you check that and resubmit?

Issue #31, opened 2019-10-03
Allow for configuring all the scylla.yaml config options


dahankzter:
Not sure how we are going to do this but extending `docker-entrypoint.py` is one option.

dahankzter:
We have at least one user that needs to inject `batch_size_fail_threshold_in_kb` so we might as well allow for all of them to be injected.

penberg:
It would be interesting to know what other Docker images do. Perhaps there's some established convention?

That said, you could introduce a generic `--server-opts` command line argument that looks something like this:

```sh
docker run scylladb/scylla --server-opts "batch_size_fail_threshold_in_kb=1024" "listen-address=0"
```

You can then either modify the YAML or pass these options to the `scylla` executable.

dahankzter:
Actually we have hooks to write scylla.yaml before the `docker-entrypoint.py` is run so that bypasses the problem I think. It doesn't help for a generic docker user but for the operator it should suffice.

dahankzter:
Is it a good idea to put this on the cluster spec @yanniszark ? That way we can set the properties in the cluster yaml and just pass it down to the scylla.yaml.

dahankzter:
I tested it and it seems to work fine. The question is if there is some aspect of lifecycle management or other that gets complicated or broken by doing this.

dahankzter:
Applying it again to change it doesn't work. I guess the reconciliation loop also needs this check and take appropriate action.

yanniszark:
@dahankzter my original idea was:

1. Add a userConfigMap field in the RackSpec.
    * The ConfigMap will contain certain well-known keys (eg scylla.yaml).
2. The ConfigMap is mounted on a temp location inside the scylla pod (eg /etc/scylla/userconfig/{files}).
3. The sidecar performs a 3-way merge of the original file (eg scylla.yaml) the user-provided file and the fields we know must have a certain value (eg we can't let the user define the broadcast_address, else things won't work correctly). So the operation is merge(merge(original, user-provided), certain-values) for every config file we support (scylla.yaml for starters).

This way the user can overwrite only the fields they want, without having to provide the whole scylla.yaml.

Now for updating:
We have 2 kinds of updates:
[...]

PR #35, opened 2019-10-28
Configure scylla.yaml through config map

dahankzter:
Allows for configuring `scylla.yaml` through a config map.

Fixes: #31 

**Checklist:**

### DIFF
--- a/examples/gke/cluster.yaml
+++ b/examples/gke/cluster.yaml
@@ -87,6 +87,7 @@ spec:
     name: <gcp_zone>
     racks:
       - name: <gcp_region>
+        scylla_config_map: "scylla-config"
         members: 2
         storage:
           storageClassName: local-raid-disks
--- a/examples/gke/cluster.yaml
+++ b/examples/gke/cluster.yaml
@@ -87,7 +87,7 @@ spec:
     name: <gcp_zone>
     racks:
       - name: <gcp_region>
-        scylla_config_map: "scylla-config"
+        scyllaConfigMap: "scyllaConfig"
         members: 2
         storage:
           storageClassName: local-raid-disks
--- a/pkg/apis/scylla/v1alpha1/cluster_types.go
+++ b/pkg/apis/scylla/v1alpha1/cluster_types.go
@@ -78,6 +78,8 @@ type RackSpec struct {
 	Placement *PlacementSpec `json:"placement,omitempty"`
 	// Resources the Scylla Pods will use.
 	Resources corev1.ResourceRequirements `json:"resources"`
+	// Scylla config map name to customize scylla.yaml
+	ScyllaConfig string `json:"scylla_config_map"`
 }
 
 type PlacementSpec struct {
--- a/pkg/apis/scylla/v1alpha1/cluster_types.go
+++ b/pkg/apis/scylla/v1alpha1/cluster_types.go
@@ -79,7 +79,7 @@ type RackSpec struct {
 	// Resources the Scylla Pods will use.
 	Resources corev1.ResourceRequirements `json:"resources"`
 	// Scylla config map name to customize scylla.yaml
-	ScyllaConfig string `json:"scylla_config_map"`
+	ScyllaConfig string `json:"scyllaConfigMap"`
 }
 
 type PlacementSpec struct {
--- a/pkg/controller/cluster/resource/resource.go
+++ b/pkg/controller/cluster/resource/resource.go
@@ -98,6 +98,11 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 		placement = &scyllav1alpha1.PlacementSpec{}
 	}
 	opt := true
+	scyllaConfigMap := r.ScyllaConfig
+	// TODO: Remove when updating the config map name is implemented
+	if scyllaConfigMap == "" {
+		scyllaConfigMap = "scyllaConfig"
+	}
 	return &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            naming.StatefulSetNameForRack(r, c),
--- a/pkg/controller/cluster/resource/resource.go
+++ b/pkg/controller/cluster/resource/resource.go
@@ -98,6 +98,10 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 		placement = &scyllav1alpha1.PlacementSpec{}
 	}
 	opt := true
+	scyllaConfigMap := r.ScyllaConfig
+	if scyllaConfigMap == "" {
+		scyllaConfigMap = "scylla-config"
+	}
 	return &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            naming.StatefulSetNameForRack(r, c),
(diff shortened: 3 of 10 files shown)


pkg/sidecar/config/config.go
```diff
@@ -94,8 +100,15 @@ func (s *ScyllaConfig) setupScyllaYAML() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to parse override options for scylla.yaml")
 	}
+	scyllaYAMLConfigMapFilteredBytes := overrideYAMLBytes
+	if len(scyllaYAMLConfigMapBytes) > 0 {
```
yanniszark: Perhaps I'm missing something.
Why do we check if the content in non-empty?
Can't we treat it uniformly `mergeYAMLs(scyllaYAMLConfigMapBytes, overrideYAMLBytes)`?

pkg/controller/cluster/resource/resource.go
```diff
@@ -151,6 +162,11 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 									MountPath: naming.SharedDirName,
 									ReadOnly:  false,
 								},
+								{
+									Name:      "scylla-config-volume",
+									MountPath: naming.ScyllaConfigDirName,
+									ReadOnly:  true,
+								},
```
yanniszark: Is it necessary to mount it in the sidecar?

pkg/sidecar/config/config.go
```diff
@@ -226,7 +239,7 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 	scyllaCmd := exec.Command(entrypointPath, args...)
 	scyllaCmd.Stderr = os.Stderr
 	scyllaCmd.Stdout = os.Stdout
-	s.logger.Info(ctx, "Scylla entrypoint command:\n %v", scyllaCmd)
+	s.logger.Info(ctx, "Scylla entrypoint", "command", scyllaCmd)
```
yanniszark: Nice fix. I didn't know that loggers aren't supposed to have newlines back then.

pkg/controller/cluster/resource/resource.go
```diff
@@ -133,6 +133,17 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 								EmptyDir: &corev1.EmptyDirVolumeSource{},
 							},
 						},
+						{
+							Name: "scylla-config-volume",
+							VolumeSource: corev1.VolumeSource{
+								ConfigMap: &corev1.ConfigMapVolumeSource{
+									LocalObjectReference: corev1.LocalObjectReference{
+										Name: "scylla-config",
```
yanniszark: Do we want to hardcode a specific name here?
What if the users wants to have multiple ConfigMaps for different Scylla Clusters?
Is that a use-case we would support?

My initial thought here was to make this a field in the RackSpec.
We can save it for a followup PR, if you feel that's better.

dahankzter: I didn't like this when I wrote it but at the time it just worked. Putting it in the RackSpec seems like a good idea.

pkg/sidecar/config/config.go
dahankzter: The new logger is adamant about even number of arguments. Before this it actually panicked which is for sure something we want to avoid.

pkg/controller/cluster/resource/resource.go
dahankzter: You mean in the init container?

yanniszark: Exactly, miswording on my part.

pkg/sidecar/config/config.go
yanniszark: Huh I should take a better look at the new logger then.

pkg/sidecar/config/config.go
dahankzter: You are right. Fixing it.

pkg/controller/cluster/resource/resource.go
dahankzter: Correct, removing it.

pkg/controller/cluster/resource/resource.go
dahankzter: I have pushed a new commit with the name configurable. Changing it does not work yet though.

dahankzter:
What do you think @yanniszark can we merge this without the possibility to rename the config map? Perhaps you can suggest a solution otherwise?

dahankzter:
Sorry! Ill add docs and examples for GKE and minikube as well before merge.

examples/gke/cluster.yaml
```diff
@@ -87,6 +87,7 @@ spec:
     name: <gcp_zone>
     racks:
       - name: <gcp_region>
+        scylla_config_map: "scylla-config"
```
yanniszark: Kubernetes Objects use camelCase so I believe it'd be better to stick to the convention.
You can read more on the Kubernetes API Conventions here:
https://github.com/kubernetes/community/blob/2a656f46492fb074cdb36d8413721db52078654b/contributors/devel/sig-architecture/api-conventions.md

pkg/apis/scylla/v1alpha1/cluster_types.go
```diff
@@ -78,6 +78,8 @@ type RackSpec struct {
 	Placement *PlacementSpec `json:"placement,omitempty"`
 	// Resources the Scylla Pods will use.
 	Resources corev1.ResourceRequirements `json:"resources"`
+	// Scylla config map name to customize scylla.yaml
+	ScyllaConfig string `json:"scylla_config_map"`
```
yanniszark: Change snake_case to camelCase.

pkg/controller/cluster/resource/resource.go
```diff
@@ -97,7 +97,11 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 	if placement == nil {
 		placement = &scyllav1alpha1.PlacementSpec{}
 	}
-
+	opt := true
+	scyllaConfigMap := r.ScyllaConfig
+	if scyllaConfigMap == "" {
+		scyllaConfigMap = "scylla-config"
```
yanniszark: I think it's better to be explicit here and not have a default value assumed inside the operator.
I know that the operator does exactly this thing in other places (eg image name) but in my experience it's preferable to be explicit.
It's better to use [MutatingWebhooks](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) for this job.

The problem here is that if the user creates a cluster without a ConfigMap, then they can't include one, unless we add the logic for it.

My suggestion would be to leave a TODO here to stop assuming a default ConfigMap once we implement the ConfigMap change action in the operator. 
What do you think?

dahankzter: I am not sure I understand. If we create one here then they can reuse that later if they want. If we don't set one at all then they can't add one.

Or have I misunderstood? The mount later isn't mandatory so a non existing config map will just wait for it to be mounted later if needed.

dahankzter: Afact the only thing they can't do as it stands now is to change the config map name but they can skip or set another initially.

examples/gke/cluster.yaml
dahankzter: Ok, fixing.

pkg/controller/cluster/resource/resource.go
dahankzter: We can use a validating webhook to enforce it?

dahankzter: I agree we should leave a TODO for now and implement a change function separately.

Issue #50, opened 2020-02-07
scylla-jmx failing to start on latest docker images on 3.2


maxfofocker:
I got a error on GKE cluster and use scylladb/scylla:3.2.1 docker image ,

log: errorVerbose: "open /usr/lib/scylla/jmx/scylla-jmx: no such file or directory

I find a similar error issue in [scylladb/scylla issues](https://github.com/scylladb/scylla/issues/5542).

I find a error point on [scylla-operator sidecar/config] (https://github.com/scylladb/scylla-operator/blob/master/pkg/sidecar/config/config.go#L30) , use it scyllaJMXPath on [L195](https://github.com/scylladb/scylla-operator/blob/master/pkg/sidecar/config/config.go#L195) and [L205](https://github.com/scylladb/scylla-operator/blob/master/pkg/sidecar/config/config.go#L205)

**Expected behavior**

I expected this path can be config from outer yaml setting or environment variable to change it.

dahankzter:
Ouch! Thanks for reporting this. There are a few things to keep in mind for this one. The operator needs to support several versions of scylla and in this case it means handling both of these transparently.
If one doesn't exists then we use the other.
Second it that we are moving it away from jmx altogether and we will replace it with an HTTP REST API call once the agent is in place.

How urgent is this for you @maxfofocker?

justdomepaul:
My system will production by the end of March.

I still test use scylla-operator or cassandra cluster to deploy my database cluster.

I want to know what time is expected to be upgrade this operator?

dahankzter:
Fixing this issue for now will be quick. The next release is not that quick but not far off either.
Is your intention to use Scylla if it works on Kubernetes otherwise go for Cassandra? Then of course we will try our best to help you but I cannot guarantee it. It works pretty well but hasn't seen much production use that we know of.

Feel free to vote for the issues or add anything you miss! All input from users are super valuable and we really can't get enough of that.

justdomepaul:
I will do my best to use scylladb works on Kubernetes.

But, I need high stability in my database cluster, 
if I still can't deploy successfully scylladb cluster, I will choice Cassandra to cover my system.

dahankzter:
Can you try now @justdomepaul @maxfofocker ?

PR #53, opened 2020-02-11
jmx: scylla-jmx location moved

dahankzter:
We now try both locations before we give up to ensure that we find
a scylla-jmx to inject Jolokia.

Fixes: #50


tzach:
@dahankzter AFAIK we have a different request of removing JMX from the operator

dahankzter:
We do indeed plan to avoid using it @tzach but in the meantime it needs to work.

Issue #32, opened 2019-10-20
Scylla wont start due to insufficent memory


tobad357:
**Describe the bug**
When scylla is coming up it crashes due to insufficient physical memory

Config
```
apiVersion: scylla.scylladb.com/v1alpha1
kind: Cluster
metadata:
  labels:
    controller-tools.k8s.io: "1.0"
  name: scylla-cluster
spec:
  version: 3.1.0
  developerMode: true
  datacenter:
    name: dc01
    racks:
      - name: rack1
        members: 7
        storage:
          capacity: 1800Gi
        resources:
          requests:
            cpu: 12
            memory: 48Gi
          limits:
            cpu: 12
            memory: 48Gi
```

This results in  env variable inside pod
```
MEMORY=49152
```
And the paramter to docker-entrypoint

```
root         30      6  0 15:35 ?        00:00:00 /opt/scylladb/python3/bin/python3 /opt/scylladb/python3/bin/../libexec/python3.7.bin -s /docker-entrypoint.py --listen-address=10.192.193.12 --broadcast-address=10.192.110.1
67 --broadcast-rpc-address=10.192.110.167 --seeds=10.192.110.167 --developer-mode=1 --smp=12 --memory=48452M```
```

The error when scylla starts up is

```
Could not initialize seastar: std::runtime_error (insufficient physical memory: needed 50805604352 available 47931835136)
```

**Expected behavior**
Scylla to not allocate more memory than what is available

**Config Files**
If relevant, paste your configuration files here (use a pastebin service, such as https://paste.fedoraproject.org/)

**Logs**
```
Could not initialize seastar: std::runtime_error (insufficient physical memory: needed 50805604352 available 47931835136)
```

**Environment:**
 - Platform: Rancker RKE with kuberneets 1.15
 - Kubernetes version: 1.15
 - Scylla version: 3.1.0
 - Scylla-operator version: v0.0-47cba43

yanniszark:
The memory is set in an environment variable inside the Pod.
https://github.com/scylladb/scylla-operator/blob/08f700bf6c70c48e3f8974cdd6a65703b3b1caaf/pkg/controller/cluster/resource/resource.go#L226-L233

The unit of measurement for this value is Mi (Mebibyte). For more information on power-of-ten vs power-of-two units of measurement, see https://en.wikipedia.org/wiki/Mebibyte.

That value is parsed, subtracted by 700Mi (empirical value) and then passed to Scylla.
[...]

PR #58, opened 2020-03-03
config: operator let's Scylla decide memory

dahankzter:
Since 3.1 the process by which Scylla decides the amount of memory
it should use works in a k8s setup as well. This patch drops the
additional heuristics for calculating the Scylla memory from the
operator itself.

Fixes: #32


yanniszark:
lgtm

Issue #57, opened 2020-02-20
Upgrade scylla minor version


dahankzter:
The idea is to relax the version condition and allow upgrading minor version but reject major version for now.

PR #59, opened 2020-03-05
Upgrade scylla version 

dahankzter:
Currently upgrades are allowed between patch versions but rejects
major and minor version upgrades.

Fixes: #57


yanniszark:
@dahankzter I took a stab at this and added my take on the [issue_57-rev](https://github.com/scylladb/scylla-operator/tree/issue_57-rev) branch.

High level description of changes:
* Treat the cluster upgrade as a long-running action that affects all Racks. If a Cluster upgrade has begun, no other actions (eg scale-up, scale-down) will be started until it is completed.
* Implemented status updating logic. 
    * If a Rack is updating (some of its Members have a new version, some an old version), we add the RackUpdating condition.
    * Each Rack has a new version field, describing its current version. If the Rack is updating, it remains at the old version until the upgrade is complete.
* In the sync loop, we check if we are in the middle of an upgrade action and resume if necessary.

I also tidied up some functions in relevant packages.
I noticed that a couple of fmt.Print were introduced in this PR.
Can we use the logger instead of print statements?

dahankzter:
Ah yes the Prints are simply debugs for me when developing, they will be removed.
I will add the logic above in the PR, I haven't checked your changes yet. Do they extend this branch or start from master?

Also the semantic version notion of "major" and "minor" are too lax for Scylla it seems. We can only safely update between "patch" versions. I will tighten the logic in the validation to reflect this.

dahankzter:
Btw you are free to make changes in this branch as well. Good way to collaborate. :)

dahankzter:
I saw it now and it starts from my head so I can easily integrate it, thx @yanniszark !

dahankzter:
There is something funny in `sync.go`:
```go
// Continue the cluster upgrade if there is one in progress.
if differentVersionRacks != 0 && differentVersionRacks != len(cluster.Spec.Datacenter.Racks) {

}
```
Shouldn't this just be removed?
What do you think @yanniszark ?

Issue #28, opened 2019-10-02
Interact with the Scylla API directly


dahankzter:
This ties into #5 and requires the Agent to be available on the nodes. We also need to generate clients from the Scylla API Swagger definitions and possibly create some easy to use wrapper code for the operations we need to use.

tzach:
This will be very good and remove the dependency on both nodetool and JMX

tzach:
Look like only `nodetool drain` and `nodetool decommission` are in use.

As fast as I can see `nodetool drain` just call `/storage_service/drain`[1], and `nodetool decommission` call `/storage_service/decommission`[2]

@amnonh please correct me if any of this operation is doing more than that (like wait for the result)

[1] https://github.com/scylladb/scylla-jmx/blob/04ea3ab7e02dbbb37eeb696f55ca4c5fe2a1ee5b/src/main/java/org/apache/cassandra/service/StorageService.java#L1036
[2] https://github.com/scylladb/scylla-jmx/blob/04ea3ab7e02dbbb37eeb696f55ca4c5fe2a1ee5b/src/main/java/org/apache/cassandra/service/StorageService.java#L916

tzach:
Two more use case to replace (both use `nodetool status`)
1. livenessCheck
https://github.com/scylladb/scylla-operator/blob/master/pkg/sidecar/checks.go#L31

2. readinessCheck
https://github.com/scylladb/scylla-operator/blob/master/pkg/sidecar/checks.go#L45

dahankzter:
Indeed @tzach we need these calls as well.

asias:
I do not think we have any node operations that are not possible without jmx. We can do everything with the restful api.  if there is any that we can't, it is "bug" we should fix it.

dahankzter:
Thx @asias then we are good to go!

dahankzter:
And with the manager installed we can do it with encrypted traffic as well.

PR #62, opened 2020-03-20
member_controller: now using scylla manager agent

dahankzter:
NB: Not hidden behind the agent feature flag. The idea is to release it when the agent docker image
becomes publicly available. 

Fixes: #28


yanniszark:
@dahankzter nice, however 1654 lines looks kinda wrong.
From the diff I can see that a scylla client is essentially vendored under pkg/scyllaclient.
Isn't there a way to just import the package instead of copying all the source code?

dahankzter:
I tried a lot but I couldn't get `dep` to play nice with some dependent libraries. Once we move to modules I think we can just delete these files and reuse the client generated for mermaid.

dahankzter:
The cluster manager also needs to discover any token the user may have configured.

dahankzter:
Latest commits adds token discovery for the scylla client inside the member controller.

yanniszark:
@dahankzter thanks for the PR!
Can you describe (or point me to some docs) what the scylla manager agent does and if it needs to coordinate with the operator in any way?

Also, I noticed you used a ConfigMap with an auth-token setting inside.
Security-sensitive data should always be in Kubernetes Secrets, never ConfigMaps.
Secrets are handled differently and many environments add extra layers of protection on top of them (think KMS). See: https://kubernetes.io/docs/concepts/configuration/secret/
For starters, we can just place the whole config in a secret instead of a ConfigMap (since it contains security-sensitive information).

It would also greatly help in the review if you could seperate all vendored code in a single commit (without anything else in it).

dahankzter:
Ah of course we should use a secret! Ill amend it here asap.

dahankzter:
I have changed the config to be mounted from a secret instead. What do you think @yanniszark? I am tempted to merge this and patch any smaller changes later.

dahankzter:
Essentially it this docs @yanniszark https://docs.scylladb.com/operating-scylla/manager/2.0/architecture/

PR #76, opened 2020-04-08
Fix generic examples

erebe:
Without those change, the default value for installing the operator for generic platform does not work

   + Operator crash loop being OOM killed with default value
   + cluster.yaml does not pass webhook validation without scyllaAgentConfig

Resolves #

**Checklist:**


dahankzter:
Looks good thx!

PR #85, opened 2020-05-07
Example cqlsh command is missing "us-"

greggerscota:
This causes the command to not run properly. This update fixes that.


dahankzter:
Thank you!

PR #95, opened 2020-05-27
Makefile fixes

mmatczuk:


Makefile
```diff
@@ -1,72 +1,80 @@
...
+GO111MODULE			:= off
+KUBEBUILDER_ASSETS	:= $(CURDIR)/bin/deps
+PATH				:= $(CURDIR)/bin/deps:$(CURDIR)/bin/deps/go/bin:$(PATH)
+GOROOT				:= $(CURDIR)/bin/deps/go
+GOVERSION			:= $(shell go version)
 
 # Run tests
+.PHONY: test
 test: fmt vet vendor
-	./bin/deps/go/bin/go test ./pkg/... ./cmd/... -coverprofile cover.out
```
dahankzter, line 19: If we don't use the downloaded Go anymore we should also avoid downloading it.

mmatczuk: We use it's in path.

dahankzter: Yes but there is a script that downloads it to `bin/deps` and that seems wasteful now.

mmatczuk: It is what it is we can remove it in another PR, this works with both project and local go.

dahankzter: Will it work? I am not sure how the github actions do it to run unit tests.

dahankzter: It seems to run fine. :)

PR #96, opened 2020-05-28
Minor gke fixes

dahankzter:
Bump GKE k8s version and switched to stackdriver monitoring.


dahankzter:
Maybe we should remove the helm2 init as well what do you think @mmatczuk ?

mmatczuk:
I think it breaks the script now you could remove it or add `|| "echo X"

dahankzter:
> I think it breaks the script now you could remove it or add `|| "echo X"

Ill remove the init I think. The init is removed from the docs already so we need just make it consistent.

Issue #111, opened 2020-06-02
cass-stress-gen.py: replace all proxy flags for cs parameters with cmd flag


mmatczuk:
In my view this cass-stress-gen.py should be migrated to Helm chart.
Parameter should be the full cs command.

dahankzter:
This is in #110

PR #115, opened 2020-06-02
scripts: cass-stress-gen.py change write parameters to --cmd

mmatczuk:
Fixes: #111


zimnx:
Inactive for more than 1 year

PR #123, opened 2020-06-08
cluster: sysctl parameters can be applied

dahankzter:
This is done via the cluster crd and supplied as a list of sysctl
key-value pairs. The sysctl parameters wil be injected into the
cluster statefulset by way of an init container.

**Checklist:**

### DIFF
--- a/pkg/controller/cluster/resource/resource.go
+++ b/pkg/controller/cluster/resource/resource.go
@@ -335,6 +335,10 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 		},
 	}
 
+	sysctlContainer := sysctlInitContainer(c.Spec.Sysctls)
+	if sysctlContainer != nil {
+		sts.Spec.Template.Spec.InitContainers = append(sts.Spec.Template.Spec.InitContainers, *sysctlContainer)
+	}
 	sts.Spec.Template.Spec.Containers = append(sts.Spec.Template.Spec.Containers, *agentContainer(c))
 	return sts
 }
@@ -378,6 +382,26 @@ func containerPorts(c *scyllav1alpha1.Cluster) []corev1.ContainerPort {
 	}
 
 	return ports
+}
+
+func sysctlInitContainer(sysctls []string) *corev1.Container {
+	if len(sysctls) == 0 {
+		return nil
+	}
+	opt := true
+	return &corev1.Container{
+		Name:            "sysctl-buddy",
+		Image:           "busybox:1.31.1",
+		ImagePullPolicy: "IfNotPresent",
+		SecurityContext: &corev1.SecurityContext{
+			Privileged: &opt,
+		},
+		Command: []string{
+			"/bin/sh",
+			"-c",
+			fmt.Sprintf("sysctl -w %s", strings.Join(sysctls, " ")),
+		},
+	}
 }
 
 func agentContainer(c *scyllav1alpha1.Cluster) *corev1.Container {
--- a/config/crds/scylla_v1alpha1_cluster.yaml
+++ b/config/crds/scylla_v1alpha1_cluster.yaml
@@ -132,6 +132,12 @@ spec:
               - version
               - repository
               type: object
+            sysctls:
+              description: 'Sysctl properties to be applied during initialization
+                given as a list of key=value pairs. Example: fs.aio-max-nr=232323'
+              items:
+                type: string
+              type: array
             version:
               description: Version of Scylla to use.
               type: string
--- a/docs/generic.md
+++ b/docs/generic.md
@@ -120,6 +120,22 @@ Checking the logs of the running scylla instances can be done like this:
 kubectl -n scylla logs simple-cluster-us-east-1-us-east-1a-0 scylla
 ```
 
+### Configure container kernel parameters
+
+Sometimes it is necessary to run the container with different kernel parameters.
+In order to support this, the Scylla Operator defines a cluster property `sysctls` that is a list of the desired key-value pairs to set.
+
+___For example___: To increase the number events available for asynchronous IO processing in the Linux kernel to N set sysctls to`fs.aio-max-nr=N`.
+
+```yaml
+racks:
+  - name: rack-name
+    scyllaConfig: "scylla-config"
+    scyllaAgentConfig: "scylla-agent-config"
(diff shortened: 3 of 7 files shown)


dahankzter:
I will add docs other than CRD as well.

pkg/controller/cluster/resource/resource.go
```diff
@@ -350,10 +350,34 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
 		},
 	}
 
+	sysctlContainer := sysctlInitContainer(c.Spec.Sysctls)
+	if sysctlContainer != nil {
+		sts.Spec.Template.Spec.InitContainers = append(sts.Spec.Template.Spec.InitContainers, *sysctlContainer)
+	}
 	sts.Spec.Template.Spec.Containers = append(sts.Spec.Template.Spec.Containers, *agentContainer(c))
```
mmatczuk, line 357: Can this go as [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/).

pkg/controller/cluster/resource/resource.go
```diff
@@ -350,10 +350,34 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
...
 	return sts
 }
 
+func sysctlInitContainer(sysctls []string) *corev1.Container {
```
mmatczuk, line 361: It would be better to call it sysctlSetContainer or vsysctlContainer to avoid confusion with k8s init containers.

pkg/controller/cluster/resource/resource.go
```diff
@@ -350,10 +350,34 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
...
+	if len(sysctls) == 0 {
+		return nil
+	}
+	opt := true
+	return &corev1.Container{
+		Name:            "sysctl-buddy",
```
mmatczuk, line 367: Is this some kind of naming scheme here?

pkg/controller/cluster/resource/resource.go
```diff
@@ -350,10 +350,34 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.Cluster, si
...
+		Image:           "busybox:1.31.1",
```
mmatczuk, line 368: Could this be yaml configurable too?

dahankzter: We can and we can even move it to the operator image if we want.

pkg/controller/cluster/resource/resource.go
dahankzter: I don't follow. They are kubernetes init containers.

pkg/controller/cluster/resource/resource.go
dahankzter: Sure.

dahankzter: No I mean why? These are k8s init containers.

pkg/controller/cluster/resource/resource.go
dahankzter: Hehe. Any suggestions?

pkg/controller/cluster/resource/resource.go
mmatczuk: Sorry my bad missed the line

pkg/controller/cluster/resource/resource.go
mmatczuk: Correct.

pkg/controller/cluster/resource/resource.go
mmatczuk: Oh that would be neat.

dahankzter: The question is if we want it yaml configurable if we move it to the operator. We can just let the operator do it and be done. Nothing more to think about.

yanniszark:
I understand that this pr introduces a way of setting sysctl values via a container.
I would also like to point out that we can set sysctl values declaratively on the podspec:
https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

This would also be better for security and auditing and seems to be the canonical way to set sysctl valuesfor pods in k8s.

dahankzter:
I did read this spec @yanniszark but I interpreted it as only setting permissions for containers to be able to set these. If it can be used to actually specify the params then that is great!

dahankzter:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: sysctl-example
spec:
  securityContext:
    sysctls:
    - name: kernel.shm_rmid_forced
      value: "0"
    - name: net.core.somaxconn
      value: "1024"
    - name: kernel.msgmax
      value: "65536"
```

You mean we could embed this in the spec somewhere @yanniszark?

dahankzter:
This is nicer @yanniszark we can set it on the pod template spec when we emit the statefulset. Good?

yanniszark:
@dahankzter the documentation isn't super clear on thst, can you launch a pod and check?
Also, it depends on which sysctl settings you want to configure. Imo, node-level settings (not linux namespaced) should be part of the node configuration via a daemonset. Copying from the docs:
> Sysctls with no namespace are called node-level sysctls. If you need to set them, you must manually configure them on each node’s operating system, or by using a DaemonSet with privileged containers.

> Use the pod securityContext to configure namespaced sysctls. The securityContext applies to all containers in the same pod.

dahankzter:
On GKE I get this @yanniszark:
```
Events:
  Type     Reason           Age   From                                                          Message
  ----     ------           ----  ----                                                          -------
  Warning  SysctlForbidden  12s   kubelet, gke-scylla-henrik-k8s-de-default-pool-64c142c4-1wk7  forbidden sysctl: "fs.aio-max-nr" not whitelisted
```
I think we can stay with this PR as is for now using the Init container. What do you think?

mmatczuk:
The init container approach is good, we will more of init containers / steps in future.

pkg/controller/cluster/resource/resource.go
mmatczuk: sysctl-init? just sysctl

mmatczuk: No name?

PR #124, opened 2020-06-08
gke: externalized cluster name

dahankzter:
The k8s cluster name can now be given on the command line but falls back to the old "scylla-demo" in case none is provided.

### DIFF
--- a/examples/gke/gke.sh
+++ b/examples/gke/gke.sh
@@ -7,29 +7,79 @@ set -e
 
 display_usage() {
 	echo "End-to-end deployment script for scylla on GKE."
-	echo "usage: $0 [GCP user] [GCP project] [GCP zone] [cluster name (optional)]"
-    echo "example: $0 yanniszark@arrikto.com gke-demo-226716 us-west1-b"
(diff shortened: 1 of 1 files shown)


examples/gke/gke.sh
```diff
@@ -1,12 +1,38 @@
...
+#########
+
 display_usage() {
 	echo "End-to-end deployment script for scylla on GKE."
-	echo "usage: $0 [GCP user] [GCP project] [GCP zone]"
+	echo "usage: $0 [GCP user] [GCP project] [GCP zone] [cluster name (optional)]"
     echo "example: $0 yanniszark@arrikto.com gke-demo-226716 us-west1-b"
 }
 
+case $# in
```
mmatczuk, line 14: If this script has any future it would be better to add proper flags.
As for validation even now it would be better to check if all variables are set instead of that.

dahankzter: Yeah that makes sense. I can do a proper arg parse thing.

dahankzter: It will make the script non backwards compatible though but that is fine I think.

Issue #121, opened 2020-06-04
cpuset is wrongly set


mmatczuk:
Machine with 8 cores, static cpu scheduler, cpu limit 7

```
        resources:
          limits:
            cpu: 7
            memory: 58G
```

We set both smp and cpuset which is redundant and cpuset is incorrect.

```
ld.so /opt/scylladb/python3/bin/../libexec/python3.7.bin -s /docker-entrypoint.py --listen-address=192.168.72.71 --broadcast-address=10.100.234.97 --broadcast-rpc-address=10.100.234.97 --seeds=10.100.234.97 --developer-mode=0 --smp=7 --cpuset=0-7
```

dahankzter:
So the problem is we set too many in `cpuset`? In this case we should set `0-6`? We parse this from running container so the it is (theoretically) possible there is something else wrong.
Seastar has:
```cc
        ("smp,c", bpo::value<unsigned>(), "number of threads (default: one per CPU)")
        ("cpuset", bpo::value<cpuset_bpo_wrapper>(), "CPUs to use (in cpuset(7) format; default: all))")
```
So using both of these values seems ok as long as they are correct I think.

mmatczuk:
cpuset is set incorectly.

mmatczuk:
Yet again cat /proc/1/status always shows all CPUs.

mmatczuk:
Interestingly kubernetes (EKS) seems not to do pin to cores 

```
[ec2-user@ip-192-168-82-124 ~]$ cat /sys/fs/cgroup/cpuset/kubepods/burstable/pod3383fd7f-aa0c-4f43-a505-8411cf7df5b5/4af130e527298120043416851f8d48ed9728fe8eb1503fd5162ff2e6ae660386/cpuset.effective_cpus 
0-7
[ec2-user@ip-192-168-82-124 ~]$ cat /sys/fs/cgroup/cpuset/kubepods/burstable/pod3383fd7f-aa0c-4f43-a505-8411cf7df5b5/4af130e527298120043416851f8d48ed9728fe8eb1503fd5162ff2e6ae660386/cpuset.cpus
0-7
```

dahankzter:
I can confirm it looks similar on GKE. Where should we get the cpuset info then?

dahankzter:
We have a very strange validation in place. It seems completely wrong:
```go
// Requests == Limits or only Limits must be set for QOS class guaranteed
requests := rack.Resources.Requests
if requests != nil {
[...]

Issue #135, opened 2020-06-17
Scylla should use qosClass: Guaranteed not Burstable


mmatczuk:
At the moment I can see:

```
  phase: Running
  podIP: 192.168.16.98
  podIPs:
  - ip: 192.168.16.98
  qosClass: Burstable
  startTime: "2020-06-17T09:46:12Z"
```

For a Pod to be given a QoS class of Guaranteed:

* Every Container in the Pod must have a memory limit and a memory request, and they must be the same.
* Every Container in the Pod must have a CPU limit and a CPU request, and they must be the same.

https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/

Some containers / initContainers have `resources: {}` which may break it.

dahankzter:
At first I read it as all the containers must have the same limits but surely that can't be the same right? It must mean that the limit and request must be the same but different containers can have different pairs of limit/request. If this is so then it should be fine I guess.
We already support setting the resources so I guess the only thing we need is to expose the QoS property.

dahankzter:
Ah so these QoSs are assigned by k8s it's not something we can request beyond adhering to the rules as given in the link. Well then we done or do you want to enforce it?

mmatczuk:
Not sure what are the implications, request is limit for scylla others have unspecified resources.

dahankzter:
I think it's a UX thing. If we enforce it then invariably some users will want to have Burstable rather than Guaranteed at some point.

PR #143, opened 2020-06-26
Cpuset incorrectly set

dahankzter:
This PR sets the resources limits of all the containers, both init- and regular containers, explicitly to ensure QoS Guaranteed.
Furthermore it makes a minor validation that the cpuset and shard (--smp) settings are optimal. It will not block deployment but logs it to inform the user that there may be a problem.

Fixes: #121, #135

### DIFF
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/ghodss/yaml"
@@ -17,6 +18,7 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/sidecar/identity"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -173,13 +175,18 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		devMode = "1"
 	}
 
+	shards, err := strconv.Atoi(options.GetSidecarOptions().CPU)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
 	args := []string{
 		fmt.Sprintf("--listen-address=%s", m.IP),
 		fmt.Sprintf("--broadcast-address=%s", m.StaticIP),
 		fmt.Sprintf("--broadcast-rpc-address=%s", m.StaticIP),
 		fmt.Sprintf("--seeds=%s", strings.Join(seeds, ",")),
 		fmt.Sprintf("--developer-mode=%s", devMode),
-		fmt.Sprintf("--smp=%s", options.GetSidecarOptions().CPU),
+		fmt.Sprintf("--overprovisioned=%d", 0),
+		fmt.Sprintf("--smp=%d", shards),
 	}
 	if cluster.Spec.Alternator.Enabled() {
 		args = append(args, fmt.Sprintf("--alternator-port=%d", cluster.Spec.Alternator.Port))
@@ -194,7 +201,9 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
-
+		if err := s.validateCpuSet(ctx, cpusAllowed, shards); err != nil {
+			return nil, errors.WithStack(err)
+		}
 		args = append(args, fmt.Sprintf("--cpuset=%s", cpusAllowed))
 	}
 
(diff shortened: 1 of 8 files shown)


mmatczuk:
As a side note if one does not specify cpuset the [docker container would add overprovisioned](https://github.com/scylladb/scylla/blob/master/dist/docker/redhat/scyllasetup.py#L91). It breaks performance and we are not overprovisioned. So please always add overprovisioned = 0.

dahankzter:
> As a side note if one does not specify cpuset the [docker container would add overprovisioned](https://github.com/scylladb/scylla/blob/master/dist/docker/redhat/scyllasetup.py#L91). It breaks performance and we are not overprovisioned. So please always add overprovisioned = 0.

Added this unconditionally.

pkg/webhook/default_server/cluster/validating/cluster_validation.go
```diff
@@ -33,16 +33,18 @@ func checkValues(c *scyllav1alpha1.Cluster) (allowed bool, msg string) {
 				return false, fmt.Sprintf("when using cpuset, you must use whole cpu cores, but rack %s has %dm", rack.Name, cores)
 			}
 
-			// Requests == Limits or only Limits must be set for QOS class guaranteed
```
mmatczuk, line 36: It implies that we'd have Guaranteed QoS only if CpuSet is enabled.
Isn't it that we should always use Guaranteed?

pkg/sidecar/config/config.go
```diff
@@ -173,13 +175,17 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
+		return nil, errors.WithStack(err)
+	}
 	args := []string{
 		fmt.Sprintf("--listen-address=%s", m.IP),
 		fmt.Sprintf("--broadcast-address=%s", m.StaticIP),
 		fmt.Sprintf("--broadcast-rpc-address=%s", m.StaticIP),
 		fmt.Sprintf("--seeds=%s", strings.Join(seeds, ",")),
 		fmt.Sprintf("--developer-mode=%s", devMode),
-		fmt.Sprintf("--smp=%s", options.GetSidecarOptions().CPU),
+		fmt.Sprintf("--smp=%d", shards),
```
mmatczuk, line 188: To be on the safe side I'd set that only if cpuset is disabled.

mmatczuk:
I fail to see overprovisioned in this patch.

pkg/sidecar/config/config.go
```diff
@@ -206,6 +214,18 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 	return scyllaCmd, nil
 }
 
+func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, shards int) error {
```
mmatczuk, line 217: How about a plain function?
Please add a comment.

pkg/sidecar/config/config.go
dahankzter: That's not what I gathered from my discussions with @denesb . SMP refers to the number of shards and each cpu can serve more than one shard afaict.

dahankzter:
> I fail to see overprovisioned in this patch.

I don't get it... https://github.com/scylladb/scylla-operator/compare/cpuset_incorrectly_set has the overprovisioned commit but it can't be seen here. Has something changed wrt how PRs works or...

pkg/sidecar/config/config.go
mmatczuk: You want shard-per-core, this would be useful if we supported shard < available cores.

But you are correct it's safer to specify smp.

mmatczuk:
Could we have a unit test to check that all limits are set correctly?

pkg/sidecar/config/config.go
dahankzter: Sure but I wanted access to the logger for that information log about less than optimal setup. It's not a problem to pass it in though.

dahankzter:
> Could we have a unit test to check that all limits are set correctly?

It depends on what you mean "set correctly". We can either log "non-optimal setup detected please consider changing...." or we can fail the deploy when the user requests cpuset since then they should know what they are doing.

pkg/webhook/default_server/cluster/validating/cluster_validation.go
dahankzter: It depends. Do we want it to be possible to run another config or not? I don't know if it's too strict or not we should hear from the field more on this. @gnumoreno what do you think? Always enforce QoS Guaranteed?

pkg/sidecar/config/config.go
denesb: > SMP refers to the number of shards and each cpu can serve more than one shard afaict.

Each CPU can host exactly one shard only, but in some cases some CPUs are reserved for interrupt processing or other side-tasks. So number of CPUS >= number of shards.

pkg/webhook/default_server/cluster/validating/cluster_validation.go
dahankzter: The question is more if any user ever will run without cpuset? What are the consequences of not applying cpuset but going for Guaranteed QoS? I think none right? In case we get Guaranteed but don't give these cores to scylla it should still be fine.

gnumoreno:
A little background
https://github.com/scylladb/scylla/issues/3336
https://github.com/scylladb/scylla/issues/2114

FYI

PR #162, opened 2020-07-28
Propose resource request update for scylla manager agent sidecar

c0ffeec0der:
Propose resource request update for scylla manager agent sidecar starting at 5 percent

Rationale:
 If the DB is pinned to max 1 CPU per pod, with scylla manager agent sidecar also at 1 CPU is quite a lot

Could we start at: cpu 50m, memory 10M; and max out to cpu 1, memory 200M?
This would be beneficial for developer testing and k8s will scale it vertically for production

Update resource.go for scylla manager agent

Resolves # None

**Checklist:**


zimnx:
LGTM
Thanks for contribution!

PR #167, opened 2020-07-30
example: include local username in example gke cluster name

zimnx:
It helps identifying cluster in shared Google Cloud accounts.


examples/gke/gke.sh
```diff
@@ -10,7 +10,7 @@ display_usage() {
 	echo "usage: $0 -u|--gcp-user [GCP user] -p|--gcp-project [GCP project] -z|--gcp-zone [GCP zone] -c|--k8s-cluster-name [cluster name (optional)]"
 }
 
-CLUSTER_NAME=scylla-demo
+CLUSTER_NAME=${USER}scylla-demo
```
ajankovic, line 13: Maybe a dash is after variable?

mmatczuk (approved):

PR #174, opened 2020-08-07
Add Darwin support, make bin/deps more complete

ofpiyush:
- Updates go to stable `1.13.15`
- Adds `requirements.txt` for ` make bin/deps`
- Adds support for Darwin to `make bin/deps`

Haven't opened an issue for minor issue.

**Checklist:**


hack/binary_deps.py
```diff
@@ -9,16 +10,25 @@
...
 import shlex
-import os
 
 log = logging.getLogger(__name__)
 
+# assume linux
 KUBEBUILDER_URL = "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_linux_amd64.tar.gz"
 KUSTOMIZE_URL = "https://github.com/kubernetes-sigs/kustomize/releases/download/v3.1.0/kustomize_3.1.0_linux_amd64"
-GO_URL = "https://dl.google.com/go/go1.13.12.linux-amd64.tar.gz"
+GO_URL = "https://dl.google.com/go/go1.13.15.linux-amd64.tar.gz"
```
mmatczuk, line 19: Would be good to extract the version info from here. I Manager we have a `.go-version` file 

```
 michal@lh:scylladb/mermaid (mmt/2064-estimate-parallelism)$ cat .go-version 
1.13.15
```

ofpiyush: There are at least 3 places where we have the go version:

1. `go.mod`
2. `Dockerfile`
3. `hacks/binary_deps.py`

I can't think of an easy solution to get all three from one source.

My current best solution is to do the `.go-version` file and add documentation for go version. Can you think of something else?

Issue #146, opened 2020-07-03
Cache iotune results


coelho:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
Cache iotune results to ensure it does not run every start. Results can be stored in the PV.

**What is use case behind this feature:**
Scylla takes a long time to start up due to running iotune every time the server starts. This can cause extended downtime if you're cold starting your cluster, or recovering from a failure.

In a bare-metal Scylla installation, this result is cached, but it is not in Kubernetes with Scylla operator.

**Additional Information:**
This may already be reported but I was unable to find a similar issue.

PR #175, opened 2020-08-12
Make io setup optional when starting scyllaDB container

erebe:
+ Related to #146 to avoid long startup time

When we provision our machines, we pre-configure them in order to get the correct configuration for scyllaDB. 
Thus running io-setup at startup takes a long time and duplicate effort in our case.

Scylla container support this option https://github.com/scylladb/scylla/issues/6587
https://github.com/scylladb/scylla/commit/fc1851cdc16f24623dcc452ef87ecfc7b97c8d0d#diff-b2c8ee866cc696977a318d279ed1d681

**It is not yet supported by latest 4.1.x docker image**, let me know how/if you want to support backward compatibility

=====

Resolves #

**Checklist:**

### DIFF
--- a/config/crd/bases/scylla.scylladb.com_clusters.yaml
+++ b/config/crd/bases/scylla.scylladb.com_clusters.yaml
@@ -486,6 +486,9 @@ spec:
             developerMode:
               description: DeveloperMode determines if the cluster runs in developer-mode.
               type: boolean
+            ioSetup:
+              description: ioSetup determines if scylladb container should do io setup at startup
+              type: boolean
             network:
               description: Networking config
               properties:
--- a/pkg/controllers/sidecar/config/config.go
+++ b/pkg/controllers/sidecar/config/config.go
@@ -166,6 +166,8 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 
 	// Check if we need to run in developer mode
 	devMode := "0"
+	// Check if we need to run io-setup
+	ioSetup := "0"
 	cluster := &v1alpha1.Cluster{}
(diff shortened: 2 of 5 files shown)


config/crd/bases/scylla.scylladb.com_clusters.yaml
```diff
@@ -486,6 +486,9 @@ spec:
             developerMode:
               description: DeveloperMode determines if the cluster runs in developer-mode.
               type: boolean
+            ioSetup:
```
zimnx, line 489: add note about which Scylla version supports this flag

pkg/controllers/sidecar/config/config.go
```diff
@@ -184,6 +189,7 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		fmt.Sprintf("--broadcast-rpc-address=%s", m.StaticIP),
 		fmt.Sprintf("--seeds=%s", strings.Join(seeds, ",")),
 		fmt.Sprintf("--developer-mode=%s", devMode),
+		fmt.Sprintf("--io-setup=%s", ioSetup),
```
zimnx, line 192: Don't append it when it's not supported for deployed Scylla version.

zimnx (changes requested):
We must reject setting this option together with Scylla version which doesn't support it. 
Please add validation check to https://github.com/scylladb/scylla-operator/blob/d75aaad0c77436611a36fc29b41c7acb34ed5cce/pkg/api/v1alpha1/cluster_validation.go

erebe:
Thanks for the review, i am on vacation right nlw. i am going to check thzt in september when i am back

erebe:
Sorry for the disturbance, I am moving the PR to https://github.com/scylladb/scylla-operator/pull/179, I forgot to use the special branch I've created to submit the PR.

PR #179, opened 2020-08-31
Make io setup optional when starting scyllaDB container

erebe:
+ Related to #146 to avoid long startup time

When we provision our machines, we pre-configure them in order to get the correct configuration for scyllaDB. 
Thus running io-setup at startup takes a long time and duplicate effort in our case.

Scylla container support this option https://github.com/scylladb/scylla/issues/6587
https://github.com/scylladb/scylla/commit/fc1851cdc16f24623dcc452ef87ecfc7b97c8d0d#diff-b2c8ee866cc696977a318d279ed1d681

**It is not yet supported by latest 4.1.x docker image**, let me know how/if you want to support backward compatibility

=====

Resolves #

**Checklist:**

### DIFF
--- a/config/crd/bases/scylla.scylladb.com_clusters.yaml
+++ b/config/crd/bases/scylla.scylladb.com_clusters.yaml
@@ -486,6 +486,9 @@ spec:
             developerMode:
               description: DeveloperMode determines if the cluster runs in developer-mode.
               type: boolean
+            ioSetup:
+              description: ioSetup determines if scylladb container should do io setup at startup
+              type: boolean
             network:
               description: Networking config
               properties:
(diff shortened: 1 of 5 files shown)


erebe:
+ Updated the PR regarding your previous comment, to not append --io-setup when the flag is set to true

Waiting for https://github.com/scylladb/scylla/issues/6587 in order to know in which version the flag will be available in docker image. For now it is not available nor in 4.2.x or 4.1.x version of docker image

mmatczuk:
IMO we need a more robust solution, we can merge it when docker supporting it is out.

ezbz:
This can be done by applying developer mode see documentation

zimnx:
Done via #379

PR #184, opened 2020-09-14
controllers/cluster/resource/resource: add VolumeMounts and Volumes options

dkropachev:
**Fixes**
https://github.com/scylladb/scylla-operator/issues/185

Idea behind the fix is to able to pass volumes and volumemounts parameters to the scylla containers
One of the application is having a share to dump corefile from all scylla containers down to it

**Checklist:**

**Tested**
Tested locally


zimnx (approved):
LGTM

Issue #173, opened 2020-08-06
Examples: extract common elements to separate folder


mmatczuk:
We should be doing more DRY there is no need to copy grafana and provisioner dir in each example.
It makes updating those resources more difficult. The concrete examples should have helm value files only.

We could try to go even further and split the examples to 

* prepare stage - different content for different providers
* deploy stage - common core, possibly decorated with helm chart to allow for easy setting of limits

mmatczuk:
When I think about it we should remove the Grafana dashboards from here and download them or use a docker image that has them.

zimnx:
Already done.

PR #188, opened 2020-09-22
Scylla Manager integration

### LINKED ISSUE #5: scylla-manager integration
Scylla Manager has many nice features that help us handle Scylla clusters, such as repairs and backup/restore. Would be nice to have the option to deploy it alongside the Operator.

**Possible solution**
Once dockerized, Scylla Manager can be deployed as an outside service that is able to communicate with the cluster. Its key-pair can be generated as a secret upon cluster creation. 

**Other Alternatives**
??

**Additional context**
This feature depends on scylla-manager working on docker.

zimnx:

Scylla Manager is optional, can be used for repair and backup scheduling.
Manager consist of:
* small Scylla cluster needed as Manager DB.
* controller watching for cluster changes.

Everything is deployed in separate namespace called `scylla-manager-system`.

Cluster CRD spec is extended with two new fields representing task
definitions.
Example:

```
kind: Cluster
spec:
  version: 4.0.0
  agentVersion: 2.0.2
  repairs:
    - dc: ["us-east-1"]
      fail_fast: true
      intensity: 17
      interval: "1d"
      name: "daily us-east-1 repair"
  backups:
    - name: "weekly backup"
      locations: ["s3:cluster-backup"]
      interval: "7d"
```

Fixes #5

**Checklist:**


docs/manager.md
```diff
@@ -0,0 +1,254 @@
...
+│ healthcheck/400b2723-eec5-422a-b7f3-236a0e10575b            │                                      │ 23 Sep 20 14:28:42 CEST (+15s) │ DONE   │
+│ healthcheck_rest/28169610-a969-4c20-9d11-ab7568b8a1bd       │                                      │ 23 Sep 20 14:29:57 CEST (+1m)  │ NEW    │
+╰─────────────────────────────────────────────────────────────┴──────────────────────────────────────┴────────────────────────────────┴────────╯
+
+```
+
+Scylla Manager by default registers recurring healhcheck tasks for Agent and for each of the enabled frontends (CQL, Alternator).
+
+In this task listing we can see CQL and REST healthchecks.
+
```
espindola, line 133: Is it possible to add the manager to an existing scylla cluster? Might be a good second example.

docs/manager.md
```diff
@@ -0,0 +1,254 @@
...
+* Kubernetes cluster
+* Scylla Operator - see [generic guide](generic.md)
+
+## Architecture
+
+Scylla Manager in K8s consist of:
+- Dedicated Scylla Cluster
+
+  Scylla Manager persists its state to a Scylla cluster.
```
espindola, line 22: Somewhat unrelated, but how big is the state?
Do we plan for a closer integration with Kubernetes in the future? If so, could it store its state in etcd?

examples/generic/cluster.yaml
```diff
@@ -100,11 +100,4 @@ spec:
           limits:
             cpu: 1
             memory: 1Gi
-        volumes:
```
espindola, line 103: Is this related?

examples/gke/manager.yaml
```diff
@@ -0,0 +1,275 @@
...
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
```
espindola, line 28: I see a lot of these verbs being duplicated. Is there any way to avoid duplicating the same text in many yaml files?

pkg/api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -40,6 +40,73 @@ func (in *AlternatorSpec) DeepCopy() *AlternatorSpec {
 	return out
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
espindola, line 43: Do we need to have autogenerated code in the repository?

pkg/cmd/manager_controller.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+
+	"github.com/scylladb/go-log"
+	"github.com/scylladb/scylla-operator/pkg/cmd/options"
+	"github.com/scylladb/scylla-operator/pkg/controllers/manager"
+	"github.com/spf13/cobra"
+
+	"go.uber.org/zap"
```
espindola, line 11: In scylla proper we normally use our own clones to avoid a left-pad like situation. The solution in here is to add the modules we use to vendor/, correct?

pkg/controllers/cluster/status.go
```diff
@@ -18,9 +18,8 @@ import (
 // It doesn't post the result to the API Server yet.
 // That will be done at the end of the sync loop.
 func (cc *ClusterReconciler) updateStatus(ctx context.Context, cluster *scyllav1alpha1.Cluster) error {
-	clusterStatus := scyllav1alpha1.ClusterStatus{
-		Racks: map[string]scyllav1alpha1.RackStatus{},
-	}
+	cluster.Status.Racks = map[string]scyllav1alpha1.RackStatus{}
```
espindola, line 21: Unrelated change?

espindola (commented):
Not sure I am able to offer a lot of feedback on this just yet.

As with the previous one, I think the main thing I noticed coming from scylla proper is the amount of text copied from somewhere else. Is there a particular reason to not use git submodules?

examples/generic/cluster.yaml
zimnx: Not, I can extract it to separate PR. It was added in #184 and it broke cluster deployment

pkg/cmd/manager_controller.go
zimnx: Yep, vendor is solution for these kind of situations. We do have vendor dir in Operator, i forgot to update it.

pkg/controllers/cluster/status.go
zimnx: It's related. Prior this PR ClusterStatus had single field, so rewriting it made sense. But now we added second field, and because of this code, it was every time zeroed out. I changed this function to modify only RackStatus and don't touch ManagerID which it isn't supposed to touch.

pkg/api/v1alpha1/zz_generated.deepcopy.go
zimnx: Keeping it here makes build process easier. It also helps with keeping builds reproducible over time.

docs/manager.md
zimnx: > Is there a particular reason to not use git submodules?

We have a plan to extract scyllaclient and mermaidclient into separate repository, then we will use Go Modules to pull it as a dependency. But since we want to move forward without waiting for it i simply copy pasted it and did necessary modifications in this repo.

zimnx:
> Is there a particular reason to not use git submodules?

We have a plan to extract scyllaclient and mermaidclient into separate repository, then we will use Go Modules to pull it as a dependency. But since we want to move forward without waiting for it i simply copy pasted it and did necessary modifications in this repo.

yanniszark:
In addition, can you expand a bit more in your commit message?
Similar to https://github.com/arrikto/oidc-authservice/pull/38:

> Using a strict SameSite policy for the session cookie is sometimes too
restrictive. For example, an application that is protected by the AuthService
may be acting as an OIDC/OAuth client to a third-party application. In that
case, the SameSite=strict policy prevents the Provider->Client auth code
redirect from working, as it originates from the Provider origin.

Make the SameSite policy of the session cookie configurable and default to
Lax, which allows cookies only from cross-site requests / redirects using the
GET method.

docs/manager.md
zimnx: It is, procedure is the same in both cases. I'll add a note.

docs/manager.md
zimnx: It depends on how much clusters you manage and how much tasks there are. We do have TTLs on task progress table which should be biggest. I don't have a numbers though.

We thought about adding different state layer, but we don't have usecase which can't be covered with current architecture. So nothing in the pipeline.

examples/gke/manager.yaml
zimnx: This file is autogenerated, base is in `config/manager/rbac/role.yaml`. We have an issue to extract these common entities to single place: #173

zimnx:
> In addition, can you expand a bit more in your commit message?

What exactly is missing? I don't want to deep dive into implementation details in commit msg, i believe example usage is enough but maybe something could be added.

espindola:
Maciej Zimnoch <notifications@github.com> writes:

>            limits:
>              cpu: 1
>              memory: 1Gi
> -        volumes:
>
> Not, I can extract it to separate PR. It was added in #184 and it broke cluster deployment

If you don't mind that would be great. Different projects do things
differently, but I personally find small PRs or patch series easier to
review.

Thanks,
Rafael

pkg/api/v1alpha1/zz_generated.deepcopy.go
espindola: Why would the build not be reproducible? Is the generator non deterministic? Is that intentional?

zimnx:
> If you don't mind that would be great. Different projects do things differently, but I personally find small PRs or patch series easier to review. Thanks, Rafael

Looks like I thought it broke cluster deployment but it didn't. Current master works fine so I removed this change from this PR.

espindola:
> > Is there a particular reason to not use git submodules?
> 
> We have a plan to extract scyllaclient and mermaidclient into separate repository, then we will use Go Modules to pull it as a dependency. But since we want to move forward without waiting for it i simply copy pasted it and did necessary modifications in this repo.

I can't speak about project priorities, but I really dislike copy and paste, so maybe I can make that refactoring my first task? Do you have any pointers?

examples/gke/manager.yaml
espindola: It is nice that it is autogenerated. It would be even nicer if it was not in the repository.

These things (copy and paste, autogenerated code) add up. I asked git to produce an archive of the master branch of both operator and seastar. The operator archive is 50 MiB. The Seastar one is 4.5 MiB. I have no doubt the seastar one would be a lot bigger if it included the auto generated files and dependencies (like boost).

Even scylla is "only" 47 MiB.

pkg/api/v1alpha1/zz_generated.deepcopy.go
zimnx: This one is deterministic (I hope so), but I can't tell that it will work the same in newer versions. Keeping autogenerated files in repository protects us from these kind of changes with low price.
It also helps with development process because it's easier to see how your code changes affects autogenerated ones. 
I don't have a good idea how to get rid of it, moving them to separate repository doesn't make sense. Not keeping it repository, breaks Go Module immutability.

examples/gke/manager.yaml
zimnx: I don't like this copy pasting too. But removing it is not trivial. Although It's planned for next release.
For autogenerated files, I refered to it in other comment.

PR #195, opened 2020-10-09
config: prefer configmap scylla yaml (#193)

### LINKED ISSUE #193: Passing a configMap with endpoint_snitch, this property is ignored and not overridden
**Describe the bug**
with scyllaOperator v.0.2.4, I am trying to pass a configMap with two parameters to override (as described at https://github.com/scylladb/scylla-operator/blob/master/docs/generic.md#configure-scylla):
```
endpoint_snitch: Ec2MultiRegionSnitch
storage_port: 7000
```
ConfigMap is called scylla-config, but the pod shows no changes:
```
[on_my_pod]>cat /etc/scylla/scylla.yaml
…
endpoint_snitch: GossipingPropertyFileSnitch
storage_port: 7000
… 
```

**To Reproduce**
1. Create file /tmp/scylla.yaml with content:
```
endpoint_snitch: Ec2MultiRegionSnitch
storage_port: 7000
```
2. Deploy configMap in K8S:
`kubectl create configmap scylla-config --from-file=/tmp/scylla.yaml -o yaml --dry-run | kubectl create -f -`

3. Rollout existing pods:
`kubectl rollout restart sts/cluster1-dc1-us-east-1a`

**Expected behavior**
File at /etc/scylla/scylla.yaml should shown overridden values. But `endpoint_snitch` is not overriden.

zimnx:
Previously operator pre-sets were overwriting user config map
which forbids changing snitch, rpc address and cluster name.
This change allows user to overwrite them on demand

Fixes #193

### DIFF
--- a/pkg/controllers/sidecar/config/config.go
+++ b/pkg/controllers/sidecar/config/config.go
@@ -58,7 +58,7 @@ func (s *ScyllaConfig) Setup(ctx context.Context) (*exec.Cmd, error) {
 	var cmd *exec.Cmd
 
 	s.logger.Info(ctx, "Setting up scylla.yaml")
-	if err = s.setupScyllaYAML(); err != nil {
+	if err = s.setupScyllaYAML(scyllaYAMLPath, scyllaYAMLConfigMapPath); err != nil {
 		return nil, errors.Wrap(err, "failed to setup scylla.yaml")
 	}
 	s.logger.Info(ctx, "Setting up cassandra-rackdc.properties")
@@ -79,15 +79,15 @@ func (s *ScyllaConfig) Setup(ctx context.Context) (*exec.Cmd, error) {
 // - cluster_name
 // - rpc_address
 // - endpoint_snitch
-func (s *ScyllaConfig) setupScyllaYAML() error {
+func (s *ScyllaConfig) setupScyllaYAML(configFilePath, configMapPath string) error {
 	// Read default scylla.yaml
-	scyllaYAMLBytes, err := ioutil.ReadFile(scyllaYAMLPath)
+	configFileBytes, err := ioutil.ReadFile(configFilePath)
 	if err != nil {
 		return errors.Wrap(err, "failed to open scylla.yaml")
 	}
 
 	// Read config map scylla.yaml
-	scyllaYAMLConfigMapBytes, err := ioutil.ReadFile(scyllaYAMLConfigMapPath)
+	configMapBytes, err := ioutil.ReadFile(configMapPath)
 	if err != nil {
 		s.logger.Info(context.Background(), "no scylla.yaml config map available")
 	}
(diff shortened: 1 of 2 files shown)


yanniszark:
@zimnx maybe a nice solution going forward would be to make the config file a template and expose the Pod's info in a context: e.g., PodIP, ClusterIP, ...

zimnx:
> @zimnx maybe a nice solution going forward would be to make the config file a template and expose the Pod's info in a context: e.g., PodIP, ClusterIP, ...

This is a nice idea, although I'm not sure about use case where these values would be useful. Config fields which needs these values are/should be filled by the Operator.

pkg/controllers/sidecar/config/config.go
```diff
@@ -79,15 +79,15 @@ func (s *ScyllaConfig) Setup(ctx context.Context) (*exec.Cmd, error) {
 // - cluster_name
 // - rpc_address
 // - endpoint_snitch
-func (s *ScyllaConfig) setupScyllaYAML() error {
+func (s *ScyllaConfig) setupScyllaYAML(scyllaYamlPath, scyllaYamlConfigMapPath string) error {
 	// Read default scylla.yaml
-	scyllaYAMLBytes, err := ioutil.ReadFile(scyllaYAMLPath)
+	scyllaYAMLBytes, err := ioutil.ReadFile(scyllaYamlPath)
```
espindola, line 84: Do we really need to have a scyllaYAMLPath and a scyllaYamlPath? For once, YAML is the common way to write acronyms in go. Second, it is not clear what the difference between the two are.

If we just need one in this function, it is fine to name parameter scyllaYAMLPath and let it hide the outer one. If we need both, we need a better name.

zimnx: renamed variables.

pkg/controllers/sidecar/config/config.go
```diff
@@ -98,22 +98,22 @@ func (s *ScyllaConfig) setupScyllaYAML() error {
 	cfg["rpc_address"] = "0.0.0.0"
 	cfg["endpoint_snitch"] = "GossipingPropertyFileSnitch"
 
-	overrideYAMLBytes, err := yaml.Marshal(cfg)
+	overrideBytes, err := yaml.Marshal(cfg)
 	if err != nil {
 		return errors.Wrap(err, "failed to parse override options for scylla.yaml")
 	}
-	scyllaYAMLConfigMapFilteredBytes, err := mergeYAMLs(scyllaYAMLConfigMapBytes, overrideYAMLBytes)
+	overwrittenBytes, err := mergeYAMLs(configFileBytes, overrideBytes)
```
espindola, line 105: The one real change in this patch is changing the first argument of mergeYAMLs in here and the following call, right?

For future changes like this, please split it in two patches in the same pull request
* Add arguments to setupScyllaYAML and use them, no functionality changed.
* Change the the priority by changing the arguments passed to mergeYAML

pkg/controllers/sidecar/config/config_test.go
```diff
@@ -111,3 +115,69 @@ func TestAllowedCPUs(t *testing.T) {
...
+		},
+		{
+			DefaultContent:    "some_key: 1",
+			UserConfigContent: "cluster_name: different_name",
+			ExpectedResult:    "cluster_name: different_name\nendpoint_snitch: GossipingPropertyFileSnitch\nrpc_address: 0.0.0.0\nsome_key: 1\n",
+		},
+		{
+			DefaultContent:    "some_key: 1",
+			UserConfigContent: "some_key: 2",
+			ExpectedResult:    "cluster_name: cluster-name\nendpoint_snitch: GossipingPropertyFileSnitch\nrpc_address: 0.0.0.0\nsome_key: 2\n",
```
espindola, line 143: Should we also have a test for "cluster_name:.." in DefaultContent?

espindola (approved):
LGTM with nits.

pkg/controllers/sidecar/config/config.go
zimnx: > The one real change in this patch is changing the first argument of mergeYAMLs in here and the following call, right?

Yes, other changes like adding function parameters were needed in order to add tests.

> For future changes like this, please split it in two patches in the same pull request
> 
>     * Add arguments to setupScyllaYAML and use them, no functionality changed.
> 
>     * Change the the priority by changing the arguments passed to mergeYAML

I agree, but in such small changes I tend to forget about such details. When everything is done, splitting them into multiple commits may double the time spent on task.

pkg/controllers/sidecar/config/config_test.go
zimnx: good idea, added.

Issue #104, opened 2020-05-29
Use trimpath


mmatczuk:
I can see lots of that full disk paths in logs `/home/henrik/go/src/github.com/scylladb/scylla-operator/pkg/scyllaclient/retry.go`

dahankzter:
Ah yes that would be nice indeed.

Issue #201, opened 2020-10-15
Can't build "docker-build" on fedora 33


espindola:
It tries to use docker, which doesn't work:

.....
Step 5/11 : RUN go mod download
 ---> Running in c770dd2bd8f2
OCI runtime create failed: this version of runc doesn't work on cgroups v2: unknown

Is there a way to use podman? I understand that we want to build a docker image, but why do we need to run something inside docker?

espindola:
And in case anyone is wondering, the reason I don't use podman-docker is because it conflicts with moby-engine which is required by kubernetes.

espindola:
Even if I remove kubernetes and install podman-docker the build still fails:

```
STEP 6: FROM alpine:3.12
Error: error building at STEP "RUN go mod download": error while running runtime: exit status 1
: exit status 125
make: *** [Makefile:77: docker-build] Error 1
```

espindola:
This works with plain docker:  See https://fedoramagazine.org/docker-and-fedora-32 for instructions.

Issue #200, opened 2020-10-15
"make fmt" not working on fedora 33


espindola:
Tested at a476d919b6b26b7c28a60f9d6f341ee7fc33a30f:

$ make fmt
go fmt ./pkg/...
go: cannot find GOROOT directory: /home/espindola/scylla/scylla-operator/bin/deps/go

And if I create that:

$ make fmt
go fmt ./pkg/...
make: go: Permission denied

Looks like make tries to execute the "go" directory, since we put deps on the PATH.

PR #202, opened 2020-10-15
Maybe use -trimpath

espindola:
I think this fixes issue #104. It is hard to say, since I can't build
without hacks because of issues #200 and #201.


zimnx:
Fix commit message to contain only description of changes and make sure to include "Fixes #104" in it to automatically close the issue.
PR Content itself looks fine.

zimnx (approved):

PR #203, opened 2020-10-16
Add a few missing dependencies

### LINKED ISSUE #200: "make fmt" not working on fedora 33

espindola:
I am not amused that our build requires downloading binaries from the
internet, but this at least fixes these targets.

Fixes #200

Signed-off-by: Rafael Ávila de Espíndola <espindola@scylladb.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,11 +60,11 @@ manifests: bin/deps controller-gen
 	kustomize build config/manager/default > examples/eks/manager.yaml
 
 # Run go fmt against code
-fmt:
+fmt: bin/deps
 	go fmt $(PKG)
 
 # Run go vet against code
-vet:
+vet: bin/deps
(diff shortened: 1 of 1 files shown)


zimnx:
I would like to get rid of internet dependencies too. It's fine to require some tooling available on local machine for development purposes. I'll create an issue for that.

zimnx (approved):

PR #206, opened 2020-10-19
actions: docs release action

zimnx:
Action which deploys documentation to Github Pages.

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -1,0 +1,34 @@
+name: Release docs
+
+on:
+  push:
+    branches:
+      - master
+      - 'branch-**'
+    paths:
+      - 'docs/**'
+
+jobs:
+  release:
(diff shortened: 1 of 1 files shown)


.github/workflows/pages.yaml
```diff
@@ -0,0 +1,34 @@
+name: Release docs
+
+on:
+  push:
+    branches:
+      - master
+      - 'branch-**'
+    paths:
+      - 'docs/**'
```
lauranovich, line 9: shouldn't the path be docs/source/something?

lauranovich (commented):
question within

.github/workflows/pages.yaml
zimnx: no, we want to update GH Pages when docs are touched, not only content, but scripts too.

lauranovich (approved):

PR #210, opened 2020-10-20
docs: explain preferred commit message format

zimnx:

Explain how commit messages should look like in contribution guide.

### DIFF
--- a/docs/source/contributing.md
+++ b/docs/source/contributing.md
@@ -120,6 +120,38 @@ git rebase -i HEAD~5
 
 Once your commit history is clean, ensure you have based on the [latest upstream](#updating-your-fork) before you open the PR.
 
+### Commit messages
+
+Please make the first line of your commit message a summary of the change that a user (not a developer) of Operator would like to read, 
+and prefix it with the most relevant directory of the change followed by a colon. 
+The changelog gets made by looking at just these first lines so make it good!
+
+If you have more to say about the commit, then enter a blank line and carry on the description. 
+Remember to say why the change was needed - the commit itself shows what was changed.
+
+Writing more is better than less. Comparing the behaviour before the change to that after the change is very useful. 
+Imagine you are writing to yourself in 12 months time when you've forgotten everything about what you just did, and you need to get up to speed quickly.
+
+If the change fixes an issue then write Fixes #1234 in the commit message. 
+This can be on the subject line if it will fit. If you don't want to close the associated issue just put #1234 and the change will get linked into the issue.
+
+Here is an example of a short commit message:
+
+```
+sidecar: log on reconcile loop - fixes #1234
+```
+
+And here is an example of a longer one:
+```
+
+api: now supports host networking (#1234)
+
+The operator CRD now has a "network" property that can be used to
+select host networking as well as setting the apropriate DNS policy.
+
+Fixes #1234
+```
+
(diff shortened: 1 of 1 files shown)


yanniszark:
I think it would be good to have a reference to this new section from the PR template.
This way, anyone opening a PR would be refered to the commit styling guide.

zimnx:
> I think it would be good to have a reference to this new section from the PR template.
> This way, anyone opening a PR would be refered to the commit styling guide.


There's a mention about Contribution guide:
```
<!-- Please take a look at our [Contributing](https://github.com/scylladb/scylla-operator/blob/master/docs/contributing.md)
```
This should be enough

docs/source/contributing.md
```diff
@@ -120,6 +120,38 @@ git rebase -i HEAD~5
 
 Once your commit history is clean, ensure you have based on the [latest upstream](#updating-your-fork) before you open the PR.
 
+### Commit messages
+
+Please make the first line of your commit message a summary of the change that a user (not a developer) of Operator would like to read,
```
espindola, line 125: This is a bit too strict. Not all changes are user visible. For example, it is good practice to split changes into multiple patches, so we can end up with a pull request that looks like:

1: Update a dependency
2: Regenerate some files
3: Implement awesome feature

Only the 3rd one is user visible.

docs/source/contributing.md
```diff
@@ -120,6 +120,38 @@ git rebase -i HEAD~5
 
 Once your commit history is clean, ensure you have based on the [latest upstream](#updating-your-fork) before you open the PR.
 
+### Commit messages
+
+Please make the first line of your commit message a summary of the change that a user (not a developer) of Operator would like to read, 
+and prefix it with the most relevant directory of the change followed by a colon.
```
espindola, line 126: s/directory/area/?

Not all areas map exactly to a directory. For example, it is probably better to say "build" than "." :-)

docs/source/contributing.md
```diff
@@ -120,6 +120,38 @@ git rebase -i HEAD~5
...
+and prefix it with the most relevant directory of the change followed by a colon. 
+The changelog gets made by looking at just these first lines so make it good!
+
+If you have more to say about the commit, then enter a blank line and carry on the description. 
+Remember to say why the change was needed - the commit itself shows what was changed.
+
+Writing more is better than less. Comparing the behaviour before the change to that after the change is very useful. 
+Imagine you are writing to yourself in 12 months time when you've forgotten everything about what you just did, and you need to get up to speed quickly.
+
+If the change fixes an issue then write Fixes #1234 in the commit message.
```
espindola, line 135: The "Fixes #1234" can also be in the merge message (i.e.: The pull request), no?

espindola (approved):
LGTM

I have a few suggestions, but they are minor.

Thanks

docs/source/contributing.md
zimnx: I think so. Commit message is automatically appended to PR description. Having it in commit msg also helps with git log grepping.

docs/source/contributing.md
zimnx: I agree, if anyone think that splitting some changes into separate commit bending this rule is beneficial, I'm all in. I don't think this needs any mention in guide.

PR #211, opened 2020-10-20
sidecar: Make it possible to pass command arguments to scyla

### LINKED ISSUE #177: Add ability to modify/replace default SCYLLA_ARGS
* Feature Request

Scylla Operator has support for modifying `scylla.yaml` but not for modifying command line arguments for `scylla` binary.

These arguments are configured by `/etc/sysconfig/scylla-server` file:

```
...
# scylla arguments
SCYLLA_ARGS="--log-to-syslog 0 --log-to-stdout 1 --default-log-level info --network-stack posix"
...
```

It'll be useful to have ability to modify them too.

SCT has a feature in test configuration to modify these arguments and some of the tests use it. Even more, SCT appends following options by default: `--blocked-reactor-notify-ms 500 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-
internal-error 1 --abort-on-ebadf 1 --enable-sstable-key-validation 1`. So, for making k8s backend complete we need to implement this feature.

dkropachev:
Addresses https://github.com/scylladb/scylla-operator/issues/177

Exposes scyllaArgs at cluster.spec 

Using scyllaArgs will make scylla startup to fail on versions that does not contain https://github.com/scylladb/scylla/pull/7458

Resolves #177

**Checklist:**
- [ ] ~~Documentation has been updated, if necessary.~~

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"github.com/blang/semver"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -51,7 +52,8 @@ type ClusterSpec struct {
 	// Sysctl properties to be applied during initialization
 	// given as a list of key=value pairs.
 	// Example: fs.aio-max-nr=232323
-	Sysctls []string `json:"sysctls,omitempty"`
+	Sysctls    []string `json:"sysctls,omitempty"`
+	ScyllaArgs string   `json:"scyllaArgs,omitempty"`
 	// Networking config
 	Network Network `json:"network,omitempty"`
 	// Repairs specifies repair task in Scylla Manager.
@@ -268,6 +270,14 @@ type ClusterList struct {
 	Items           []Cluster `json:"items"`
 }
 
+// Version of scylla docker starting from which passing arguments via entry-point is supported
+var ScyllaVersionThatSupportsArgsText = "4.2.0"
+var ScyllaVersionThatSupportsArgs semver.Version
+
 func init() {
+	var err error
+	if ScyllaVersionThatSupportsArgs, err = semver.Parse(ScyllaVersionThatSupportsArgsText); err != nil {
+		panic(err)
+	}
 	SchemeBuilder.Register(&Cluster{}, &ClusterList{})
 }
--- a/pkg/controllers/sidecar/config/config.go
+++ b/pkg/controllers/sidecar/config/config.go
@@ -3,9 +3,12 @@ package config
 import (
 	"context"
 	"fmt"
+	"github.com/blang/semver"
 	"io/ioutil"
+	"k8s.io/utils/pointer"
 	"os"
 	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -156,6 +159,32 @@ func loadProperties(fileName string, logger log.Logger) *properties.Properties {
 	return p
 }
 
+var scyllaArgumentsRegexp = regexp.MustCompile(`--([^= ]+)(="[^"]+"|=[^ ]+|[ \t]+"[^"]+"|[ \t]+[^-][^-]?[^ ]*|)`)
+
+func convertScyllaArguments(scyllaArguments string) map[string]string {
+	output := make(map[string]string)
+	for _, value := range scyllaArgumentsRegexp.FindAllStringSubmatch(scyllaArguments, -1) {
+		if value[2] == "" {
+			output[value[1]] = ""
+		} else if value[2][0] == '=' {
+			output[value[1]] = strings.TrimSpace(value[2][1:])
+		} else {
+			output[value[1]] = strings.TrimSpace(value[2])
+		}
+	}
+	return output
+}
+
+func appendScyllaArguments(ctx context.Context, s *ScyllaConfig, scyllaArgs string, scyllaFinalArgs map[string]*string) {
+	for argName, argValue := range convertScyllaArguments(scyllaArgs) {
+		if existing := scyllaFinalArgs[argName]; existing == nil {
+			scyllaFinalArgs[argName] = pointer.StringPtr(strings.TrimSpace(argValue))
+		} else {
+			s.logger.Info(ctx, fmt.Sprintf("ScyllaArgs: argument '%s' is ignored, it is already in the list", argName))
+		}
+	}
+}
+
 func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 	m := s.member
 	// Get seeds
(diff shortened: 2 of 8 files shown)


zimnx:
> Using scyllaArgs will make scylla startup to fail on versions that does not contain [scylladb/scylla#7458](https://github.com/scylladb/scylla/pull/7458)

Which version of Scylla is going to support it? 
Either we should add validation if used Scylla versions supports that, or at least log warning that it may not work.

dkropachev:
> > Using scyllaArgs will make scylla startup to fail on versions that does not contain [scylladb/scylla#7458](https://github.com/scylladb/scylla/pull/7458)
> 
> Which version of Scylla is going to support it?
> Either we should add validation if used Scylla versions supports that, or at least log warning that it may not work.

Hopefully it will work on 4.3

pkg/controllers/sidecar/config/config.go
```diff
@@ -178,17 +179,22 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
+	seedsStr := strings.Join(seeds, ",")
+	overprovisionedStr := "0"
+	smpStr := strconv.Itoa(shards)
+	args := map[string]*string{
+		"listen-address":        &m.IP,
+		"broadcast-address":     &m.StaticIP,
+		"broadcast-rpc-address": &m.StaticIP,
+		"seeds":                 &seedsStr,
+		"developer-mode":        &devMode,
+		"overprovisioned":       &overprovisionedStr,
```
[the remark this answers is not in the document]
dkropachev: I have switched to "k8s.io/utils/pointer", since we are already have it.

pkg/controllers/sidecar/config/config.go
```diff
@@ -203,10 +209,39 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		if err := s.validateCpuSet(ctx, cpusAllowed, shards); err != nil {
 			return nil, errors.WithStack(err)
 		}
-		args = append(args, fmt.Sprintf("--cpuset=%s", cpusAllowed))
+		args["cpuset"] = &cpusAllowed
+	}
+
+	if cluster.Spec.ScyllaArgs != nil {
+		tmp := regexp.MustCompile(`--([^= ]+)(="[^"]+"|=[^ ]+|[ \t]+"[^"]+"|[ \t]+[^-][^-]?[^ ]*|)`).FindAllStringSubmatch(*cluster.Spec.ScyllaArgs, -1)
```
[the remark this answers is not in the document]
dkropachev: switched ScyllaArgs to map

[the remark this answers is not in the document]
dkropachev: > If you make ScyllaArgs a slice, you won't have to worry about slicing.
> For validation we have a validating webhook which rejects CRD with invalid fields. Check `cluster_webhook.go`
Problem here that you don't have list of valid fields, unless you run scylla --help.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -51,7 +51,8 @@ type ClusterSpec struct {
 	// Sysctl properties to be applied during initialization
 	// given as a list of key=value pairs.
 	// Example: fs.aio-max-nr=232323
-	Sysctls []string `json:"sysctls,omitempty"`
+	Sysctls    []string `json:"sysctls,omitempty"`
+	ScyllaArgs *string  `json:"scyllaArgs,omitempty"`
```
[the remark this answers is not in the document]
dkropachev: It would be better to have it as Map, but unfortunatelly controller-get v0.3.0 does not support nillable values in the map, whcih does not allow us to have flag arguments there.
Currently I've move it back to string.

pkg/controllers/sidecar/config/config.go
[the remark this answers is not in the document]
dkropachev: Turned out that controller-get v0.3.0 does not support nillable values in the map, which does not allow us to have flag arguments there, so I have move it back to string and add tests to that part of the logic.

pkg/controllers/sidecar/config/config.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	pointer2 "k8s.io/utils/pointer"
```
zimnx, line 7: pointer2? package is called `pointer`, i don't see name conflict in mports

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -51,7 +51,8 @@ type ClusterSpec struct {
 	// Sysctl properties to be applied during initialization
 	// given as a list of key=value pairs.
 	// Example: fs.aio-max-nr=232323
-	Sysctls []string `json:"sysctls,omitempty"`
+	Sysctls    []string `json:"sysctls,omitempty"`
+	ScyllaArgs string   `json:"scyllaArgs,omitempty"`
```
zimnx, line 55: now it's required, but should be optional

pkg/controllers/sidecar/config/config_test.go
```diff
@@ -186,3 +186,59 @@ func writeTempFile(t *testing.T, namePattern, content string) string {
 
 	return tmp.Name()
 }
+
+func TestScyllaArguments(t *testing.T) {
```
zimnx, line 190: Split it to multiple tests. Failure in one case shouldn't cause failure in cases where regexp works fine.
You can use table tests: https://github.com/golang/go/wiki/TableDrivenTests

examples/eks/manager.yaml
```diff
@@ -0,0 +1,275 @@
+apiVersion: v1
```
zimnx, line 1: master is missing this update after manager merge. Please extract it to separate commit

pkg/controllers/sidecar/config/config.go
```diff
@@ -156,6 +158,22 @@ func loadProperties(fileName string, logger log.Logger) *properties.Properties {
 	return p
 }
 
+var scyllaArgumentsRegexp = regexp.MustCompile(`--([^= ]+)(="[^"]+"|=[^ ]+|[ \t]+"[^"]+"|[ \t]+[^-][^-]?[^ ]*|)`)
+
+func convertScyllaArguments(scyllaArguments string) map[string]*string {
+	output := make(map[string]*string)
```
zimnx, line 164: nil pointers aren't added to this map, you can keep `map[string]string`. Later there's unnecessary dereferece:  `pointer2.StringPtr(strings.TrimSpace(*argValue))`

examples/eks/manager.yaml
dkropachev: removed

pkg/controllers/sidecar/config/config.go
dkropachev: fixed

pkg/controllers/sidecar/config/config_test.go
dkropachev: done

pkg/api/v1alpha1/cluster_types.go
dkropachev: scylla.scylladb.com_clusters.yaml resulted in 
```
          required:
          - agentVersion
          - datacenter
          - version

```
So it is not required on scheme level, also I've checked it by deploying cluster with no scyllaArgs and it passed thru

pkg/controllers/sidecar/config/config.go
dkropachev: fixed

pkg/controllers/sidecar/config/config.go
```diff
@@ -203,10 +222,29 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		if err := s.validateCpuSet(ctx, cpusAllowed, shards); err != nil {
 			return nil, errors.WithStack(err)
 		}
-		args = append(args, fmt.Sprintf("--cpuset=%s", cpusAllowed))
+		args["cpuset"] = &cpusAllowed
+	}
+
+	if len(cluster.Spec.ScyllaArgs) > 0 {
+		for argName, argValue := range convertScyllaArguments(cluster.Spec.ScyllaArgs) {
```
zimnx, line 229: It doesn't work for older Scylla versions, it doesn't even work for released versions. We definitely need some kind of warning or info log level message about missing support.

dkropachev: Validation has been added.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -268,6 +270,12 @@ type ClusterList struct {
 	Items           []Cluster `json:"items"`
 }
 
+// Version of scylla docker starting from which passing arguments via entry-point is supported
+var ScyllaVersionThatSupportsArgsText = "4.2.0"
+var ScyllaVersionThatSupportsArgs *semver.Version
```
zimnx, line 275: why pointer? Every usage dereference it. Use value.

zimnx (commented):
last comments

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -268,6 +270,12 @@ type ClusterList struct {
...
+
 func init() {
+	tmp, _ := semver.Parse(ScyllaVersionThatSupportsArgsText)
```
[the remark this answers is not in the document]
dkropachev: fixed

pkg/api/v1alpha1/cluster_types.go
dkropachev: fixed

zimnx (approved):
LGTM

PR #219, opened 2020-11-02
api: rename Cluster kind to ScyllaCluster in missing places

### LINKED ISSUE #144: Consider using less common kind
* Feature Request

I would suggest to replace wide meaning kind name  `kind: Cluster` to something less common.

Most of the operators using own kinds, eg:
- Elasticsearch operator uses `kind: Elasticsearch`
- Redis operator uses `kind: RedisCluster`
- PostgreSQL operator uses `kind: postgresql`
- Cassandra operator uses `kind: CassandraDatacenter`

Scylla-operator should use something simmilar.

Using so wide meaning kinds is not common in Kubernetes.

**Additional Information:**

zimnx:
Problems solved:
 - webhooks were watching old "Cluster" resource
- controller wasn't able to update cluster status
- controller watch on owned StatefulSet events was registered on old "Cluster" resource
    
Fixes #144

### DIFF
--- a/config/manager/rbac/role.yaml
+++ b/config/manager/rbac/role.yaml
@@ -1,8 +1,10 @@
+
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: role
+  name: manager-role
 rules:
 - apiGroups:
   - ""
@@ -10,33 +12,8 @@ rules:
   - events
   verbs:
   - create
-  - update
-  - patch
-- apiGroups:
-  - apps
-  resources:
-  - statefulsets
-  verbs:
-  - get
-  - list
-- apiGroups:
-  - scylla.scylladb.com
-  resources:
-  - clusters
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups:
-  - scylla.scylladb.com
-  resources:
-  - clusters/status
-  verbs:
   - patch
   - update
-  - watch
-  - get
-  - list
 - apiGroups:
   - ""
   resources:
@@ -52,3 +29,28 @@ rules:
   verbs:
   - get
   - list
+- apiGroups:
+  - apps
+  resources:
+  - statefulsets
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - scylla.scylladb.com
+  resources:
+  - scyllaclusters
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - scylla.scylladb.com
+  resources:
+  - scyllaclusters/status
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
--- a/examples/eks/cluster.yaml
+++ b/examples/eks/cluster.yaml
@@ -40,7 +40,7 @@ rules:
   - apiGroups:
       - scylla.scylladb.com
     resources:
-      - clusters
+      - scyllaclusters
     verbs:
       - get
 
@@ -74,7 +74,7 @@ subjects:
 
 # Scylla Cluster
 apiVersion: scylla.scylladb.com/v1alpha1
-kind: Cluster
+kind: ScyllaCluster
 metadata:
   name: scylla-cluster
   namespace: scylla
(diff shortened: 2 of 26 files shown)


espindola:
Please open one issue for adding regression tests for each of the problems you found.

Some of the files in here are generated, no? Please split the patch in

* Change the non generated files
* Change to the generated files

In the second patch, say which command was used to regenerate them.

zimnx:
> Please open one issue for adding regression tests for each of the problems you found.
> 
Added issue to implement basic scenario first. I wasn't able to spawn any node using master due to above problems. https://github.com/scylladb/scylla-operator/issues/220

> Some of the files in here are generated, no? Please split the patch in
> 
>     * Change the non generated files
> 
>     * Change to the generated files
> 
> 
> In the second patch, say which command was used to regenerate them.

done

examples/eks/cluster.yaml
```diff
@@ -40,7 +40,7 @@ rules:
   - apiGroups:
       - scylla.scylladb.com
     resources:
-      - clusters
+      - scyllaclusters
```
espindola, line 43: This file is a copy and paste from the one in config, no? It seems to have some arbitrary dependencies:

````
$ diff -u  ./config/manager/cluster/cluster.yaml examples/eks/cluster.yaml
... 
 kind: Cluster
 metadata:
-  name: cluster
-  namespace: scylla-manager
+  name: scylla-cluster
+  namespace: scylla
```

examples/generic/cluster.yaml
```diff
@@ -40,7 +40,7 @@ rules:
   - apiGroups:
       - scylla.scylladb.com
     resources:
-      - clusters
+      - scyllaclusters
```
espindola, line 43: and this one.

examples/gke/cluster.yaml
```diff
@@ -40,7 +40,7 @@ rules:
   - apiGroups:
       - scylla.scylladb.com
     resources:
-      - clusters
+      - scyllaclusters
```
espindola, line 43: and this one.

pkg/api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -21,7 +21,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
espindola, line 24: According to the file name this file is generated.

examples/eks/cluster.yaml
zimnx: `cluster.yaml` files are not generated. These are examples what user should provide and these were manually crafted or copied from generic if they suits.

zimnx:
@espindola I found that autogeneration stopped working some time ago or wasn't there in some files. I fixed the issues, extracted to separate commits. PTAL

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -33,14 +33,11 @@ func (r *ScyllaCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
espindola, line 36: Is this change part of running "make manifests"?

config/manager/rbac/role.yaml
```diff
@@ -6,19 +6,25 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
-- resources:
```
espindola: This file is modified in 3 of the 4 patches. It is auto generated, correct? If so, it should be modified in only one patch, no?

espindola (commented):
Thanks a lot for fixing the Makefile.

I think this needs to be rebased on top of current master.

Auto generated files are still modified in multiple patches. Why?

config/manager/rbac/role.yaml
zimnx: done, somehow missed it during rebasing and squashing

pkg/api/v1alpha1/cluster_webhook.go
zimnx: no, i manually removed this comment. It was generated by the `kubebuilder` during project initializaiton.

espindola (approved):
LGTM

PR #233, opened 2020-11-09
update manager and agent version to 2.2 to support GCP bakcp

gcarmin:

Resolves #

**Checklist:**

### DIFF
--- a/examples/gke/manager.yaml
+++ b/examples/gke/manager.yaml
@@ -182,7 +182,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: scylladb/scylla-manager:2.1.2
+        image: scylladb/scylla-manager:2.2.0
         imagePullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


examples/gke/manager.yaml
```diff
@@ -182,7 +182,7 @@ spec:
           valueFrom:
```
zimnx, line 182: manager.yaml is generated based on https://github.com/scylladb/scylla-operator/blob/978b5fd5f0c2751f97c1a567c960d2b6a84f83d6/config/manager/cluster/cluster.yaml#L73
and 
https://github.com/scylladb/scylla-operator/blob/5688a6fb99b49b2c93913c5195c9eefc30782a3e/config/manager/manager/kustomization.yaml#L9

zimnx (changes requested):

examples/gke/manager.yaml
gcarmin: ok - I'll update both.
thanks

PR #236, opened 2020-11-10
docs: update theme automatically

dgarcia360:
Installs the latest documentation theme version on every documentation build when there are no breaking changes. 
This will speed up the development of the theme, since it would allow us to release updates without having to submit a PR per repository. https://github.com/scylladb/sphinx-scylladb-theme/issues/75

Also, the ``gh-branch`` existing contents are deleted before publishing the new docs. By doing so, we can make sure there are no conflicts with documentation that was generated previously. https://github.com/scylladb/scylla-monitoring/issues/1087

### DIFF
--- a/docs/_utils/deploy.sh
+++ b/docs/_utils/deploy.sh
@@ -5,16 +5,21 @@
 
 set -eu -o pipefail
 
-# Clone repo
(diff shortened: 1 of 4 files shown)


tzach:
LGTM

zimnx (approved):

PR #246, opened 2020-11-13
Make it possible to control timezone on scylla nodes

### LINKED ISSUE #239: Make it possible to control timezone on scylla nodes
* Feature Request
It would be great to have option on cluster that would control time zone on the nodes
Article that describes how to do that:
https://medium.com/@yildirimabdrhm/kubernetes-timezone-management-8cc139b01f9d

Enable control over tz on scylla nodes

**Additional Information:**

dkropachev:
Add timezone option to the cluster specification that links /etc/localtime on container to the particular timezone file on the host

Resolves #239

**Checklist:**

### DIFF
--- a/examples/generic/cluster.yaml
+++ b/examples/generic/cluster.yaml
@@ -83,6 +83,7 @@ metadata:
 spec:
   version: 4.0.0
   agentVersion: 2.0.2
+  timeZone: GMT
   developerMode: true
   datacenter:
     name: us-east-1
@@ -90,7 +91,7 @@ spec:
       - name: us-east-1a
         scyllaConfig: "scylla-config"
         scyllaAgentConfig: "scylla-agent-config"
-        members: 3
+        members: 1
         storage:
           capacity: 5Gi
         resources:
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -54,6 +54,13 @@ type ClusterSpec struct {
 	// Example: fs.aio-max-nr=232323
 	Sysctls    []string `json:"sysctls,omitempty"`
 	ScyllaArgs string   `json:"scyllaArgs,omitempty"`
+
+	// Timezone property links /etc/localtime on scylla containers to the host timezone file.
+	// Host timezone file path is calculated of Timezone value,
+	//   if Timezone starts with '/' it is assumed to be full path the the timezone file.
+	// Otherwise host file path is calculated by prepending Timezone with '/usr/share/zoneinfo/'.
+	Timezone string `json:"timeZone,omitempty"`
+
 	// Networking config
 	Network Network `json:"network,omitempty"`
 	// Repairs specifies repair task in Scylla Manager.
--- a/pkg/controllers/cluster/resource/resource.go
+++ b/pkg/controllers/cluster/resource/resource.go
@@ -350,6 +350,37 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.ScyllaClust
 	if sysctlContainer != nil {
 		sts.Spec.Template.Spec.InitContainers = append(sts.Spec.Template.Spec.InitContainers, *sysctlContainer)
 	}
+
+	if c.Spec.Timezone != "" {
+		var timezone string
+		hostPath := corev1.HostPathFile
+		if c.Spec.Timezone[0] == '/' {
+			timezone = c.Spec.Timezone
+		} else {
+			timezone = "/usr/share/zoneinfo/" + c.Spec.Timezone
+		}
+
+		sts.Spec.Template.Spec.Containers[0].VolumeMounts = append(
+			sts.Spec.Template.Spec.Containers[0].VolumeMounts,
+			corev1.VolumeMount{Name: "host-timezone", MountPath: "/etc/localtime", ReadOnly: true})
+
+		// /etc/localtime is getting overridden by tmpfiles whenever scylla container is started.
+		// It is getting linked to /usr/share/zoneinfo/UTC, so we need to mount ther too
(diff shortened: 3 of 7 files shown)


examples/generic/cluster.yaml
```diff
@@ -83,14 +83,15 @@ metadata:
 spec:
   version: 4.0.0
   agentVersion: 2.0.2
+  timeZone: GMT
```
zimnx, line 86: nit: I think for example UTC is the best one

examples/generic/cluster.yaml
```diff
@@ -83,14 +83,15 @@ metadata:
...
   developerMode: true
   datacenter:
     name: us-east-1
     racks:
       - name: us-east-1a
         scyllaConfig: "scylla-config"
         scyllaAgentConfig: "scylla-agent-config"
-        members: 3
+        members: 1
```
zimnx, line 94: let's not change this one

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -54,6 +54,13 @@ type ClusterSpec struct {
 	// Example: fs.aio-max-nr=232323
 	Sysctls    []string `json:"sysctls,omitempty"`
 	ScyllaArgs string   `json:"scyllaArgs,omitempty"`
+
+	// Timezone property links /etc/localtime on scylla containers to the host timezone file.
+	// Host timezone file path is calculated of Timezone value,
+	//   if Timezone starts with '/' it is assumed to be full path the the timezone file.
```
zimnx, line 60:
```suggestion
	// if Timezone starts with '/' it is assumed to be full path the the timezone file.
```

zimnx: Some examples in comment section would be great

pkg/controllers/cluster/resource/resource.go
```diff
@@ -350,6 +350,37 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.ScyllaClust
...
+			timezone = c.Spec.Timezone
+		} else {
+			timezone = "/usr/share/zoneinfo/" + c.Spec.Timezone
+		}
+
+		sts.Spec.Template.Spec.Containers[0].VolumeMounts = append(
+			sts.Spec.Template.Spec.Containers[0].VolumeMounts,
+			corev1.VolumeMount{Name: "host-timezone", MountPath: "/etc/localtime", ReadOnly: true})
+
+		// /etc/localtime is getting overridden by tmpfiles whenever scylla container is started.
```
zimnx, line 367: So if above is overwriten why do we mount it in first place?

pkg/controllers/cluster/resource/resource.go
```diff
@@ -350,6 +350,37 @@ func StatefulSetForRack(r scyllav1alpha1.RackSpec, c *scyllav1alpha1.ScyllaClust
...
+			sts.Spec.Template.Spec.Containers[0].VolumeMounts,
+			corev1.VolumeMount{Name: "host-timezone", MountPath: "/usr/share/zoneinfo/UTC", ReadOnly: true})
+
+		sts.Spec.Template.Spec.Volumes = append(
+			sts.Spec.Template.Spec.Volumes,
+			corev1.Volume{
+				Name: "host-timezone",
+				VolumeSource: corev1.VolumeSource{
+					HostPath: &corev1.HostPathVolumeSource{Path: timezone, Type: &hostPath},
+				}})
```
zimnx, line 381:
```suggestion
		scyllaVolumeMounts := sts.Spec.Template.Spec.Containers[0].VolumeMounts
		scyllaVolumeMounts = append(scyllaVolumeMounts,
			corev1.VolumeMount{
				Name: "host-timezone", 
				MountPath: "/usr/share/zoneinfo/UTC", 
				ReadOnly: true,
			})
		volumes := sts.Spec.Template.Spec.Volumes
		volumes = append(volumes,
			corev1.Volume{
				Name: "host-timezone",
				VolumeSource: corev1.VolumeSource{
					HostPath: &corev1.HostPathVolumeSource{
						Path: timezone, 
						Type: pointer.StringPtr(corev1.HostPathFile),
					},
				}})
```

zimnx:
We have some docs for CRD, please update them. You can also make a section about how to change the timezone.

pkg/controllers/cluster/resource/resource.go
dkropachev: I think that rewriting of the timezone is result of a leftovers that were not removed from scylla docker image by mistake.
But despite that there are two reasons to do that:
1. It does not break anything in current docker image
2. It will make whole thing work when/if scylla image get rid of tmpfiles

pkg/api/v1alpha1/cluster_types.go
dkropachev: added

pkg/controllers/cluster/resource/resource.go
dkropachev: Results in:
pkg/controllers/cluster/resource/resource.go:378:61: cannot use pointer.StringPtr("k8s.io/api/core/v1".HostPathFile) (type *string) as type *"k8s.io/api/core/v1".HostPathType in field value
pkg/controllers/cluster/resource/resource.go:378:84: cannot use "k8s.io/api/core/v1".HostPathFile (type "k8s.io/api/core/v1".HostPathType) as type string in argument to pointer.StringPtr

Also construction like this won't update sts.Spec.Template.Spec.Containers[0].VolumeMounts:

```

VolumeMounts := sts.Spec.Template.Spec.Containers[0].VolumeMounts
append(VolumeMounts, ...)

```

dkropachev:
bad solution, let's redo when decision on how to do that is made at https://github.com/scylladb/scylla-operator/issues/239

PR #250, opened 2020-11-16
controller: added replace dead node feature (#48)

### LINKED ISSUE #48: Replace a dead node
This operation requires the use of `replace_address_first_boot`. This means we probably need a tag that the operator can use to identify this special scenario.

zimnx:
Dead node can be replaced using special label added to member service.

When label "scylla/replace" with empty string value is added to member
service, Operator will remove PVC attached to dead node, delete dead pod
and remove member service.
New pod will be scheduled on different node, and scylla will be run with
additional parameter `--replace-address-first-boot` together with IP
address of deleted node.

Fixes #48

**Checklist:**


pkg/controllers/cluster/service.go
```diff
@@ -44,10 +44,16 @@ func (cc *ClusterReconciler) syncMemberServices(ctx context.Context, c *scyllav1
 		}
 		for _, pod := range podlist.Items {
 			memberService := resource.MemberServiceForPod(&pod, c)
-			_, err := controllerutil.CreateOrUpdate(ctx, cc.Client, memberService, serviceMutateFn(ctx, memberService, cc.Client))
+			op, err := controllerutil.CreateOrUpdate(ctx, cc.Client, memberService, serviceMutateFn(ctx, memberService, cc.Client))
 			if err != nil {
 				return errors.Wrapf(err, "error syncing member service %s", memberService.Name)
 			}
+			switch op {
+			case controllerutil.OperationResultCreated:
```
mmatczuk, line 52: consider on case with ,

docs/source/node_operations.md
```diff
@@ -0,0 +1,73 @@
...
+   Datacenter: us-east-1
+   =====================
+   Status=Up/Down
+   |/ State=Normal/Leaving/Joining/Moving
+   --  Address        Load       Tokens       Owns    Host ID                               Rack
+   UN  10.43.125.110  74.62 KB   256          ?       8ebd6114-969c-44af-a978-87a4a6c65c3e  us-east-1a
+   UN  10.43.231.189  91.03 KB   256          ?       35d0cb19-35ef-482b-92a4-b63eee4527e5  us-east-1a
+   UN  10.43.191.172  74.77 KB   256          ?       1ffa7a82-c41c-4706-8f5f-4d45a39c7003  us-east-1a
+   ```
+1. Run the repair on the cluster to make sure that the data is synced with the other nodes in the cluster. You can use Scylla Manager to run the repair.
```
mmatczuk, line 73: Add example or link

examples/eks/operator.yaml
```diff
@@ -1564,6 +1564,11 @@ spec:
                     description: ReadyMembers is the number of ready members in the specific Rack
                     format: int32
                     type: integer
+                  replace_addresses:
```
mmatczuk, line 1567: Shouldn't this be replace_address_first_boot?

mmatczuk (approved):

PR #269, opened 2020-11-24
docs: Fixed LATEST_VERSION should be available for all the steps

dgarcia360:
The variable LATEST_VERSION was set in the last step "deploy".
This PR makes the variable globally available since the step "Build docs" needs to read it too.

Related issue: scylladb/sphinx-scylladb-theme#74


dgarcia360:
@lauranovich @tzach please review

PR #285, opened 2020-12-01
alternator: Add write-isolation parameter with default 'always'

zimnx:
Alternator since 4.0+ requires additional parameter. Unfortunately it refuses to start even when mounted scylla.yaml contain desired value. 

**Checklist:**

### DIFF
--- a/config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
+++ b/config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
@@ -45,6 +45,8 @@ spec:
                   description: Port on which to bind the Alternator API
                   format: int32
                   type: integer
+                write_isolation:
+                  type: string
               type: object
             automaticOrphanedNodeCleanup:
               description: AutomaticOrphanedNodeCleanup controls if automatic orphan node cleanup should be performed.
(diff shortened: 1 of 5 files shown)


dorlaor:
'Always' is too restrictive and slow. better do the rmw-only

zimnx:
Good point. Fixed in #286

nyh:
> 'Always' is too restrictive and slow. better do the rmw-only

I don't understand this statement... "Always" isn't restrictive - it is the most general choice which doesn't make any assumptions on the workload and is always correct. Yes, it is significantly slower than "rmw-only" for pure-write workloads - e.g., injection of a lot of new data into the database.

Dor, you asked to *not* have a default for the write isolation mode, and asked that Alternator users must always consider the options and chose for themselves - and that Alternator shouldn't start at all without the user choosing one of these options. Did you change your mind, and now believe tht rmw-only should be the default? Should we make it the default in Alternator itself, instead of one of the upper layers like Operator?

dorlaor:
I wanted to start w/ a simple solution that works w/ good performance out
of the
box. Later, we can think about exposing this option but for the
immediate present,
we should just make it work out of the box (w/ good perf).

PR #294, opened 2020-12-11
Major upgrade procedure

### LINKED ISSUE #51: Upgrade scylla major version
Currently we employ a check that the scylla version of the target state is the same as the previous.
This prohibits an upgrade and we need to device a way to allow the user to upgrade.

zimnx:
Current patch version upgrade procedure is ran when cluster is being
upgraded to next patch version.
New major upgrade procedure handles all others upgrades.

Procedure:
* Check if the cluster has schema agreement (using API call)
* Take `system` and `system_schema` tables snapshot on all nodes in
parallel.

For each node:

* Drain node
* Backup the data - snapshot of all data keyspaces
* Update Scylla image by restarting Pod
* Validate if node is up and version is updated via API call
* Clear data snapshot

After last node:

* Delete `system` and `system_schema` table snapshots on all nodes in
parallel

Fixes #51


mmatczuk:
@zimnx can we start putting the stolen packages to a separate repo?

zimnx:
@mmatczuk sure thing, I plan to move scyllaclient, managerclient, and most of utils packages to separate repo. Since this is a major refactor, i would like to do it in one batch, as a part of separate PR. Once this is done I'm not going to accept any source code copying.

mmatczuk:
The commit 8c78f9e1db268cc034180e6d6ab3e66c7861d015 changes go.mod please change msg.

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+
+// ErrEventRejected is the error returned when the state machine cannot process
+// an event in the state that it is in.
+var ErrEventRejected = errors.New("event rejected")
+
+const (
+	// Default represents the default state of the system.
+	Default StateType = ""
+
+	// NoOp represents a no-op event.
```
mmatczuk, line 20: Which is?

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/pkg/errors"
+)
+
+// ErrEventRejected is the error returned when the state machine cannot process
+// an event in the state that it is in.
+var ErrEventRejected = errors.New("event rejected")
+
+const (
+	// Default represents the default state of the system.
+	Default StateType = ""
```
mmatczuk, line 18: By default FMS has empty set of states.

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// NoOp represents a no-op event.
+	NoOp EventType = "NoOp"
+)
+
+// StateType represents an extensible state type in the state machine.
+type StateType string
```
mmatczuk, line 25: I'd suggest rename to State, the current state looks like it's implementation specific and should not be exported.

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+
+// EventType represents an extensible event type in the state machine.
+type EventType string
```
mmatczuk, line 28: Again this looks like event

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+
+// EventContext represents the context to be passed to the action implementation.
```
mmatczuk, line 30: As far as I can tell it's unused.

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+// state, or an error if the event can't be handled in the given state.
+func (s *StateMachine) getNextState(event EventType) (StateType, error) {
+	if state, ok := s.States[s.Current]; ok {
+		if state.Events != nil {
+			if next, ok := state.Events[event]; ok {
+				return next, nil
+			}
+		}
+	}
+	return Default, ErrEventRejected
```
mmatczuk, line 74: Looks like state did not changed why not return state?

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+	Events Events
+}
+
+type Hook func(ctx context.Context, currentState, nextState StateType, event EventType) error
+
+// States represents a mapping of states and their implementations.
+type States map[StateType]State
+
+// StateMachine represents the state machine.
+type StateMachine struct {
```
mmatczuk, line 51: You are exporting mutex protected fields directly, I'd suggest to use builder or at least a constructor.

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+	// Current represents the current state.
+	Current StateType
+
+	// States holds the configuration of states and events handled by the state machine.
+	States States
+
+	// TransitionHook is called on every state transition.
+	TransitionHook Hook
+
+	mutex sync.Mutex
```
mmatczuk, line 61: std naming is mu, and it's placed above a group of fields it protects.

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+		return err
+	}
+	if nextEvent == NoOp {
+		return nil
+	}
+
+	return s.sendEvent(ctx, nextEvent)
+}
+
+func (s *StateMachine) sendEvent(ctx context.Context, event EventType) error {
```
mmatczuk, line 94: It seems that this shall be mutex protected because state can be changed on the flight.
Maybe this is a single threaded FSM and the mutex shall be removed altogether?

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+			return errors.Wrapf(ErrEventRejected, "unknown state %q for event %q", nextState, event)
+		}
+
+		if s.TransitionHook != nil {
+			if err := s.TransitionHook(ctx, s.Current, nextState, event); err != nil {
+				return err
+			}
+		}
+		// Transition over to the next state.
+		s.Current = nextState
```
mmatczuk, line 114: Mutex?

pkg/util/fsm/fsm.go
```diff
@@ -0,0 +1,127 @@
...
+			if next, ok := state.Events[event]; ok {
+				return next, nil
+			}
+		}
+	}
+	return Default, ErrEventRejected
+}
+
+// Transition triggers current state action and sends event to the state machine.
+func (s *StateMachine) Transition(ctx context.Context) error {
```
mmatczuk, line 78: After reading sendEvent it just makes the machine going.

mmatczuk:
The commit 78a9f6ffc7d47c0307c9a775934509d00539a1b3 adding FSM is missing docs and tests.

Worse this code is coppied from https://venilnoronha.io/a-simple-state-machine-framework-in-go without a notice.

I wonder why would you implement fsm not take one from a library.
After taking a look it seems that the state of FSM in go is substandard.

If you are interested in FSMs read https://hoverbear.org/blog/rust-state-machine-pattern/

I'd suggest to use input token channel like here https://github.com/hashicorp/raft/blob/e3c5b666287bb8dfe4131ae8759eacc75bbb39c0/api.go#L90 it would allow you do decouple read path (the FSM) from write path (Actions).

I'd suggest you implement sth like that https://github.com/eugene-babichenko/rust-fsm/blob/master/src/lib.rs#L123
FSM has current state, state graph, it takes a stream of events and a transition function.
Much simpler and cleaner design.

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -1618,6 +1618,35 @@ spec:
                 - name
                 type: object
               type: array
+            upgrade:
+              description: UpgradeStatus contains state of ongoing upgrade procedure.
```
mmatczuk, line 1622: UpgradeStatus or upgrade?

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -1618,6 +1618,35 @@ spec:
...
+              properties:
+                currentNode:
+                  description: CurrentNode node under upgrade.
+                  type: string
+                currentRack:
+                  description: CurrentRack rack under upgrade.
+                  type: string
+                dataSnapshotTag:
```
mmatczuk, line 1630: SystemSnapshotTag or dataSnapshotTag

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -1618,6 +1618,35 @@ spec:
...
+                  description: SystemSnapshotTag snapshot tag of data keyspaces.
+                  type: string
+                fromVersion:
+                  description: FromVersion reflects from which version ScyllaCluster is being upgraded.
+                  type: string
+                state:
+                  description: State reflect current state machine state.
```
mmatczuk, line 1637: ?

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -1618,6 +1618,35 @@ spec:
...
+                  type: string
+                systemSnapshotTag:
+                  description: SystemSnapshotTag snapshot tag of system keyspaces.
+                  type: string
+                toVersion:
+                  description: ToVersion reflects to which version ScyllaCluster is being upgraded.
```
mmatczuk, line 1643: Simplify pls.

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -1618,6 +1618,35 @@ spec:
...
+                dataSnapshotTag:
+                  description: SystemSnapshotTag snapshot tag of data keyspaces.
+                  type: string
+                fromVersion:
+                  description: FromVersion reflects from which version ScyllaCluster is being upgraded.
+                  type: string
+                state:
+                  description: State reflect current state machine state.
+                  type: string
+                systemSnapshotTag:
```
mmatczuk, line 1639: Why do we need the snapshots?

examples/common/operator.yaml
```diff
@@ -1633,6 +1633,35 @@ spec:
                 - name
                 type: object
               type: array
+            upgrade:
```
mmatczuk, line 1636: Same comments, perhaps we can copy the file not to manually replicate it?

pkg/naming/constants.go
```diff
@@ -22,6 +22,10 @@ const (
 	// ReplaceLabel express the intent to replace pod under the specific member.
 	ReplaceLabel = "scylla/replace"
 
+	// NodeMaintenanceLabel means that node is under maintenance. Readiness check will always fail
```
mmatczuk, line 25: Line too long?

pkg/controllers/cluster/actions/replace.go
```diff
@@ -129,7 +129,7 @@ func podReady(pod *corev1.Pod) bool {
 }
 
 const (
-	retryInterval     = time.Second
+	retryInterval     = 200 * time.Millisecond
```
mmatczuk, line 132: Why?

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+	Cluster        *scyllav1alpha1.ScyllaCluster
+	ScyllaClient   *scyllaclient.Client
+	ClusterSession cqlSession
+
+	ipMapping map[string]string
+
+	currentRack *scyllav1alpha1.RackSpec
+	currentNode *corev1.Pod
+
+	cc         client.Client
```
mmatczuk, line 50: I'd consider moving Client to scylla package

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+
+	oldVersion, err := semver.Parse(fromVersion)
+	if err != nil {
+		a.logger.Info(ctx, "Invalid from semantic version", "version", fromVersion)
+		return majorUpgradeProcedure
+	}
+	newVersion, err := semver.Parse(a.Cluster.Spec.Version)
+	if err != nil {
+		a.logger.Info(ctx, "Invalid to semantic version", "version", fromVersion)
+		return majorUpgradeProcedure
```
mmatczuk, line 127: How this is supposed to work?

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+	return ActionSuccess, nil
+}
+
+func (a *ClusterVersionUpgrade) fsm() *fsm.StateMachine {
+	state := BeginUpgrade
+	if a.Cluster.Status.Upgrade != nil {
+		state = fsm.StateType(a.Cluster.Status.Upgrade.State)
+	}
+
+	return &fsm.StateMachine{
```
mmatczuk, line 861: It looks to me that the FSM is a plain list. Maybe it would be easier to drop the FSM and just proceed with the list of steps.

I cannot see support for ActionFailure, there is not cleanup path?

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+
+	return ActionSuccess, nil
+}
+
+var systemKeyspaces = []string{"system", "system_schema"}
+
+func (a *ClusterVersionUpgrade) backupKeyspaces(ctx context.Context, tag string, keyspaces []string, hosts ...string) error {
+	if hosts == nil {
+		for _, r := range a.Cluster.Spec.Datacenter.Racks {
+			for i := 0; i < int(r.Members); i++ {
```
mmatczuk, line 285: Maybe you would like to add forEachMember helper function?

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+				host := naming.MemberServiceName(r, a.Cluster, i)
+				hosts = append(hosts, a.ipMapping[host])
+			}
+		}
+	}
+
+	err := parallel.Run(len(hosts), parallel.NoLimit, func(i int) error {
```
mmatczuk, line 292: You can just return it.

mmatczuk: Applies to other instances too

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+		return ActionFailure, errors.Wrap(err, "system backup")
+	}
+
+	return ActionSuccess, nil
+}
+
+func versionFromImage(image string) (string, error) {
+	parts := strings.Split(image, ":")
+	if len(parts) != 2 {
+		return "", errors.Errorf("can't extract version from image %q", image)
```
mmatczuk, line 359: Do not write can't failed and so on in errors

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+	}
+
+	if err := a.cc.Delete(ctx, pod); err != nil {
+		return ActionFailure, errors.Wrap(err, "delete pod")
+	}
+
+	return ActionSuccess, nil
+}
+
+// TODO: configurable in cluster spec
```
mmatczuk, line 686: Maybe you want to add it?

pkg/controllers/cluster/actions/upgrade_version.go
```diff
@@ -3,54 +3,966 @@ package actions
...
+				},
+			},
+			FinishUpgrade: {
+				Action: a.finishUpgrade,
+			},
+		},
+	}
+}
+
+func contains(arr []string, v string) bool {
```
mmatczuk, line 961: This little sh** deserves it's own package...

mmatczuk:
The commit a0d4b448de93ad15ca2f796da2608f34bbf8d3b4 is a massive bloat. It contains some surprising knits I find impossible to tell what they do like https://github.com/scylladb/scylla-operator/pull/294/commits/a0d4b448de93ad15ca2f796da2608f34bbf8d3b4#diff-4e02da97ee19f45e95b02561d8b9ae230be86de08e271d10c40adacff7aa0914R97
I'd strongly recommend splitting it to smaller comments / steps at last to remove the pretyfing from the main thing.

mmatczuk:
I'd suggest to drop the FSM idea for now and replace it with a plain list of functions.

pkg/controllers/cluster/actions/upgrade_version.go
zimnx: > I cannot see support for ActionFailure, there is not cleanup path?

`ActionFailure fsm.Event = fsm.NoOp `
failed action will be retried on next reconcilation iteration. All actions except beginUpgrade should be idempotent, so upgrade is always retried, and there's no actual failure path. 


> It looks to me that the FSM is a plain list. Maybe it would be easier to drop the FSM and just proceed with the list of steps.

FSM isn't flat, there two loops:
FindNextNode->EnableMaintenanceMode->[...]->ClearDataBackup->FindNextNode
and 
FindNextRack->[...]->RestoreUpgradeStrategy->FindNextRack

pkg/controllers/cluster/actions/upgrade_version.go
zimnx: Totally agree, I'm surprised there's no such function in `strings` package. I guess they are waiting for generics.

examples/common/operator.yaml
zimnx: this file is an output of:
`kustomize build config/operator/default > examples/common/operator.yaml`

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
zimnx: In order to restore tables in case of failure. Rollback requires restoring system tables to state prior upgrade.

pkg/naming/constants.go
zimnx: 96 chars, not so so

pkg/controllers/cluster/actions/upgrade_version.go
zimnx: this is k8s apiserver client

pkg/controllers/cluster/actions/upgrade_version.go
zimnx: I'll change name to "generic upgrade" procedure. For example upgrading to `nightly` shouldn't be rejected, and it would be rejected because it's not semantic, the same way Scylla versions aren't. Logic whether given version transition should be allowed or not, should be in webhook where change can be rejected with aproperiate message back to the user.

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
zimnx: what do you mean?

PR #356, opened 2021-02-01
integration tests stability fixes

zimnx:
Image parsing tests were relying on global value, and they were failing when any integration tests failed. 
Two replace integration tests were flaky.


tnozicka (approved):
/lgtm

PR #358, opened 2021-02-02
Helm example

zimnx:


tnozicka (approved):
/lgtm
not a helm expert though

PR #360, opened 2021-02-03
config: remove operator memory limit (#357) 

### LINKED ISSUE #357: Operator OOM due to too low resource limits in example
**Describe the bug**

Operator might OOM due to too harsh memory limit in Operator example.
We should reevaluate the limits. 

**Environment:**
 - Platform: GKE
 - Scylla-operator version: e.g.: master

zimnx:
Measurements show Operator can use 50MB of RSS memory
when multpile racks are added to Scylla cluster.
Although it scales with number of objects, so no limit
is better. User may define his own limits if needed.

Selection_225

Fixes #357

### DIFF
--- a/config/operator/operator/operator.yaml
+++ b/config/operator/operator/operator.yaml
@@ -41,9 +41,6 @@ spec:
         imagePullPolicy: IfNotPresent
         name: manager
         resources:
-          limits:
-            cpu: 100m
-            memory: 30Mi
           requests:
             cpu: 100m
             memory: 20Mi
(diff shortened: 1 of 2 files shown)


config/operator/operator/operator.yaml
```diff
@@ -43,7 +43,7 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
+            memory: 100Mi
```
tnozicka, line 46: any reason not to go best effort and use only requests? the memory depends on how many objects live in the cluster which varies by size.

zimnx: Good point

tnozicka (approved):
/lgtm

PR #364, opened 2021-02-04
helm: release process

zimnx:
Current approach to Helm Chart releasing doesn't allow to test remote repository.
Even though our Helm Charts are not released yet, people are using it without
any explicit message that these are latest not stable builds.
I suggest to add non-public development bucket where dev repository is stored.

In addition I propose to have to separate directories in each prod/dev bucket:
- stable - where stable version will be kept, published manually after QA approval
- latest - where unstable latest version will be kept, published via CI

This way users will be able to choose wheter they want to use stable
Helm chart tied with stable Operator branch, or if they want to try bleeding edge
latest builds available for promotion branches.

Stable repository will be available via:
```
helm repo add scylla https://scylla-operator-charts.storage.googleapis.com/stable
```
and latest via:
```
helm repo add scylla-nightly https://scylla-operator-charts.storage.googleapis.com/latest
```

To build and publish latest Helm chart devs with write access can use:
```
make helm-publish
```
or for testing purposes:
```
make dev-helm-publish
```
The only difference between those commands are target buckets.

Version and release channel can be overwrtten using env variables.
```
make helm-publish HELM_CHANNEL=stable HELM_APP_VERSION=v1.0.0 HELM_CHART_VERSION=v1.0.0
```

Action publishes images:
- v1.0.0-39-gffb19b8-nightly - daily on schedule, uses Operator nightly by default
- v1.0.0-39-gffb19b8-latest - on every master push, uses Operator latest by default

Prefix is taken from current branch state.


.github/workflows/go.yaml
```diff
@@ -73,5 +73,28 @@ jobs:
         working-directory: ${{ env.working-directory }}
         run: make docker-push
 
+  publish-helm:
```
tnozicka, line 76: probably best to `if` on job level.

might worth to condition it on branch name to avoid republishing on periodics and in case we create `release-1.x` branches for the future that will also run CI.

helm/scylla-manager/Chart.yaml
```diff
@@ -1,12 +1,12 @@
 apiVersion: v2
-appVersion: v1.0.0
 name: scylla-manager
-version: 1.0.0
 description: Scylla Manager automates database operations.
+version: "0.0.0" # overwritten during publishing
+appVersion: "0.0.0" # overwritten during publishing
```
tnozicka, line 5: nit: does appVersion require any convention for format? `as it was previously `v1.0.0`, maybe `v0.0.0`

scripts/release-helm.sh
```diff
@@ -0,0 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -euo pipefail
+
+CHANNEL="${CHANNEL:-"$(scripts/release-channel.sh)"}"
```
tnozicka, line 21: would it be better to substitute these var on makefile level than the script calling 2 tiny scrips?
I suspect the relative path won't work if this script is invoked in different working directory

scripts/release-helm.sh
```diff
@@ -0,0 +1,72 @@
...
+VERSION="${VERSION:-"$(scripts/version.sh)"}"
+
+if [[ "$CHANNEL" == "nightly" ]]; then
+  LATEST_TAG=$(git describe --tags --always --abbrev=0)
```
tnozicka, line 25: same here, I have a PR that consistently sets these in a makefile and we could reuse that logic if you want https://github.com/scylladb/scylla-operator/pull/365/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R19

helm/scylla-manager/Chart.yaml
zimnx: appVersion doesn't need semantic, `version` does. Both versions are accepted, with `v` and w/o

Makefile
```diff
@@ -2,14 +2,11 @@ all: build
 
 SHELL :=/bin/bash -euEo pipefail
 
-IMAGE_REF		?= docker.io/scylladb/scylla-operator:latest
+IMAGE_LATEST_TAG ?= latest
```
tnozicka, line 5: nit: would it be better to call it just `IMAGE_TAG`? would still keep the meaning if someone overrides the value on invocation

Makefile
```diff
@@ -222,11 +222,28 @@ generate:
 	controller-gen object:headerFile="hack/boilerplate.go.txt" paths="$(GO_PACKAGES)"
 .PHONY: generate
 
+HELM_CHANNEL ?=latest
+HELM_CHARTS=scylla-operator scylla-manager scylla
```
tnozicka, line 226: nit: conventionally global makefile variables are kept together before defining target so people can easily see what is available when opening the makefile

Makefile
```diff
@@ -222,11 +222,28 @@ generate:
...
+
+# Build Helm charts and publish them in Development GCS repo
+dev-helm-publish: export REPOSITORY ?=https://scylla-operator-charts-dev.storage.googleapis.com/$(HELM_CHANNEL)
```
tnozicka, line 229: nit: given every env var is make var, I'd still prefix the names

Makefile
```diff
@@ -222,11 +222,28 @@ generate:
...
+dev-helm-publish: export BUCKET ?=gs://scylla-operator-charts-dev/$(HELM_CHANNEL)
+dev-helm-publish:
+	@$(MAKE) helm-publish
```
tnozicka, line 232: `dev-helm-publish: helm-publish`

Makefile
```diff
@@ -222,11 +222,28 @@ generate:
...
-helm-release:
-	mkdir -p $(HELM_LOCAL_REPO)
-	gsutil rsync -d $(HELM_BUCKET) $(HELM_LOCAL_REPO)
-	helm package helm/scylla-operator helm/scylla helm/scylla-manager -d $(HELM_LOCAL_REPO)
-	helm repo index $(HELM_LOCAL_REPO) --url $(HELM_REPOSITORY) --merge $(HELM_LOCAL_REPO)/index.yaml
-	gsutil rsync -d $(HELM_LOCAL_REPO) $(HELM_BUCKET)
-.PHONY: helm-release
+helm-publish: export LOCAL_REPO ?=helm/repo/$(HELM_CHANNEL)
+helm-publish: export BUCKET ?=gs://scylla-operator-charts/$(HELM_CHANNEL)
+helm-publish: export REPOSITORY ?=https://scylla-operator-charts.storage.googleapis.com/$(HELM_CHANNEL)
```
tnozicka, line 238: I think I'd still make some of these global. late evaluation (`=`) will help you easily override them on the other target while the rest resolves correctly. You can then avoid a global export by doing just the export part in this target

Makefile
```diff
@@ -222,11 +222,28 @@ generate:
...
+helm-publish: export APP_VERSION ?=$(IMAGE_LATEST_TAG)
+helm-publish: export CHART_VERSION ?=$(GIT_TAG)-$(APP_VERSION)
+helm-publish:
+	mkdir -p $(LOCAL_REPO)
+	gsutil rsync -d $(BUCKET) $(LOCAL_REPO)
+
+	@$(foreach chart,$(HELM_CHARTS),helm package helm/$(chart) --destination $(LOCAL_REPO) --app-version $(APP_VERSION) --version $(CHART_VERSION); )
```
tnozicka, line 245: the `;` on expansion will ignore errors I think, macros with last empty line avoid that, or some version with `set -e`

tnozicka: or you can make them individual target but that's harder, although parallel

tnozicka (commented):
```
HELM_CHANNEL=stable APP_VERSION=v1.0.0 CHART_VERSION=v1.0.0 make helm-publish
```
might be better done as `make target VAR=aa`

.github/workflows/go.yaml
```diff
@@ -146,13 +149,21 @@ jobs:
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
-    - uses: actions/checkout@v2
     # TODO: use the same image sha from images step - has to match what we used for e2e
-    - name: Publish latest image
+    - name: Publish Docker image
```
tnozicka, line 153: tiny nit: s/Docker/container/

Makefile
```diff
@@ -86,6 +91,15 @@ define build-package
 
 endef
 
+# $1 - chart name
+# $2 - destination dir
+# $3 - app version
+# $4 - chart version
+define package-helm
+	helm package $(HELM_CHARTS_DIR)/$(1) --destination $(2) --app-version $(3) --version $(4)
```
tnozicka, line 99: nit: might be safer to quote the variable expansions

Makefile
```diff
@@ -149,7 +163,11 @@ update-deps-overrides:
 	cp "$(tmp_dir)"/{updated,current}/deps.diff
 .PHONY: update-deps-overrides
 
-verify: verify-govet verify-gofmt
+verify-helm:
+	@$(foreach chart,$(HELM_CHARTS),helm lint helm/$(chart); )
```
tnozicka, line 167: this will ignore errors
```
helm lint helm/chart_0; helm lint helm/chart_1; helm lint helm/chart_2; 
```

tnozicka: if you are suppressing the execution output, might be nice to at least have an echo saying you are linting

Makefile
```diff
@@ -227,11 +245,19 @@ generate:
 	controller-gen object:headerFile="hack/boilerplate.go.txt" paths="$(GO_PACKAGES)"
 .PHONY: generate
 
+# Build Helm charts and publish them in Development GCS repo
+dev-helm-publish: HELM_REPOSITORY=https://scylla-operator-charts-dev.storage.googleapis.com/$(HELM_CHANNEL)
```
tnozicka, line 249: nit: might be easier for code completion if named `helm-publish-dev`

tnozicka:
few comments, most importantly the foreach expansion
I appreciate setting this up for CI promotion :+1:

Makefile
zimnx: Helm itself is verbose about it

tnozicka (approved):
/lgtm

PR #366, opened 2021-02-05
Add support for discovering rack name directly from nodes

geobeau:

The current system (one statefulset per rack) makes managing tens of racks impractical.
With this change it's possible to make the sidecar get the rack name directly from a label on the node.

None open

**Checklist:**
- [ ] Add testing

### DIFF
--- a/pkg/cmd/options/operator.go
+++ b/pkg/cmd/options/operator.go
@@ -1,7 +1,10 @@
 package options
 
 import (
+	"os"
+
 	"github.com/pkg/errors"
+	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/spf13/cobra"
 )
 
@@ -14,6 +17,7 @@ type OperatorOptions struct {
 	*CommonOptions
 	Image                  string
 	EnableAdmissionWebhook bool
+	RackFromNode           bool
 }
 
 func GetOperatorOptions() *OperatorOptions {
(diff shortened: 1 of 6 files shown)


yanniszark:
> The current system (one statefulset per rack) makes managing tens of racks impractical.

@geobeau using a StatefulSet per rack is an explicit design decision, in order to be able to enforce a desired topology placement (e.g., x members on zone-a, y members on zone-b, etc). Using a single StatefulSet provides a placement guarrantee only for a single topology domain (although I've heard something landed in 1.18). So, for example, you can place a StatefulSet in a region, but the StatefulSets can't guarrantee how its Pods are going to spread out across Nodes. That's why the operator uses multiple StatefulSets.

To reinforce this point, CockroachDB had this problem and had to write their own scheduler extender:
https://kubernetes.io/blog/2020/12/21/writing-crl-scheduler/

geobeau:
Thank you for your answer. I understand better. I'll expose a bit more our use case:

We run our Kubernetes clusters on bare-metal servers. Here the availability zone is the rack, if we reserve 100 nodes for the kubernetes cluster, the servers will be spread from 10 to 40 racks. It means that if a node die, it will be replaced by a server from another rack just because we don't have a node on the same rack for this cluster.

So if we use the current system, we either:
- have something that manage the STS (like another operator)
- work around our rack and create "virtual availability zones"

If we could get the sidecar to detect the rack (like with this PR), it would simplifie the system for us:
- We can specify prefferedPodAntiAffinity to have the STS spread the pods of a cluster as good as it can
- We can just create a big STS and not care about the zones

zimnx:
It's not that easy to change rack in Scylla because Scylla is DC/Rack aware, and data is replicated based on these information depending which replication strategy is selected for keyspace. Here you can find some information: https://www.scylladb.com/tech-talk/how-to-be-successful-with-scylla/

> It means that if a node die, it will be replaced by a server from another rack just because we don't have a node on the same rack for this cluster.

In this scenario and with this patch, node rack assignment would change and Scylla would have to reshard and restream data to keep right replication on each topology change.  
Also it would cause incositencies between ScyllaCluster topology spec and actual toplogy status.

geobeau:
Can those inconsistencies between the operator static topology and the actual scylla topology create issues? 

Today we operate Cassandra this way (almost 1-n per rack, tens racks) so we understand the tradeoff well. Rebalancing the data on failure is okay since it’s a rare occurrence.

geobeau:
Will reopen it later after discussing more on the implementation

PR #371, opened 2021-02-08
helm: allow to customize ServiceAccount in scylla Chart (#369)

### LINKED ISSUE #369: Allow to customize Scylla cluster RBAC in Helm chart
Current Helm chart always create RBAC and doesn't allow for any customization. Users may want to create rbac rules on their own, or at least provide annotation customization.

zimnx:
Fixes #369


tnozicka (commented):
What is the purpose of committing binary file named `scylla-1.0.0.tgz` to git?

zimnx:
> What is the purpose of committing binary file named `scylla-1.0.0.tgz` to git?

Good question, this is what Helm is doing during packaging. Turns out it also support symlinks to dependencies, so I removed binary file and now symlink is added.

vponomaryov:
@zimnx , @tnozicka hi, here is the PR to solve the mentioned problem related to that "tgz" archive: https://github.com/scylladb/scylla-operator/pull/381

tnozicka (approved):

PR #374, opened 2021-02-09
replace: eliminate conflict during cleanup of PVC finalizers (#350) 

### LINKED ISSUE #350: docs: stable redirection broke branch links
Links which were used in Release Notes, and other user facing communication doesn't work after #342 

These two should be redirected to the same content (at this moment), but first one returns 404
https://operator.docs.scylladb.com/v1.0/migration.html
https://operator.docs.scylladb.com/stable/migration.html

We cannot use `stable` links in release notes, because once tag is changed, old release notes will use new documentation page.

zimnx:
Operator updated PVC which is subject for removal, so this is expected
that update may conflict.
To avoid conflicts instead updating whole resource, patch is used.

Fixes #350

### DIFF
--- a/pkg/controllers/cluster/actions/replace.go
+++ b/pkg/controllers/cluster/actions/replace.go
@@ -14,6 +14,7 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


pkg/controllers/cluster/actions/replace.go
```diff
@@ -174,25 +175,28 @@ func (a *RackReplaceNode) replaceNode(ctx context.Context, state *State, member
 	)
 
 	a.Logger.Debug(ctx, "Waiting for PVC deletion", "member", member.Name, "pvc", pvc.Name)
-	if err := wait.PollImmediate(time.Second, time.Minute, func() (done bool, err error) {
-		p := &corev1.PersistentVolumeClaim{}
-		if err := cc.Get(ctx, naming.NamespacedName(naming.PVCNameForPod(member.Name), member.Namespace), p); err != nil {
-			if apierrors.IsNotFound(err) {
-				a.Logger.Debug(ctx, "PVC deleted", "member", member.Name, "pvc", pvc.Name)
-				return true, nil
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
tnozicka, line 178: if you want finalizers gone, would it be better to use a PATCH call?

pkg/controllers/cluster/actions/replace.go
```diff
@@ -174,25 +175,28 @@ func (a *RackReplaceNode) replaceNode(ctx context.Context, state *State, member
...
-		// We want to delete pod anyway, so it's better to delete pvc immediately.
-		pvcCopy := p.DeepCopy()
-		pvcCopy.SetFinalizers([]string{})
-		if err := cc.Update(ctx, pvcCopy); err != nil {
-			return false, errors.Wrap(err, "failed to update pvc")
-		}
+			// Remove finalizer, which will wait until pod is deleted.
+			// We want to delete pod anyway, so it's better to delete pvc immediately.
+			pvcCopy := p.DeepCopy()
+			pvcCopy.SetFinalizers([]string{})
```
tnozicka, line 192: might worth to check is there are any finalizers, otherwise you may not need the api call

tnozicka (approved):
lgtm

PR #378, opened 2021-02-10
Fix CI promotion and few nits

tnozicka:
Fixes https://github.com/scylladb/scylla-operator/runs/1871687731?check_suite_focus=true#step:2:10


zimnx (approved):
LGTM

PR #379, opened 2021-02-11
Skip io-tuning if it has already been done

geobeau:

Right now, every-time the scylla container restart we have to redo the IO tuning.
Since it's quite long it can slow down operations on the cluster (like a simple rolling restart).
This change makes the sidecar skip io tuning if the properties file exists.

Alternative to #179 

**Checklist:**

### DIFF
--- a/pkg/api/v1/cluster_types.go
+++ b/pkg/api/v1/cluster_types.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1
 
 import (
-	"github.com/blang/semver"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -334,14 +333,6 @@ type ScyllaClusterList struct {
 	Items           []ScyllaCluster `json:"items"`
 }
 
-// Version of scylla docker starting from which passing arguments via entry-point is supported
-var ScyllaVersionThatSupportsArgsText = "4.2.0"
-var ScyllaVersionThatSupportsArgs semver.Version
-
 func init() {
-	var err error
-	if ScyllaVersionThatSupportsArgs, err = semver.Parse(ScyllaVersionThatSupportsArgsText); err != nil {
-		panic(err)
-	}
 	SchemeBuilder.Register(&ScyllaCluster{}, &ScyllaClusterList{})
 }
--- a/pkg/api/v1/cluster_validation.go
+++ b/pkg/api/v1/cluster_validation.go
@@ -4,7 +4,6 @@ import (
 	"reflect"
 	"strconv"
 
-	"github.com/blang/semver"
 	"github.com/pkg/errors"
 	"github.com/scylladb/go-set/strset"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -42,9 +41,9 @@ func checkValues(c *ScyllaCluster) error {
 	}
 
 	if len(c.Spec.ScyllaArgs) > 0 {
-		version, err := semver.Parse(c.Spec.Version)
-		if err == nil && version.LT(ScyllaVersionThatSupportsArgs) {
-			return errors.Errorf("ScyllaArgs is only supported starting from %s", ScyllaVersionThatSupportsArgsText)
+		version := NewScyllaVersion(c.Spec.Version)
+		if !version.SupportFeatureUnsafe(ScyllaVersionThatSupportsArgs) {
+			return errors.Errorf("ScyllaArgs is only supported starting from %s", ScyllaVersionThatSupportsArgs)
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


pkg/controllers/sidecar/config/config.go
```diff
@@ -256,6 +258,12 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		}
 	}
 
+	if _, err := os.Stat(scyllaIOPropertiesPath); err == nil {
+		s.logger.Info(ctx, "Scylla IO properties are already set, skipping io tuning (feature only supported for Scylla >= 3.3)")
```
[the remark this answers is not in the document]
geobeau: Done

pkg/api/v1/cluster_types.go
```diff
@@ -334,14 +334,28 @@ type ScyllaClusterList struct {
...
 	var err error
+	if ScyllaDefaultVersion, err = semver.Parse(ScyllaDefaultVersionText); err != nil {
+		panic(err)
+	}
 	if ScyllaVersionThatSupportsArgs, err = semver.Parse(ScyllaVersionThatSupportsArgsText); err != nil {
 		panic(err)
 	}
+	if ScyllaVersionThatSupportsDisablingIOTuning, err = semver.Parse(ScyllaVersionThatSupportsDisablingIOTuningText); err != nil {
+		panic(err)
+	}
```
zimnx, line 359:
```suggestion
var (
    ScyllaDefaultVersion = semver.MustParse("4.2.0")
    ScyllaVersionThatSupportsArgs = semver.MustParse( "4.2.0")
    ScyllaVersionThatSupportsDisablingIOTuning = semver.MustParse("4.3.0")
)

func init() {

```

pkg/controllers/sidecar/config/config.go
```diff
@@ -244,18 +246,28 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
-		} else if version.LT(v1.ScyllaVersionThatSupportsArgs) {
-			s.logger.Info(ctx, "This scylla version does not support ScyllaArgs. ScyllaArgs is ignored", "version", cluster.Spec.Version)
+		if version.LT(v1.ScyllaVersionThatSupportsArgs) {
+			s.logger.Info(ctx, "This scylla version does not support ScyllaArgs. ScyllaArgs is ignored", "version", version)
 		} else {
 			appendScyllaArguments(ctx, s, cluster.Spec.ScyllaArgs, args)
 		}
 	}
 
+	if _, err := os.Stat(scyllaIOPropertiesPath); err == nil && version.GTE(v1.ScyllaVersionThatSupportsDisablingIOTuning) {
```
zimnx, line 265:
```
func scyllaSupportsDisabledIOSetup(v string) bool {
    version, err := semver.Parse(v)
    if err != nil {
        return false
    }
    _, err := os.Stat(scyllaIOPropertiesPath)
    return err == nil && version.GTE(v1.ScyllaVersionThatSupportsDisablingIOTuning) 
}
```
This way we don't need this error prone default version assumption.

zimnx (changes requested):
Thanks, two comments about minor refactor needed.

pkg/controllers/sidecar/config/config.go
geobeau: I added a wrapper on top of the version instead

pkg/api/v1/cluster_validation.go
```diff
@@ -42,9 +41,9 @@ func checkValues(c *ScyllaCluster) error {
 	}
 
 	if len(c.Spec.ScyllaArgs) > 0 {
-		version, err := semver.Parse(c.Spec.Version)
-		if err == nil && version.LT(ScyllaVersionThatSupportsArgs) {
-			return errors.Errorf("ScyllaArgs is only supported starting from %s", ScyllaVersionThatSupportsArgsText)
+		version := NewScyllaVersion(c.Spec.Version)
+		if version.SupportFeatureUnsafe(ScyllaVersionThatSupportsArgs) {
+			return errors.Errorf("ScyllaArgs is only supported starting from %s", ScyllaVersionThatSupportsArgs)
```
zimnx, line 46: This is a regression. 
```suggestion
		version := NewScyllaVersion(c.Spec.Version)
		if !version.SupportFeatureUnsafe(ScyllaVersionThatSupportsArgs) {
			return errors.Errorf("ScyllaArgs is only supported starting from %s", ScyllaVersionThatSupportsArgs)
```

zimnx (changes requested):
One last thing

pkg/api/v1/cluster_validation.go
geobeau: Should be good now

zimnx: It is still the same, perhaps you forgot to amend your commit :)

geobeau: Ah I modified the one in config.go 😅 . Thanks!

zimnx (approved):
LGTM, thanks

PR #381, opened 2021-02-11
helm: remove dependency of 'scylla-manager' chart on 'scylla' one

vponomaryov:
Remove 'helm/scylla-manager/Chart.lock' file which sets explicit
dependency on the another, existing in this repo, chart called 'scylla'
Also, remove redundant tgz-archived copy of that 'another chart'

Reason: it is just unnecessary duplication

Resolves #

**Checklist:**


zimnx:
`scylla` chart is a dependency of `scylla-manager`, SM needs it in order to operate, and chart creates Scylla cluster dedicated to SM. Binary tar file is removed in https://github.com/scylladb/scylla-operator/pull/371

PR #387, opened 2021-02-15
helm: use git tag as chart version 

zimnx:
Special case is when nightly job is running. Because in nighly charts we
want to use nightly Operator image, suffix differentiate it from latest
build.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -151,7 +151,9 @@ jobs:
     - name: Determine promotion tag for scheduled job
       if: ${{ github.event_name == 'schedule' }}
       run: |
-        echo "IMAGE_TAG=nightly" | tee -a ${GITHUB_ENV}
+        IMAGE_TAG=nightly
+        echo "IMAGE_TAG=${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
+        echo "HELM_CHART_VERSION_SUFFIX=-${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
     - name: Login to DockerHub
       uses: docker/login-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yaml
```diff
@@ -174,4 +175,5 @@ jobs:
       env:
         HELM_CHANNEL: latest
         HELM_APP_VERSION: ${{ env.IMAGE_TAG }}
+        HELM_CHART_VERSION_SUFFIX: ${{ env.HELM_CHART_VERSION_SUFFIX }}
```
tnozicka, line 178: you've already made it env on L155.

but maybe this could be
```
HELM_CHART_VERSION_SUFFIX: -${{ env.IMAGE_TAG }}
```

.github/workflows/go.yaml
```diff
@@ -152,6 +152,7 @@ jobs:
       if: ${{ github.event_name == 'schedule' }}
       run: |
         echo "IMAGE_TAG=nightly" | tee -a ${GITHUB_ENV}
+        echo "HELM_CHART_VERSION_SUFFIX=-${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
```
tnozicka, line 155: the env var isn't available here yet (needs next step) but you could make it so:
```
IMAGE_TAG=nightly
echo "IMAGE_TAG=${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
echo "HELM_CHART_VERSION_SUFFIX=-${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
```

tnozicka (approved):
lgtm, thx

PR #391, opened 2021-02-16
actions: fix tag trigger pattern

zimnx:
Pattern doesn't support matching a group of alternatives. To workaround
it, each alternative become a separate pattern.

Additionally, patterns matching dot character doesn't need special
escaping because it's not considered as a special character. It works
fine when dot is escaped, but this behavior it's not documented.


tnozicka (approved):
/lgtm

Issue #362, opened 2021-02-03
E2E tests


zimnx:
E2E tests provides more value than integration test we have. 

These could be run using CI to validate release candidates before passing them to QA.

PR #393, opened 2021-02-16
Add e2e suite

tnozicka:
Adds e2e suite and hooks it into CI/CD.

Resolves https://github.com/scylladb/scylla-operator/issues/362

Prerequisites:
- [x] https://github.com/scylladb/scylla-operator/pull/444

Future work:
- run in parallel


test/e2e/framework/util.go
```diff
@@ -0,0 +1,186 @@
...
+
+	return event.Object.(*corev1.ServiceAccount), nil
+}
+
+func WaitForObjectDeletion(ctx context.Context, dynamicClient dynamic.Interface, resource schema.GroupVersionResource, namespace, name string, uid types.UID) error {
+	fieldSelector := fields.OneTermEqualSelector("metadata.name", name).String()
+	lw := &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			options.FieldSelector = fieldSelector
+			return dynamicClient.Resource(resource).Namespace(namespace).List(context.TODO(), options)
```
zimnx, line 72: why not use context from parameter? I know some deletions must be blocking (like cleaning up after a failure), but such case should be handled on caller side.

test/e2e/framework/util.go
```diff
@@ -0,0 +1,186 @@
...
+				APIGroup:  corev1.GroupName,
+				Kind:      rbacv1.ServiceAccountKind,
+				Namespace: sa.Namespace,
+				Name:      sa.Name,
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     "scylla-manager-cluster-member",
```
zimnx, line 145: scylla-manager?

test/e2e/framework/util.go
```diff
@@ -0,0 +1,186 @@
...
+
+func DumpEventsInNamespace(ctx context.Context, c kubernetes.Interface, namespace string) {
+	events, err := c.CoreV1().Events(namespace).List(ctx, metav1.ListOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	By(fmt.Sprintf("Found %d events.", len(events.Items)))
+	// Sort events by their first timestamp
+	sortedEvents := events.Items
+	if len(sortedEvents) > 1 {
+		sort.Sort(byFirstTimestamp(sortedEvents))
```
zimnx, line 163: nit: you can inline less function, it's short enough.
sort.Slice(sortedEvents, func(){...})

test/e2e/set/scyllacluster/config.go
```diff
@@ -0,0 +1,17 @@
+package scyllacluster
+
```
zimnx, line 2: maybe rename to consts.go

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -0,0 +1,88 @@
...
+		err := framework.SetupScyllaClusterSA(ctx, f.KubeClient().CoreV1(), f.KubeClient().RbacV1(), f.Namespace(), scyllaCluster.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		scyllaCluster, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, scyllaCluster, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to deploy")
+		waitCtx1, waitCtx1Cancel := contextForRollout(ctx, scyllaCluster)
+		defer waitCtx1Cancel()
+		scyllaCluster, err = waitForScyllaClusterState(waitCtx1, f.ScyllaClient().ScyllaV1(), scyllaCluster.Namespace, scyllaCluster.Name, scyllaClusterRolledOut)
```
zimnx, line 42: some lines are too long

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -0,0 +1,88 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f.KubeClient(), scyllaCluster)
```
zimnx, line 45: Function name could be improved, it's not clear what is verified here.

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -0,0 +1,88 @@
...
+
+		framework.By("replacing a non-seed node")
```
zimnx, line 47: nit: r -> R

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -0,0 +1,88 @@
...
+		pod, err := f.KubeClient().CoreV1().Pods(f.Namespace()).Get(
+			ctx,
+			fmt.Sprintf("%s-%s-%s-"+
```
zimnx, line 50: nit: This could be a function nonSeedNode(r, c)

.gitignore
```diff
@@ -1,4 +1,5 @@
 /scylla-operator
+/scylla-operator-tests
```
zimnx, line 2: 20MB Binary is commited as a part of "api: Move v1 api into group folder", and later this ignore is added which doesn't have affect on previous commits.

zimnx (commented):
Comments inline, i still need to run it

Makefile
```diff
@@ -175,10 +179,49 @@ verify-helm:
...
 .PHONY: verify-helm
 
-verify: verify-govet verify-gofmt verify-helm
+define run-codegen
+	GOPATH=$(GOPATH) $(GO) run "$(CODEGEN_PKG)/cmd/$(1)" --go-header-file='$(CODEGEN_HEADER_FILE)' $(2)
+
+endef
+
+define run-deepcopy-gen
+	$(call run-codegen,deepcopy-gen,--input-dirs='github.com/scylladb/scylla-operator/pkg/api/scylla/v1' --output-file-base='zz_generated.deepcopy' --bounding-dirs='github.com/scylladb/scylla-operator/pkg/api/' $(1))
```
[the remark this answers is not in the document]
tnozicka: not without changing the semantics. on the other hand even GH wraps lines as most editors do

Makefile
```diff
@@ -175,10 +179,49 @@ verify-helm:
...
+
+endef
+
+define run-client-gen
+	$(call run-codegen,client-gen,--clientset-name=versioned --input-base="./" --input='github.com/scylladb/scylla-operator/pkg/api/scylla/v1' --output-package='github.com/scylladb/scylla-operator/pkg/client/scylla/clientset' $(1))
```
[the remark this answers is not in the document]
tnozicka: this one follows the convention based on the API group name, maybe we can adjust the other to reflect CQL or DB in it's name in the future

pkg/api/scylla/v1/register.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var (
+	GroupName = "scylla.scylladb.com"
+	// GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+	// Install is a function which adds this version to a scheme
+	Install = schemeBuilder.AddToScheme
+
+	// SchemeGroupVersion generated code relies on this name
```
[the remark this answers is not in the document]
tnozicka: scylla clientset

.github/actions/setup-minikube/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+      sudo chmod +x /usr/local/bin/minikube
+      sudo curl --fail -L "https://dl.k8s.io/release/$( curl --fail -L https://dl.k8s.io/release/stable.txt )/bin/linux/amd64/kubectl" -o /usr/local/bin/kubectl
+      sudo chmod +x /usr/local/bin/kubectl
+  - name: Configure minikube
+    shell: bash
+    run: |
+      set -x
+      sudo minikube config set driver none
+      sudo minikube config set embed-certs true
+      sudo ip link set docker0 promisc on
```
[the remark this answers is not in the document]
tnozicka: not sure about TRUNCATE per say, but it solved connectivity using service network
it installs minikube onto the same machine (no VM) so the shells are the same ones

.github/workflows/go.yaml
```diff
@@ -108,31 +121,120 @@ jobs:
...
     - uses: actions/checkout@v2
-    - uses: docker/build-push-action@v2
       with:
-        context: .
-        file: ./Dockerfile
-        platforms: linux/amd64
-        push: false
+        path: ${{ env.git_repo_path }}
+        fetch-depth: 0  # also fetch tags
+    - name: Install podman
```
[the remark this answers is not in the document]
tnozicka: the docker version present doesn't support experimental `--squash` flag

.github/workflows/go.yaml
```diff
@@ -108,31 +121,120 @@ jobs:
...
+      run: |
+        echo "SCYLLA_OPERATOR_TESTS_FLAKE_ATTEMPTS=5" | tee -a ${GITHUB_ENV}
+    - name: Run e2e
+      env:
+        ARTIFACTS_DIR: ${{ runner.temp }}/e2e-artifacts
+      run: |
+        set -x
+        timeout 45m docker run --rm --entrypoint=/usr/bin/scylla-operator-tests -v "/root/.kube/config:/root/.kube/config" '${{ env.image_repo_ref }}:pr' run
+    - name: Dump cluster state
+      if: ${{ always() }}
```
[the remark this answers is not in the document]
tnozicka: yes, the default condition is "run if the previous one succeeded"

.github/workflows/go.yaml
```diff
@@ -108,31 +121,120 @@ jobs:
...
+    - name: Tolerate flakes on promotion jobs
+      if: ${{ github.event_name != 'pull_request' }}
+      run: |
+        echo "SCYLLA_OPERATOR_TESTS_FLAKE_ATTEMPTS=5" | tee -a ${GITHUB_ENV}
+    - name: Run e2e
+      env:
+        ARTIFACTS_DIR: ${{ runner.temp }}/e2e-artifacts
+      run: |
+        set -x
+        timeout 45m docker run --rm --entrypoint=/usr/bin/scylla-operator-tests -v "/root/.kube/config:/root/.kube/config" '${{ env.image_repo_ref }}:pr' run
```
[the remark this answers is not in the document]
tnozicka: I don't feel strong about the timeout. The current thinking was that you set it just slightly higher then what it needs, so if you introduce some bug having a big slowdown like 2x, you'd find out. In the case of adding a new test you can always raise the timeout in the same PR.
When we enable parallel runs the timeout should stay mostly constant following the longest Describe.

test/e2e/framework/util.go
tnozicka: good catch, copy&paste mistake

test/e2e/framework/util.go
tnozicka: are you sure? this the rolebinding for the pod and it seemed to work so far

test/e2e/framework/util.go
tnozicka: I don't really have a preference there, so why not :)

test/e2e/set/scyllacluster/config.go
tnozicka: I'd prefer not to lock it to consts only, some expression types will need vars

test/e2e/set/scyllacluster/scyllacluster_replace.go
tnozicka: yeah, but it least to me it kind of still looks better then every simple arg on it's line:
```
scyllaCluster, err = waitForScyllaClusterState(
    waitCtx1,
    f.ScyllaClient().ScyllaV1(),
    scyllaCluster.Namespace,
    scyllaCluster.Name,
    scyllaClusterRolledOut,
)
```
90s+ editors are pretty good with wrapping lines
e.g. Kubernetes e2e uses the same https://github.com/kubernetes/kubernetes/blob/303fd4aca2ced102f2698beaadea4f1e1417ec5b/test/e2e/apps/deployment.go#L351

.gitignore
tnozicka: Oh, must have got there during the hundreds of sync commits when I was fighting GH actions to get it working. Thanks for noticing.

test/e2e/framework/util.go
zimnx: It works by coincidence. CI script installs manager, and manager manifest deploys small scylla cluster with SA, role and role binding. This **scylla-manager-cluster**-member role is binded with SA used by the cluster under test. IMO we shouldn't depend here on role availability in CI, but instead create if from yaml in the test framework logic.

tnozicka: Probably, the naming is a bit confusing in the current layout
manager-role?
https://github.com/scylladb/scylla-operator/blob/c5cfd75e383303909116aa9e368c8757dafb5232/config/operator/rbac/role.yaml#L7

tnozicka: (also that clusterrole should have been prefixed with scylla or scylla-operator given it's in the global namespace)

tnozicka: can you get me a link to which one you mean, the closest I see is `simple-cluster-member` (also good naming for non-namesapced obj) https://github.com/scylladb/scylla-operator/blob/a2b891ba3dff8c95a7bd8d627c665f04c02d4650/examples/generic/cluster.yaml#L13

zimnx: Naming isn't best I agree. Exactly this one `simple-cluister-member`. This role provides permission for sidecar to list services etc. 
Role name doesn't matter, but it must be binded with "<cluster.Spec.Name>-member"  SA.

tnozicka: the issue is also coupling a deploy manifest to an example scylla - I suppose I'll have to parse it out with yq until we get manifests sorted out

zimnx: Or we can merge #451 first :D

tnozicka: on the other hand you'll know you have done #451 correctly by being able to remove all the hacks from this PR. and also know your changes work because you'll have e2e coverage

.github/workflows/go.yaml
```diff
@@ -108,31 +121,120 @@ jobs:
...
+        set -x
+         unlz4 ~/operatorimage.tar.lz4 - | docker load
+         # docker looses the registry part on save/load
+         docker tag "$( echo "${image_repo_ref}:pr" | sed -E -e 's~[^/]+/(.*)~\1~' )" "${image_repo_ref}:pr"
+         docker images '${{ env.image_repo_ref }}:pr'
+    - name: Setup go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.go_version }}
+    - uses: ./go/src/github.com/scylladb/scylla-operator/.github/actions/setup-minikube
```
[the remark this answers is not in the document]
tnozicka: now that I've tried again (and it failed), I recalled that GH actions suck regarding var substitution and this is one of those contexts where it isn't supported :(

```
The workflow is not valid. .github/workflows/go.yaml (Line: 174, Col: 13): Unrecognized named-value: 'env'. Located at position 1 within expression: env.git_repo_path
```
https://github.com/scylladb/scylla-operator/actions/runs/636537964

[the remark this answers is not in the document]
zimnx: I meant to give step a string name, now CI displays:
Selection_231
where it could show "Setup minikube" or something like that

zimnx (approved):
LGTM, awesome job!

Issue #375, opened 2021-02-10
Document image publishing


tnozicka:
When we are done setting up the CI we should document which image tags are automatically published and from which branch.

EDIT: also add status badge to the README.

zimnx:
+ same thing about Helm charts

Issue #187, opened 2020-09-22
Support for Operator upgrade


zimnx:
There should be safe a procedure for Operator upgrade. 
Currently it's adviced to delete everything and start from scratch. It shouldn't be that aggressive.

yanniszark:
@zimnx in addition, operator upgrade usually means sidecar update as well.
The sidecar is part of the operator.

PR #409, opened 2021-02-17
v1.1 documentation

zimnx:


tzach:
LGTM
We also change the "latest" tag, but it can be done with a follow up commit

zimnx:
> LGTM
> We also change the "latest" tag, but it can be done with a follow up commit

We do have a follow up for documentaion about image names: https://github.com/scylladb/scylla-operator/issues/375

docs/source/upgrade.md
```diff
@@ -2,6 +2,83 @@
...
+
+1. Get name of StatefulSet managing Scylla Operator
+   ```shell
+   kubectl --namespace scylla-operator-system get sts --selector="control-plane=controller-manager"
+
+   NAME                                 READY   AGE
+   scylla-operator-controller-manager   1/1     95m
+   ```
+
+1. Change readyness probe and used container image by applying following patch:
```
tnozicka, line 18: we may not be there yet, but I'd have just expected a line saying to update the operator deployment like `kubectl apply -f ./deploy/scylla-operator/statefulset.yaml` and be done

docs/source/upgrade.md
```diff
@@ -2,6 +2,83 @@
...
+
+1. Collect all StatefulSets managing Scylla Pods
+    ```shell
+    $ kubectl --namespace scylla get sts --selector="app=scylla,scylla/cluster=simple-cluster"
+
+    NAME                                  READY   AGE
+    simple-cluster-us-east-1-us-east-1a   3/3     107m
+    ```
+    Depending how much racks you have, you may have multiple StatefulSets.
+1. Update each Statefulset by applying following patch:
```
tnozicka, line 61: These statefulsets are managed by the operator. Why would we ask users to do the work manually instead of our operator upgrading the operands automatically?

tnozicka (commented):
I guess my comments are more meant for the future as v1.1 probably won't change

docs/source/upgrade.md
zimnx: Yeah, it's a consequence of having single manifest with everything.

docs/source/upgrade.md
zimnx: Current operator is not able to detect that sidecar image is different #187. Maybe we should include it 1.1. WDYT @tzach ?

tzach: Make sense
Agree with @tnozicka we should try to use standard K8s procedures. 
How much of an effort is #187?

zimnx: Should be straightforward, the only tricky part is upgrading racks sequentially, but it's not the first time we are doing it. I expect to have PR by the end of this week.

docs/source/upgrade.md
```diff
@@ -2,6 +2,48 @@
...
+   kubectl --namespace scylla-operator-system get sts --selector="control-plane=controller-manager"
+
+   NAME                                 READY   AGE
+   scylla-operator-controller-manager   1/1     95m
+   ```
+
+1. Change probes and used container image by applying following patch:
+    ```yaml
+    apiVersion: apps/v1
+    kind: StatefulSet
```
tnozicka, line 21: patch isn't GVK aware, I'd assume you only need the spec part

zimnx: right, removed.

tnozicka (approved):
/lgtm

PR #430, opened 2021-02-23
Lower cp sidecar resources

### LINKED ISSUE #427: operator creates an initContainer for cp with 1 full core
**Describe the bug**
The init container for cp has unexpectedly high resource requests. Although it runs just a while and then exits, kubelet has to reserve those resources for restarts, so it takes whole core for requests.

https://github.com/scylladb/scylla-operator/blob/6bdcfeee47124eebd8be1a39e1fc8de1115bec86/pkg/controllers/cluster/resource/resource.go#L201-L210

**Expected behavior**
Tiny resource usage, cpu req like "10m", no limits.

tnozicka:
1 full core for cp is too much :) We have only 2 cores available for the e2e suite.

Resolves #427


zimnx (approved):
LGTM, please add `Fixes #xxx` to commit message too before merging

zimnx (approved):

PR #441, opened 2021-02-26
helm: fix ServiceAccount Name for Scylla Chart

niels-s:

Previously the `serviceAccount.name` wasn't used. This change aligns the
behavior with the scylla-operator and scylla-manager Helm Charts. So that the value when supplied is used to name the ServiceAccount.

**Checklist:**

### DIFF
--- a/helm/scylla/templates/_helpers.tpl
+++ b/helm/scylla/templates/_helpers.tpl
@@ -54,5 +54,9 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Create the name of the service account to use
 */}}
 {{- define "scylla.serviceAccountName" -}}
-{{- printf "%s-%s" (include "scylla.fullname" .) "member" }}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "scylla.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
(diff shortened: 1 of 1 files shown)


zimnx:
Service Account name used by the Scylla cluster is generated based on cluster name:
https://github.com/scylladb/scylla-operator/blob/8307864754ef0ee0695b1049c6bfaa0541d3ca8e/pkg/naming/names.go#L46-L48

So current value is correct. Helm shouldn't allow to provide different name until SA name becomes customizable in Operator.

niels-s:
Ah gotcha, but then we should remove the`serviceAccount.name` value from the values.yml file?

niels-s:
I'll close this PR since you indicate it doesn't make sense right now. But I assume I can open another PR to remove `serviceAccount.name` from the [values.yml file](https://github.com/scylladb/scylla-operator/blob/master/helm/scylla/values.yaml#L24) since it's confusing.

zimnx:
> I'll close this PR since you indicate it doesn't make sense right now. But I assume I can open another PR to remove `serviceAccount.name` from the [values.yml file](https://github.com/scylladb/scylla-operator/blob/master/helm/scylla/values.yaml#L24) since it's confusing.

That would be great :)

PR #442, opened 2021-02-26
helm: impl option to add additional rules to Scylla ClusterRole

### LINKED ISSUE #440: Implement an optional value for the Scylla Helm Chart to add additional rules to ClusterRoleBinding
* Feature Request

Implement an optional value as part of the Scylla Helm Chart so we can extend the rules defined for the ClusterRoleBinding generated by the Scylla Helm Chart.

In our use case, we are running our K8s Cluster using [Kinvolk's Lokomotive](https://github.com/kinvolk/lokomotive) which by default restricts the permissions a ServiceAccount receives when deploying on to the cluster. So we need to add additional rules to the ClusterRoleBinding, otherwise, the Statefulsets generated by the Operator don't have enough permissions to start any pods. We will receive the following error:

```
Warning  FailedCreate      6s (x13 over 27s)  statefulset-controller  create Pod scylla-ewr-1-ewr-1-0 in StatefulSet scylla-ewr-1-ewr-1 failed error: pods "scylla-ewr-1-ewr-1-0" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.initContainers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].securityContext.capabilities.add: Invalid value: "SYS_NICE": capability may not be added]
```

To resolve this we need to add a rule to give more permission to the service account so it's allowed to add the capabilities. In our case that's:

```
  - apiGroups: [ 'policy' ]
    resources: [ 'podsecuritypolicies' ]
    verbs: [ 'use' ]
    resourceNames:
      - zz-privileged
```

Curious to here your thoughts on this use case

niels-s:

Implements the option to add additional rules to the ClusterRole generated by the Scylla Helm Chart. 

This is a draft PR, to get a discussion going if this is something of interest to the Scylla team to support, in the meanwhile I can use this as a draft to keep moving forward until this is resolved.

Resolves #440 

**Checklist:**

### DIFF
--- a/helm/scylla/templates/rbac.yaml
+++ b/helm/scylla/templates/rbac.yaml
@@ -36,6 +36,9 @@ rules:
       - scyllaclusters
     verbs:
       - get
+  {{- if .Values.rbac.clusterRole.rules }}
+    {{- toYaml .Values.rbac.clusterRole.rules | nindent 2 }}
+  {{- end }}
 
 ---
 {{- if .Values.serviceAccount.create -}}
--- a/helm/scylla/values.yaml
+++ b/helm/scylla/values.yaml
@@ -23,6 +23,11 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+rbac:
+  clusterRole:
+    # Define additional rules to add to the ClusterRole
+    rules: []
+
 alternator:
   # Allows to enable Alternator (DynamoDB compatible API) frontend
   enabled: false
--- a/helm/scylla/values.schema.json
+++ b/helm/scylla/values.schema.json
@@ -13,6 +13,22 @@
                 }
             }
         },
+        "rbac": {
+            "type": "object",
+            "properties": {
+                "clusterRole": {
+                    "type": "object",
+                    "properties": {
+                        "rules": {
+                            "type": "array",
+                            "items": {
+                                "type": "object"
+                            }
+                        }
+                    }
+                }
+            }
+        },
         "alternator": {
             "type": "object",
             "properties": {


helm/scylla/values.yaml
```diff
@@ -23,6 +23,11 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+rbac:
+  clusterRole:
+    # Define additional rules to add to the ClusterRole
+    rules: []
```
niels-s, line 29: I'm open to discussing alternative naming schemes, I think this would offer the best extensibility for the future.

niels-s:
@zimnx I guess the integration test failing is a flaky test? Don't think this is related to my changes. I can't rerun the test myself, perhaps you could give it a kick?

helm/scylla/templates/rbac.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 # Role for scylla members.
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
```
niels-s, line 4: @zimnx I also ran into an issue with the `ClusterRole`. I want to deploy multiple Scylla Clusters on the same k8s cluster. However, using the `ClusterRole` which receives the same name prevents me from doing so.
I believe a normal Role and RoleBinding are more appropriate for the Scylla deployment, any thoughts?

niels-s: nvm, I've used `fullnameOverride` to overcome that problem, also the change didn't seem to work. So reverting.

tnozicka (commented):
I wonder if we need to integrate this or just use cluster role aggregation https://github.com/scylladb/scylla-operator/issues/440#issuecomment-788690347

niels-s:
Closing this in favor of https://github.com/scylladb/scylla-operator/pull/467

PR #444, opened 2021-02-26
Convert unit tests to golang native framework

tnozicka:
a) we should use consistent framework
b) these were failing with -count=3 after bumping deps
```
 You may only call BeforeEach from within a Describe, Context or When
```
https://github.com/scylladb/scylla-operator/pull/393/checks?check_run_id=1986747025#step:6:68


zimnx (approved):
lgtm

PR #451, opened 2021-03-01
Installation changes

### LINKED ISSUE #408: Operator shouldn't run as StatefulSet
**Describe the bug**
Running as StatefulSet reduces availability because it uses pods with identity. If a node crashes or network splits, the operator pods gets stuck in `phase: Pending` until manual intervention or node lifecycle controller deleting it after about 5 minutes.

**Expected behavior**
Deployment with recreate strategy would be a better choice with similar behaviour, but ideally we would run 2 replicas with leader election.

zimnx:

- Scylla Operator and Manager Controller are deployed as Deployments instead of as StatefulSet. 
- Scylla Operator and Manager Controller are deployed in 2 replicas and they use leader election.
- New deploy directory containing static manifest files for each resource needed for installation.
- Naming of resources in manifests/charts was improved.
- Resource limit of ScyllaOperator was removed, resource request was reduced.

Resolves #408 

**Checklist:**


config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -122,6 +127,13 @@ spec:
                     description: RackSpec is the desired state for a Scylla Rack.
                     properties:
                       agentResources:
+                        default:
+                          limits:
```
tnozicka: I'd not set limits by default anymore

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -33,10 +33,11 @@ spec:
           description: ClusterSpec defines the desired state of Cluster
           properties:
             agentRepository:
-              description: Repository to pull the agent image from. Defaults to "scylladb/scylla-manager-agent".
+              default: scylladb/scylla-manager-agent
```
tnozicka: we default the other image to docker.io/... we should get consistent one way or the other

tnozicka (commented):
Would you mind splitting these 12 commits into 2 PRs? One to do in place defaulting and the other for moving to deploy folder and dropping customize as a result?

tnozicka:
I support the direction this is heading.

zimnx:
> Would you mind splitting these 12 commits into 2 PRs? One to do in place defaulting and the other for moving to deploy folder and dropping customize as a result?

Done, extracted webhook changes to separate PR #466

Makefile
```diff
@@ -232,15 +228,10 @@ deploy: manifests cert-manager
...
 # Generate manifests e.g. CRD, RBAC etc.
 manifests:
-	cd config/operator/operator && kustomize edit set image controller=$(IMAGE_REF)
-	cd config/manager/manager && kustomize edit set image operator=$(IMAGE_REF)
+	controller-gen crd paths="$(GO_PACKAGES)" output:crd:artifacts:config=deploy/operator/
 
-	controller-gen $(CRD_OPTIONS) paths="$(GO_PACKAGES)" output:crd:dir=config/operator/crd/bases \
-	rbac:roleName=manager-role output:rbac:artifacts:config=config/operator/rbac \
-	webhook output:webhook:artifacts:config=config/operator/webhook
+	kubectl kustomize deploy/operator > examples/common/operator.yaml
```
tnozicka, line 233: I was hoping we'd make a split so the deploy folder is about deploying the operator and manager and then examples would be related to different ScyllaCluster objects

deploy/operator/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
tnozicka, line 1: I was hoping we could get rid of kustomize entirely. I am fine creating "all-in-one" yaml for deploying it all using http (otherwise `kubectl apply -f deploy/operator/` just works) but foreach and echo can do the same job without a kustomize config

deploy/operator/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
```
tnozicka: we should use full dns prefix for every label, but I am not sure we even use that

deploy/operator/namespace.yaml
```diff
@@ -0,0 +1,6 @@
...
+  name: scylla-operator-system
```
tnozicka: `scylla-operator`?

deploy/operator/role.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
```
tnozicka: redundant

deploy/operator/role.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: rbac.authorization.k8s.io/v1
...
-  name: manager-role
+  name: scylla-operator-manager-role
```
tnozicka: `scylla-operator`?

deploy/operator/role_binding.yaml
```diff
@@ -1,12 +1,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: manager-rolebinding
+  name: scylla-operator-manager-rolebinding
```
tnozicka: it's a rolebinding by type
`scylla-operator`?

deploy/operator/service.yaml
```diff
@@ -2,8 +2,8 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: webhook-service
-  namespace: system
+  name: scylla-operator-webhook-service
```
tnozicka: already a service by type
`scylla-operator-webhook`?

deploy/operator/statefulset.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
tnozicka: I assume this is changed in of of the next commits so it's kind of hard to review multiple times. I see why you did that as you are keeping the changes separate but as we are essentially redesigning how it works, would you mind just reducing those 10 commits to say about 3 (moving the manifests, generated changes+deletions of the old generated files, deps)?

tnozicka:
hm, I wonder why GH isn't running CI checks on this PR

deploy/manager/deployments.yaml
```diff
@@ -1,88 +1,61 @@
...
+          imagePullPolicy: IfNotPresent
+          name: scylla-manager
           resources:
-            limits:
-              cpu: 500m
-              memory: 500Mi
             requests:
-              cpu: 500m
-              memory: 500Mi
+              cpu: 100m
```
tnozicka, line 33: nit: i'd just start with `10m`

deploy/manager/deployments.yaml
```diff
@@ -1,88 +1,61 @@
...
           command:
             - /usr/bin/scylla-manager
-          args:
-            - --config-file=/mnt/etc/scylla-manager/scylla-manager.yaml
           env:
             - name: POD_NAMESPACE
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+          image: scylladb/scylla-manager:2.2.0
```
tnozicka, line 28: nit:full ref

deploy/manager/deployments.yaml
```diff
@@ -1,88 +1,61 @@
...
+          imagePullPolicy: IfNotPresent
```
tnozicka, line 29: nit: I'd drop the pull policy, it's usually enforced or defaulted at cluster level

deploy/manager/deployments.yaml
```diff
@@ -1,88 +1,61 @@
----
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: system
----
 apiVersion: apps/v1
```
tnozicka, line 1: this should go into the generated commit

deploy/manager/deployments.yaml
```diff
@@ -97,18 +70,11 @@ spec:
...
-          image: operator:latest
+          image: docker.io/scylladb/scylla-operator:latest
           imagePullPolicy: IfNotPresent
           name: scylla-manager-controller
           resources:
-            limits:
-              cpu: 100m
-              memory: 30Mi
             requests:
               cpu: 100m
```
tnozicka, line 78: nit: lets' get to 10m

deploy/operator/deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+kind: Deployment
+metadata:
+  labels:
+    app: scylla-operator
+  name: scylla-operator
+  namespace: scylla-operator
+spec:
+  replicas: 2
+  strategy:
+    type: Recreate
```
tnozicka, line 11: rolling update

deploy/operator/tls.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: scylla-operator
+spec:
+  dnsNames:
+    - scylla-operator.scylla-operator.svc
+    - scylla-operator.scylla-operator.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: scylla-operator-selfsigned-issuer
+  secretName: webhook-server-cert
+---
```
tnozicka, line 14: nit: one object per file

pkg/api/v1/cluster_types.go
```diff
@@ -236,7 +236,7 @@ type RackSpec struct {
 	// Resources the Scylla container will use.
 	Resources corev1.ResourceRequirements `json:"resources"`
 	// AgentResources which Agent container will use.
-	// +kubebuilder:default:={requests: {cpu: "50m", memory: "10M"}, limits: {cpu: "1", memory: "200M"}}
+	// +kubebuilder:default:={requests: {cpu: "50m", memory: "10M"}}
```
tnozicka, line 239: I am all in for requests only, but doesn't our current validation prevent that?

Makefile
```diff
@@ -206,51 +224,38 @@ help:
...
+generate-crd:
+	$(CONTROLLER_GEN) crd paths="$(GO_PACKAGES)" output:crd:stdout | tee $(CRD_LOCATIONS) 1>/dev/null
+.PHONY: generate-crd
 
-# Generate manifests e.g. CRD, RBAC etc.
-manifests:
-	cd config/operator/operator && kustomize edit set image controller=$(IMAGE_REF)
-	cd config/manager/manager && kustomize edit set image operator=$(IMAGE_REF)
+generate-example-operator:
+	$(call concat-manifest,$(OPERATOR_DEPLOY_DIR),$(OPERATOR_EXAMPLE))
```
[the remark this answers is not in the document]
tnozicka: I suppose there are docs and scripts coupled to the old way, so we can still generate the old deployment files also into examples and clean that up next time.

tnozicka (commented):
Thanks for the split! Good direction, few comments.

install-dependencies.sh
```diff
@@ -61,7 +61,7 @@ then
     sudo apt-get install -y "${debian_packages[@]}"
 fi
 
-go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1 sigs.k8s.io/kustomize/kustomize/v3@v3.5.1
+go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1
```
tnozicka, line 64: I'd split these away from the vendor+mod changes

pkg/api/v1/cluster_types.go
zimnx: I double checked and AgentResources aren't validated

Makefile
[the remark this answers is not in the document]
zimnx: Exactly, docs are referencing files in existing examples, that's why i want to still keep them here. Later once we will start working on documentation, we can rephrase them to different location.

pkg/api/v1/cluster_types.go
tnozicka: oh, agent

Makefile
```diff
@@ -206,51 +224,38 @@ help:
...
 
-# Generate code
-generate:
-	controller-gen object:headerFile="hack/boilerplate.go.txt" paths="$(GO_PACKAGES)"
+generate-deepcopy:
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="$(GO_PACKAGES)"
+.PHONY: generate-deepcopy
+
+# Generate code and examples
+generate: generate-crd generate-deepcopy generate-examples
```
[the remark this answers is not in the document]
zimnx: I'll change prefix to `update` and leave adding of `verify` as a followup.

deploy/manager/20_manager_deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+          requests:
+            cpu: 10m
+            memory: 20Mi
+        volumeMounts:
+        - mountPath: /mnt/etc/scylla-manager
+          name: scylla-manager-config-volume
+      volumes:
+      - configMap:
+          name: scylla-manager-config
+        name: scylla-manager-config-volume
```
tnozicka, line 41: name: it's already a volume by kind

deploy/operator/10_role_binding.yaml
```diff
@@ -1,12 +1,12 @@
...
-  name: manager-rolebinding
+  name: scylla-operator
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: manager-role
+  name: scylla-operator
 subjects:
 - kind: ServiceAccount
   name: default
```
tnozicka, line 11: also a dedicated SA

hack/ci-deploy.sh
```diff
@@ -43,23 +43,20 @@ kubectl apply -f"${deploy_dir}"/scylla-operator-member.clusterrole.yaml
```
tnozicka, line 43: I can't comment on the previous lines but L28 should be gone and the cp above should use the new manifests

hack/ci-deploy.sh
```diff
@@ -43,23 +43,20 @@ kubectl apply -f"${deploy_dir}"/scylla-operator-member.clusterrole.yaml
...
 kubectl wait --for condition=established crd/scyllaclusters.scylla.scylladb.com
-kubectl -n scylla-operator-system rollout status --timeout=5m statefulset.apps/scylla-operator-controller-manager
+kubectl -n scylla-operator rollout status --timeout=5m deployment.apps/scylla-operator
 
 # ScyllaCluster doesn't support changes and is immutable so we always have to delete/create it.
 kubectl delete --ignore-not-found=true --grace-period=0 --cascade=foreground -f"${deploy_dir}"/manager.yaml
 kubectl create -f"${deploy_dir}"/manager.yaml
-object=$( kubectl -n scylla-manager-system get scyllacluster/scylla-manager-cluster -o json | jq '.metadata.uid = "" | .metadata.resourceVersion = "" | .metadata.generation = 0 | .spec.datacenter.racks[0].resources = {"requests":{"cpu":"10m","memory":"100Mi"},"limits":{"cpu":"200m","memory":"500Mi"}}')
+object=$( kubectl -n scylla-manager get scyllacluster/scylla-manager-cluster -o json | jq '.metadata.uid = "" | .metadata.resourceVersion = "" | .metadata.generation = 0 | .spec.datacenter.racks[0].resources = {"requests":{"cpu":"10m","memory":"100Mi"},"limits":{"cpu":"200m","memory":"500Mi"}}')
```
tnozicka, line 51: ideally this would be gone and have the correct resources initially

hack/ci-deploy.sh
```diff
@@ -43,23 +43,20 @@ kubectl apply -f"${deploy_dir}"/scylla-operator-member.clusterrole.yaml
...
+wait-for-object-creation scylla-manager statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
 
 # Wait for the rest to be ready and running
-kubectl -n scylla-manager-system rollout status --timeout=5m statefulset.apps/scylla-manager-controller
-kubectl -n scylla-manager-system rollout status --timeout=5m statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
+kubectl -n scylla-manager rollout status --timeout=5m deployment.apps/scylla-manager-controller
+kubectl -n scylla-manager rollout status --timeout=5m statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
 
 # Allow users to use scyllacluster CRs
 kubectl apply --server-side -f - <<EOF
```
tnozicka, line 62: we should have manifests also for these

tnozicka (commented):
Thanks, looks good overall!

Would you mind adding a README into the deploy folder with `kubectl apply` instruction that users can just copy&paste?

I know you've just moved a lot of these files, and didn't choose the naming so the comments are meant in general.

Operator and manager need a PDB.

PR #455, opened 2021-03-02
cluster: clear rack status when rack STS doesn't exist (#418)

### LINKED ISSUE #418: Deleting STS does not initiate cluster creation
After deleting sts cluster is not recreated .
Behavior was noticed in Operator v1.0.0 while in v.0.3.0 it was working just fine .

Step to reproduce:
Delete existing sts

➜ k delete sts scylla-cluster-us-central1-node -n scylla
statefulset.apps "scylla-cluster-us-central1-node" deleted
➜   k get sts -n scylla
No resources found in scylla namespace.
➜   k get sts -n scylla
No resources found in scylla namespace.
➜ k get sts -n scylla
No resources found in scylla namespace.

zimnx:
Rack StatefulSet is created when rack is defined in Spec, and it's
Status is not available. Because Operator was not clearing Rack Status when
STS was missing, stale record blocked creating new one.

Fixes #418

### DIFF
--- a/pkg/controllers/cluster/status.go
+++ b/pkg/controllers/cluster/status.go
@@ -57,8 +57,9 @@ func (cc *ClusterReconciler) updateStatus(ctx context.Context, cluster *scyllav1
 		}
 		// Get corresponding StatefulSet from lister
 		err := cc.Get(ctx, naming.NamespacedName(naming.StatefulSetNameForRack(rack, cluster), cluster.Namespace), sts)
-		// If it wasn't found, continue
+		// If it wasn't found, clear rack status and continue
 		if apierrors.IsNotFound(err) {
+			delete(cluster.Status.Racks, rack.Name)
 			continue
 		}
 		// If we got a different error, requeue and log it
--- a/pkg/controllers/cluster/status_test.go
+++ b/pkg/controllers/cluster/status_test.go
@@ -1,0 +1,34 @@
+// Copyright (C) 2021 ScyllaDB
+
+package cluster
+
+import (
+	"context"
+	"testing"
+
+	"github.com/scylladb/scylla-operator/pkg/test/unit"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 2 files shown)


pkg/controllers/cluster/status.go
```diff
@@ -57,8 +57,9 @@ func (cc *ClusterReconciler) updateStatus(ctx context.Context, cluster *scyllav1
 		}
 		// Get corresponding StatefulSet from lister
 		err := cc.Get(ctx, naming.NamespacedName(naming.StatefulSetNameForRack(rack, cluster), cluster.Namespace), sts)
-		// If it wasn't found, continue
+		// If it wasn't found, clear rack status and continue
 		if apierrors.IsNotFound(err) {
+			delete(cluster.Status.Racks, rack.Name)
```
tnozicka, line 62: This kind of shows how our current reconciliation is edge triggered instead of level triggered.

e.g. if there would be a new statefulset created in the meantime it could miss the event but I don't thing we can do much better before we go to level based reconciliation with hashes

pkg/controllers/cluster/status_test.go
```diff
@@ -0,0 +1,34 @@
...
+	scheme := runtime.NewScheme()
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+	cc := &ClusterReconciler{
+		Client: fake.NewFakeClientWithScheme(scheme),
+	}
+
+	if err := cc.updateStatus(context.Background(), cluster); err != nil {
+		t.Errorf("expected non nil err, got %s", err)
```
tnozicka, line 28:
```suggestion
		t.Errorf("expected nil, got %s", err)
```

tnozicka (approved):

pkg/controllers/cluster/status.go
zimnx: Agree, current approach is very error prone.

tnozicka (approved):
/lgtm

PR #456, opened 2021-03-02
[1.1] Backports 

zimnx:
Backport of:
#409 
#455


tnozicka:
I guess this is fine with docs but backports are usually done as 1 PR for backport, so you can revert them separately if something were to go wrong. (PRs are safer/simpler to revert as individual commits didn't went through CI.)

PR #462, opened 2021-03-04
cluster: fix source of rack version and members in cluster status

zimnx:


pkg/controllers/cluster/status.go
```diff
@@ -82,7 +82,7 @@ func (cc *ClusterReconciler) updateStatus(ctx context.Context, cluster *scyllav1
 			if err != nil {
 				return errors.WithStack(err)
 			}
-			rackStatus.Version, err = naming.ScyllaVersion(sts.Spec.Template.Spec.Containers)
+			rackStatus.Version, err = naming.ScyllaVersion(firstRackMember.Spec.Containers)
```
tnozicka, line 85: should it be the last one? so the update is complete at that point, not only starting?

tnozicka: disregard, statefulset reverse order

tnozicka (approved):
/lgtm

PR #465, opened 2021-03-05
cluster: fix source of rack version in cluster status

zimnx:
Take version from the first pod in a rack, which is upgraded as a last
one.


tnozicka (approved):
/lgtm

PR #466, opened 2021-03-05
Webhooks cleanup

zimnx:
* Mutating webhook was removed in favor of default values in CRD OpenAPI schema.
* Convertion webhook was removed from the CRD - it wasn't used because we support a single CRD version.
* ScyllaCluster CRD apiVersion was bumped to v1.
* Validating webhook apiVersion was bumped to v1.
* controller-gen used for CRD generation was bumped to 0.4.1.
* controller-runtime dependency bumped to latest 0.8.3


pkg/api/v1/cluster_types.go
```diff
@@ -32,14 +32,18 @@ type ClusterSpec struct {
...
+	//+optional
+	Repository string `json:"repository,omitempty"`
 	// Alternator designates this cluster an Alternator cluster
 	Alternator *AlternatorSpec `json:"alternator,omitempty"`
-	// Version of Scylla Manager Agent to use. Defaults to "latest".
-	AgentVersion *string `json:"agentVersion"`
-	// Repository to pull the agent image from. Defaults to "scylladb/scylla-manager-agent".
-	AgentRepository *string `json:"agentRepository,omitempty"`
+	// Version of Scylla Manager Agent to use.
+	AgentVersion string `json:"agentVersion"`
```
tnozicka, line 42: should we default to latest?

pkg/api/v1/cluster_types.go
```diff
@@ -77,10 +81,12 @@ const (
 type GenericUpgradeSpec struct {
 	// FailureStrategy specifies which logic is executed when upgrade failure happens.
 	// Currently only Retry is supported.
+	// +kubebuilder:default:="Retry"
```
tnozicka, line 84: +optional

pkg/api/v1/cluster_types.go
```diff
@@ -77,10 +81,12 @@ const (
...
 	FailureStrategy GenericUpgradeFailureStrategy `json:"failureStrategy,omitempty"`
 	// PollInterval specifies how often upgrade logic polls on state updates.
 	// Increasing this value should lower number of requests sent to apiserver, but it may affect
 	// overall time spent during upgrade.
+	// +kubebuilder:default:="1s"
```
tnozicka, line 89: This and onward should have `+optional` when omit empty https://github.com/kubernetes/community/blob/86550d6/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

pkg/api/v1/cluster_types.go
```diff
@@ -110,17 +119,20 @@ type RepairTaskSpec struct {
 	// For Scylla clusters that *do not support row-level repair*, intensity can be a decimal between (0,1).
 	// In that case it specifies percent of shards that can be repaired in parallel on a repair master node.
 	// For Scylla clusters that are row-level repair enabled, setting intensity below 1 has the same effect as setting intensity 1.
+	// +kubebuilder:default:="1"
 	Intensity *string `json:"intensity,omitempty" mapstructure:"intensity,omitempty"`
 	// Parallel the maximum number of Scylla repair jobs that can run at the same time (on different token ranges and replicas).
 	// Each node can take part in at most one repair at any given moment. By default the maximum possible parallelism is used.
 	// The effective parallelism depends on a keyspace replication factor (RF) and the number of nodes.
 	// The formula to calculate it is as follows: number of nodes / RF, ex. for 6 node cluster with RF=3 the maximum parallelism is 2.
+	// +kubebuilder:default:=0
```
tnozicka, line 128: we can probably get rid of the pointers globally, as you have already done somewhere, when there is a default

tnozicka:
nice cleanup! few nits; would be great if you could group the generated changes into a separate commit for review

pkg/api/v1/cluster_types.go
```diff
@@ -190,14 +231,19 @@ type RackSpec struct {
 	// Storage describes the underlying storage that Scylla will consume.
 	Storage StorageSpec `json:"storage"`
 	// Placement describes restrictions for the nodes Scylla is scheduled on.
+	// +optional
 	Placement *PlacementSpec `json:"placement,omitempty"`
 	// Resources the Scylla container will use.
 	Resources corev1.ResourceRequirements `json:"resources"`
 	// AgentResources which Agent container will use.
+	// +kubebuilder:default:={requests: {cpu: "50m", memory: "10M"}, limits: {cpu: "1", memory: "200M"}}
```
tnozicka, line 239: limits?

tnozicka: xref https://github.com/scylladb/scylla-operator/pull/451#discussion_r589452428

install-dependencies.sh
```diff
@@ -61,7 +61,7 @@ then
     sudo apt-get install -y "${debian_packages[@]}"
 fi
 
-go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 sigs.k8s.io/kustomize/kustomize/v3@v3.5.1
+go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1 sigs.k8s.io/kustomize/kustomize/v3@v3.5.1
```
tnozicka, line 64: doesn't seem to belong in vendor commit

zimnx: done

tnozicka: looks like I was a bit ambiguous there - I was pointing out to the script `install-dependencies.sh` that shouldn't be in the same commit as vendor changes. At least I am used to a rule that vendor changes go separately. so I was thinking of 2 commits:
1. go.{mod,sum} vendor/
2. install-dependencies.sh or together with some other related changes

config/operator/webhook/service.yaml
```diff
@@ -7,6 +7,6 @@ metadata:
 spec:
   ports:
     - port: 443
-      targetPort: 443
+      targetPort: 9443
```
tnozicka, line 10: +100 for unprivileged port

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
```diff
@@ -15,1701 +15,1792 @@ spec:
...
-                        type: object
-                      agentVolumeMounts:
-                        description: AgentVolumeMounts to be added to Agent container.
-                        items:
-                          description: VolumeMount describes a mounting of a Volume within a container.
+                      description: RackSpec is the desired state for a Scylla Rack.
+                      properties:
+                        agentResources:
+                          default:
+                            limits:
```
tnozicka, line 129: I thought agent resources aren't enforces so we don't need to set limits here, or not?

config/operator/webhook/service.yaml
tnozicka: but I don't see a corresponding listen change:
a) there is some magic
b) we don't setup validation in e2e and that's why it passed

tnozicka (commented):
looks great, thanks!

some last nits

config/operator/crd/bases/scylla.scylladb.com_scyllaclusters.yaml
zimnx: It's removed in other PR, i wanted to have change which directly translates existing default to markers, and in other commit remove it.

config/operator/webhook/service.yaml
zimnx: I bumped controller-runtime version, and they changed default port, so a) :)

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -142,19 +178,24 @@ type BackupTaskSpec struct {
 	Location []string `json:"location" mapstructure:"location,omitempty"`
 	// RateLimit a list of megabytes (MiB) per second rate limits expressed in the format [<dc>:]<limit>.
 	// The <dc>: part is optional and only needed when different datacenters need different upload limits.
-	// Set to 0 for no limit (default 100).
+	// Set to 0 for no limit.
```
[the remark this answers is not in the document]
zimnx: good catch. done

tnozicka (approved):
/lgtm

PR #467, opened 2021-03-05
helm: Add aggregationRule to ClusterRoles (#440)

### LINKED ISSUE #440: Implement an optional value for the Scylla Helm Chart to add additional rules to ClusterRoleBinding

niels-s:
Add aggregrationRule definitions to each of Helm chart defined
ClusterRoles so end users have an option to extend the rules of the
ClusterRole for their own needs.

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

Resolves #440 

**Checklist:**


tnozicka:
Thanks, this is in line with what I had in mind. It needs to be changed in more files though. We have open PRs to simplify and reorganize our manifests, so if you wouldn't mind waiting a bit, it would be easier to change after it lands.

niels-s:
Sure 👍 keep me posted when the reorganization is done

niels-s:
@tnozicka I assume you are referring to the work in #498?

tnozicka:
That's the one. It should hopefully wire verify target for generated files as well so it will tell you which ones you missed, if any.

niels-s:
Hey @tnozicka, I'm testing these changes on my cluster and I don't think the aggregation rules actually work how we envisioned it. Because I notice the original rules defined in the ClusterRole are now replaced by the rules from one of my label rules for aggregation. 

So in able for this to work with aggregation rules, we should abstract away the rules needed for the manager/operator/cluster into separate ClusterRole and label it accordingly. And make sure we have a ClusterRole with an empty ruleset so the k8s cluster can aggregate all the rules into it, and that one should be used to create the ClusterRoleBinding, etc.

If you look at the [docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) you can see they start out with an empty ClusterRole which will be used for aggregation. 

Curious to hear your thoughts on how you would like to proceed, if you can come back to me I'm able to make the necessary changes and prepare my own deployments to match them.

tnozicka:
hm, that would be unexpected:
```
The default user-facing roles use ClusterRole aggregation. This lets you, as a cluster administrator, include rules for custom resources, such as those served by CustomResourceDefinitions or aggregated API servers, to extend the default roles.
```
We are using it e.g. in https://github.com/scylladb/scylla-operator/blob/74caca42dd9d27650e9fe2ce8079b815f2835928/hack/ci-deploy.sh#L64-L103 and the default edit rules are still present

zimnx:
> hm, that would be unexpected:
> 
> ```
> The default user-facing roles use ClusterRole aggregation. This lets you, as a cluster administrator, include rules for custom resources, such as those served by CustomResourceDefinitions or aggregated API servers, to extend the default roles.
> ```
> 
> We are using it e.g. in
> 
> https://github.com/scylladb/scylla-operator/blob/74caca42dd9d27650e9fe2ce8079b815f2835928/hack/ci-deploy.sh#L64-L103
> 
> and the default edit rules are still present

Yesterday I tried to remove read/write permissions for ScyllaClusters from Operator ClusterRole, add label matching and added this special lable to scylla-view and scylla-edit roles to merge them into Operator role, and indeed i ended up with rule containing only stuff that was merged to it. Base ruleset were cleared out. So I guess it only works when base ruleset is empty, and all the rest is merged using labels.
I wonder if it's intended or it's a k8s bug.

tnozicka:
I guess I've misinterpreted the `extend` in kube docs. I've went over the kubernetes controller for aggregation and it calculates the rules from the selector and replaces anything that was there before. The standard rules worked because they are now actually stored in `system:aggregate-to-view`, not `view`.

niels-s:
Thanks for looking into it, didn't find the time yet to come around and test myself. Any ideas on how the team wants to move forward?

tnozicka:
np, I think we'd follow the example set by kube - so empty destination ClusterRole and the real perms in `-base` or something.

niels-s:
@tnozicka and @zimnx I've rebased with the latest master and changed the aggregation rules accordingly, could you have a look at it? Thanks

tnozicka:
thanks, this will need also updating the `deploy` folder - best to wait for https://github.com/scylladb/scylla-operator/pull/519 to land which will generate the content for you, should land soon

niels-s:
Yeah, I noticed that one already. Also saw it separates the service account for the manager deploy so I'll probably need to make some changes anyway. I'll hold on 👍

niels-s:
@tnozicka I've rebased the PR and updated all the necessary examples/deploys I hope could you verify the naming is also as desired? Would love to get some feedback and get it merged since it's a feature we depend on 🙏 thanks

niels-s:
Seems like I need to fix the e2e tests, looks like it's not deploying the roles properly, I'll have a look next week Monday. But if you have any pointers in the meanwhile how to resolve it, that would be a nice starting point 👍

tnozicka:
hm, the artifacts already expired :( when re-trigger it will give you a dump of all the namespaces (yamls, logs) to find out why the operator didn't deploy

tnozicka:
```

**Comment by tnozicka** [comment]
given the roles growing, naming wise I'd aim for this pattern with clusterroles:
```
scylladb:controller:operator (deploy/operator/00_clusterrole.yaml)
scylladb:controller:aggregate-to-operator (deploy/operator/00_clusterrole_def.yaml)

scylladb:controller:manager-controller
scylladb:controller:aggregate-to-manager-controller

scyllacluster-member
scylladb:aggregate-to-scyllacluster-member
```
scyllacluster-member is the only one user needs to bind manually (for now) so I'd keep the public name.

**Comment by niels-s** [comment]
I've updated the names accordingly and also renamed the files according to your hint in the previous commit. I didn't touch the service account name for the operator since it's using a template helper to define the name. So I waited to see if you want to change that and remove the name option? 

I still need to fix the e2e tests so I'll look into that next

**Comment by tnozicka** [comment]
```
{"L":"FATAL","T":"2021-04-28T08:50:05.510Z","M":"unable to create manager controller","error":"discover manager: failed to list member services: services is forbidden: User \"system:serviceaccount:scylla-manager:scylla-manager-controller\" cannot list resource \"services\" in API group \"\" in the namespace \"scylla-manager\": RBAC: clusterrole.rbac.authorization.k8s.io \"scylla-manager-controller\" not found","_trace_id":"SpSPtxpLQGmhYv0R_UnT2A","errorStack":"github.com/scylladb/scylla-operator/pkg/controllers/manager.New\n\tgithub.com/scylladb/scylla-operator/pkg/controllers/manager/manager_controller.go:52\ngithub.com/scylladb/scylla-operator/pkg/cmd/operator.NewManagerControllerCmd.func1\n\tgithub.com/scylladb/scylla-operator/pkg/cmd/operator/manager_controller.go:52\ngithub.com/spf13/cobra.(*Command).execute\n\tgithub.com/spf13/cobra@v1.1.1/command.go:854\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\tgithub.com/spf13/cobra@v1.1.1/command.go:958\ngithub.com/spf13/cobra.(*Command).Execute\n\tgithub.com/spf13/cobra@v1.1.1/command.go:895\nmain.main\n\tgithub.com/scylladb/scylla-operator/cmd/scylla-operator/scylla-operator.go:32\nruntime.main\n\truntime/proc.go:225\nruntime.goexit\n\truntime/asm_amd64.s:1371\n"}
```

**Comment by tnozicka** [comment]
```
scylla-manager-controller                              ClusterRole/scylla-manager-controller                                              6m30s                                                                                      scylla-manager/scylla-manager-controller
scylla-operator                                        ClusterRole/scylladb:controller:operator                                           6m44s                                                                                      scylla-operator/scylla-operator
```
the clusterrolebindings should follow the same naming pattern and it looks like the manager-controller is bound the a wrong clusterrole

tnozicka:
otherwise the structure lgtm, you need to fix the binding, rebase and split generated changes into a separate commit. thanks

niels-s:
@tnozicka I addressed your remarks and splitsed the generated changes into separate commits. The tests look to be green, but when I look at `Test Integration` step even though it's marked as successful, when checking the output there are 7 failing tests that I don't think are related to this PR.
When I checked the same step for a commit on master it has the same issue: https://github.com/scylladb/scylla-operator/runs/2456771212

tnozicka:
No worries about the integration - the result is intentionally ignored as it was flaky and we are migrating to e2e. e2e and units is what matters.

Just squash the 2 generated commits into one (say called `make update`) and you are good to go. Thanks.

niels-s:
@tnozicka great well rebased and changed the commits. Ready for final review 🙏 😛

tnozicka (approved):
/lgtm
thanks @niels-s

PR #468, opened 2021-03-05
actions: trigger CI for every PR

zimnx:
Single star doesn't match target branches containing '/' character.
Double star matches all branches.


tnozicka (approved):
lgtm

PR #487, opened 2021-03-10
Promote the tested image

tnozicka:
We should promote the same image that was tested, not to build a new one.

Fixes promotion issues like https://github.com/scylladb/scylla-operator/runs/2078750039?check_suite_focus=true#step:6:32


zimnx (approved):
LGTM, thanks

PR #492, opened 2021-03-11
Add Slack notifications

tnozicka:
If a run fails (that isn't a PR) it will notify our Slack channel. This is especially important for promotions where we need to rerun the job otherwise we would be missing an image on GH or other artifact.


zimnx (approved):
LGTM, thanks

PR #494, opened 2021-03-12
Deploy e2e with validating admission webhook

tnozicka:
CI e2e wasn't deploying the validation webhook config, this PR fixes it. We don't need to handle the other webhooks as these are just going away in the next PR https://github.com/scylladb/scylla-operator/pull/466


tnozicka:
we have that already, I was checking a wrong dir

PR #495, opened 2021-03-12
Don't fail workflow on flaky integration suite

tnozicka:
Don't fail the workflow on integration suite failure which isn't a required check. We should remove it when we deflake the tests and make them required or transfer them to another suite.

Resolves #

**Checklist:**


zimnx (approved):
LGTM, thanks

PR #502, opened 2021-03-16
helm: bump default scylla storage capacity to 10Gi (#501)

### LINKED ISSUE #501: Default values of Scylla Helm chart cannot be deployed
**Describe the bug**

Scylla Helm chart cannot be deployed using default values because developerMode is disabled and Scylla requires at least 10GB of storage. Chart use 5GB as default.

zimnx:
Scylla refuses to start when developerMode is disabled and disk is smaller than 10Gi.

Fixes #501

### DIFF
--- a/helm/scylla/values.yaml
+++ b/helm/scylla/values.yaml
@@ -55,7 +55,7 @@ racks:
     members: 3
     # Storage definition
     storage:
-      capacity: 5Gi
+      capacity: 10Gi
     # Scylla container resource definition
     resources:
        limits:


tnozicka:
Can we do 10GiB to start with? It would be easiest for those just wanting to try it out and anyone will raise it for production anyways.

tnozicka:
(setting the milestone to 1.2 which is the current master, 1.1 would be for the backport)

tnozicka (approved):
lgtm, thanks!

PR #503, opened 2021-03-16
[1.1] helm: bump default scylla storage capacity to 10Gi (#501)

### LINKED ISSUE #501: Default values of Scylla Helm chart cannot be deployed

zimnx:
Scylla refuses to start when developerMode is disabled and disk is
smaller than 10Gi.

Fixes #501

### DIFF
--- a/helm/scylla/values.yaml
+++ b/helm/scylla/values.yaml
@@ -55,7 +55,7 @@ racks:
     members: 3
     # Storage definition
     storage:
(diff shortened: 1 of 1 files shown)


tnozicka (approved):
lgtm

PR #498, opened 2021-03-16
Deployment changes

zimnx:
Reopen of: https://github.com/scylladb/scylla-operator/pull/451

**Description of your changes:**

- Scylla Operator and Manager Controller are deployed as Deployments instead of as StatefulSet. 
- Scylla Operator and Manager Controller are deployed in 2 replicas and they use leader election.
- New deploy directory containing static manifest files for each resource needed for installation.
- Naming of resources in manifests/charts was improved.
- Resource limit of ScyllaOperator was removed, resource request was reduced.

**Which issue is resolved by this Pull Request:**
Resolves #408 

**Checklist:**
- [ ] Documentation has been updated, if necessary.
- [x] Image has been built (`make docker-build`) on the last commit.

PR #504, opened 2021-03-16
helm: Fix ServiceAccount for Scylla Manager chart

niels-s:
The deploymentrelied on the "default" Service Account, but
didn't bind the right ClusterRole when another Service Account is used
as defined by the template helper `scylla-manager.serviceAccountName`.

**Checklist:**

### DIFF
--- a/helm/scylla-manager/templates/manager_deployment.yaml
+++ b/helm/scylla-manager/templates/manager_deployment.yaml
@@ -21,6 +21,7 @@ spec:
       labels:
         {{- include "scylla-manager.labels" . | nindent 8 }}
     spec:
+      serviceAccountName: {{ include "scylla-manager.serviceAccountName" . }}
       containers:
       - name: {{ .Chart.Name }}
         image: {{ .Values.image.repository }}/scylla-manager:{{ .Values.image.tag | default .Chart.AppVersion }}
--- a/helm/scylla-manager/templates/manager_serviceaccount.yaml
+++ b/helm/scylla-manager/templates/manager_serviceaccount.yaml
@@ -1,0 +1,13 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "scylla-manager.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "scylla-manager.labels" . | nindent 4 }}
+  {{- with .Values.serviceAccount.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+{{- end }}
--- a/Makefile
+++ b/Makefile
@@ -295,6 +295,7 @@ define generate-manager-manifests-prod
 	mv '$(3)'/scylla-manager/templates/controller_serviceaccount.yaml '$(2)'/10_controller_serviceaccount.yaml
 	mv '$(3)'/scylla-manager/templates/controller_pdb.yaml '$(2)'/10_controller_pdb.yaml
 	mv '$(3)'/scylla-manager/templates/manager_service.yaml '$(2)'/10_manager_service.yaml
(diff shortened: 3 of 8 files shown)


niels-s:
It seems like this is already handled in #498 so it's probably easiest to wait until those changes land in master/ release candidate

niels-s:
@tnozicka or @zimnx I've rebased this PR with the latest master. Seems like the service account was still missing from the manager deployment. In this case, I reuse the service account created for the controller part, I guess it's fine for now, or we could introduce another ServiceAccount specifically for the manager, let me know what you think 👍

niels-s:
Hey Team, I've rebased so could one of you check this out? It's a very small change. Thanks

helm/scylla-manager/templates/manager_deployment.yaml
```diff
@@ -21,6 +21,7 @@ spec:
       labels:
         {{- include "scylla-manager.labels" . | nindent 8 }}
     spec:
+      serviceAccountName: {{ include "scylla-manager.serviceAccountName" . }}
```
zimnx, line 24: I noticed this ServiceAccount is not created by any template. 

You also have to execute `make update` to update the autogenerated files.

niels-s: Yes, noticed that in #467 😅 I've should have updated everything now 👍

niels-s:
Like with #467 the e2e will probably fail, so I'll have to fix that too. Will have a look on Monday

niels-s:
@zimnx could you do another round of reviews?

tnozicka (commented):
I don't see why it should have a ClusterRole - manager doesn't need any cluster-level permissions, actually any at all I think.

Can you describe the usecase you are trying to solve? I don't think it needs privileged pods like in the case of https://github.com/scylladb/scylla-operator/issues/440

niels-s:
@tnozicka, I think it a mistake on my part. A regular role should be sufficient to I guess. I need to give more privileges to the service account on our cluster since the manager container runs as root, but by default, we configure our workload to `runAsNonRoot`. To get things working, I copy/pasted the approach from the controller. 
But I'll give it a try and see if I can get it working with a regular role instead.

Another solution would be not to run the container as root.

niels-s:
I've removed the CusterRole changes, for now, so we can get the SA changes approved and merged first. This change will allow us to deploy the chart as is because we have an admission controller which blocks any deploys that rely on the default service account.

I've opened #570 to discuss it further but already tested that adding a `runAsUser: 999` works as expected.

Could you perhaps add a label to this PR to get the checks green?

helm/scylla-manager/templates/manager_serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.controllerServiceAccount.create -}}
```
tnozicka, line 1: @zimnx would we just always create the SA account with the fixed name? we own the namespace

zimnx: It's a common pattern in Helm (https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L131, https://github.com/bitnami/charts/blob/master/bitnami/cassandra/values.yaml#L60), and it's autogenerated in the chart scaffold. 
Although I'm not enterly sure what's the usecase for disabling creation of serviceAccount. Perhaps adding some custom stuff like image secrets, but these could be added to the SA template.
I'm fine with always creating it, but that would be a regression compared to 1.1.

zimnx, line 1:
```suggestion
{{- if .Values.serviceAccount.create -}}
```

niels-s: Whoops thanks 🔪 👍

niels-s:
Fixed @zimnx comment, rebased and reorganized the commit to using the same approach as my previous PR, and group the generated files in a single commit.

Ready for a final round of reviews?

zimnx (approved):
LGTM, thanks for contribution!

tnozicka (approved):

Issue #484, opened 2021-03-10
Linkerd sidecars fails due to invalid resolv.conf on generated pods


flotothemoon:
**Describe the bug**
When applying linkerd service mesh to pods generated with this operator, the linkerd sidecars error during startup because of an invalid resolv.conf file. This also causes Scylla manager to fail (since it relies on a Scylla cluster). This may be due to a conflict in sidecar injection between Scylla operator and Linkerd. This might be a variant/regression of a bug that was reported and fixed earlier (#176 Pods generated by the Scylla operator cause runtime errors in linkerd sidecars). 

**To Reproduce**
The steps to me are identical to that in the referenced issue - simply use `linkerd inject` (or manually annotate generated pods with `linkerd.io/inject: enabled`). See the referenced issue #176.

**Expected behavior**
Linkerd should operate transparently, meshing the Scylla generated pods without any errors.

**Config Files**
- 

**Logs**
```
time="2021-03-10T11:05:47Z" level=info msg="running version stable-2.9.4"
time="2021-03-10T11:05:47Z" level=info msg="Found pre-existing key: /var/run/linkerd/identity/end-entity/key.p8"
time="2021-03-10T11:05:47Z" level=info msg="Found pre-existing CSR: /var/run/linkerd/identity/end-entity/csr.der"
[     0.000824s]  INFO ThreadId(01) linkerd2_proxy::rt: Using single-threaded proxy runtime
thread 'main' panicked at 'system DNS config must be valid: ResolveError { kind: Io(Custom { kind: Other, error: "Error parsing resolv.conf: InvalidOption(1)" }), backtrack: None }', linkerd/app/core/src/dns.rs:21:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Environment:**
 - Platform: minikube
 - Kubernetes version: 1.20 
 - Scylla version: 4.2.0
 - Scylla-operator version: master

**Additional context**
-

flotothemoon:
@tzach sorry to ping directly, looking to get this resolved or move on. Is this an issue with Scylla, the operator or some other setup? Any ideas for a workaround if there is no easy fix in sight?

tnozicka:
> error during startup because of an invalid resolv.conf file

Can you exec into the sidecar that is failing and attach the invalid resolv.conf? Also possibly some other random pod that is working to have a baseline and see the diff.

[...]

PR #512, opened 2021-03-19
Always default to cluster first DNS

tnozicka:
Kubernetes defaults to `ClusterFirst` DNSPolicy, yet we were defaulting to `Default` which breaks some sidecars like linkerd that use cluster DNS. Defaulting to the newer `ClusterFirstWithHostNet` as default should cover all cases.

Likely https://github.com/scylladb/scylla-operator/issues/484 and https://scylladb-users.slack.com/archives/CCC9J41FW/p1616127037002500

### DIFF
--- a/pkg/api/scylla/v1/cluster_types.go
+++ b/pkg/api/scylla/v1/cluster_types.go
@@ -205,13 +205,11 @@ type Network struct {
 }
 
 func (s Network) GetDNSPolicy() corev1.DNSPolicy {
-	if s.DNSPolicy == "" {
-		if s.HostNetworking {
-			return corev1.DNSClusterFirstWithHostNet
-		}
-		return corev1.DNSDefault
+	if len(s.DNSPolicy) != 0 {
+		return s.DNSPolicy
 	}
-	return s.DNSPolicy
+
+	return corev1.DNSClusterFirstWithHostNet
 }
 
 // DatacenterSpec is the desired state for a Scylla Datacenter.


pkg/api/scylla/v1/cluster_types.go
```diff
@@ -205,13 +205,11 @@ type Network struct {
...
-		if s.HostNetworking {
-			return corev1.DNSClusterFirstWithHostNet
-		}
-		return corev1.DNSDefault
+	if len(s.DNSPolicy) != 0 {
+		return s.DNSPolicy
 	}
-	return s.DNSPolicy
+
+	return corev1.DNSClusterFirstWithHostNet
```
zimnx, line 212: shouldn't we set DNSClusterFirstWithHostNet for hostNetworking setups and DNSClusterFirst for others?

tnozicka: not based on
https://github.com/scylladb/scylla-operator/blob/6c9209d48b551558872b9d0587767672c658d5f0/vendor/k8s.io/api/core/v1/types.go#L2578-L2586
and trying it out

tnozicka: but let me check kubelet code to be 100% sure

tnozicka: https://github.com/kubernetes/kubernetes/blob/7735cf7f2674627283db081d61a2c45f4b70ee8f/pkg/kubelet/network/dns/dns.go#L275-L285

zimnx (approved):
LGTM

PR #519, opened 2021-03-23
deploy: generate manifests from Helm (#482)

### LINKED ISSUE #482: Make sure helm charts don't get out of sync with our deployment manifests
* Feature Request

Make sure helm charts don't get out of sync with our deployment manifests. Like https://github.com/scylladb/scylla-operator/pull/480

When editing manifests it's easy to overlook a change that wasn't reflected in helm manifests.

**Additional Information:**
As the helm manifests are templated we can't easily use the same source. We can create a verify/update job to keep the diff between a regular and helm manifest committed as generated file. When changing a manifest without a corresponding helm edit, or the other way around, verify won't pass until both manifests are properly edited or the diff is updated.

zimnx:
Static manifests are generated from Helm charts.
This will guarantee that static manifests are in sync with Helm charts.

Fixes #482


tnozicka (dismissed):
~/lgtm~ (sorry, had this opened from yesterday - I went only over the schema so far)

deploy/operator/10_certificate.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
-  name: scylla-operator-serving-cert
+  name: scylla-operator-selfsigned
```
tnozicka, line 4: scylla-operator-serving-cert

Makefile
```diff
@@ -243,22 +243,67 @@ verify-crd:
 	$(diff) '$(tmp_file)' '$(CRD_PATH)' || (echo 'File $(CRD_PATH) is not up-to date. Please run `make update-crd` to update it.' && false)
 .PHONY: verify-crd
 
-# $1 - target dir
-define generate-dev-manifests
-	mkdir -p '$(1)' && \
-	cp deploy/manager/prod/*.yaml '$(1)' && \
-	yq eval '.spec.cpuset=false | .spec.datacenter.racks[] = (.spec.datacenter.racks[] | .resources.limits.cpu="200m" | .resources.limits.memory="200Mi" | .resources.requests.cpu="10m" | .resources.requests.memory="100Mi" )' -i '$(1)'/10_scylla_scyllacluster.yaml
+# $1 - name
+define generate-namespace
```
tnozicka, line 247: I don't think we want to inline manifests into the makefile. If it's not coming from helm, why don't we just save the file in the target directory?

Makefile
```diff
@@ -243,22 +243,67 @@ verify-crd:
...
-update-deploy-dev:
-	$(call generate-dev-manifests,deploy/manager/dev)
-.PHONY: update-deploy-dev
+# $1 - name
+# $2 - chart path
+# $3 - values path
+# $4 - target dir
+define generate-manifests
+	mkdir -p '$(4)'
+	$(file > $(4)/00_namespace.yaml,$(generate-namespace))
```
tnozicka, line 260: `$(file >` can you elaborate on what this does? I am not familiar with such syntax in the makefile, couldn't find buildin function `file` in make reference either.

Makefile
```diff
@@ -243,22 +243,67 @@ verify-crd:
...
 endef
 
-update-deploy-dev:
-	$(call generate-dev-manifests,deploy/manager/dev)
-.PHONY: update-deploy-dev
+# $1 - name
+# $2 - chart path
+# $3 - values path
+# $4 - target dir
+define generate-manifests
```
tnozicka, line 258: generate-manifests-from-helm

Makefile
```diff
@@ -243,22 +243,67 @@ verify-crd:
...
+
+update-deploy-manager: update-deploy-manager-dev update-deploy-manager-prod
+.PHONY: update-deploy-manager
+
+verify-deploy-manager: verify-deploy-manager-dev verify-deploy-manager-prod
+.PHONY: verify-deploy-manager
+
+update-deploy-operator:
+	$(call generate-manifests,scylla-operator,helm/scylla-operator,deploy/operator.yaml,deploy/operator)
+	ln -sf ../../$(CRD_PATH) deploy/operator/00_scylla.scylladb.com_scyllaclusters.yaml
```
tnozicka, line 292: the symlink can be manual

tnozicka (commented):
first pass, but I'd not generate the directory from scratch. How about creating the dir and some files like namespace or symlinks manually and just wiring updates to the other files?

Makefile
```diff
@@ -243,22 +243,67 @@ verify-crd:
...
+# $2 - chart path
+# $3 - values path
+# $4 - target dir
+define generate-manifests
+	mkdir -p '$(4)'
+	$(file > $(4)/00_namespace.yaml,$(generate-namespace))
+	helm template '$(1)' '$(2)' --namespace '$(1)' --values '$(3)' | ./hack/split-helm.sh '$(4)'
+endef
+
+update-deploy-manager-prod:
```
tnozicka, line 264: nit: Although your version is better for parallelism, I'd be fine if we merge those into just `update-deploy` for simplicity. I am fine with whichever you prefer. Just wanted to write it down so you know you have the option.

deploy/manager_dev.yaml
```diff
@@ -0,0 +1,46 @@
+# Scylla Manager image
```
tnozicka, line 1: I'd keep the heml vars somewhere else so it doesn't confuse users under deploy/ dir which isn't really connected to helm. We may change the source in the future, split it, or not.

Makefile
```diff
@@ -243,22 +243,67 @@ verify-crd:
...
-	$(call generate-dev-manifests,deploy/manager/dev)
-.PHONY: update-deploy-dev
+# $1 - name
+# $2 - chart path
+# $3 - values path
+# $4 - target dir
+define generate-manifests
+	mkdir -p '$(4)'
+	$(file > $(4)/00_namespace.yaml,$(generate-namespace))
+	helm template '$(1)' '$(2)' --namespace '$(1)' --values '$(3)' | ./hack/split-helm.sh '$(4)'
```
tnozicka, line 261: Would it be better to pipe it into `yq` and use the structured file split?

tnozicka: or `helm template --output-dir` to tmpdir and mv / rename files ?

Makefile
zimnx: no longer used, but it allows to save templatized definition in file. https://www.gnu.org/software/make/manual/html_node/File-Function.html

deploy/manager_dev.yaml
zimnx: moved to helm/deploy/

Makefile
zimnx: i wasn't able to use yq, but i replaced awk script with Helm `--output-dir` and couple of seds and finds.

tnozicka (commented):
Would this be better
https://github.com/tnozicka/scylla-operator/commit/b5a3675e862da19794bdc2c3adb1a22a8cf1a1c3
then forcing it all into runlevel 10?

Makefile
```diff
@@ -244,22 +244,89 @@ verify-crd:
...
+
+	@leftovers=$$( find '$(3)'/scylla-manager/ -mindepth 1 -type f ) && [[ "$${leftovers}" == "" ]] || \
+	( echo -e "Internal error: Unhandled helm files: \n$${leftovers}" && false )
+endef
+
+update-deploy: tmp_dir:=$(shell mktemp -d)
+update-deploy:
+	$(call generate-operator-manifests,helm/deploy/operator.yaml,deploy/operator,$(tmp_dir))
+	$(call generate-manager-manifests,helm/deploy/manager_prod.yaml,deploy/manager/prod,$(tmp_dir))
+	$(call generate-manager-manifests,helm/deploy/manager_dev.yaml,deploy/manager/dev,$(tmp_dir))
```
tnozicka, line 311: This won't verify e.g. the manual `00_namespace.yaml` file matches (or any future manual files) - would it be safer to cp `deploy/manager/prod` into  `deploy/manager/dev` and just `yq` edit the resources?

Makefile
```diff
@@ -244,22 +244,89 @@ verify-crd:
...
+.PHONY: update-deploy
+
+verify-deploy: tmp_dir :=$(shell mktemp -d)
+verify-deploy: tmp_dir_generate :=$(shell mktemp -d)
+verify-deploy:
+	mkdir -p $(tmp_dir)/{operator,manager}
```
tnozicka, line 317: nit `/{operator,manager/{prod,dev}}`

Makefile
tnozicka: also `helm/deploy/manager_prod.yaml` and `helm/deploy/manager_dev.yaml` could get out of sync

deploy/manager/prod/10_manager_configmap.yaml
```diff
@@ -7,7 +7,7 @@ data:
   scylla-manager.yaml: |-
     http: :5080
     logger:
-      level: info
+      level: debug
```
tnozicka, line 10: nit: info

deploy/manager/prod/50_scyllacluster.yaml
```diff
@@ -4,23 +4,28 @@ metadata:
...
 spec:
-  agentVersion: 2.2.0
+  version: 4.3.2
+  agentVersion: 2.2.4
+  repository: docker.io/scylladb/scylla
+  agentRepository: docker.io/scylladb/scylla-manager-agent
+  developerMode: true
+  cpuset: true
+  automaticOrphanedNodeCleanup: false
+  network:
```
tnozicka, line 14: nit: can we get rid of those omitempty values?

tnozicka (commented):
Thanks for updating it. The last outstanding issue is handling dev manifests, otherwise this lgtm.

tnozicka:
@zimnx would you mind picking the updates in https://github.com/tnozicka/scylla-operator/commit/9ef22bd934632ccb5b61ca486c201bf1aa95a493 and splitting them to the appropriate commits?

tnozicka (approved):
/lgtm
thanks!

PR #529, opened 2021-03-29
Let scylla listen on 0.0.0.0

tnozicka:
Scylla now listens on all interfaces except just the ClusterIP. This will allow people to connect to the localhost and allow service meshes to redirect the traffic there.

The comment for `--listen-address` saying `Never specify 0.0.0.0; it is always wrong.` is likely wrong as it predates `--broadcast-address` flag. (Pending confirmation from Scylla engineering.)

Resolves https://github.com/scylladb/scylla-operator/issues/484


zimnx (approved):
LGTM, thanks

PR #539, opened 2021-04-06
Get rid of busybox image

### LINKED ISSUE #431: hardcoded busybox image
I have just noticed scylla cluster requires  busybox image which is “hardcoded” and do appear in cluster as
image: busybox:1.31.1
There is a need to support loading busybox image from our own repository

tnozicka:
We already have the sidecar image pulled on the machine so there is no need to pull another one, reusing is faster. It also avoid having another image that needs to be configurable.

Resolves #431


tnozicka:
I don't think this is cover in our standard test suite so it at least needs to be tested manually.

tnozicka:
Added the e2e test.

zimnx (approved):
LGTM, thanks

PR #543, opened 2021-04-12
e2e: speedup scylla bootstrap

zimnx:


zimnx:
compared to last `master` it saves ~7mins. I'll retrigger it couple of times to check flakyness.

tnozicka:
That looks good, but to have a better sense of how much the args help, can you bump scylla version separately?

I am fine with using args tuning but we should make sure there is one e2e test that runs without any extra args, to make sure it rolls out. If it's only a one the price won't be that big.

zimnx:
> That looks good, but to have a better sense of how much the args help, can you bump scylla version separately?

Appending args to Scylla image is available since 4.2, so tests validating update from 4.1 -> 4.2 wouldn't pass.

> I am fine with using args tuning but we should make sure there is one e2e test that runs without any extra args, to make sure it rolls out. If it's only a one the price won't be that big.

Good point, added single smoke test using 'slow' cluster.

tnozicka:
> Appending args to Scylla image is available since 4.2, so tests validating update from 4.1 -> 4.2 wouldn't pass.

I meant a prerequisite PR, so you first merge it and then build this PR on top and do benchmarks.

zimnx:
This trick introduced flakyness, so i'll revisit it later.

PR #545, opened 2021-04-14
member service: ensure service is well deleted when pod is decommission

ashangit:
When a pod is decommission a DecommissionLabel set at true is added by the sidecar operator
on the service. Then the operator delete this service during cleanup stage. But the pod
can still be alive which leads to the recreation of this service in syncMemberServices without the DecommissionLabel
Due to that the service cannot be the service can be never reclaimed

Resolves #


zimnx:
Solved by #534

PR #549, opened 2021-04-16
sidecar/config: fix ScyllaArgs parsing

zimnx:
Parsing of combination of different value typeswas broken, because part of next argument was considered as value.


tnozicka (approved):
/lgtm

PR #551, opened 2021-04-16
makefile: use the same image tag for manifests and images

zimnx:


Makefile
```diff
@@ -261,6 +261,8 @@ endef
 define generate-operator-manifests
 	$(call generate-manifests-from-helm,scylla-operator,helm/scylla-operator,$(1),$(3))
 
+    $(YQ) eval -i -P '.spec.template.spec.containers[0].image = "$(IMAGE_REF)"' '$(3)'/scylla-operator/templates/deployment.yaml
```
tnozicka, line 264: wouldn't we rather change the default in the helm template and this would get it for free?

tnozicka (commented):
also please split generated changes

helm/deploy/manager_prod.yaml
```diff
@@ -5,7 +5,6 @@ image:
 
 controllerImage:
   repository: docker.io/scylladb
-  tag: latest
```
tnozicka, line 8: would it be better to add `<unspecified>`? so we get an explicit failure and not default to latest if the substitution breaks

tnozicka (dismissed):
/lgtm
thanks

tnozicka (approved):

PR #553, opened 2021-04-19
e2e: Dump test namespace content

tnozicka:
When something goes wrong in the CI we need to dump information for debugging, like for
https://github.com/scylladb/scylla-operator/runs/2376373176?check_suite_focus=true
Given CI has limited resources, we can't use `--delete-namespace-policy=OnSuccess` and `hack/ci-gather-artifacts.sh` as someone running a real cluster or individual test would. Leaving the namespaces around might lead to resource exhaustion and failing future tests as a result.
This PR makes sure we collect testing namespace dump before deleting it.

### DIFF
--- a/test/e2e/framework/dump.go
+++ b/test/e2e/framework/dump.go
@@ -1,0 +1,222 @@
+package framework
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


test/e2e/framework/dump.go
```diff
@@ -0,0 +1,214 @@
...
+			return err
+		}
+
+		err = ioutil.WriteFile(filepath.Join(resourceDir, fmt.Sprintf("%s.yaml", obj.GetName())), data, 0666)
+		if err != nil {
+			return err
+		}
+
+		switch gvr.String() {
+		case corev1.SchemeGroupVersion.WithResource("pods").String():
```
zimnx, line 97: nit: case body is too long, extracting it to DumpPod would reduce lenght of entire function and indentation

tnozicka: done

tnozicka:
(clean rebase)

zimnx (approved):

PR #556, opened 2021-04-21
Tool which generates release notes based on merge commits between two git refs

zimnx:
This tool automatically generates release notes based on merged PRs between two dates.
PRs are grouped by kinds using their labels.


zimnx:
Example run: `./gen-release-notes --merged-from "$(git show -s v1.1.0-alpha.0 --format='%cI')" --merged-to "$(git show -s v1.2.0-alpha.0 --format='%cI')"  --maintainers zimnx,tnozicka --release 1.2.0`
Example output  https://gist.github.com/zimnx/cbd935aec7de74bec45c419858456fef

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	toCommit, err := commitFromTag(r, tagTo)
+	if err != nil {
+		return nil, err
+	}
+
+	mergeBaseCommits, err := toCommit.MergeBase(fromCommit)
```
tnozicka, line 154: Doesn't the start commit already have to me a merge-base? I was assuming the start commit is already a direct predecessor of the end commit, otherwise the PR diff would be 2 ways. 

- for diff between `v1.1.0` and `v1.1.1` this is fine
- when we want a diff between `v1.2.0` and the previous release base, I think we'd rather find the merge-base externally

pkg/cmd/utils/gen-release-notes/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+	"time"
+
+	"github.com/google/go-github/v35/github"
+	"golang.org/x/oauth2"
+)
+
+func githubClient(ctx context.Context) *github.Client {
+	var httpClient *http.Client
+	if os.Getenv("GH_TOKEN") != "" {
```
tnozicka, line 17: wire it as a flag

pkg/cmd/utils/gen-release-notes/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+		ts := oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: os.Getenv("GH_TOKEN")},
+		)
+		httpClient = oauth2.NewClient(ctx, ts)
+	}
+
+	return github.NewClient(httpClient)
+}
+
+type TimeFlag struct {
```
tnozicka, line 27: I don't see it wired

pkg/cmd/utils/gen-release-notes/render.go
```diff
@@ -0,0 +1,112 @@
...
+			{Title: "Other", Kinds: []string{"cleanup"}},
+		},
+	}
+
+	t := template.New("release-notes").Funcs(map[string]interface{}{
+		"FilterByLabelKinds": filterByLabelKinds,
+		"Uncategorized":      uncategorized,
+		"FormatPR":           formatPR,
+	})
+	t = template.Must(t.Parse(releaseNotesTemplate))
```
tnozicka, line 77: error handling

tnozicka:
few comments, but good job with generating nice release notes!

pkg/cmd/utils/gen-release-notes/render.go
```diff
@@ -0,0 +1,112 @@
...
+  {{- end }}
+{{- end }} 
+
+---
+`
+
+func formatPR(pr *github.PullRequest) string {
+	user := pr.GetUser()
+	details := fmt.Sprintf("[#%d](%s),[@%s](%s)", pr.GetNumber(), pr.GetHTMLURL(), user.GetLogin(), user.GetHTMLURL())
+	return fmt.Sprintf("* %s (%s)", strings.TrimSpace(pr.GetTitle()), details)
```
tnozicka, line 44: you need to remove `[v1.1]` prefixes

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,183 @@
...
+
+		n, err := parsePullRequestNumber(c.Message)
+		if err != nil {
+			if errors.Is(ErrNotAPullRequest, err) {
+				continue
+			}
+			return nil, err
+		}
+
+		pr, _, err := ghClient.PullRequests.Get(ctx, o.OrganizationName, o.RepositoryName, n)
```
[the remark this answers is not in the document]
zimnx: Unfortunately library doesn't allow to query for time, api is very limited. I changed to listing each page with PRs in descending order of creation.

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,183 @@
...
+			return nil, err
+		}
+		pullRequests = append(pullRequests, pr)
+	}
+
+	return pullRequests, nil
+}
+
+func isMergeCommit(commit *object.Commit) bool {
+	return commit.NumParents() > 1
```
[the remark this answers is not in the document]
zimnx: I'll leave it as it is.

pkg/cmd/utils/gen-release-notes/generate.go
zimnx: changed to sanity check

pkg/cmd/utils/gen-release-notes/helpers.go
zimnx: good catch, it's a leftover

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,193 @@
...
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&o.ReleaseName, "release-name", o.ReleaseName, "Name of the release.")
+	cmd.Flags().StringVar(&o.RepositoryPath, "repository-path", o.RepositoryPath, "Path to the git repository.")
+	cmd.Flags().StringVar(&o.RefStart, "ref-start", o.RefStart, "PRs merged after before this ref won't be considered.")
```
tnozicka, line 57: The comment should answer the question of "What is ref-start?". "First commit reference." You can describe the behaviour after if needed.

tnozicka: naming wise I'd prefer `start-ref` to `ref-start`, similarly you'd use `start-number`, not `number-start`

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,193 @@
...
+	cmd.Flags().StringVar(&o.RefEnd, "ref-end", o.RefEnd, "PRs merged before before this ref won't be considered.")
+	cmd.Flags().StringVar(&o.GithubToken, "github-token", o.GithubToken, "GitHub token used in request authentication.")
+
+	return cmd
+}
+
+func pullRequestsFromMergeCommits(ctx context.Context, ghClient *github.Client, o *GenerateOptions) ([]*github.PullRequest, error) {
```
tnozicka, line 64: shouldn't this be a method of `GenerateOptions` given you are passing the whole struct along?

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,193 @@
...
+	repo, err := git.PlainOpen(o.RepositoryPath)
```
tnozicka, line 65: or you could do this in .Complete() and have a method that takes git repo struct

cmd/gen-release-notes/gen-release-notes.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"flag"
+	"fmt"
+	"math/rand"
+	"os"
+	"runtime"
+	"time"
+
+	cmd "github.com/scylladb/scylla-operator/pkg/cmd/utils/gen-release-notes"
```
tnozicka, line 14: i'd not nest it under utils of it's not nested in `/cmd/`

pkg/cmd/utils/gen-release-notes/helpers.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/google/go-github/v35/github"
+	"golang.org/x/oauth2"
+)
+
+func githubClient(ctx context.Context, token string) *github.Client {
```
tnozicka, line 13: this is what o.complete would do

pkg/cmd/utils/gen-release-notes/options.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if err := renderReleaseNotes(o.Out, o.ReleaseName, pullRequests); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func NewGitGenerateOptions(streams genericclioptions.IOStreams) *GenerateOptions {
```
tnozicka, line 48: "constructors" are conventionally done right after struct declaration

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,193 @@
...
+func commitFromRevision(r *git.Repository, revName string) (*object.Commit, error) {
+	rev, err := r.ResolveRevision(plumbing.Revision(revName))
+	if err != nil {
+		return nil, fmt.Errorf("resolve revision %q, %w", revName, err)
+	}
+
+	hash := plumbing.NewHash(rev.String())
+	commit, err := r.CommitObject(hash)
+	if err != nil {
+		return nil, fmt.Errorf("commit object, %w", err)
```
tnozicka, line 149: nit: error chains are split with `: `

pkg/cmd/utils/gen-release-notes/render.go
```diff
@@ -0,0 +1,130 @@
...
+{{ FormatPR $pr }}
+  {{- end }}
+{{- end }} 
+
+---
+`
+
+var branchPrefixRe = regexp.MustCompile(`^(\[[^]].*])?(.*)`)
+
+func cleanupPRTitle(title string) (string, error) {
```
tnozicka, line 44: can you write a unit test for it?

pkg/cmd/utils/gen-release-notes/render.go
```diff
@@ -0,0 +1,130 @@
...
+			{Title: "Features", Kinds: []string{"feature"}},
+			{Title: "Bugs", Kinds: []string{"bug"}},
+			{Title: "Documentation", Kinds: []string{"documentation"}},
+			{Title: "Flaky/Failing Tests", Kinds: []string{"failing-test", "flake"}},
+			{Title: "Other", Kinds: []string{"cleanup"}},
+		},
+	}
+
+	t := template.New("release-notes").Funcs(map[string]interface{}{
+		"FilterByLabelKinds": filterByLabelKinds,
```
tnozicka, line 88: I guess you could just create a map kinds->prs and save some logic in the template but I don't feel strong

tnozicka (commented):
Can we use some already existing commits to wire a verify/update job for the output?

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,193 @@
...
+
+	mergeBases, err := endCommit.MergeBase(startCommit)
+	if err != nil {
+		return nil, fmt.Errorf("merge base %w", err)
+	}
+	if len(mergeBases) != 1 || mergeBases[0].Hash.String() != startCommit.Hash.String() {
+		return nil, errors.New("start ref is not direct predecessor of end ref")
+	}
+
+	commitIter, err := r.Log(&git.LogOptions{From: endCommit.Hash})
```
tnozicka, line 174: is this listing it in reverse order?

pkg/cmd/utils/gen-release-notes/generate.go
```diff
@@ -0,0 +1,193 @@
...
+
+		pullRequestNumbers[n] = struct{}{}
+	}
+
+	var pullRequests []*github.PullRequest
+	page := 0
+	for len(pullRequestNumbers) != 0 {
+		prs, resp, err := ghClient.PullRequests.List(ctx, o.OrganizationName, o.RepositoryName, &github.PullRequestListOptions{
+			State: "closed",
+			ListOptions: github.ListOptions{
```
tnozicka, line 94: Would descending sorting by creation date give you better chances at going through less pages?

pkg/cmd/utils/gen-release-notes/generate.go
zimnx: Yes simillar to `git log`

pkg/cmd/utils/gen-release-notes/generate.go
zimnx: That's exactly what it does. This sorting order is by default.

zimnx: Replaced with graphql.

zimnx:
Example run: `./gen-release-notes --start-ref v1.3.0-alpha.1 --end-ref scylla/master --release-name 1.3.0-rc.0 --github-token xxx`
Example output: https://gist.github.com/zimnx/70b5594751d6bdd4df8487a80be7cdd4

pkg/cmd/utils/gen-release-notes/generate.go
tnozicka: Would it be better to sort PRs in the order they landed? Is that how kube release notes do it?

zimnx: Added sorting by mergedAt

pkg/cmd/releasenotes/options.go
```diff
@@ -0,0 +1,116 @@
...
+		)
+		httpClient = oauth2.NewClient(ctx, ts)
+	}
+
+	o.ghClient = githubql.NewClient(httpClient)
+
+	return nil
+}
+
+func (o *GenerateOptions) Run(ctx context.Context) error {
```
tnozicka, line 69: ideally this would be located in a file named `releasenotes.go` or `releasenotes_run.go` as klog prints the file name `options.go` is pretty generic

pkg/cmd/releasenotes/options.go
```diff
@@ -0,0 +1,116 @@
...
+
+	pullRequestNumbers := make(map[int]struct{})
+	for _, c := range commits {
+		if !isMergeCommit(c) {
+			continue
+		}
+
+		n, ok := parsePullRequestNumber(c.Message)
+		if !ok {
+			continue
```
tnozicka, line 83: at least an info log

pkg/cmd/releasenotes/options.go
```diff
@@ -0,0 +1,116 @@
...
+		return commits[i].Committer.When.Before(commits[j].Committer.When)
+	})
+	startDate, endDate := commits[0].Committer.When, commits[len(commits)-1].Committer.When
+
+	prs, err := listPullRequests(ctx, o.ghClient, o.OrganizationName, o.RepositoryName, startDate, endDate)
+	if err != nil {
+		return err
+	}
+
+	filteredPRs := prs[:0]
```
tnozicka, line 99: Is it taking en empty slice from an existing array? How is this better then `var filteredPRs [PullRequest]`?

pkg/cmd/releasenotes/options.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, pr := range prs {
+		if _, ok := pullRequestNumbers[int(pr.Number)]; ok {
```
tnozicka, line 101: Would it be better to filter the other way around? Making a map from GH pull request and iterating through the ones parsed from commits and looking those up in the map?

If there was a case where you have parsed a PR number which wasn't listed from GH query, this wouldn't fail while it should have.

pkg/cmd/releasenotes/options.go
```diff
@@ -0,0 +1,116 @@
...
+	return &GenerateOptions{
+		IOStreams:        streams,
+		OrganizationName: "scylladb",
+		RepositoryName:   "scylla-operator",
+		RepositoryPath:   ".",
+	}
+}
+
+func (o GenerateOptions) Validate() error {
+	if o.ReleaseName == "" {
```
tnozicka, line 42: also verify release name format - so it doesn't start with `v` as you are using it for the docker image ref

pkg/cmd/releasenotes/render.go
```diff
@@ -0,0 +1,130 @@
...
+    {{- range $pr := $section.PullRequests }}
+{{ FormatPR $pr }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+
+---
+`
+
+var branchPrefixRe = regexp.MustCompile(`^(\[[^]].*])?(.*)`)
```
tnozicka, line 35: Interesting that this works, I'd say it's missing escapes but maybe there is some rule saving it
```suggestion
var branchPrefixRe = regexp.MustCompile(`^(\[[^\]]+\])?(.*)`)
```

pkg/cmd/releasenotes/render.go
```diff
@@ -0,0 +1,130 @@
...
+
+func cleanupPRTitle(title string) (string, error) {
+	m := branchPrefixRe.FindStringSubmatch(strings.TrimSpace(title))
+	if len(m) != 3 {
+		return "", fmt.Errorf("%q PR title cannot be parsed", title)
```
tnozicka, line 40: nit `can't parse PR title %q`

pkg/cmd/releasenotes/render.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+type releaseNoteData struct {
+	Sections []section
+	Release  string
+}
+
+func renderReleaseNotes(out io.Writer, release string, pullRequests []PullRequest) error {
+	sections := []section{
+		{"API Changes", filterByLabelKinds(pullRequests, []string{"api-change"})},
```
tnozicka, line 75: If we add a new label for kind/ we might miss it here. How about creating a map from kind->[]PullRequest for all of them, then a helper map to give the categories readable names and those that don't have a readable names go into "Others"? The other option would be to fail and explicitly require the new category to be mapped into Others or given a readable name by updating the release note tool but I think "Others" would be fine.

tnozicka: uncategorized would go into others with a warning log

pkg/cmd/releasenotes/render.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func uncategorized(prs []PullRequest) []PullRequest {
+	var filtered []PullRequest
+	for _, pr := range prs {
+		if len(pr.Labels.Nodes) == 0 {
```
tnozicka, line 107: uncategorized means it doesn't have `kind/*` label, not any

pkg/cmd/releasenotes/render_test.go
```diff
@@ -0,0 +1,47 @@
...
+		t.Run(test.Name, func(t *testing.T) {
+			t.Parallel()
+
+			out, err := cleanupPRTitle(test.Title)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if out != test.ExpectedTitle {
+				t.Errorf("wrong title, got %q, expected %q", out, test.ExpectedTitle)
```
tnozicka, line 43: nit: the usual test line goes like `expected title %q, got %q`

pkg/cmd/releasenotes/options.go
zimnx: It creates slice with the same storage and It filters it in place. Doesn't require additional allocation.

pkg/cmd/releasenotes/github.go
```diff
@@ -0,0 +1,103 @@
...
+	Search struct {
+		PageInfo struct {
+			HasNextPage githubql.Boolean
+			EndCursor   githubql.String
+		}
+		Nodes []PRNode
+	} `graphql:"search(type: ISSUE, first: 100, after: $searchCursor, query: $query)"`
+}
+
+const searchTimeFormat = "2006-01-02T15:04:05Z"
```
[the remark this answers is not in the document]
zimnx: There're couple of [constants](https://golang.org/pkg/time/#pkg-constants) used there to describe format, this one already uses 24h format (`3` or `03` is for 12h, `15` for 24h). Although I changed it to RFC3339 because it is also accepted by GH.

pkg/cmd/releasenotes/helpers.go
```diff
@@ -0,0 +1,99 @@
...
+
+	mergeBases, err := endCommit.MergeBase(startCommit)
+	if err != nil {
+		return nil, fmt.Errorf("merge base %w", err)
+	}
+	if len(mergeBases) != 1 || mergeBases[0].Hash.String() != startCommit.Hash.String() {
+		return nil, errors.New("start ref is not direct predecessor of end ref")
+	}
+
+	commitIter, err := r.Log(&git.LogOptions{From: endCommit.Hash})
```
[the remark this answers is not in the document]
zimnx: It works other way around, default order is DFS from the provided commit. So we need to start at the end and go deeper in tree.

pkg/cmd/releasenotes/render.go
zimnx: Goland reports it's rendudant escaping, and tests proves that it works fine without `\`

pkg/cmd/releasenotes/options.go
zimnx: good catch, thanks!

pkg/cmd/releasenotes/options.go
zimnx: Although it's not required anymore

pkg/cmd/releasenotes/github.go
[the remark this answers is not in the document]
tnozicka: +1 for RFC3339

pkg/cmd/releasenotes/helpers.go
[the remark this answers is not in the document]
tnozicka: you are right, commits are connected through parrents, I got fixated on the previous ordering discussion

pkg/cmd/releasenotes/render.go
tnozicka: what about the rest? you are trying to match anything that isn't `]` but I read your regex as: `[` + a char that isn't `]` + any char + `]` - so why the `.*`?

pkg/cmd/releasenotes/options.go
tnozicka: sharing the storage while you are also modifying it seems error prone

pkg/cmd/releasenotes/render.go
zimnx: Right iteration is better, this is fixed in the updated version. I was reffering to escaping `]`.

tnozicka: oh, sorry, I was only responding to the comments in mail, should have checked the updated version

pkg/cmd/releasenotes/helpers_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	for i := range ts {
+		test := ts[i]
+		t.Run(test.Name, func(t *testing.T) {
+			t.Parallel()
+
+			number, ok := parsePullRequestNumber(test.Message)
+			if test.ExpectedSuccess != ok {
+				t.Errorf("Expected parsing failure, got %d", number)
```
tnozicka, line 46: the message no longer matches the condition

pkg/cmd/releasenotes/helpers.go
```diff
@@ -0,0 +1,104 @@
...
+
+	mergeBases, err := endCommit.MergeBase(startCommit)
+	if err != nil {
+		return nil, fmt.Errorf("merge base: %w", err)
+	}
+	if mergeBases[0].Hash.String() != startCommit.Hash.String() {
+		return nil, errors.New("start ref is not direct predecessor of end ref")
+	}
+	if len(mergeBases) != 1 {
+		return nil, errors.New("ref ranges containing several merge bases are not supported")
```
tnozicka, line 82: nit: `%d` to say how many

pkg/cmd/releasenotes/releasenotes.go
```diff
@@ -0,0 +1,68 @@
...
+	"k8s.io/klog/v2"
+)
+
+func (o *GenerateOptions) Run(ctx context.Context) error {
+	commits, err := listCommitsBetween(o.RepositoryPath, o.StartRef, o.EndRef)
+	if err != nil {
+		return err
+	}
+
+	requiredPullRequestsNumbers := make(map[int]struct{})
```
tnozicka, line 20: shouldn't this be just an array?

pkg/cmd/releasenotes/releasenotes.go
```diff
@@ -0,0 +1,68 @@
...
+	for i := range prs {
+		numberToPR[int(prs[i].Number)] = &prs[i]
+	}
+
+	var filteredPRs []PullRequest
+	var errs []error
+	for prNumber := range requiredPullRequestsNumbers {
+		pr, found := numberToPR[prNumber]
+		if !found {
+			errs = append(errs, fmt.Errorf("%d PR not found in GitHub", prNumber))
```
tnozicka, line 54: nit: `PR #%d`

pkg/cmd/releasenotes/releasenotes.go
```diff
@@ -0,0 +1,68 @@
...
+		} else {
```
tnozicka, line 55: `continue` after the error and you can remove the else nesting

pkg/cmd/releasenotes/render.go
```diff
@@ -0,0 +1,152 @@
...
+
+	prsByKind, uncategorized := splitByKind(pullRequests)
+
+	sections := []section{
+		{"API Changes", extractByKind(prsByKind, []string{"api-change"})},
+		{"Features", extractByKind(prsByKind, []string{"feature"})},
+		{"Bugs", extractByKind(prsByKind, []string{"bug"})},
+		{"Documentation", extractByKind(prsByKind, []string{"documentation"})},
+		{"Flaky/Failing Tests", extractByKind(prsByKind, []string{"failing-test", "flake"})},
+		{"Other", flatten(prsByKind)},
```
tnozicka, line 88: guess we should unify the singular/plural form. If we say *bugs* we should use *others* or the other way around

pkg/cmd/releasenotes/render.go
```diff
@@ -0,0 +1,152 @@
...
+
+func flatten(m map[string][]PullRequest) []PullRequest {
+	var prs []PullRequest
+	for _, v := range m {
+		prs = append(prs, v...)
+	}
+	return prs
+}
+
+func extractByKind(prsByKind map[string][]PullRequest, kinds []string) []PullRequest {
```
tnozicka, line 120: tiny nit: guess `kind ...string` would make the call site nicer

tnozicka (commented):
thanks for the updates, some last nits but otherwise you are ready to land

pkg/cmd/releasenotes/releasenotes.go
```diff
@@ -0,0 +1,69 @@
...
+	for i := range prs {
+		numberToPR[int(prs[i].Number)] = &prs[i]
+	}
+
+	var filteredPRs []PullRequest
+	var errs []error
+	for prNumber := range requiredPullRequestsNumbers {
+		pr, found := numberToPR[prNumber]
+		if !found {
+			errs = append(errs, fmt.Errorf("#%d PR not found in GitHub", prNumber))
```
tnozicka, line 54: Can you switch the order as I've suggested before? `PR #5 not found` sounds more like a sentence compared to  `#5 PR not found`.

pkg/cmd/releasenotes/helpers_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	for i := range ts {
+		test := ts[i]
+		t.Run(test.Name, func(t *testing.T) {
+			t.Parallel()
+
+			number, ok := parsePullRequestNumber(test.Message)
+			if test.ExpectedSuccess != ok {
+				t.Errorf("Expected sucess:%v, got result:%v", test.ExpectedSuccess, ok)
```
tnozicka, line 46: spelling nit: success

tnozicka:
thanks, please fix the spelling and this is ready to land. good work!

tnozicka (approved):
/lgtm
thanks!

PR #573, opened 2021-04-29
sidecar-injection: avoid -EBUSY on cp -a of scylla-operator

kjetijor:

sidecar attempts to copy scylla-operator binary into shared, if a process of scylla-operator is already running with that binary, this will fail with with -EBUSY. Work around by letting cp retry on failure, with unlink and retry.

```
cp: cannot create regular file '/mnt/shared/scylla-operator': Text file busy
```

Resolves #N/A

### DIFF
--- a/pkg/controllers/cluster/resource/resource.go
+++ b/pkg/controllers/cluster/resource/resource.go
@@ -201,7 +201,7 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, sidecarI
 							Command: []string{
 								"/bin/sh",
 								"-c",
-								fmt.Sprintf("cp -a /usr/bin/scylla-operator %s", naming.SharedDirName),
+								fmt.Sprintf("cp -fa /usr/bin/scylla-operator %s", naming.SharedDirName),
 							},
 							Resources: corev1.ResourceRequirements{
 								Requests: corev1.ResourceList{


zimnx:
According to [k8s documentation](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#understanding-init-containers), containers are started only after every initContainer is completed. 
```
A Pod can have multiple containers running apps within it, but it can also have one or more init containers, which are run before the app containers are started.
[...]
Init containers always run to completion.
```
I wonder how it can be busy. Are you abel to reproduce it and collect more data?

kjetijor:
> I wonder how it can be busy. Are you abel to reproduce it and collect more data to better understand the root cause?

I suspect that the initContainer resource have somehow ended up updated.

The state I'm ending up in is:
 - scylla and scylla-manager container is running fine
 - the sidecar injection initcontainer is crash-loop-restarting

I delete the pod, have it get re-created, it looks fine, "some hours later, I'm back in the same situation.

Now - I'll make an argument that regardless of how I end up in this situation - you'll still need this, given that you *do* update the initContainer resource in the ClusterReconciler.

kjetijor:
So - I suspect what's happening is that the reconciler runs, but makes no changes. Essentially the STS diff is the version number & date.

zimnx:
Could you submit logs from when this restart happens to you? We are updating initCointainer only when sidecar image update is required, which should happen only during operator version upgrade. Empty updates of pod template shouldn't result in pod restarts, at least in newer version of k8s. 

In general I still don't understand how this file can be busy. Even when pod template in sts is changed, sts controller first tears down the pod, and when all containers are down, it recreates the pod. So none container should be running when initContainer is executed. 


Which kubernetes distribution and which version are you using?

kjetijor:
@zimnx Thanks for playing. I'm going to close this and go speak to my fabric owner, there's a "docker garbage collector", removing the exit-0 init container, which has kubernetes go say - hey - I didn't run my init container, let's run it.  Whic is "not your problem(tm)", and all kinds of potentially bad.

PR #574, opened 2021-04-30
Add optional SecurityContext to Helm Charts

### LINKED ISSUE #570: Run containers (operator/controller/manager/member) as non root
* Feature Request

Update the helm charts for the operator/manager/cluster to deploy the statefulsets/deployments to run as a non root user using a security context like for example:

```
securityContext:
  runAsUser: 999
```

We use a [Lokomotive Kubernetes](https://github.com/kinvolk/lokomotive) distribution which likes to promote a security approach of least privilege, so by default, the containers aren't allowed to run as root. 
I'm not familiar with the details of the deployment but it would be a good approach to run the containers as a specific user instead of root.

**Additional Information:**

Currently, without defining additional permission to the default set of rules in the ClusterRole the pods won't be able to start because they lack the permissions to start the pods as root user.
```
Error: container has runAsNonRoot and image will run as root
```

niels-s:
This gives the option to pass an optional `securityContext` to the
operator, manager, and manager controller deployments so for example the
containers can be run using as a non-root user.

Resolves #570

### DIFF
--- a/helm/scylla-manager/templates/controller_deployment.yaml
+++ b/helm/scylla-manager/templates/controller_deployment.yaml
@@ -18,6 +18,9 @@ spec:
         {{- include "scylla-manager.controllerLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "scylla-manager.controllerServiceAccountName" . }}
+      {{- with .Values.controllerSecurityContext }}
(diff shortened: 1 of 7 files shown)


tnozicka:
thanks, structurally this lgtm

zimnx (approved):
LGTM, thanks

Issue #507, opened 2021-03-17
Automatically replace scylla node that looses storage


tnozicka:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
When a node is decommissioned, the local storage is lost as well. Currently it requires a manual action by annotating the service to trigger replacement, otherwise the new pod is stuck on join as it doesn't have replace-address-first-boot set.

**What is use case behind this feature:**
Stability - operator should be able to run scylla without any user intervention.

**Additional Information:**
One option is to write down a file in an init container, check its presents and generate the replace-address-first-boot. Maybe there is a more sophisticated way to get the same information directly from scylla.

tnozicka:
The case of a kubernetes node being decommissioned is covered by `AutomaticOrphanedNodeCleanup` although there is a race if the scylla node wouldn't be bootstrapped yet which would get stuck on scylla not replacing a node that's not in gossip.

There are other cases though AWS EC2 instances loose local disks when the instance is stopped, hibernated or a local drive fails. Same goes for GCP. Those won't result in a Kubernetes node removal so `AutomaticOrphanedNodeCleanup` won't help here and it can get stuck, needed a manual intervention to trigger a node replacement.

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2469.

PR #584, opened 2021-05-07
Document manual seed node replacement procedure

zimnx:


docs/source/nodeoperations/replace_node.md
```diff
@@ -68,4 +70,37 @@ _This procedure is for replacing one dead node. To replace more than one dead no
...
+
+1. Mark pod with replace label to tell Operator that this node must be replaced.
+    ```
+    kubectl -n scylla label svc <pod-name> scylla/replace=""
+    ```
+   Operator will remove and recreate pod, service and pvc. Seed nodes are hardcoded (first two pods in rack), 
+   the recreated service will contain a seed label.
+   This causes Scylla to fail to start because seed nodes cannot be replaced.
+
+1. Remove the seed label again, and manually delete the pod, in order to allow the sidecar to pick up the state we want.
```
tnozicka, line 91: Wonder if we could avoid the double unlabel + double delete with:
1. degrade/unlabel
2. delete the PVC
3. delete the Pod
4. mark it as a seed again
5. rolling restart

zimnx: We would have to have #507, because SVC would keep the same IP and new (empty) storage would be attached.

docs/source/nodeoperations/scylla_upgrade.md
```diff
@@ -89,14 +89,14 @@ Upgrade may get stuck on `validate_upgrade` stage. This happens when Scylla Pod
 
 To continue with upgrade, first turn off operator by scaling Operator replicas to zero:
 ```bash
-kubectl -n scylla-operator-system scale sts scylla-operator-controller-manager --replicas=0
+kubectl -n scylla-operator scale deployment.apps/scylla-operator --replicas=0
 ```
 Then user have to manually resolve issue with Scylla by checking what is the root cause of a failure in Scylla container logs.
 If needed data and system keyspaces SSTable snapshots are available on the node. You can check ScyllaCluster status for their names.
 
 Once issue is resolved and Scylla Pod is up and running (Pod is in Ready state), scale Operator back to one replica:
```
tnozicka, line 97: 2 replicas

tnozicka (approved):
/lgtm
thanks

PR #597, opened 2021-05-11
Implement seedless mode

### LINKED ISSUE #292: Two first nodes of every rack are seed ones
**Describe the bug**
Currently we have following logic on setting seed flag on the nodes:
```
	if strings.HasSuffix(pod.Name, "-0") || strings.HasSuffix(pod.Name, "-1") {
		labels[naming.SeedLabel] = ""
	}
```

It leads to marking two first nodes of every rack as seed nodes.
As result you can't replace them.

### LINKED ISSUE #295: Nodes are getting different list of seeds
**Describe the bug**
Currently we gather seed information from services when sidecar is startingup, problem with this approach is that if service for any seed is not yet created you endup with different list of seeds on those seeds that were booted/initialized before others.
For example in three node cluster, which by current logic ending up with two seeds, on first and seconds node, can have scylal running with following seed configuration:

```
node1 - /usr/bin/scylla ... --seed-provider-parameters seeds=10.105.30.69 --broadcast-address 10.105.30.69
node2 - /usr/bin/scylla ... --seed-provider-parameters seeds=10.105.30.69,10.105.39.190 --broadcast-address 10.105.39.190
```
Which means that first node knows only it self as a seed, while second one know both seeds.

Less pron to happen on GKE, probably due to the more time to bring node up.

**To Reproduce**
1. Deploy scylla operator
2. Deploy scylla cluster of 3 nodes
3. Check seeds on 1 node: kubectl exec -ti -n scylla simple-cluster-us-east-1-us-east-1a-0 -- ps aux | grep scylla
4. Check seeds on 2 node: kubectl exec -ti -n scylla simple-cluster-us-east-1-us-east-1a-1 -- ps aux | grep scylla
5. Check seeds on 3 node: kubectl exec -ti -n scylla simple-cluster-us-east-1-us-east-1a-2 -- ps aux | grep scylla

**Expected behavior**
Seed list is the same on all nodes

### LINKED ISSUE #405: Replace a seed node

zimnx:
From Scylla 4.3 seed concept is relaxed, and any node can be provided as
contact point during bootstrap.
From now on, sidecar will pick any UN node as a contact point if any is available,
itself when there's only single Pod, or first ever created Pod when all nodes
are down.

Fixes #292
Fixes #295
Fixes #405

### DIFF
--- a/pkg/sidecar/identity/member_test.go
+++ b/pkg/sidecar/identity/member_test.go
@@ -1,0 +1,133 @@
+// Copyright (C) 2021 ScyllaDB
+
+package identity
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/scylladb/scylla-operator/pkg/controller/helpers"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestMember_GetSeeds(t *testing.T) {
+	createPodAndSvc := func(name, ip string, creationTimestamp time.Time) (*corev1.Pod, *corev1.Service) {
+		pod := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: "namespace",
+				Labels: map[string]string{
+					naming.ClusterNameLabel: "my-cluster",
+				},
+				CreationTimestamp: metav1.NewTime(creationTimestamp),
+			},
+		}
+		svc := &corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: "namespace",
+			},
+			Spec: corev1.ServiceSpec{
+				ClusterIP: ip,
+			},
+		}
+		return pod, svc
+	}
+
+	now := time.Now()
+	firstPod, firstService := createPodAndSvc("pod-0", "1.1.1.1", now)
+	secondPod, secondService := createPodAndSvc("pod-1", "2.2.2.2", now.Add(time.Second))
+	thirdPod, thirdService := createPodAndSvc("pod-2", "3.3.3.3", now.Add(2*time.Second))
+
+	ts := []struct {
+		name        string
+		memberName  string
+		memberIP    string
+		objects     []runtime.Object
+		expectSeed  string
+		expectError error
+	}{
+		{
+			name:        "error when no pods are found",
+			memberName:  firstPod.Name,
+			memberIP:    firstService.Spec.ClusterIP,
+			objects:     []runtime.Object{},
+			expectError: fmt.Errorf("internal error: can't find any pod for this cluster, including itself"),
+		},
+		{
+			name:       "bootstraps with itself when cluster is empty",
+			memberName: firstPod.Name,
+			memberIP:   firstService.Spec.ClusterIP,
+			objects:    []runtime.Object{firstPod, firstService},
+			expectSeed: firstService.Spec.ClusterIP,
+		},
+		{
+			name:       "bootstrap with first created UN node",
+			memberName: firstPod.Name,
+			memberIP:   firstService.Spec.ClusterIP,
+			objects:    []runtime.Object{firstPod, firstService, markPodReady(secondPod), secondService, markPodReady(thirdPod), thirdService},
+			expectSeed: secondService.Spec.ClusterIP,
+		},
+		{
+			name:       "bootstrap only with UN node",
+			memberName: firstPod.Name,
+			memberIP:   firstService.Spec.ClusterIP,
+			objects:    []runtime.Object{firstPod, firstService, secondPod, secondService, markPodReady(thirdPod), thirdService},
+			expectSeed: thirdService.Spec.ClusterIP,
+		},
+		{
+			name:       "bootstrap with first created Pod when all are down",
+			memberName: firstPod.Name,
+			memberIP:   firstService.Spec.ClusterIP,
+			objects:    []runtime.Object{firstPod, firstService, secondPod, secondService, thirdPod, thirdService},
+			expectSeed: secondService.Spec.ClusterIP,
+		},
+	}
+
+	for i := range ts {
+		test := ts[i]
+		t.Run(test.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			member := Member{
+				Cluster:   "my-cluster",
+				Namespace: "namespace",
+				Name:      test.memberName,
+				StaticIP:  test.memberIP,
+			}
+
+			fakeClient := fake.NewSimpleClientset(test.objects...)
+			seed, err := member.GetSeed(ctx, fakeClient.CoreV1())
+			if !reflect.DeepEqual(err, test.expectError) {
+				t.Errorf("expected error %v, got %v", test.expectError, err)
+			}
+			if seed != test.expectSeed {
+				t.Errorf("expected seed %v, got %v", test.expectSeed, seed)
+			}
+		})
+	}
+}
+
+func markPodReady(pod *corev1.Pod) *corev1.Pod {
+	p := pod.DeepCopy()
+	cond := helpers.GetPodCondition(p.Status.Conditions, corev1.PodReady)
+	if cond != nil {
+		cond.Status = corev1.ConditionTrue
+		return p
+	}
+
+	p.Status.Conditions = append(p.Status.Conditions, corev1.PodCondition{
+		Type:   corev1.PodReady,
+		Status: corev1.ConditionTrue,
+	})
+
+	return p
+}
--- a/test/e2e/set/scyllacluster/scyllacluster_replace.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_replace.go
@@ -22,12 +22,12 @@ var _ = g.Describe("ScyllaCluster replace", func() {
 
 	f := framework.NewFramework("scyllacluster")
 
-	g.It("should replace a non-seed node", func() {
+	g.It("should replace a node", func() {
 		ctx, cancel := context.WithTimeout(context.Background(), testTimout)
 		defer cancel()
 
 		sc := scyllaclusterfixture.BasicScyllaCluster.ReadOrFail()
-		sc.Spec.Datacenter.Racks[0].Members = 3
+		sc.Spec.Datacenter.Racks[0].Members = 2
 
 		framework.By("Creating a ScyllaCluster")
 		err := framework.SetupScyllaClusterSA(ctx, f.KubeClient().CoreV1(), f.KubeClient().RbacV1(), f.Namespace(), sc.Name)
@@ -44,10 +44,10 @@ var _ = g.Describe("ScyllaCluster replace", func() {
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
 
-		framework.By("Replacing a non-seed node")
+		framework.By("Replacing a node #0")
 		pod, err := f.KubeClient().CoreV1().Pods(f.Namespace()).Get(
 			ctx,
-			getFirstNonSeedNodeName(sc),
+			getNodeName(sc, 0),
 			metav1.GetOptions{},
 		)
 		o.Expect(err).NotTo(o.HaveOccurred())
--- a/pkg/controller/scyllacluster/resource/resource.go
+++ b/pkg/controller/scyllacluster/resource/resource.go
@@ -54,11 +54,6 @@ func MemberService(sc *scyllav1.ScyllaCluster, rackName, name string, oldService
 	labels[naming.DatacenterNameLabel] = sc.Spec.Datacenter.Name
 	labels[naming.RackNameLabel] = rackName
 
-	// If Member is seed, add the appropriate label
-	if strings.HasSuffix(name, "-0") || strings.HasSuffix(name, "-1") {
-		labels[naming.SeedLabel] = ""
-	}
-
 	// Copy the old replace label, if present.
 	var replaceAddr string
 	var hasReplaceLabel bool
--- a/pkg/controller/scyllacluster/sync_services.go
+++ b/pkg/controller/scyllacluster/sync_services.go
@@ -138,15 +138,6 @@ func (scc *Controller) syncServices(
 			continue
 		}
 
-		_, isSeed := svc.Labels[naming.SeedLabel]
-		if isSeed {
-			klog.ErrorS(nil, "Seed node replace is not supported",
-				"ScyllaCluster", klog.KObj(sc),
-				"Service", klog.KObj(svc),
-			)
-			continue
-		}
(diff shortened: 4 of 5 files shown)


pkg/controllers/sidecar/identity/member.go
```diff
@@ -63,27 +66,43 @@ func Retrieve(ctx context.Context, name, namespace string, kubeclient kubernetes
 	}, nil
 }
 
-func (m *Member) GetSeeds(ctx context.Context, kubeClient kubernetes.Interface) ([]string, error) {
-	var services *corev1.ServiceList
-	var err error
+// GetSeed return a single node contact point used for bootstrapping.
```
tnozicka, line 69: nit: s/return/returns

pkg/controllers/sidecar/identity/member.go
```diff
@@ -63,27 +66,43 @@ func Retrieve(ctx context.Context, name, namespace string, kubeclient kubernetes
...
+// It is enough to return a single node because bootstrap process will fail if any of the nodes in cluster is down.
```
tnozicka, line 70: nit: `A single node is enough because ...`

pkg/controllers/sidecar/identity/member.go
```diff
@@ -63,27 +66,43 @@ func Retrieve(ctx context.Context, name, namespace string, kubeclient kubernetes
...
+func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (string, error) {
```
tnozicka, line 71: If all nodes have to be UN for a new one to join, would it work if we just return the fixed DNS name for first pod in the statefulset?
If it has to be a fixed "external" IP would a fixed service for the first node work?

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -47,7 +47,7 @@ var _ = g.Describe("ScyllaCluster replace", func() {
 		framework.By("Replacing a non-seed node")
 		pod, err := f.KubeClient().CoreV1().Pods(f.Namespace()).Get(
 			ctx,
-			getFirstNonSeedNodeName(sc),
+			getNodeName(sc, podIdx),
```
tnozicka, line 50: Can we just replace those nodes in sequence instead of running the whole setup again?

pkg/controllers/sidecar/identity/member.go
zimnx: That's what I tried at the beginning, but it doesn't work. DNS resolves to PodIP on the same pod, and to service clusterIP on other pods. Scylla cannot bootstrap using PodIP, it has to use one of the node broadcast address. 
Fixed IP also means that fixed node is not going to be replacable since it would try to connect to itself. 
To avoid those problems why I pick up any UN node, or itself (serviceIP) if it is the first node in the cluster.

tnozicka: hm, the service DNS resolves to ClusterIP for me
```
[root@basic-us-east-1-us-east-1a-0 /]# ping basic-us-east-1-us-east-1a-0.test.svc
PING basic-us-east-1-us-east-1a-0.test.svc.cluster.local (10.97.130.36) 56(84) bytes of data.
...
[root@basic-us-east-1-us-east-1a-0 /]# ip a
...
12: eth0@if13: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:11:00:06 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.17.0.6/16 brd 172.17.255.255 scope global eth0
```
you are right about replacing itself - i guess we could just get all of the node service names there

pkg/controllers/sidecar/identity/member.go
```diff
@@ -63,27 +66,43 @@ func Retrieve(ctx context.Context, name, namespace string, kubeclient kubernetes
...
+			naming.ClusterNameLabel: m.Cluster,
+		}).String(),
+	})
+	if err != nil {
+		return "", err
+	}
 
-	sel := fmt.Sprintf("%s,%s=%s", naming.SeedLabel, naming.ClusterNameLabel, m.Cluster)
+	// Return static IP of itself to bootstrap first node.
+	if len(pods.Items) == 1 && pods.Items[0].Name == m.Name {
```
tnozicka, line 82: if a pod is evicted in the meantime, it won't be listed here - that doesn't meant it's alone. I suppose the desired member count from ScyllaCluster would be more fitting

pkg/controllers/sidecar/identity/member.go
```diff
@@ -63,27 +66,43 @@ func Retrieve(ctx context.Context, name, namespace string, kubeclient kubernetes
...
-	for retryCount := 0; ; retryCount++ {
-		services, err = kubeClient.CoreV1().Services(m.Namespace).List(ctx, metav1.ListOptions{LabelSelector: sel})
-		if err == nil && len(services.Items) > 0 {
-			break
-		}
-		if retryCount > 5 {
-			return nil, errors.New(fmt.Sprintf("failed to get seeds, error: %+v, len(services): %d", err, len(services.Items)))
+	// If at least one pod is ready, return the IP of service.
+	for _, pod := range pods.Items {
+		if podutils.IsPodReady(&pod) && pod.Name != m.Name {
```
tnozicka, line 88: Pod can go unready just after this line so you can still potentially give a service to unready. Would be better to give scylla a bigger list of seeds and let it fallback to another one if it can't connect?

pkg/controllers/sidecar/identity/member.go
```diff
@@ -63,27 +66,43 @@ func Retrieve(ctx context.Context, name, namespace string, kubeclient kubernetes
...
+			return svc.Spec.ClusterIP, nil
 		}
-		time.Sleep(1000 * time.Millisecond)
 	}
 
-	seeds := []string{}
-	for _, svc := range services.Items {
-		seeds = append(seeds, svc.Spec.ClusterIP)
+	// When all nodes are down, return service IP of the first created pod.
+	sort.Slice(pods.Items, func(i, j int) bool {
```
tnozicka, line 98: this can return a service for the same pod, right?

zimnx: Yes. Although i don't think it matters. New pods are not scheduled when any of the previous nodes is down. And when node is already boostraped, seed list doesn't matter.
@asias Could you confirm the latter?

zimnx:
Rebased on top of #534

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,108 @@
...
+		t.Run(test.Name, func(t *testing.T) {
+			member := Member{
+				Cluster:   "my-cluster",
+				Namespace: "namespace",
+				Name:      test.MemberName,
+				StaticIP:  test.MemberIP,
+			}
+
+			clientFake := fake.NewSimpleClientset(test.Objects...)
+			seed, err := member.GetSeed(context.Background(), clientFake.CoreV1())
```
tnozicka, line 84: nit: cancelable context for each run with defer

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			if seed != test.ExpectedSeed {
+				t.Fatalf("Expected %v seed, got %v", test.ExpectedSeed, seed)
```
tnozicka, line 89:
```suggestion
				t.Errorf("Expected seed %v, got %v", test.ExpectedSeed, seed)
```

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,108 @@
...
+			Objects:      []runtime.Object{firstPod, firstService, markPodReady(secondPod), secondService},
+			ExpectedSeed: secondService.Spec.ClusterIP,
+		},
+		{
+			Name:         "bootstrap with first created Pod when all are down",
+			MemberName:   firstPod.Name,
+			Objects:      []runtime.Object{firstPod, firstService, secondPod, secondService},
+			ExpectedSeed: secondService.Spec.ClusterIP,
+		},
+	}
```
tnozicka, line 71: test case: no pod listed => errror

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,108 @@
...
+			Name:         "bootstraps with itself when cluster is empty",
+			MemberName:   firstPod.Name,
+			MemberIP:     firstService.Spec.ClusterIP,
+			Objects:      []runtime.Object{firstPod, firstService},
+			ExpectedSeed: firstService.Spec.ClusterIP,
+		},
+		{
+			Name:         "bootstrap with any UN node",
+			MemberName:   firstPod.Name,
+			Objects:      []runtime.Object{firstPod, firstService, markPodReady(secondPod), secondService},
```
tnozicka, line 62: nit: might be best to always set the MemberIP

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestMember_GetSeeds(t *testing.T) {
+	createPodAndSvc := func(name, ip string, creationTimestamp time.Time) (*corev1.Pod, *corev1.Service) {
```
tnozicka, line 18: nit: separate functions might be cleaner if the args are not connected

pkg/sidecar/identity/member_test.go
tnozicka: don't you need at least 3 pods to test the sorting since it sorts only the "otherPods"

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if seed != test.ExpectedSeed {
+				t.Fatalf("Expected %v seed, got %v", test.ExpectedSeed, seed)
+			}
+		})
+	}
+}
+
+func markPodReady(pod *corev1.Pod) *corev1.Pod {
+	p := pod.DeepCopy()
+	for i := range p.Status.Conditions {
```
tnozicka, line 97: `GetPodCondition`

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -44,44 +44,46 @@ var _ = g.Describe("ScyllaCluster replace", func() {
...
-
-		framework.By("Waiting for the ScyllaCluster to re-deploy")
-		waitCtx3, waitCtx3Cancel := contextForRollout(ctx, sc)
-		defer waitCtx3Cancel()
-		sc, err = waitForScyllaClusterState(waitCtx3, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, scyllaClusterRolledOut)
-		o.Expect(err).NotTo(o.HaveOccurred())
-
-		verifyScyllaCluster(ctx, f.KubeClient(), sc)
+		for podIdx := 0; podIdx < int(sc.Spec.Datacenter.Racks[0].Members); podIdx++ {
+			framework.By("Replacing a %d node", podIdx)
```
tnozicka, line 48:
```suggestion
			framework.By("Replacing a node #%d", podIdx)
```

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -44,44 +44,46 @@ var _ = g.Describe("ScyllaCluster replace", func() {
...
-		time.Sleep(10 * time.Second)
-
-		framework.By("Waiting for the ScyllaCluster to re-deploy")
-		waitCtx3, waitCtx3Cancel := contextForRollout(ctx, sc)
-		defer waitCtx3Cancel()
-		sc, err = waitForScyllaClusterState(waitCtx3, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, scyllaClusterRolledOut)
-		o.Expect(err).NotTo(o.HaveOccurred())
-
-		verifyScyllaCluster(ctx, f.KubeClient(), sc)
+		for podIdx := 0; podIdx < int(sc.Spec.Datacenter.Racks[0].Members); podIdx++ {
```
tnozicka, line 47: wonder if it would be enough to replace node `#0`, both used to be seed nodes

pkg/sidecar/identity/member_test.go
zimnx: they do share the name, it's a private function only in this test so i'll leave it as it is.

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,126 @@
...
+	secondPod, secondService := createPodAndSvc("pod-1", "2.2.2.2", now.Add(time.Second))
+	thirdPod, thirdService := createPodAndSvc("pod-2", "3.3.3.3", now.Add(2*time.Second))
+
+	ts := []struct {
+		name        string
+		memberName  string
+		memberIP    string
+		objects     []runtime.Object
+		expectSeed  string
+		expectError bool
```
tnozicka, line 53: nit: make it `error` and DeepEqual the expected value

pkg/sidecar/identity/member_test.go
tnozicka: test case: second pod is not ready but third is ready => third pod

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,126 @@
...
+			defer cancel()
+
+			member := Member{
+				Cluster:   "my-cluster",
+				Namespace: "namespace",
+				Name:      test.memberName,
+				StaticIP:  test.memberIP,
+			}
+
+			clientFake := fake.NewSimpleClientset(test.objects...)
```
tnozicka, line 98: nit: usually called `fakeClient`

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+		})
+	}
+}
+
+func markPodReady(pod *corev1.Pod) *corev1.Pod {
+	p := pod.DeepCopy()
+	cond := helpers.GetPodCondition(p.Status.Conditions, corev1.PodReady)
+	if cond != nil {
+		cond.Status = corev1.ConditionTrue
```
tnozicka, line 117: nit: return and you avoid the else block

pkg/sidecar/identity/member_test.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			fakeClient := fake.NewSimpleClientset(test.objects...)
+			seed, err := member.GetSeed(ctx, fakeClient.CoreV1())
+			if err == nil && test.expectError != nil {
+				t.Errorf("expected error but got none")
+			}
+			if err != nil && test.expectError == nil {
+				t.Errorf("did not expect error but got: %v", err)
+			}
```
tnozicka, line 113:
```suggestion
			if !reflect.DeepEqual(err, test.expectError) {
				t.Errorf("expected error %v, got %v", , test.expectError, err)
			}
```

tnozicka (commented):
last nit, lgtm otherwise

tnozicka (approved):

tnozicka:
/lgtm
thanks!

PR #599, opened 2021-05-18
Added webhook cleanup step to upgrade procedure

zimnx:


docs/source/upgrade.md
```diff
@@ -17,6 +17,11 @@ Sidecar image is going to be upgraded automatically, so a rolling restart of you
     ```
     kubectl delete namespace scylla-operator-system --wait=true
     ```
+1. Remove old webhooks:
+    ```
+    kubectl delete MutatingWebhookConfiguration scylla-operator-mutating-webhook-configuration
+    kubectl delete ValidatingWebhookConfiguration scylla-operator-validating-webhook-configuration
```
tnozicka, line 23: We still have a validating webhook, right?

zimnx: We do, but we use different name for the object.

tnozicka: you are right, I didn't check the name

tnozicka (dismissed):
/lgtm

tnozicka (approved):

PR #603, opened 2021-05-20
Use REST API for schema agreement check

zimnx:
CQL may require authentication, and there's no way to provide
credentials to Operator. For schema agreement we can consult REST API
where authentication will be managed by the Operator.


pkg/scyllaclient/client.go
```diff
@@ -288,6 +288,19 @@ func (c *Client) IsNativeTransportEnabled(ctx context.Context, host string) (boo
 	return resp.Payload, nil
 }
 
+func (c *Client) SchemaAgreement(ctx context.Context) (bool, error) {
```
tnozicka, line 291: nit: start with a verb like `HasSchemaAgreement` so the caller can better identify what it does and what the returned bool means

tnozicka:
you need `go mod tidy && go mod vendor` https://github.com/scylladb/scylla-operator/pull/603/checks?check_run_id=2631264947#step:4:168

tnozicka (approved):
/lgtm
thanks!

PR #605, opened 2021-05-24
Provision autogenerated auth token for Scylla API 

zimnx:
From now on Agent requires a proper authorization token for requests.
Token is provisioned by the Operator as a Secret for each ScyllaCluster.
If user already enabled authentication by filling it in optional Agent
config Secret, its value will be used, if not, a random autogenerated string will be saved.

### DIFF
--- a/pkg/controllers/sidecar/sidecar_controller.go
+++ b/pkg/controllers/sidecar/sidecar_controller.go
@@ -18,11 +18,13 @@ package sidecar
 import (
 	"context"
 	"fmt"
+	"path"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/scylladb/go-log"
 	"github.com/scylladb/scylla-operator/pkg/cmd/operator/options"
+	"github.com/scylladb/scylla-operator/pkg/controllers/helpers"
 	"github.com/scylladb/scylla-operator/pkg/controllers/sidecar/config"
 	"github.com/scylladb/scylla-operator/pkg/controllers/sidecar/identity"
 	"github.com/scylladb/scylla-operator/pkg/naming"
@@ -106,10 +108,18 @@ func New(ctx context.Context, mgr manager.Manager, logger log.Logger) (*MemberRe
 		return nil, errors.Wrap(err, "get scylla address")
 	}
 
-	cfg := scyllaclient.DefaultConfig(host.String())
-	if err := cfgutil.ParseYAML(&cfg, naming.ScyllaClientConfigDirName+"/"+naming.ScyllaClientConfigFileName); err != nil {
+	authToken, err := helpers.GetAgentAuthToken(ctx, kubeClient.CoreV1(), member.Cluster, member.Namespace)
+	if err != nil {
+		return nil, errors.Wrap(err, "get auth token")
+	}
+
+	cfg := scyllaclient.DefaultConfig(authToken, host.String())
+	if err := cfgutil.ParseYAML(&cfg, path.Join(naming.ScyllaClientConfigDirName, naming.ScyllaClientConfigFileName)); err != nil {
 		return nil, errors.Wrap(err, "parse scylla agent config")
 	}
+	// Auth token from secret is preferred.
+	cfg.AuthToken = authToken
+
 	scyllaClient, err := scyllaclient.NewClient(cfg, logger.Named("scylla_client"))
 	if err != nil {
 		return nil, errors.Wrap(err, "create scylla client")
--- a/test/e2e/set/scyllacluster/scyllacluster_auth.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_auth.go
@@ -1,0 +1,155 @@
+// Copyright (C) 2021 ScyllaDB
+
+package scyllacluster
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	"github.com/scylladb/go-log"
+	"github.com/scylladb/scylla-operator/pkg/controllers/helpers"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/scyllaclient"
+	scyllaclusterfixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scyllacluster"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var _ = g.Describe("ScyllaCluster authentication", func() {
+	defer g.GinkgoRecover()
+
+	f := framework.NewFramework("scyllacluster")
+
+	g.It("agent requires authentication", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimout)
+		defer cancel()
+
+		sc := scyllaclusterfixture.BasicScyllaCluster.ReadOrFail()
+		sc.Spec.Datacenter.Racks[0].Members = 1
+
+		framework.By("Creating a ScyllaCluster")
+		err := framework.SetupScyllaClusterSA(ctx, f.KubeClient().CoreV1(), f.KubeClient().RbacV1(), f.Namespace(), sc.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to deploy")
+		waitCtx1, waitCtx1Cancel := contextForRollout(ctx, sc)
+		defer waitCtx1Cancel()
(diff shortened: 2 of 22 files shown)


test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,59 @@
...
+
+		framework.By("Waiting for the ScyllaCluster to deploy")
+		waitCtx1, waitCtx1Cancel := contextForRollout(ctx, sc)
+		defer waitCtx1Cancel()
+		sc, err = waitForScyllaClusterState(waitCtx1, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, scyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f.KubeClient(), sc)
+
+		framework.By("Unauthorized request is rejected")
```
tnozicka, line 45: nit:  by `Rejecting an unauthorized request`

pkg/controllers/cluster/resource/authtoken.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+		Type:      corev1.SecretTypeOpaque,
+		Immutable: pointer.BoolPtr(true),
+		Data: map[string][]byte{
+			naming.ScyllaAgentAuthTokenFileName: data,
+		},
+	}, nil
+}
+
+func GetAgentAuthToken(ctx context.Context, client corev1client.CoreV1Interface, clusterName, namespace string) (string, error) {
```
tnozicka, line 41: nit: I think getters should go to controller/helpers

pkg/controllers/cluster/resource/authtoken.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, fmt.Errorf("marshal auth token: %w", err)
+	}
+
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      naming.AgentAuthTokenSecretName(c.Name),
+			Namespace: c.Namespace,
+		},
+		Type:      corev1.SecretTypeOpaque,
+		Immutable: pointer.BoolPtr(true),
```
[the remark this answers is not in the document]
tnozicka: or at some point it may need to be rotated

tnozicka (commented):
The other option would be to merge agent secret and the user config secret in the operator and create 1 secret to mount into the pod. This would also be resistant to an unparsable user config and the token secret wouldn't have to marshal and could be a plain field. Both options look good though.

pkg/controllers/sidecar/sidecar_controller.go
```diff
@@ -106,10 +108,18 @@ func New(ctx context.Context, mgr manager.Manager, logger log.Logger) (*MemberRe
...
 
-	cfg := scyllaclient.DefaultConfig(host.String())
-	if err := cfgutil.ParseYAML(&cfg, naming.ScyllaClientConfigDirName+"/"+naming.ScyllaClientConfigFileName); err != nil {
+	authToken, err := resource.GetAgentAuthToken(ctx, kubeClient.CoreV1(), member.Cluster, member.Namespace)
+	if err != nil {
+		return nil, errors.Wrap(err, "get auth token")
+	}
+
+	cfg := scyllaclient.DefaultConfig(authToken, host.String())
+	if err := cfgutil.ParseYAML(&cfg, path.Join(naming.ScyllaClientConfigDirName, naming.ScyllaClientConfigFileName)); err != nil {
```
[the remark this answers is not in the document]
zimnx: right now it may only contain custom timeout. It's optional and ParseYAML function handles missing file. It's kinda error prone

tnozicka (dismissed):
pls add the e2e and you are good to go

one nit are the error messages - when printed in logs `can't get a token from secret: <context>` gives a better hint than `get token from secret: <context>`. `can't` is slightly better then `failed to` because it doesn't rule out a retry

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,59 @@
...
+		authToken := ""
+		cfg := scyllaclient.DefaultConfig(authToken, hosts...)
+
+		client, err := scyllaclient.NewClient(cfg, log.NopLogger)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		_, err = client.Status(ctx, hosts[0])
+		o.Expect(err).To(o.HaveOccurred())
+		o.Expect(scyllaclient.StatusCodeOf(err)).To(o.Equal(http.StatusUnauthorized))
+	})
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,163 @@
...
+
+		gen, err = restartRack(ctx, f.KubeClient().AppsV1(), sc.Spec.Datacenter.Racks[0], sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to pick up token change")
+
+		rackGenerationUpgraded = func(sc *scyllav1.ScyllaCluster) (bool, error) {
+			stss, err := getStatefulSetsForScyllaCluster(ctx, f.KubeClient().AppsV1(), sc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			return stss[sc.Spec.Datacenter.Racks[0].Name].Status.ObservedGeneration == gen, nil
```
tnozicka, line 140: `observed generation != reconciled generation` (racy)

tnozicka: wonder if we can trigger it through the scylla cluster, although the reconciliation there is minimalistic

the rollout condition for a statefulset is
```
sts.Status.ObservedGeneration >= sts.Generation &&
sts.Status.CurrentRevision == sts.Status.UpdateRevision &&
sts.Status.ReadyReplicas == *sts.Spec.Replicas
```

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,163 @@
...
+})
+
+func checkAgentAuth(ctx context.Context, hosts []string, authToken string, f func(err error)) {
+	cfg := scyllaclient.DefaultConfig(authToken, hosts...)
+
+	client, err := scyllaclient.NewClient(cfg, log.NopLogger)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	_, err = client.Status(ctx, hosts[0])
+	f(err)
```
tnozicka, line 162: would it be cleaner to just return the error and handle it on the call site?
`func getScyllaClientStatus(ctx context.Context, hosts []string, authToken string) (Status, error)`

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,163 @@
...
+
+		framework.By("Initiating a rolling restart")
+
+		gen, err = restartRack(ctx, f.KubeClient().AppsV1(), sc.Spec.Datacenter.Racks[0], sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to pick up token change")
+
+		rackGenerationUpgraded = func(sc *scyllav1.ScyllaCluster) (bool, error) {
+			stss, err := getStatefulSetsForScyllaCluster(ctx, f.KubeClient().AppsV1(), sc)
```
tnozicka, line 138: also you'd need to wait for the statefulset itself, this will requeue only on scyllacluster changes

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,159 @@
...
+		}{
+			AuthToken: "42",
+		}
+		agentConfigData, err := yaml.Marshal(agentConfig)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		agentConfigSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "agent-config",
+				Namespace: f.Namespace(),
```
tnozicka, line 81: nit: this isn't needed as the client will always overwrite it

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -0,0 +1,159 @@
...
+
+		_, err = f.KubeClient().CoreV1().Secrets(f.Namespace()).Create(ctx, agentConfigSecret, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		_, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Patch(
+			ctx,
+			sc.Name,
+			types.MergePatchType,
+			[]byte(fmt.Sprintf(
+				`{"metadata":{"uid":"%s"},"spec":{"datacenter":{"racks":[{"name":"%s","scyllaAgentConfig":"%s","members":1,"resources":{"limits":{"cpu":"1","memory":"1Gi"},"requests":{"cpu":"10m","memory":"100Mi"}},"storage":{"capacity":"100Mi"}}]}}}`,
```
tnozicka, line 97: Why are you patching the resources with the agent config?

zimnx: Without full rack spec i get:
`admission webhook "webhook.scylla.scylladb.com" denied the request: set cpu, memory resource limits for rack us-east-1a`
Looks like patch overwrites entire rack.

zimnx: good idea! done

tnozicka (approved):
/lgtm
thanks!

Issue #582, opened 2021-05-06
Unify scylla version for v1.3 release


tnozicka:
With the beginning of the release cycle, the current latest scylla version is `4.4` which is also the one we should bump the operator to. Latest z-stream is `4.4.1` which we should set up as the primary version in manifests, e2e and QA test. As there is the nodetool bug we'd bump to a later z-stream with the fix during the release cycle, at least in the affected tests.

cc: @zimnx @vponomaryov @tzach

tnozicka:
e2e tests are switched already https://github.com/scylladb/scylla-operator/pull/579

vponomaryov:
@tnozicka better to add links for bugs you are referring to.
It is public place and not all of the interested people are aware about [1], [2] and [3].

[1] https://github.com/scylladb/scylla-tools-java/commit/3eca0e35
[2] https://github.com/scylladb/scylla/issues/7930
[3] https://github.com/scylladb/scylla-tools-java/pull/232

tnozicka:
@vponomaryov we are bumping in https://github.com/scylladb/scylla-operator/pull/609 - hopefully QA tests with the defaults, otherwise please bump yours.

PR #609, opened 2021-05-26
Default scylla to 4.4.2 and manager to 2.3.0

tnozicka:
Pin scylla to 4.4.2 and manager to 2.3.0

Fixes https://github.com/scylladb/scylla-operator/issues/582

### DIFF
--- a/helm/scylla-manager/values.yaml
+++ b/helm/scylla-manager/values.yaml
@@ -7,7 +7,7 @@ fullnameOverride: ""
 image:
   repository: scylladb
   pullPolicy: IfNotPresent
-  tag: 2.2.4
+  tag: 2.3.0
 
 # Allows to customize Scylla Manager Controller image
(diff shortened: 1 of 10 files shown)


helm/scylla-manager/values.yaml
```diff
@@ -7,7 +7,7 @@ fullnameOverride: ""
 image:
   repository: scylladb
   pullPolicy: IfNotPresent
-  tag: 2.2.4
+  tag: 2.3.0
```
zimnx, line 10: This should be reflected in our deploy folder but isn't. 
Looks like there's a not needed version override in `helm/deploy/manager_prod.yaml`.

tnozicka: removed the override

zimnx (approved):
LGTM, please rebase

PR #615, opened 2021-05-27
[WIP] Remove non-OCI instructions from the Dockerfile

tnozicka:
SHELL instruction is not in the OCI spec, some builder may ignore it:
```
WARN[0066] SHELL is not supported for OCI image format, [/bin/bash -euEo pipefail -c] will be ignored. Must use `docker` format 
```


tnozicka:
Forgot we are stuck with `dash` otherwise, and `dash` doesn't know about `-o pipefail`. in 2021... way to go Ubuntu

PR #633, opened 2021-06-09
Add full support for HostNetwork pod IPs 

geobeau:

This PR add full support for HostNetworking on scylla clusters.

The reason why HostNetworking is needed is when the clients are not inside the same Kubernetes clusters and that internal IPs are not routable outside of the cluster.

The previously proposed solution was to only override the rpc_address so that clients can connect with another address than the peer address. https://github.com/scylladb/scylla-operator/commit/a457bd5df65bbbc51368115a3ac2da934cb36ff8
However this doesn't work when joining an external cluster as nodes are using the peer address.

The main difficulty was to keep all the operations (notably replace) working. The way it was solve was to add a label on the member service with the real IP of the pod.
Since the IP is assigned only after the pod is scheduled, we need to listen for changes on scylla managed pods. 

Resolves #

### DIFF
--- a/pkg/controllers/cluster/actions/replace.go
+++ b/pkg/controllers/cluster/actions/replace.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/scylladb/go-log"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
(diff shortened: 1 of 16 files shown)


Ashtonian:
Any update on this?

tnozicka:

PR #634, opened 2021-06-09
Add e2e test for automatic orphaned volume cleanup

tnozicka:
Adds e2e coverage for automatic orphaned volume cleanup.

### DIFF
--- a/pkg/controller/helpers/mergepatch.go
+++ b/pkg/controller/helpers/mergepatch.go
@@ -1,0 +1,28 @@
+package helpers
+
+import (
+	"fmt"
+
+	jsonpatch "github.com/evanphx/json-patch"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func GenerateMergePatch(original runtime.Object, modified runtime.Object) ([]byte, error) {
+	originalJSON, err := runtime.Encode(unstructured.UnstructuredJSONScheme, original)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode original to JSON: %v", err)
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/controller/helpers/mergepatch.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+
+	jsonpatch "github.com/evanphx/json-patch"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func GenerateMergePatch(original runtime.Object, modified runtime.Object) ([]byte, error) {
+	if original == nil {
```
zimnx, line 12: these are interface objects so plain nil comparison doesn't work. I think you can skip these checks

test/e2e/set/scyllacluster/scyllacluster_pv.go
```diff
@@ -0,0 +1,112 @@
...
+		sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Patch(
+			ctx,
+			sc.Name,
+			types.MergePatchType,
+			[]byte(`{"metadata": {"annotations": {"foo": "bar"} } }`),
+			metav1.PatchOptions{},
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the PVC to be replaced")
```
zimnx, line 94: entire scylla node should be replaced, we could check that Pod UUID is different

tnozicka: although PVC has the finalizer waiting for pod deletion

zimnx: right, this implies pod deletion

zimnx (dismissed):
LGTM, one nit

zimnx (approved):

PR #640, opened 2021-06-15
Explicitly set gcloud project in GKE example commands

zimnx:
When default GCP project is different than set by CLI parameter script
gets stuck on checking cluster readiness because it can't find cluster
status.

### DIFF
--- a/examples/gke/gke.sh
+++ b/examples/gke/gke.sh
@@ -79,9 +79,12 @@ then
   exit 1
 fi
 
-GCP_REGION=${GCP_ZONE:0:$((${#GCP_ZONE}-2))}
+gcloud() {
+  command gcloud "$@" --project "${GCP_PROJECT}" --zone "${GCP_ZONE}"
+}
 
-CLUSTER_VERSION="$(gcloud container get-server-config --zone ${GCP_ZONE} --format "value(validMasterVersions[0])")"
+GCP_REGION="${GCP_ZONE:0:$((${#GCP_ZONE}-2))}"
+CLUSTER_VERSION=$(gcloud container get-server-config --format "value(validMasterVersions[0])")
 
 check_prerequisites() {
     echo "Checking if kubectl is present on the machine..."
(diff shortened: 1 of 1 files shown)


examples/gke/gke.sh
```diff
@@ -81,7 +81,7 @@ fi
 
 GCP_REGION=${GCP_ZONE:0:$((${#GCP_ZONE}-2))}
 
-CLUSTER_VERSION="$(gcloud container get-server-config --zone ${GCP_ZONE} --format "value(validMasterVersions[0])")"
+CLUSTER_VERSION="$(gcloud container get-server-config --project "${GCP_PROJECT}" --zone="${GCP_ZONE}" --format "value(validMasterVersions[0])")"
```
tnozicka, line 84: might worth to use the `=` consistently as `--flag="${VAR}", it's also safer

tnozicka (commented):
one nit, rest lgtm

if you were looking for a bigger change you could create an alias inside the script like
```
alias gcloud='gcloud --project="'${GCP_PROJECT}'" --zone="'${GCP_ZONE}'"'
```
and use only `gcloud` after

zimnx:
Very good idea with an alias. PTAL

examples/gke/gke.sh
```diff
@@ -79,9 +79,12 @@ then
   exit 1
 fi
 
-GCP_REGION=${GCP_ZONE:0:$((${#GCP_ZONE}-2))}
+gcloud() {
+  command gcloud "$@" --project "${GCP_PROJECT}" --zone "${GCP_ZONE}"
```
tnozicka, line 83: nit: I don't think gcloud is that sensitive but I'd put the `"$@"` at the end, I know some programs (although unfortunate) require flags before args

tnozicka (dismissed):

examples/gke/gke.sh
tnozicka: guess that can go the other way too so let's keep what's working ;)

zimnx: exactly:
```
++ gcloud --project skilled-adapter-452 --zone us-east1-b container get-server-config --format 'value(validMasterVersions[0])'                                                                                  
ERROR: (gcloud) Invalid choice: 'us-east1-b'.                                                                                                                                                                                                                                                                                                                                                                                    
Maybe you meant:
  gcloud config get-value
  gcloud config list
  gcloud config set
  gcloud config unset
  gcloud container get-server-config

To search the help text of gcloud commands, run:
  gcloud help -- SEARCH_TERMS

zimnx:
clean rebase

tnozicka (approved):

PR #658, opened 2021-06-28
Use RHEL UBI images

tnozicka:
Changes our base image to `registry.access.redhat.com/ubi8/ubi-minimal`. We also get rid of Dash workarounds and go back to OCI compatible builds.


slivne:
Please don;t merge till tzach approves licensing

tnozicka:
> Please don;t merge till tzach approves licensing

ack, hopefully that won't be an issue since UBI is freely redistributable and we need it for RHEL / OCP certification.

/assign @tzach

PR #664, opened 2021-06-29
Claim kube version >= 1.19.10

tnozicka:
We need [statefulset controller fix for handling PVCs](https://github.com/kubernetes/kubernetes/pull/98732), v1.19.10 is the first one to have it. It's also contained in `v1.20.0-alpha.1` and higher so we don't need to specify any particular z-stream there.

cc @vponomaryov

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 ![](logo.png)
 
 ## Deploying the Operator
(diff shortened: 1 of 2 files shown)


zimnx (dismissed):
lgtm, thanks

tnozicka:
(clean rebase)

zimnx (approved):

Issue #665, opened 2021-06-29
scylla-manager uses wrong dns name


vponomaryov:
**Describe the bug**
Scylla-manager pod start up but fail to resolve DNS name and become non-operational because of this.
There are lots of the following errors in the scylla-manager logs:
```
E0628 19:23:12.475318       1 manager/controller.go:131] syncing key 'scylla/sct-cluster' failed: Get "http://scylla-manager.scylla-manager.svc/api/v1/clusters": dial tcp: lookup scylla-manager.scylla-manager.svc on 10.87.240.10:53: no such host
```

Looks like the svc name is incorrect. It uses "scylla-manager" in namespace part, but namespace is "scylla-manager-system".

List of namespaces:
```
NAME                     STATUS   AGE

...

scylla                   Active   4h18m
scylla-manager-system    Active   4h26m
scylla-operator-system   Active   4h27m
```

**To Reproduce**
1. Deploy Scylla operator
2. Deploy scylla-manager
3. See error

**Expected behavior**
Scylla-manager should start up correctly.

**Config Files**
If relevant, upload your configuration files here using GitHub, there is no need to upload them to any 3rd party services

**Logs**
Please, provide `kubectl get events`, `kubectl logs -n scylla pod-name` and other relevant information.
[kubernetes-fdff4382.tar.gz](https://github.com/scylladb/scylla-operator/files/6732868/kubernetes-fdff4382.tar.gz)

**Environment:**
 - Platform: GKE
 - Kubernetes version: 1.19.11
 - Scylla version: 4.4.3
 - Scylla-operator version: v1.4.0-alpha.0-34-gf13771d

**Additional context**

PR #666, opened 2021-06-29
Fix scylla manager DNS name in the URL

tnozicka:
We claim global namespace `scylla-manager` but unfortunately as the helm charts allow it to be configurable, some still have the old namespace name `scylla-manager-system`. This PR makes sure the manager URL works in both environments and referencing the same namespace is more flexible and correct.

Resolves https://github.com/scylladb/scylla-operator/issues/665

### DIFF
--- a/pkg/cmd/operator/manager_controller.go
+++ b/pkg/cmd/operator/manager_controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
 	"github.com/scylladb/scylla-operator/pkg/leaderelection"
 	"github.com/scylladb/scylla-operator/pkg/mermaidclient"
+	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/signals"
 	"github.com/scylladb/scylla-operator/pkg/version"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


pkg/cmd/operator/manager_controller.go
```diff
@@ -121,7 +121,7 @@ func (o *ManagerControllerOptions) Complete() error {
 	}
 
 	// TODO: Use https and wire certs.
-	managerClient, err := mermaidclient.NewClient("http://scylla-manager.scylla-manager.svc/api/v1", &http.Transport{})
+	managerClient, err := mermaidclient.NewClient("http://scylla-manager/api/v1", &http.Transport{})
```
zimnx, line 124: How about we search for a service with particular label in every namespace?

tnozicka: the manager controller and manager are always in the same namespace though, right?

zimnx: yeah, namespace is the same. Although name can be overwriten in helm values file: https://github.com/scylladb/scylla-operator/blob/d2b52a604f4327107d8b1e2fa1dd7e0a7fb76d83/helm/scylla-manager/templates/_helpers.tpl#L15

tnozicka: we should configurations for stuff that doesn't need it - there is no value in configuring resource names in a namespace we fully own

tnozicka: > service with particular label in every namespace
sounds a bit like an overkill to search in every namespace (if there can be thousands) and the service labels are also configurable and not fixed

zimnx: having #667 lets just extract const values for these names in `naming/names.go` and use them here

tnozicka:
just a clean rebase, going to push the changes separately for better review

pkg/cmd/operator/manager_controller.go
tnozicka: made it const

zimnx (dismissed):

tnozicka:
(clean rebase)

zimnx (approved):

PR #667, opened 2021-06-29
Force fixed names and namespaces for system components in helm

tnozicka:
We fully own the `scylla-operator` and `scylla-manager` system namespaces so we don't need to make the names configurable and templating those brings issues that can break the deployment. These namespaces are singletons and there will always be just one instance for each - that's why they don't match the regular helm pattern where you template the namespace name so it can deploy to multiple namespaces. We also own the whole namespace, contrary to the instance type charts that deploy into a namespace with other objects, so there can't be name collisions needing the templating.

We need fixed names so we can easily:
- talk to the deployed components, like from the manager-controller to the manager through the service DNS name
- gather artifacts when we have a "must-gather" command
- make sure the deployment always works, we only test one combination of names and namespaces in the CI like in https://github.com/scylladb/scylla-operator/issues/665
- ...

We need to mention this in the upgrade guide so people fix the names (if changed).

### DIFF
--- a/helm/scylla-manager/templates/NOTES.txt
+++ b/helm/scylla-manager/templates/NOTES.txt
@@ -1,8 +1,8 @@
 
 The Scylla Manager has been installed. Check its status by running:
 
-  kubectl -n {{ .Release.Namespace }} get pods -l "app.kubernetes.io/name={{ template "scylla-manager.name" . }}"
-  kubectl -n {{ .Release.Namespace }} get pods -l "app.kubernetes.io/name={{ template "scylla-manager.name" . }}-controller"
+  kubectl -n scylla-manager get pods -l "app.kubernetes.io/name=scylla-manager"
+  kubectl -n scylla-manager get pods -l "app.kubernetes.io/name=scylla-manager-controller"
 
 Visit https://github.com/scylladb/scylla-operator for tutorials on how to
 create and configure Scylla clusters using the Scylla Operator and set up monitoring.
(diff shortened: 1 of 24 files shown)


zimnx (dismissed):
lgtm, thanks

tnozicka:
(clean rebase)

zimnx (approved):

j14s:
I realize this is a closed ticket/pr but I didn't want to neccisarrily create a issue for this, but I I just ran into this. 
There are many organizations that have strictures and standards around the naming of namespace in clusters. This choice to hardcode the namespace instead of making it configurable means that ScyllaDB becomes a no-go for a number of organizations. I do understand wanting to standardize for testing and making your life easier to produce this great service, but making it hard coded means if I want to use it I have to maintain a copy that is heavily modified if I want to use it.

steve-gray:
@tnozicka - I agree with @j14s here, you've conflated two issues:

- Ownership of the resources by the operator and ensuring there's just one.

- The clusters internal naming and topology from a Kubernetes operator perspective.

The "right thing" to do, and what every other operator (Jaeger, Strimzi, Grafana, Minio....) is to permit these names to be set by the cluster operators as they see fit. If you need those references/names, then the downstream resources that leverage them should be made configurable. 

_Scylla Operator is not special in regards to the concerns you've raised, and the reasons given for hard-coding them are faulty to the core._

You own the software in the namespace, but the cluster operator chooses the namespace. This is the way.

Issue #672, opened 2021-07-02
Remove forced ownership


tnozicka:
v1.3 and previous versions didn't set selectors correctly for some cases. Given the Three laws of controllers v1.4 (that respects them) had to release those objects. That got us to a state where we couldn't update them because we don't own them and we can't adopt them because they don't match our selector.

We should set `forceOwnership` parameter back to `false` in v1.5.

Choraden:
There are 4 places where operator forces ownership 

controller/scyllacluster/:
- sync_serviceaccounts.go
- sync_rolebindings.go
- sync_agent_config.go
- sync_pdb.go

Unfortunately the scyllacluster_sa e2e test needs the first 2 to force ownership. 
e.g. 
`syncing key 'e2e-test-scyllacluster-tzr76-5nppk/basic-7554m' failed: can't sync serviceaccounts: can't apply service account: serviceAccount "e2e-test-scyllacluster-tzr76-5nppk/basic-7554m-member" isn't controlled by us`

`"Event occurred" object="e2e-test-scyllacluster-8f75c-2bpdc/basic-89lj9-member" fieldPath="" kind="RoleBinding" apiVersion="rbac.authorization.k8s.io/v1" type="Warning" reason="UpdateRoleBindingFailed" message="Failed to update RoleBinding e2e-test-scyllacluster-8f75c-2bpdc/basic-89lj9-member: roleBinding \"e2e-test-scyllacluster-8f75c-2bpdc/basic-89lj9-member\" isn't controlled by us"
`

So it looks like for now we can disable forcing ownership only in pdb and secrets.

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2463.

PR #673, opened 2021-07-02
Fix operator upgrade bugs

tnozicka:
Creating the ScyllaCluster with scylla-operator v1.3 and updating the scylla-operator uncovered few corner cases like objects with invalid selectors or a nil segfault.

v1.3 and previous versions didn't set selectors correctly for some cases. Given the Three laws of controllers v1.4/master (that respects them) had to release those objects. That got us to a state where we couldn't update them because we don't own them and we can't adopt them because they don't match our selector.

`forceOwnership` flag allows to migrate from such state and we'll switch it back to `false` in v1.5 (https://github.com/scylladb/scylla-operator/issues/672).

### DIFF
--- a/deploy/operator/00_clusterrole_def.yaml
+++ b/deploy/operator/00_clusterrole_def.yaml
@@ -66,6 +66,7 @@ rules:
   - watch
   - create
   - update
+  - patch
 - apiGroups:
   - ""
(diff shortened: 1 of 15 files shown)


zimnx (approved):
LGTM

PR #668, opened 2021-06-30
Support observedGeneration API concept

tnozicka:
**Description of your changes:**
Add support for `scyllacluster.status.observedGeneration` to help clients to distinguish that
the reported status reflects the most recent desired status.

https://github.com/kubernetes/community/blame/f710ed530e5f42bcafaebb6bc4cfdb7a829f1630/contributors/devel/sig-architecture/api-conventions.md#L467-L470

PR #674, opened 2021-07-02
Add support for manually triggering ScyllaCluster rollout

tnozicka:
This PR add an API filed `ForceRedeploymentReason` that allows user to trigger a ScyllaCluster rollout manually. This maintains HA and respects the PDB, contrary to having to delete all pods manually or at once. This has been previously handled by a hack of manually forcing rollouts for all StatefulSets, but given `kubectl rollout restart` adds an annotation we'd now reconcile and remove it, which would do a second rollout. This also had to be done for each StatefulSet/Rack separately and couldn't be done for the whole cluster.

Manually triggering cluster rollout is useful e.g. for reloading scylla config.

Requires:
- [x] https://github.com/scylladb/scylla-operator/pull/668
- [x] https://github.com/scylladb/scylla-operator/pull/684

### DIFF
--- a/examples/common/operator.yaml
+++ b/examples/common/operator.yaml
@@ -3066,6 +3066,10 @@ spec:
               developerMode:
                 description: developerMode determines if the cluster runs in developer-mode.
                 type: boolean
+              forceRedeploymentReason:
+                description: forceRedeploymentReason can be used to force a rolling
+                  update of all racks by providing a unique string.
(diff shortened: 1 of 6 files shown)


zimnx (dismissed):
LGTM, thanks

tnozicka:
(clean rebase)

zimnx (approved):

PR #681, opened 2021-07-08
Fix GH action to work with new minikube 1.22.0

tnozicka:
Minikube https://github.com/kubernetes/minikube/pull/11576 broke our install script, this PR reacts to the change and renames the config value.

This is currently blocking all CI.
```
# minikube config set embed-certs true

❌  Exiting due to MK_CONFIG_SET: find settings for "embed-certs" value of "true": property name "embed-certs" not found
```


zimnx (approved):
lgtm, thanks

PR #683, opened 2021-07-08
Fix controller leader election lock names

### LINKED ISSUE #676: Operator options are not compatible between v1.3 and v1.4-alpha versions causing upgrade failure
**Describe the bug**
If we deploy operator v1.3.0 using appropriate Helm chart and then upgrade it using v1.4.0-alpha Helm chart we get following error in the scylla logs for existing scylla cluster:

```
Error: unknown flag: --secret-name
Usage:
   sidecar [flags]

Flags:
      --cpu string             number of cpus to use
  -h, --help                   help for sidecar
      --log-level string       verbosity of the logs. Possible values, in descending order of verbosity: fatal, error, info, debug (default "debug")
      --pod-name string        name of the pod
      --pod-namespace string   namespace of the pod

{"L":"ERROR","T":"2021-07-07T10:50:32.725Z","M":"Root command: a fatal error occured","error":"unknown flag: --secret-name"}
```

Part of scylla pod spec:
```
spec:
  affinity:
    ...
  containers:
  - command:
    - /mnt/shared/scylla-operator
    - sidecar
    - --secret-name=sct-cluster-auth-token
    - --service-name=$(SERVICE_NAME)
    - --cpu-count=$(CPU_COUNT)
    - --loglevel=2
    env:
    - name: SERVICE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: CPU_COUNT
      valueFrom:
        resourceFieldRef:
          containerName: scylla
          divisor: "1"
          resource: limits.cpu
    image: scylladb/scylla:4.4.2
...
    name: scylla
...
```

**To Reproduce**
1. Deploy operator using v1.3.0 Helm chart
2. Deploy a Scylla cluster
3. Upgrade operator using v1.4.0-alpha.0-50-gb19ea00 Helm chart
4. Wait for scylla operator upgrade
5. Wait for scylla cluster upgrade
6. See error

**Expected behavior**
No errors appear, rolling scylla cluster upgrade happens successfully.

**Logs**
Please, provide `kubectl get events`, `kubectl logs -n scylla pod-name` and other relevant information.
[kubernetes-aa06900d.tar.gz](https://github.com/scylladb/scylla-operator/files/6778259/kubernetes-aa06900d.tar.gz)

**Environment:**
 - Platform: EKS
 - Kubernetes version: v1.20.4-eks-6b7464
 - Scylla version: 4.4.2
 - Scylla-operator version: e.g.: v1.4.0-alpha.0-50-gb19ea00

**Additional context**

zimnx:
1.3 controllers used different LE lock names.
During rolling upgrade to 1.4 two controllers were able to win LE and hence reconcile Scylla clusters.
Lock names are fixed to names used in 1.3.

Fixes #676

### DIFF
--- a/pkg/cmd/operator/manager_controller.go
+++ b/pkg/cmd/operator/manager_controller.go
@@ -142,10 +142,13 @@ func (o *ManagerControllerOptions) Run(streams genericclioptions.IOStreams, comm
 		cancel()
 	}()
 
+	// Lock names cannot be changed, because it may lead to two leaders during rolling upgrades.
+	const lockName = "scylla-manager-controller-lock"
+
 	return leaderelection.Run(
 		ctx,
 		commandName,
-		commandName+"-locks",
+		lockName,
 		o.Namespace,
 		o.kubeClient,
 		o.LeaderElectionLeaseDuration,
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -141,10 +141,13 @@ func (o *OperatorOptions) Run(streams genericclioptions.IOStreams, commandName s
 		cancel()
 	}()
 
+	// Lock names cannot be changed, because it may lead to two leaders during rolling upgrades.
+	const lockName = "scylla-operator-lock"
+
 	return leaderelection.Run(
 		ctx,
 		commandName,
-		commandName+"-locks",
+		lockName,
 		o.Namespace,
 		o.kubeClient,
 		o.LeaderElectionLeaseDuration,


tnozicka (dismissed):
Thanks for finding that out!

Technically, I think we can change it eventually, these were just incompatible reconciliation approaches. When the reconciliation approach is unchanged we should be able to tolerate 2 controllers running for a short period of time. Leader election doesn't guarantee at most one anyways, like with clock skew.

/lgtm

tnozicka (approved):

zimnx:
> Thanks for finding that out!
> 
> Technically, I think we can change it eventually, these were just incompatible reconciliation approaches. When the reconciliation approach is unchanged we should be able to tolerate 2 controllers running for a short period of time. Leader election doesn't guarantee at most one anyways, like with clock skew.
> 
> /lgtm

That's right, although it may lead to two restarts, because each operator would try to change the sidecar image to it's own. It wouldn't be fatal as long as both reconcile entire template.

tnozicka:
> That's right, although it may lead to two restarts, because each operator would try to change the sidecar image to it's own. It wouldn't be fatal as long as both reconcile entire template.

yep, but if both use the hash approach and the same annotation they would reconcile eventually.

tnozicka (approved):

PR #695, opened 2021-07-13
Chain validation errors

### LINKED ISSUE #608: Validation should chain the errors and return all of them
* Feature Request

Validation errors should be chained and all of them should be returned, not only single validation failure.

Example usage is in
https://github.com/kubernetes/kubernetes/blob/2a0ef026b6b4a6d14660113fcfeab19ed2bd1cfa/pkg/apis/apps/validation/validation.go#L560-L565

UX, consistency behaviour with the ecosystem
n:**

rzetelskik:
- Added chaining of error messages in validations.
 
Resolves #608

### DIFF
--- a/pkg/admissionreview/handler.go
+++ b/pkg/admissionreview/handler.go
@@ -8,6 +8,7 @@ import (
 
 	admissionv1 "k8s.io/api/admission/v1"
 	v1 "k8s.io/api/admission/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -89,14 +90,22 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		responseAdmissionReview.Response = &v1.AdmissionResponse{
 			UID:     requestedAdmissionReview.Request.UID,
 			Allowed: funcErr == nil,
-			Result: &metav1.Status{
-				Message: func() string {
-					if funcErr == nil {
-						return ""
-					}
-					return funcErr.Error()
-				}(),
-			},
+			Result: func() *metav1.Status {
+				s, ok := funcErr.(apierrors.APIStatus)
+				if ok {
+					sts := s.Status()
+					return &sts
+				}
+
+				return &metav1.Status{
+					Message: func() string {
+						if funcErr == nil {
+							return ""
+						}
+						return funcErr.Error()
+					}(),
+				}
+			}(),
 		}
 		responseObj = responseAdmissionReview
 
--- a/pkg/api/validation/cluster_validation.go
+++ b/pkg/api/validation/cluster_validation.go
@@ -1,14 +1,15 @@
 package validation
 
 import (
+	"fmt"
 	"reflect"
 	"strconv"
 
-	"github.com/pkg/errors"
 	"github.com/scylladb/go-set/strset"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	"github.com/scylladb/scylla-operator/pkg/semver"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


pkg/api/validation/cluster_validation.go
```diff
@@ -25,151 +26,171 @@ var (
...
 }
 
-func ValidateScyllaClusterUpdate(new, old *scyllav1.ScyllaCluster) error {
-	err := ValidateScyllaCluster(new)
-	if err != nil {
-		return err
+func ValidateScyllaClusterUpdate(new, old *scyllav1.ScyllaCluster) field.ErrorList {
+	allErrs := ValidateScyllaCluster(new)
+	if allErrs.ToAggregate() != nil {
+		return allErrs
```
tnozicka, line 137: Why not chain the errors in this case?

pkg/cmd/operator/webhooks.go
```diff
@@ -230,10 +230,10 @@ func validate(ar *admissionv1.AdmissionReview) error {
 	case scyllav1.GroupVersion.WithResource("scyllaclusters"):
 		switch ar.Request.Operation {
 		case admissionv1.Create:
-			return validation.ValidateScyllaCluster(obj.(*scyllav1.ScyllaCluster))
+			return validation.ValidateScyllaCluster(obj.(*scyllav1.ScyllaCluster)).ToAggregate()
```
tnozicka, line 233: maybe we should plumb it into the webhook and set `StatusDetails` with `apierrors.NewInvalid`

tnozicka (commented):
Looks good overall, please change the unit tests to DeepEqual the validation errors so the we have the actual stings checked out and easily validated.

Would `Chain validation errors` be a better commit name?

pkg/api/validation/cluster_validation.go
rzetelskik: chained

pkg/admissionreview/handler.go
```diff
@@ -89,14 +91,21 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-			Result: &metav1.Status{
-				Message: func() string {
-					if funcErr == nil {
-						return ""
-					}
-					return funcErr.Error()
-				}(),
-			},
+			Result: func() *metav1.Status {
+				s, ok := funcErr.(*errors.StatusError)
```
tnozicka, line 95: APIStatus interface

pkg/cmd/operator/webhooks.go
```diff
@@ -228,17 +232,18 @@ func validate(ar *admissionv1.AdmissionReview) error {
 
 	switch gvr {
 	case scyllav1.GroupVersion.WithResource("scyllaclusters"):
+		var allErrs field.ErrorList
```
tnozicka, line 235: nit : errs (allErrs kind of implies you'd be placing multiple errs into it)

pkg/admissionreview/handler.go
rzetelskik: fixed

pkg/cmd/operator/webhooks.go
rzetelskik: fixed

pkg/admissionreview/handler.go
```diff
@@ -89,14 +90,22 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-					if funcErr == nil {
-						return ""
-					}
-					return funcErr.Error()
-				}(),
-			},
+			Result: func() *metav1.Status {
+				s, ok := funcErr.(errors.APIStatus)
+				if ok {
+					sts := s.Status()
```
tnozicka, line 96: guess this should check nil first

pkg/api/validation/cluster_validation_test.go
```diff
@@ -68,7 +68,7 @@ func TestValidateScyllaCluster(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			err := validation.ValidateScyllaCluster(test.obj)
+			err := validation.ValidateScyllaCluster(test.obj).ToAggregate()
```
tnozicka, line 71: this test should deepequal the fieldErrors

pkg/admissionreview/handler.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	admissionv1 "k8s.io/api/admission/v1"
 	v1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
rzetelskik: fixed

pkg/admissionreview/handler.go
rzetelskik: I'm not sure if you're referring to s or sts, but i think that if s was nil, then ok would be false, and s.Status() can't return nil as its return type is metav1.Status, not *metav1.Status.

pkg/api/validation/cluster_validation_test.go
rzetelskik: fixed

pkg/admissionreview/handler.go
tnozicka: `funcErr` or `s` could be nil

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,120 @@ func TestValidateScyllaCluster(t *testing.T) {
...
+			obj:               validCluster,
+			expectedErrorList: field.ErrorList{},
 		},
 		{
-			name:    "two racks with same name",
-			obj:     sameName,
-			allowed: false,
+			name:              "two racks with same name",
+			obj:               sameName,
+			expectedErrorList: field.ErrorList{field.Duplicate(field.NewPath("spec", "datacenter", "racks").Key("test-rack").Child("name"), "test-rack")},
```
tnozicka, line 58: i think, in this case, it would be more beneficial to compare the errors string to see if the make sense

you could compare Aggregates

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,120 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			err := validation.ValidateScyllaCluster(test.obj)
-			if test.allowed {
-				if err != nil {
-					t.Errorf("Wrong value returned from checkValues function. Message: '%s'", err)
-				}
-			} else {
-				if err == nil {
-					t.Errorf("Wrong value returned from checkValues function. Message: '%s'", err)
-				}
+			errList := validation.ValidateScyllaCluster(test.obj)
```
tnozicka, line 77: `.ToAggregate()`

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,120 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			err := validation.ValidateScyllaClusterUpdate(test.new, test.old)
-			if test.allowed {
-				if err != nil {
-					t.Errorf("Wrong value returned from checkTransitions function. Message: '%s'", err)
-				}
-			} else {
-				if err == nil {
-					t.Errorf("Wrong value returned from checkTransitions function. Message: '%s'", err)
-				}
+			errList := validation.ValidateScyllaClusterUpdate(test.new, test.old)
```
tnozicka, line 157: `.ToAggregate()`

pkg/admissionreview/handler.go
rzetelskik: I believe it wouldn't enter the block in either case: https://play.golang.org/p/5aD5HvZ5gyQ

tnozicka: https://play.golang.org/p/Z_4vh29ZrEJ

pkg/api/validation/cluster_validation_test.go
rzetelskik: I've changed comparing ErrorLists to comparing Aggregates but using DeepEqual still requires creating actual field.Errors, i.e. errorString is not deep equal with field.Error. I don't really see any benefit in comparing aggregates instead of comparing ErrorLists, but creating field.Errors literals is probably better than using the functions. Let me know what you think

pkg/admissionreview/handler.go
rzetelskik: Do we really care about explicitly assigning (*T)(nil) though? In this case checking for nil would require using reflect, right?

pkg/api/validation/cluster_validation_test.go
tnozicka: yeah, you're right, the real type is still hidden in the error interface. My motivation is to see the error strings that will be shown to the users. It looks like we won't get it at once so how about we do `expectedErrorList` as you had before and add `expectedErrorString` field which you'd compare to `.ToAggregate.String()`?

pkg/admissionreview/handler.go
tnozicka: I thought you'd just check s after casting but I haven't double checked

rzetelskik: For the usual cases (untyped nil) checking `ok` is enough, as it checks whether funcErr is not nil and implements the interface. For (*T)(nil) we would have to use reflect, but then it would probably make sense to check other comparisons in this function to `reflect` ones.
There is a painfully long conversation about this [here](https://groups.google.com/g/golang-nuts/c/wnH302gBa4I).
I'd just stick with the type assertion.

tnozicka: you're right, we are casting into an interface so the nil check wouldn't work, I think it's fine then and the main nil value is already handled in the cast

pkg/api/validation/cluster_validation_test.go
rzetelskik: Changed to checking `expectedErrorList` and `expectedErrorString`

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
 		{
-			name:    "two racks with same name",
-			obj:     sameName,
-			allowed: false,
+			name: "two racks with same name",
+			obj:  sameName,
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeDuplicate, Field: "spec.datacenter.racks[test-rack].name", BadValue: "test-rack"},
+			},
+			expectedErrorString: `spec.datacenter.racks[test-rack].name: Duplicate value: "test-rack"`,
```
tnozicka, line 63: this implies a map, while racks are an array

pkg/api/validation/cluster_validation.go
```diff
@@ -25,151 +26,168 @@ var (
...
-				if requests.Memory().MilliValue() != limits.Memory().MilliValue() {
-					return errors.Errorf("when using cpuset, memory requests must be the same as memory limits in rack %s", rack.Name)
-				}
-			} else {
-				// Copy the limits
-				rack.Resources.Requests = limits.DeepCopy()
-			}
-		}
+	for _, rack := range spec.Datacenter.Racks {
+		allErrs = append(allErrs, ValidateScyllaClusterRackSpec(rack, rackNames, spec.CpuSet, fldPath.Child("datacenter", "racks").Key(rack.Name))...)
```
tnozicka, line 60: s/Key/Index

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
 		{
-			name:    "invalid intensity in repair task spec",
-			obj:     invalidIntensity,
-			allowed: false,
+			name: "invalid intensity in repair task spec",
+			obj:  invalidIntensity,
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeInvalid, Field: "spec.repairs[]", BadValue: "100Mib", Detail: "invalid intensity, it must be a float value"},
+			},
+			expectedErrorString: `spec.repairs[]: Invalid value: "100Mib": invalid intensity, it must be a float value`,
```
tnozicka, line 71: index, subfield

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			name:    "non-unique names in manager tasks spec",
-			obj:     nonUniqueManagerTaskNames,
-			allowed: false,
+			name: "invalid intensity in repair task spec && non-unique names in manager tasks spec",
+			obj:  nonUniqueManagerTaskNames,
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeInvalid, Field: "spec.repairs[task-name]", BadValue: "", Detail: "invalid intensity, it must be a float value"},
+				&field.Error{Type: field.ErrorTypeDuplicate, Field: "spec.backups[task-name]", BadValue: "task-name"},
+			},
+			expectedErrorString: `[spec.repairs[task-name]: Invalid value: "": invalid intensity, it must be a float value, spec.backups[task-name]: Duplicate value: "task-name"]`,
```
tnozicka, line 80: array, not a map

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-					t.Errorf("Wrong value returned from checkValues function. Message: '%s'", err)
-				}
+			errList := validation.ValidateScyllaCluster(test.obj)
+			if !reflect.DeepEqual(errList, test.expectedErrorList) {
+				t.Errorf("expected and actual error lists differ: %s", cmp.Diff(test.expectedErrorList, errList))
+			}
+
+			errStr := ""
+			if errList.ToAggregate() != nil {
+				errStr = errList.ToAggregate().Error()
```
tnozicka, line 93: nit: save the aggregate into a var

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			old:     unit.NewSingleRackCluster(3),
-			new:     placementChanged(unit.NewSingleRackCluster(3)),
-			allowed: false,
+			name: "rackPlacement changed",
+			old:  unit.NewSingleRackCluster(3),
+			new:  placementChanged(unit.NewSingleRackCluster(3)),
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeForbidden, Field: "spec.datacenter.racks[test-rack]", BadValue: "", Detail: "changes in placement are currently not supported"},
+			},
+			expectedErrorString: "spec.datacenter.racks[test-rack]: Forbidden: changes in placement are currently not supported",
```
tnozicka, line 160: array

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			old:     unit.NewSingleRackCluster(3),
-			new:     storageChanged(unit.NewSingleRackCluster(3)),
-			allowed: false,
+			name: "rackStorage changed",
+			old:  unit.NewSingleRackCluster(3),
+			new:  storageChanged(unit.NewSingleRackCluster(3)),
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeForbidden, Field: "spec.datacenter.racks[test-rack]", BadValue: "", Detail: "changes in storage are currently not supported"},
+			},
+			expectedErrorString: "spec.datacenter.racks[test-rack]: Forbidden: changes in storage are currently not supported",
```
tnozicka, line 169: array

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			old:     unit.NewSingleRackCluster(3),
-			new:     resourceChanged(unit.NewSingleRackCluster(3)),
-			allowed: false,
+			name: "rackResources changed",
+			old:  unit.NewSingleRackCluster(3),
+			new:  resourceChanged(unit.NewSingleRackCluster(3)),
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeForbidden, Field: "spec.datacenter.racks[test-rack]", BadValue: "", Detail: "changes in resources are currently not supported"},
+			},
+			expectedErrorString: "spec.datacenter.racks[test-rack]: Forbidden: changes in resources are currently not supported",
```
tnozicka, line 178: array

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-			old:     unit.NewSingleRackCluster(3),
-			new:     rackDeleted(unit.NewSingleRackCluster(3)),
-			allowed: false,
+			name: "rack deleted",
+			old:  unit.NewSingleRackCluster(3),
+			new:  rackDeleted(unit.NewSingleRackCluster(3)),
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeForbidden, Field: "spec.datacenter.racks", BadValue: "", Detail: "racks [test-rack] not found, you cannot remove racks from the spec"},
+			},
+			expectedErrorString: "spec.datacenter.racks: Forbidden: racks [test-rack] not found, you cannot remove racks from the spec",
```
tnozicka, line 187: array

rzetelskik:
addressed all the comments, ready for review

pkg/api/validation/cluster_validation_test.go
```diff
@@ -40,129 +43,164 @@ func TestValidateScyllaCluster(t *testing.T) {
...
 	}{
 		{
-			name:    "same as old",
-			old:     unit.NewSingleRackCluster(3),
-			new:     unit.NewSingleRackCluster(3),
-			allowed: true,
+			name:              "same as old",
+			old:               unit.NewSingleRackCluster(3),
+			new:               unit.NewSingleRackCluster(3),
+			expectedErrorList: field.ErrorList{},
```
tnozicka, line 114: nit: also explicitly set `expectedErrorString` so the intent is clear (applies globally)

tnozicka (commented):
last nit, lgtm otherwise

pkg/api/validation/cluster_validation_test.go
rzetelskik: Thanks, missed it. Fixed

tnozicka (dismissed):
/lgtm
thanks!

zimnx (approved):

zimnx:
LGTM, please rebase

zimnx (approved):

PR #706, opened 2021-07-16
Fix date ranges for listing PRs in gen-release-notes

tnozicka:
GitHub date ranges are not inclusive and `gen-release-notes` was failing to find the first PR.
```
$ go run ./cmd/gen-release-notes --start-ref=$( git merge-base v1.3.0 v1.2.0 ) --end-ref=v1.3.0 --release-name=1.3.0 --github-token=${GITHUB_TOKEN}
Error: PR #648 not found in GitHub
```

### DIFF
--- a/pkg/cmd/releasenotes/options.go
+++ b/pkg/cmd/releasenotes/options.go
@@ -5,7 +5,6 @@ package releasenotes
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"strings"
 
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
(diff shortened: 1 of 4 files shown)


pkg/cmd/releasenotes/options.go
```diff
@@ -39,32 +39,38 @@ func NewGitGenerateOptions(streams genericclioptions.IOStreams) *GenerateOptions
 	}
 }
 
-func (o GenerateOptions) Validate() error {
+func (o *GenerateOptions) Validate() error {
```
zimnx, line 42: nit: validate should be read-only

pkg/cmd/releasenotes/options.go
```diff
@@ -39,32 +39,38 @@ func NewGitGenerateOptions(streams genericclioptions.IOStreams) *GenerateOptions
...
 
+	if len(o.GithubToken) == 0 {
+		errs = append(errs, fmt.Errorf("github-token can't be empty"))
+	}
+
 	return errors.NewAggregate(errs)
 }
 
 func (o *GenerateOptions) Complete(ctx context.Context) error {
 	var httpClient *http.Client
```
zimnx, line 69: nit: without the condition this var could be initialized inline in L73

pkg/cmd/releasenotes/options.go
tnozicka: shouldn't all receivers be consistent?

pkg/cmd/releasenotes/options.go
tnozicka: done

pkg/cmd/releasenotes/options.go
zimnx: it doesn't implement any interface so no ambiguity can happen. maybe consistency is bettter.

zimnx (approved):
LGTM, thanks

PR #711, opened 2021-07-20
[WIP] Add automatic replacement of dead nodes

### LINKED ISSUE #507: Automatically replace scylla node that looses storage

rzetelskik:
- Add an initContainer detecting dead nodes needing replacement
- Add setting "replace-address-first-boot" flag for auto-detected dead nodes in sidecar
- Add an e2e tests for auto-replacement of dead nodes

Resolves #507

**Requires:**
- #721

### DIFF
--- a/pkg/cmd/operator/auto_replace.go
+++ b/pkg/cmd/operator/auto_replace.go
@@ -1,0 +1,190 @@
+package operator
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path"
+
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/cmdutil"
+	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/scyllaclient"
+	"github.com/scylladb/scylla-operator/pkg/version"
+	"github.com/spf13/cobra"
+	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	apierrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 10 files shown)


rzetelskik:
@zimnx @tnozicka Please review.

pkg/cmd/operator/auto_replace.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil
+	}
+
+	// Create dummy cluster for helper functions
+	sc := &scyllav1.ScyllaCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      pod.Labels[naming.ClusterNameLabel],
+		Namespace: o.Namespace,
+	}}
+
+	scyllaClient, _, err := scyllaclient.GetScyllaClientForScyllaCluster(ctx, o.kubeClient.CoreV1(), sc)
```
tnozicka, line 161: hm, if we are taking just to the other nodes we could do it from the operator

pkg/cmd/operator/auto_replace.go
```diff
@@ -0,0 +1,190 @@
...
+	for i := range podList.Items {
+		p := &podList.Items[i]
+		if p.Name != pod.Name {
+			foundOther = true
+			break
+		}
+	}
+	if !foundOther {
+		// First node in cluster skips bootstrap, can't replace other node
+		klog.Infof("first node in cluster, skipping")
```
tnozicka, line 151: if there are 3 nodes and the first one looses storage, how does this help?

tnozicka (commented):
I am not sure if this is the right approach or not, but I'd like to see a proposal first with different alternatives considered. Might save time on coding side.

rzetelskik:
Closing this as it was submitted before considering #721. Will open a new PR if need be.

PR #715, opened 2021-07-22
Add PodDisruptionBudget for webhook server

tnozicka:
Adds PodDisruptionBudget for webhook server so it can't be evicted if the other pods are down.
Also switches the operator PDB to use more flexible `minAvailable`.


zimnx (approved):
LGTM, thanka

PR #721, opened 2021-07-23
[WIP] Node Autoreplacement proposal

rzetelskik:
- Add enhancement proposal for Node Autoreplacement

### DIFF
--- a/enhancements/proposals/node-autoreplacement.md
+++ b/enhancements/proposals/node-autoreplacement.md
@@ -1,0 +1,128 @@
+# Node Autoreplacement
+
+## Summary
+
+Node Autoreplacement enhancement extends Scylla Operator with a functionality of automated replacement of nodes which lost their storage, which, at this point, renders them getting stuck on joining the cluster.
+
+This proposal suggests creating a module responsible for triggering a node replacement in such circumstances.
+As a side effect, introducing such functionality would allow for simplifying the existing `OrphanedPVController` by
+reducing its task to evicting orphaned PVs.
+
+## Motivation
+
+Currently, when a node looses its storage, it gets stuck on joining a cluster, since Scylla produces a following runtime error:
+```
+Startup failed: std::runtime_error (A node with address <ip> already exists, cancelling join. Use replace_address if you want to replace this node.)
+```
+
+Successfully restarting the node requires a manual intervention of a human operator by annotating the
+service corresponding to the affected node with a `scylla/replace` label. The main goal of this proposal is to improve
+Scylla Operator's stability in running Scylla independently of its administrators' interventions.
+
+### Goals
+
+- Successfully discovering and autoreplacing nodes which lost their storage without requiring user intervention.
+
+### Non-Goals
+
+- Removing orphaned PVs. This is left for `OrphanedPVController`.
+
+## Proposal
+
+Extend ScyllaCluster controller with code responsible for checking whether a given Scylla node:
+1. is **not** a first bootstrapping node in the cluster
+2. already exists in gossip
+
+In case both of these conditions are true, a dedicated ConfigMap is created/updated.
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: <service_name>-autoreplacement-config
+  ...
+data:
+  autoreplacement.config: |
+    replace-address-first-boot: <clusterIP>    
+```
+
+The config is later appended to the generic config during Scylla setup, which effectively triggers running Scylla with `replace-address-first-boot` flag.
+The ConfigMap takes precedence over the generic config. In case `replace-address-first-boot` is already defined there, it is overwritten.
+
+The ConfigMap is optional and unavailable during the first start of a pod. It is not live reloaded and only read during the entrypoint setup, after every restart of a pod.
+The ScyllaCluster controller talks to the nodes whenever a generated ConfigMap differs from the existing one, which essentially comes down to two cases: after first boot and after a change to corresponding service's `ClusterIP`.
+
+It is necessary to check whether a node is first in the cluster, as the first node in the cluster skips bootstrap, and therefore the flag can't be set.
+Checking for the IP already existing in gossip is necessary as it is not possible to replace a node that does not.
+Setting `replace-address-first-boot` results in a desired output, i.e. a node which lost its storage will be replaced during bootstrap.
+Otherwise, setting the flag is safe as Scylla ignores it after bootstrap.
+
+Running a manual node replacement does not affect the autoreplacement process.
+
+Introducing this enhancement additionally allows for removing replacement functionality from `OrphanedPVController`. The controller would be simplified and only be responsible for removing the orphaned PVs.
+
+### User Stories
+
+#### Running ScyllaCluster on AWS/GCP
+
+The user is running a ScyllaCluster on AWS/GCP cluster with local storage. One of the instances is subject to an event resulting in erasing the local storage (see [AWS Instance store lifetime](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-lifetime) and [GCP Local SSD data persistance](https://cloud.google.com/compute/docs/disks/local-ssd#data_persistence))  .
+The Scylla node scheduled on this instance loses its storage, even though a PVC bound to it is not removed. The Scylla node gets replaced automatically.
+
+### Notes/Constraints/Caveats [Optional]
+
+- It would be preferred to use [Immutable ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable). It protects the system from any accidental updates or unwanted user interventions that could cause the cluster to misbehave. 
+- In case there is only one Scylla node in a cluster, the node can't be replaced, as `replace-address` and `replace-address-first-boot` are not allowed for seed nodes, and the singular node is its own seed.
+
+### Risks and Mitigations
+
+- The proposed design is based on a fact that `replace-address-first-boot` flag is ignored after bootstrap. Any changes to this behaviour or side effects introduced could result in an undefined behaviour of the enhancement. To mitigate, we could try to have it covered by tests in Scylla.
+
+## Design Details
+
+To check whether the Scylla node is not the first to bootstrap in a cluster and already exists in gossip, the controller uses [`ScyllaClient's Status`](https://github.com/scylladb/scylla-operator/blob/689c0a0a82aeea0fef72b2cd11af0333d1d2e180/pkg/scyllaclient/client.go#L98).
+Client is created with default config. Controller authenticates with the already known token. It uses all nodes of a given cluster as a pool of hosts. The conditions for creating the ConfigMap are met if the node's IP is among the returned hosts.
+
+`Sidecar` accesses the ConfigMap through a dedicated Volume. 
+
+### Test Plan
+
+- E2E test plan:
+  - Create a ScyllaCluster with 2 nodes
+  - Wait for the ScyllaCluster to deploy
+  - Remove first node's PVC
+  - Remove first node's Pod
+  - Await PVC's and Pod's deletion
+  - Wait for the ScyllaCluster to reconcile and become fully available again
+  - Verify ScyllaCluster
+
+### Upgrade / Downgrade Strategy
+
+No upgrade/downgrade action is needed, no manifests are changed.
+
+### Version Skew Strategy
+
+There might occur a version skew between a Scylla Operator and a ScyllaCluster deployed by a Scylla Operator of a different version.
+
+1. ScyllaCluster deployed with an unupgraded Scylla Operator. Upgraded Scylla Operator running: ScyllaCluster would ignore the ConfigMap.
+2. ScyllaCluster deployed with upgraded Scylla Operator. Unupgraded Scylla Operator running: no ConfigMaps would be created.
+
+In both cases the Node Autoreplacement feature would fail. The administrator would have to use the manual node replacement as a fallback.
+
+## Implementation History
+
+- 1.5, enhancement proposal created
+
+## Drawbacks
+
+- Controller needs to make API calls to Scylla nodes. However, it would only do so on first start of each node or if `ClusterIP` of a corresponding service changed. 
+
+## Alternatives
+
+- Alternatives to extending the controller:
+    - A new InitContainer could be added to the StatefulSet's `PodTemplateSpec`, although it seems more cumbersome compared to extending Scylla Operator's controller, as it would require performing many operations that the controller does anyway. 
+    - This functionality could be added to the existing sidecar, since triggering the replacement is dependent on the sidecar anyway. However, sidecar is meant to be removed in the future and such approach impedes the separation of concerns even further.
+- Instead of merging the `ConfigMap` with the general Scylla config, a different approach (mapping to a file, creating an environment variable) could be used to trigger setting the `replace-address-first-boot` flag.
+- `ScyllaClient` makes multiple API calls in order to get a status of each host. While this enhancement only needs to know whether a given node is in gossip, a simpler `ScyllaClient` method could be added.
+
+## Infrastructure Needed [optional]
+
+None.


enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,114 @@
...
+- Successfully discovering and autoreplacing nodes which lost their storage without requiring user intervention.
+
+### Non-Goals
+
+- Removing orphaned PVs. This is left for `OrphanedPVController`.
+
+## Proposal
+
+Create an InitContainer with a dedicated Scylla Operator command responsible for checking whether a given node is
+1. **not** first in the cluster
```
tnozicka, line 33: first in the sense of having index 0, or first bootstrapping node?

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,114 @@
...
+
+### User Stories
+
+#### Story 1
+
+The user is running a ScyllaCluster and removes a PVC bound to one of Scylla nodes. The node gets automatically replaced. 
+
+#### Story 2
+
+The user is running a ScyllaCluster on AWS/GCP cluster with a local storage. One of the instances gets restarted, which results in erasing the local storage.
```
tnozicka, line 53: I think restart keeps the local storage, suspend, hibernate and others don't

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,114 @@
...
+N/A
+
+### Risks and Mitigations
+
+- The proposed design is based on a fact that `replace-address-first-boot` flag is ignored after bootstrap. Any changes to this behaviour or side effects introduced could result in an undefined behaviour of the enhancement.
+
+
+## Design Details
+
+- The InitContainer uses the scylla-operator image.
```
tnozicka, line 67: I'd have expected of a description of who does it talk to to get the gossip information. If it's talking to the other nodes and not the local one, why should the code should live in a sidecar compared to the operator?

To which port does it talk to and how does it handle authentication?

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Implementation History
+
+- 1.5, enhancement proposal created
+
+## Drawbacks
+
+Introducing an additional InitContainer, as well as a new shared Volume is required.
+
+## Alternatives
```
tnozicka, line 108: one alternative could be a detection from the local filesystem - either reading scylla files or creating a marker file when the node appears in bootstrap or becomes ready for the first time. also possibly talking to the local REST API.

enhancements/proposals/node-autoreplacement.md
rzetelskik: Explicit restart of an instance with local SSD is technically not allowed, but I've done the following:
1. Deploy a GKE cluster with 2 nodes
2. Wait for it to deploy
3. ssh to one of the instances and run `sudo shutdown -h now`
4. Wait for it to get back up again
And it resulted with losing storage and the pod logging `Startup failed: std::runtime_error (A node with address 10.55.254.64 already exists, cancelling join. Use replace_address if you want to replace this node.)`

enhancements/proposals/node-autoreplacement.md
rzetelskik: first bootstrapping node, see https://github.com/scylladb/scylla/blob/b503ec36c22856cf9e1d02ff0055767001be6f66/service/storage_service.cc#L240

tnozicka: the answer was meant to be included in the document so it's not ambiguous

enhancements/proposals/node-autoreplacement.md
tnozicka: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html
https://cloud.google.com/compute/docs/disks/local-ssd#data_persistence

rzetelskik:
@tnozicka I've changed the approach described in the proposal

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+
+- The proposed design is based on a fact that `replace-address-first-boot` flag is ignored after bootstrap. Any changes to this behaviour or side effects introduced could result in an undefined behaviour of the enhancement.
+
+## Design Details
+
+To check whether the node is not the first to bootstrap in a cluster and already exists in gossip, the controller uses [`ScyllaClient's Status`](https://github.com/scylladb/scylla-operator/blob/689c0a0a82aeea0fef72b2cd11af0333d1d2e180/pkg/scyllaclient/client.go#L98).
+Client is created with default config. Controller authenticates with the already known token. It communicates with all nodes of a given cluster. The conditions for creating the `ConfigMap` are met if the node's IP is among the returned hosts.
+
+`Sidecar` accesses the `ConfigMap` through a dedicated `Volume`.
+
```
tnozicka, line 83: details I am missing mentioned here:
- is the configmap optional or required
- is the CM live reloaded
- will any of it result in spec update / rollout
- possible races
- in which cases the controller talks to the nodes, what is the condition to not ask anymore
- how is the CM (holding the IP) updated when a node is replaced manually (with the replace label)
- what happens when a manual replace is already in progress
- what happens if a user config contains the replace-address-first-boot option (not that we should support it but the behaviour should be defined - like our takes precedence)

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+### Test Plan
+
+- e2e/integration tests
```
tnozicka, line 86: we don't have integration tests, just e2e

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+     2.
+          - Create a `ScyllaCluster` with 2 nodes
+          - Wait for `ScyllaCluster` to deploy
+          - Turn on maintenance mode, Exec `systemctl stop scylla && rm -rf /var/lib/scylla` in first node's Pod
+          - Remove first node's Pod
+          - Await Pod's deletion
+          - Wait for the `ScyllaCluster` to redeploy
+          - Verify `ScyllaCluster`
+    
+More fitting e2e tests could be created once the tests are run against a real distribution instead of Minikube.
```
tnozicka, line 104: How does it matter for this feature? I see the boundary as "a PV gets deleted", the node relation ship matters for the OrphanedPVController

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+### Test Plan
+
+- e2e/integration tests
+     1.
+          - Create a `ScyllaCluster` with 2 nodes
+          - Wait for the `ScyllaCluster` to deploy
+          - Remove first node's PVC
+          - Remove first node's Pod
+          - Await PVC's and Pod's deletion
+          - Wait for the `ScyllaCluster` to redeploy
```
tnozicka, line 93: nit: redeploy seems slightly misleading here, `Wait for the ScyllaCluster to reconcile and become fully available again`

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Goals
+
+- Successfully discovering and autoreplacing nodes which lost their storage without requiring user intervention.
+
+### Non-Goals
+
+- Removing orphaned PVs. This is left for `OrphanedPVController`.
+
+## Proposal
```
tnozicka, line 30: explicitly mention what happens to a scyllacluster with 1 node (already bootstrapped)

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+          - Wait for the `ScyllaCluster` to deploy
+          - Remove first node's PVC
+          - Remove first node's Pod
+          - Await PVC's and Pod's deletion
+          - Wait for the `ScyllaCluster` to redeploy
+          - Verify `ScyllaCluster`
+     2.
+          - Create a `ScyllaCluster` with 2 nodes
+          - Wait for `ScyllaCluster` to deploy
+          - Turn on maintenance mode, Exec `systemctl stop scylla && rm -rf /var/lib/scylla` in first node's Pod
```
tnozicka, line 98: is this simulating hot swapping disks? would a scylla process be stopped in that case? otherwise this would look pretty similar to the first case where a PVC is deleted and it get new and empty disk.

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+More fitting e2e tests could be created once the tests are run against a real distribution instead of Minikube.
+
+### Upgrade / Downgrade Strategy
+
+This enhancement adds a `ConfigMap` per `ScyllaCluster`'s Pod and extends `StatefulSetSpec` with a dedicated `Volume` to access the config.
+It does not cause any drawbacks on a downgrade.
+
+### Version Skew Strategy
+
+This enhancement only requires changes in the Scylla Operator image. There is no version skew involved.
```
tnozicka, line 113: this means operator N+1 vs. scyllacluster deployed with operator N and vice versa. Given you are changing the podSpec and creating a new object (CM) there is a skew in how the deployment looks like in each version. Describe what happens when the new operator sees the old deployment and vice versa.

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+          - Remove first node's Pod
+          - Await Pod's deletion
+          - Wait for the `ScyllaCluster` to redeploy
+          - Verify `ScyllaCluster`
+    
+More fitting e2e tests could be created once the tests are run against a real distribution instead of Minikube.
+
+### Upgrade / Downgrade Strategy
+
+This enhancement adds a `ConfigMap` per `ScyllaCluster`'s Pod and extends `StatefulSetSpec` with a dedicated `Volume` to access the config.
```
tnozicka, line 108: This is meant in relation to an admin managing the scylla-operator. So something like: `No upgrade/downgrade action is needed, no manifests are changed.`

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+
+This enhancement only requires changes in the Scylla Operator image. There is no version skew involved.
+
+## Implementation History
+
+- 1.5, enhancement proposal created
+
+## Drawbacks
+
+- Controller needs to make API calls to Scylla nodes.
```
tnozicka, line 121: if there is a mitigation, like it's very infrequent, it's worth mentioning it as well

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Alternatives to extending the controller:
+    - A new InitContainer could be added to the StatefulSet's `PodTemplateSpec`, although it seems more cumbersome compared to extending Scylla Operator's controller, as it would require performing many operations that the controller does anyway. 
+    - This functionality could be added to the existing sidecar, since triggering the replacement is dependent on the sidecar anyway. However, sidecar is meant to be removed in the future and such approach impedes the separation of concerns even further.
+- Instead of merging the `ConfigMap` with the general Scylla config, a different approach (mapping to a file, creating an environment variable) could be used to trigger setting the `replace-address-first-boot` flag.
+- `ScyllaClient` makes multiple API calls in order to get a status of each host. While this enhancement only needs to know whether a given node is in gossip, a simpler `ScyllaClient` method could be added.
+
+## Infrastructure Needed [optional]
+
+N/A
```
tnozicka, line 133: None.

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+kind: ConfigMap
+metadata:
+  name: <podName>-autoreplacement-config
+  ...
+data:
+  autoreplacement.config: |
+    replace-address-first-boot: <clusterIP>    
+```
+
+The config is later merged with generic config during Scylla setup, which effectively triggers running Scylla with `replace-address-first-boot` flag.
```
tnozicka, line 48: mention the config order

enhancements/proposals/node-autoreplacement.md
tnozicka: also what happens if the scyllacluster has lost the storage before the operator was upgraded

tnozicka:
I think the approach looks viable overall, needs a few details that should be flashed out in the proposal.

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Story 2
+
+The user is running a ScyllaCluster on AWS/GCP cluster with a local storage. One of the instances is subject to an event resulting in erasing the local storage (see [AWS Instance store lifetime](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-lifetime) and [GCP Local SSD data persistance](https://cloud.google.com/compute/docs/disks/local-ssd#data_persistence))  .
+A node scheduled on this instance loses its storage, even though a PVC bound to it is not removed. The node gets replaced automatically.
+
+### Notes/Constraints/Caveats [Optional]
+
+- It would be preferred to use [Immutable ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable).
+- Having introduced this feature, we might consider removing the existing functionality of [replacing dead nodes manually](https://operator.docs.scylladb.com/master/nodeoperations/replace_node.html).
```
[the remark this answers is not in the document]
rzetelskik: With this approach we would run Scylla with `replace-address-first-boot` set at all times after the node has bootstrapped, so I don't see any cases in which setting the label manually could change anything. Do you have anything specific in mind?

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,133 @@
...
+Setting `replace-address-first-boot` results in a desired output, i.e. a node which lost its storage will be replaced during bootstrap.
+Otherwise, setting the flag is safe as Scylla ignores it after bootstrap.
+
+Introducing this enhancement additionally allows for removing replacement functionality from `OrphanedPVController`. The controller would be simplified and only be responsible for removing the orphaned PVs.
+
+### User Stories
+
+#### Story 1
+
+The user is running a ScyllaCluster and looses a PVC bound to one of Scylla nodes. The node gets automatically replaced.
```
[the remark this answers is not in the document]
rzetelskik: I've removed that user story as it's almost a duplicate of the AWS/GCP one essentially

enhancements/proposals/node-autoreplacement.md
rzetelskik: You're right. I've removed the second test.

rzetelskik:
@tnozicka I've addressed all your comments. I'm not entirely sure about how this enhancement could be affected by running the manual replacement. I don't see any potential issues/races atm but I'm happy to discuss

enhancements/proposals/node-autoreplacement.md
```diff
@@ -0,0 +1,128 @@
...
+data:
+  autoreplacement.config: |
+    replace-address-first-boot: <clusterIP>    
+```
+
+The config is later appended to the generic config during Scylla setup, which effectively triggers running Scylla with `replace-address-first-boot` flag.
+The ConfigMap takes precedence over the generic config. In case `replace-address-first-boot` is already defined there, it is overwritten.
+
+The ConfigMap is optional and unavailable during the first start of a pod. It is not live reloaded and only read during the entrypoint setup, after every restart of a pod.
+The ScyllaCluster controller talks to the nodes whenever a generated ConfigMap differs from the existing one, which essentially comes down to two cases: after first boot and after a change to corresponding service's `ClusterIP`.
```
tnozicka, line 52: Won't the ConfigMaps differ through out the whole bootstrap process? That's currently like 1-2 minutes per node. Making manager API calls on every sync during that time is gonna be extremely costly. We need some condition that's based on the API object - e.g. could we only create the CM for nodes that are ready (and update always)? Guess we might not need to investigate the gossip at all then because a node ready implies presence in gossip.

enhancements/proposals/node-autoreplacement.md
tnozicka: I don't think I saw a mention of precedence with manual replace using the label

rzetelskik:
Closing as stale

PR #736, opened 2021-08-03
Add proposal about improving UX around disks setup and benchmark

zimnx:


enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,186 @@
...
+
+Scylla Operator is going to reconcile Disk Provisioner and NodeTuner DaemonSet based on values in available ScyllaOperatorConfig.
+It's going to create a NodeTuner DaemonSet even when no Scylla Pods are running.
+
+If `ScyllaOperatorConfig.DiskSetup.SetupLocalDisks` is true NodeTuner DaemonSet is going to discover disks attached 
+to the instance and will setup RAID0, format to XFS and will mount it in the location specified in `ScyllaOperatorConfig.DiskSetup.LocalDisksHostPath`.
+
+When new disk appears in specified location, Disk Provisioner will expose them as PersistentVolumes.
+
+Node Tuner DaemonSet is going to have a special finalizer in PodTemplateSpec. Operator will add a special label to each Node
```
tnozicka, line 144: I wonder if we can find out a more elegant solution that doesn't include the sidecar and giving up perms to node objects. Worst case the operator could watch where the pod landed and create an empty CM if that's a node without iosetup.

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,186 @@
...
+running Node Tuner Pod.
+When Node Tuner Pod is evicted, Operator will first remove label from the Node, and then clear the finalizer.
```
tnozicka, line 146: if a DS pod is evicted a new one is created in its place, so eviction isn't the event to trigger on

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,186 @@
...
+Scylla Pods are going to check if Node where they're running is labelled and if so,
+wait until ConfigMap with `iosetup-<pv-uid>` is created in the same namespace.
+
+#### User provided Disk Provisioner
+
+User may have his own Disk Provisioner. In such case, he should disable installing our provisioner by setting
```
tnozicka, line 152: `he might want to disable` - the other option is to have 2 of them and non-intersecting node selectors

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,186 @@
...
+`DiskSetup.InstallDiskProvisioner` to `false`.
+If `DiskSetup.SetupLocalDisk` is enabled, he should also sync `DiskSetup.LocalDisksHostPath` with his provisioner config.
```
tnozicka, line 154: nit: `s/sync/make sure the paths are identical

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Test Plan
+
+Due to lack of local storage available in e2e it won't be possible to test the disk setup. This will have to be covered by the QA test suite.
```
tnozicka, line 158: when we have a basic gce e2e job we'll cover the PV provisioner, just the disks will be setup in dirs like https://github.com/scylladb/scylla-operator/pull/694

for the rest we should setup periodics where we can afford the real disks

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,186 @@
...
+NodeTuner DaemonSet will have system admin privileges, although it won't be exposed to end users.
+Data loss is possible in case of a bug in disk formatting logic.
+
+## Design Details
+
+### API
+ScyllaOperatorConfig is going to be extended with following API:
+
+```
+
```
tnozicka, line 93: there need to be status fields communication the operand's state

tnozicka (commented):
had few comments but I am not fully decided myself. I wonder if it would make sense if we should support iotune on nodes that don't have the DaemonSet, e.g. on demand and slower but to save the result.

Proposal could use some details around configmap naming or selectors.

enhancements/proposals/disk_setup.md
zimnx: I don't like it either, i think i have a better idea. 
I would like to keep a list of tuned nodes in ScyllaOperatorConfig status. Scylla Pods could check if node they are running (taken from Pod.spec.nodeName) is on the list, and if optimizations are enabled. If so, they would wait until ConfigMap is created.

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,225 @@
...
+Because for disk benchmarking not only PV access is required, but also CPU resources, we are going to inject 
+commands into Scylla Pod. We cannot use Scylla to run it because in one case an additional parameter is needed. 
+Details are described in separate paragraph.
+
+### User Stories
+
+#### Default Cloud Provider setup
+
+On currently supported Cloud Providers (EKS, GKE) default instances have not formatted and unmounted local drives.
+When user install Scylla Operator disks will be configured and exposed as PV automatically. Setting `local-raid-disks`
```
rzetelskik, line 66:
```suggestion
When user installs Scylla Operator, disks will be configured and exposed as PV automatically. Setting `local-raid-disks` 
```

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,225 @@
...
+on their types of machines.
+
+The first two of those require root access to host machine, third one requires read access to Pod storage.
+These are going to be run by DaemonSet described in [perftune.md](perftune.md), because it 
+has root access to host machine, and it's running on every k8s node having Scylla Pods, so it will
+be able to access both local PVs and network attached PVs. 
+The third one requires access to Pod storage and ability to create ConfigMaps. 
+
+For setting up the disks we are going to execute `scylla_raid_setup` script. 
+This will create a RAID0 array and will format partition to XFS.
```
rzetelskik, line 52: I believe this section needs a description of how the DaemonSet is going to recognize which cloud it's running on and whether we support it. #760 has recently proven that currently it might be impossible to make the disk setup cloud agnostic.

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,225 @@
...
+### Precomputed benchmark
+
+`scylla_io_setup` script which contains logic for detecting and calculating precomputed values for known instances requires
+special parameter (--ami) in order to skip actual benchmark on AWS and to use precomputed values. 
+This parameter has unfortunate behavior when AWS instances are *not* in recommended set (i3, i3en, and more) - it just exits with 1 code.  
+So before running this script with this parameter we have to be sure that we are on AWS and instance is recommended.  
+
+ScyllaNodeTuner Pod is going to detect if we are running on AWS, and if instance type is one of the recommened types 
+(we will hardcode the list based on `scylla_util`). If so an `--ami` parameter is added to `scylla_io_setup`.
+GKE and other Clouds and deployments doesn't require any additional flags.
```
rzetelskik, line 186:
```suggestion
GKE and other Clouds and deployments don't require any additional flags.  
```

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,225 @@
...
+`scylla_io_setup` script which contains logic for detecting and calculating precomputed values for known instances requires
+special parameter (--ami) in order to skip actual benchmark on AWS and to use precomputed values. 
+This parameter has unfortunate behavior when AWS instances are *not* in recommended set (i3, i3en, and more) - it just exits with 1 code.  
+So before running this script with this parameter we have to be sure that we are on AWS and instance is recommended.  
+
+ScyllaNodeTuner Pod is going to detect if we are running on AWS, and if instance type is one of the recommened types 
+(we will hardcode the list based on `scylla_util`). If so an `--ami` parameter is added to `scylla_io_setup`.
+GKE and other Clouds and deployments doesn't require any additional flags.  
+
+If `ScyllaNodeTuner.DiskBenchmarkOptions.IgnorePrecomputed` is `true`, instead running `scylla_io_setup` ScyllaNodeTuner will direcly execute `iotune`.
```
rzetelskik, line 188:
```suggestion
If `ScyllaNodeTuner.DiskBenchmarkOptions.IgnorePrecomputed` is `true`, instead of running `scylla_io_setup` ScyllaNodeTuner will direcly execute `iotune`.
```

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### User provided Disk Provisioner
+
+User may have his own Disk Provisioner. In such case, he might want disable our provisioner by setting
```
rzetelskik, line 192:
```suggestion
User may have his own Disk Provisioner. In such case, he might want to disable our provisioner by setting
```

enhancements/proposals/disk_setup.md
```diff
@@ -0,0 +1,225 @@
...
+`ScyllaNodeTuner.PVProvisionerDisabled` to `true`.
+If `ScyllaNodeTuner.DiskSetupDisabled` is not disabled, he should also make sure that his Disk Provisioner is using `ScyllaNodeTuner.LocalDiskPath`.
```
rzetelskik, line 194: > `ScyllaNodeTuner.DiskSetupDisabled` is not disabled 

That's a little confusing

PR #740, opened 2021-08-03
Add automatic update of CRD schemas in helm charts' schemas

### LINKED ISSUE #631: scylla manager chart doesn't allow to provide placement for it's scylla subchart
**Describe the bug**
Currently it is not possible to tell scylla manager chart where it's scylla must be scheduled.

If we try to provide following config in values.yaml file:

```
scylla:
  developerMode: true
  scyllaImage:
    tag: 4.3.0
  agentImage:
    tag: 2.2.1
  datacenter: manager-dc
  racks:
  - name: manager-rack
    members: 1
    placement:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
          - matchExpressions:
            - key: cloud.google.com/gke-nodepool
              operator: In
              values:
              - scylla-pool
    storage:
      capacity: 5Gi
    resources:
      limits:
        cpu: 1
        memory: 200Mi
      requests:
        cpu: 1
        memory: 200Mi
```

where "placement" field for manager's scylla subchart is specified then we get following error:

```
Error: unable to build kubernetes objects from release manifest: error validating "": 
    error validating data: ValidationError(ScyllaCluster.spec.datacenter.racks[0].placement.nodeAffinity): 
        unknown field "nodeAffinity" in com.scylladb.scylla.v1.ScyllaCluster.spec.datacenter.racks.placement.nodeAffinity
```

Also, it is not possible to update placement of an existing "scyllacluster" CRD which gets created by manager helm chart with following error:

```
Error from server (rack manager-rack: changes in placement are not currently supported): 
    admission webhook "webhook.scylla.scylladb.com" denied the request: 
        rack manager-rack: changes in placement are not currently supported
```

**To Reproduce**
1. Deploy operator
2. Deploy scylla manager using helm chart with "placement" field for scylla subchart.
3. See error

**Expected behavior**
It is expected that all the scylla subchart values are supported by scylla-manager helm chart.

**Config Files**
If relevant, upload your configuration files here using GitHub, there is no need to upload them to any 3rd party services

**Logs**
Please, provide `kubectl get events`, `kubectl logs -n scylla pod-name` and other relevant information.

**Environment:**
 - Platform: GKE
 - Kubernetes version: 1.19.9-gke.1400
 - Scylla version: 4.4.2
 - Scylla-operator version: e.g.: v1.3.0-beta.0-0-gc4dcd8d

**Additional context**

rzetelskik:
- Add a Makefile recipe for updating values.schema.yaml files for `scylla` and `scylla-manager` charts with autogenerated CRD's schema
- Update values.schema.yaml files for `scylla` and `scylla-manager`

Resolves #631

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,6 +247,38 @@ verify-crd:
 	$(diff) '$(tmp_file)' '$(CRD_PATH)' || (echo 'File $(CRD_PATH) is not up-to date. Please run `make update-crd` to update it.' && false)
 .PHONY: verify-crd
 
+# $1 - target file
+define generate-helm-schema-scylla
+	$(YQ) eval -j '{"$$schema": "http://json-schema.org/schema#"} * (.spec.versions[] | select(.name == "v1") | \
+	 .schema.openAPIV3Schema.properties.spec) | \
+	 .properties.racks=.properties.datacenter.properties.racks | \
+	 .properties.datacenter={"type": "string"}' $(CRD_PATH) > '$(1)'
+endef
+
+# $1 - Scylla schema
+# $2 - target file
+define generate-helm-schema-scylla-manager
+	$(YQ) eval-all -j 'select(fi==0).properties.scylla = ( \
+		select(fi==1) | del(.$$schema) \
+	) | select(fi==0)' '$(HELM_CHARTS_DIR)/scylla-manager/values.schema.template.json' '$(1)' > '$(2)'
+endef
+
+update-helm-schemas:
+	$(call generate-helm-schema-scylla,'$(HELM_CHARTS_DIR)/scylla/values.schema.json')
+	$(call generate-helm-schema-scylla-manager,'$(HELM_CHARTS_DIR)/scylla/values.schema.json','$(HELM_CHARTS_DIR)/scylla-manager/values.schema.json')
+.PHONY: update-helm-schemas
+
+verify-helm-schemas: tmp_dir:=$(shell mktemp -d)
+verify-helm-schemas:
+	mkdir -p $(tmp_dir)/{scylla,scylla-manager}
+
+	$(call generate-helm-schema-scylla,'$(tmp_dir)/scylla/values.schema.json')
+	$(diff) '$(tmp_dir)/scylla/values.schema.json' '$(HELM_CHARTS_DIR)/scylla/values.schema.json'
+
+	$(call generate-helm-schema-scylla-manager,'$(HELM_CHARTS_DIR)/scylla/values.schema.json','$(tmp_dir)/scylla-manager/values.schema.json')
+	$(diff) '$(tmp_dir)/scylla-manager/values.schema.json' '$(HELM_CHARTS_DIR)/scylla-manager/values.schema.json'
+.PHONY: verify-helm-schemas
+
 # $1 - name
 # $2 - chart path
 # $3 - values path
@@ -400,10 +432,10 @@ verify-codegen:
 	$(call run-informer-gen,--verify-only)
 .PHONY: verify-codegen
 
-verify: verify-gofmt verify-codegen verify-crd verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint
+verify: verify-gofmt verify-codegen verify-crd verify-helm-schemas verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint
 .PHONY: verify
 
-update: update-gofmt update-codegen update-crd update-helm-charts update-deploy update-examples
+update: update-gofmt update-codegen update-crd update-helm-schemas update-helm-charts update-deploy update-examples
 .PHONY: update
 
 test-unit:
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -54,7 +54,7 @@ jobs:
     - name: Install tools
       run: |
         set -x
-        go install github.com/mikefarah/yq/v4@v4.6.1
+        go install github.com/mikefarah/yq/v4@v4.11.2
     - name: Verify
       run: make verify --warn-undefined-variables
 
--- a/helm/scylla-manager/values.schema.json
+++ b/helm/scylla-manager/values.schema.json
@@ -1,204 +1,2540 @@
 {
-    "$schema": "http://json-schema.org/schema#",
-    "type": "object",
-    "properties": {
-        "affinity": {
-            "type": "object"
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "properties": {
+    "affinity": {
+      "type": "object"
+    },
+    "controllerAffinity": {
+      "type": "object"
+    },
+    "controllerImage": {
+      "type": "object",
+      "properties": {
+        "pullPolicy": {
+          "type": "string"
         },
-        "controllerAffinity": {
-            "type": "object"
+        "repository": {
+          "type": "string"
         },
-        "controllerImage": {
-            "type": "object",
-            "properties": {
-                "pullPolicy": {
-                    "type": "string"
-                },
-                "repository": {
-                    "type": "string"
-                },
-                "tag": {
-                    "type": "string"
-                }
+        "tag": {
+          "type": "string"
+        }
+      }
+    },
+    "controllerResources": {
+      "type": "object",
+      "properties": {
+        "requests": {
+          "additionalProperties": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "integer"
+              }
+            ]
+          },
+          "type": "object"
+        },
+        "limits": {
+          "additionalProperties": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "integer"
+              }
+            ]
+          },
+          "type": "object"
+        }
+      }
+    },
+    "controllerSecurityContext": {
+      "type": "object"
+    },
+    "image": {
+      "type": "object",
+      "properties": {
+        "pullPolicy": {
+          "type": "string"
+        },
+        "repository": {
+          "type": "string"
+        },
+        "tag": {
+          "type": "string"
+        }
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -246,6 +246,13 @@ verify-crd:
 	$(diff) '$(tmp_file)' '$(CRD_PATH)' || (echo 'File $(CRD_PATH) is not up-to date. Please run `make update-crd` to update it.' && false)
 .PHONY: verify-crd
 
+# $1 - Helm charts dir
+define update-helm-schemas
+	@crd_schema=$$( $(YQ) eval -j '.spec.versions[0].schema.openAPIV3Schema | del(..|.description?)' $(CRD_PATH) ); \
```
tnozicka, line 251: What happens if we add v2alpha1 to the CRD? I guess you should delete the description in all versions (and select the correct one) and it may not be enough in the future still. yq is missing `--argfile` argument so the env will eventually grow over the system limit to pass through. Funny enough I was swearing at that today when generating OLM manifests but I got away with higher reduction (`$(YQ) eval 'del(.spec.versions[].schema.openAPIV3Schema.properties)`).

Makefile
```diff
@@ -246,6 +246,13 @@ verify-crd:
...
+	crd_schema=$${crd_schema} $(YQ) eval -i -j 'env(crd_schema) * select(fileIndex == 0)' $(1)/scylla/values.schema.json; \
```
tnozicka, line 252: why selecting the fileindex when there is only 1 file? is there a reason why we can't fully generate the file?

Makefile
```diff
@@ -326,13 +333,18 @@ define set-default-app-version
...
 	$(call set-default-app-version,helm/scylla-operator,$(IMAGE_TAG))
 	$(call set-default-app-version,helm/scylla-manager,$(IMAGE_TAG))
 .PHONY: update-helm-charts
 
 verify-helm-charts: tmp_dir:=$(shell mktemp -d)
 verify-helm-charts:
-	cp -r helm/scylla-{operator,manager} '$(tmp_dir)'
+	cp -r helm/scylla helm/scylla-{operator,manager} '$(tmp_dir)'
+
+	$(call update-helm-schemas, $(tmp_dir))
```
tnozicka, line 345: would it be better to have a separate targets and verify it by creating the schema file from scratch?

Makefile
rzetelskik: We could easily generate the schema for `scylla` chart and I'm going to use this approach. As for the `scylla-manager` schema, I don't see a nice way to generate the entire file besides generating it from values, but that sort of defeats its purpose. I suppose we could have a `template.values.schema.json` file that we could merge with the generated `scylla` schema and output to `values.schema.json`. Wdyt?

Makefile
rzetelskik: That was the first approach I was going for but then `verify-helm-charts` compares the entire directories anyway. Should we just have a separate target that would compare the schemas first? Should `verify-helm-charts` exclude schemas?

Makefile
rzetelskik: Changed to `(.spec.versions[] | select(.name == "v1")`. This is UB (well, it will most likely generate an invalid json) in case there is two versions with the same name. Figured this is better than using `any` (or any other way to do get the same result in `yq` as `any` is missing). The target is now generating the entire file and not using variables.

Makefile
```diff
@@ -246,6 +246,41 @@ verify-crd:
...
+	 .properties.racks=.properties.datacenter.properties.racks |\
+	 .properties.datacenter={"type": "string"}' $(CRD_PATH) > '$(1)'
+endef
+
+# $1 - Scylla schema
+# $2 - target file
+define generate-helm-scylla-manager-schema
+	$(YQ) eval-all -j 'select(fi==0).properties.scylla = (select(fi==1) |\
+	del(.$$schema)) |\
+	select(fi==0)' $(HELM_CHARTS_DIR)/scylla-manager/.template.values.schema.json '$(1)' > '$(2)'
```
zimnx, line 263: maybe filename without a dot for better dev experience

rzetelskik:
@zimnx thanks, I've addressed your comments

Makefile
```diff
@@ -247,6 +247,40 @@ verify-crd:
...
+	$(YQ) eval -j '{"$$schema": "http://json-schema.org/schema#"} * (.spec.versions[] | select(.name == "v1") | \
+	 .schema.openAPIV3Schema.properties.spec) | \
+	 .properties.racks=.properties.datacenter.properties.racks | \
+	 .properties.datacenter={"type": "string"}' $(CRD_PATH) > '$(1)'
+endef
+
+# $1 - Scylla schema
+# $2 - target file
+define generate-helm-schema-scylla-manager
+	$(YQ) eval-all -j 'select(fi==0).properties.scylla = (select(fi==1) | \
```
tnozicka, line 261: nit: formatting and quotes
```
	$(YQ) eval-all -j 'select(fi==0).properties.scylla = ( \
		select(fi==1) | del(.$$schema) \
	) | select(fi==0)' '$(HELM_CHARTS_DIR)/scylla-manager/template.values.schema.json' '$(1)' > '$(2)'
```

Makefile
```diff
@@ -247,6 +247,40 @@ verify-crd:
...
+update-helm-schemas:
+	$(call generate-helm-schema-scylla,'$(HELM_CHARTS_DIR)/scylla/values.schema.json')
+	$(call generate-helm-schema-scylla-manager,'$(HELM_CHARTS_DIR)/scylla/values.schema.json','$(HELM_CHARTS_DIR)/scylla-manager/values.schema.json')
+.PHONY: update-helm-schemas
+
+verify-helm-schemas: tmp_dir:=$(shell mktemp -d)
+verify-helm-schemas:
+	mkdir -p $(tmp_dir)/{scylla,scylla-manager}
+
+	cp '$(HELM_CHARTS_DIR)/scylla/values.schema.json' '$(tmp_dir)/scylla/values.schema.json'
```
tnozicka, line 275: why do we need a pre-copy of the file we are writing into?

Makefile
```diff
@@ -247,6 +247,40 @@ verify-crd:
...
+	$(call generate-helm-crd-schema,'$(tmp_dir)/scylla/values.schema.json')
```
tnozicka, line 276: `Makefile:275: warning: undefined variable 'generate-helm-crd-schema'`
https://github.com/scylladb/scylla-operator/pull/740/checks?check_run_id=3281663722#step:5:18

helm/scylla-manager/template.values.schema.json
```diff
@@ -0,0 +1,133 @@
+{
```
tnozicka, line 1: should be in the other commit

Makefile
rzetelskik: leftovers from previous version that went unnoticed, sorry

Makefile
rzetelskik: fixed

Makefile
```diff
@@ -247,6 +247,38 @@ verify-crd:
...
+	 .properties.racks=.properties.datacenter.properties.racks | \
+	 .properties.datacenter={"type": "string"}' $(CRD_PATH) > '$(1)'
+endef
+
+# $1 - Scylla schema
+# $2 - target file
+define generate-helm-schema-scylla-manager
+	$(YQ) eval-all -j 'select(fi==0).properties.scylla = ( \
+		select(fi==1) | del(.$$schema) \
+	) | select(fi==0)' '$(HELM_CHARTS_DIR)/scylla-manager/template.values.schema.json' '$(1)' > '$(2)'
```
tnozicka, line 263: nit: value.schema.template.json (for ordering)

Makefile
```diff
@@ -247,6 +247,38 @@ verify-crd:
...
+	$(call generate-helm-schema-scylla,'$(HELM_CHARTS_DIR)/scylla/values.schema.json')
+	$(call generate-helm-schema-scylla-manager,'$(HELM_CHARTS_DIR)/scylla/values.schema.json','$(HELM_CHARTS_DIR)/scylla-manager/values.schema.json')
+.PHONY: update-helm-schemas
+
+verify-helm-schemas: tmp_dir:=$(shell mktemp -d)
+verify-helm-schemas:
+	mkdir -p $(tmp_dir)/{scylla,scylla-manager}
+
+	$(call generate-helm-schema-scylla,'$(tmp_dir)/scylla/values.schema.json')
+	$(diff) '$(HELM_CHARTS_DIR)/scylla/values.schema.json' '$(tmp_dir)/scylla/values.schema.json'
```
tnozicka, line 276: nit: the diff should be the other way around (diff expected, got)

Makefile
```diff
@@ -247,6 +247,38 @@ verify-crd:
...
+
+	$(call generate-helm-schema-scylla-manager,'$(tmp_dir)/scylla/values.schema.json','$(tmp_dir)/scylla-manager/values.schema.json')
```
tnozicka, line 278: `$(HELM_CHARTS_DIR)/scylla/values.schema.json`

Makefile
```diff
@@ -247,6 +247,38 @@ verify-crd:
...
+	$(diff) '$(HELM_CHARTS_DIR)/scylla-manager/values.schema.json' '$(tmp_dir)/scylla-manager/values.schema.json'
```
tnozicka, line 279: nit: the diff should be the other way around

helm/scylla-manager/template.values.schema.json
tnozicka: I guess I missed that it isn't generated - it should be in the `Add automatic update of CRD schemas in helm charts' schemas` then

rzetelskik:
@tnozicka thanks, fixed

tnozicka (dismissed):
/lgtm
thanks!

rzetelskik:
rebased on current master

zimnx (approved):

PR #741, opened 2021-08-04
Explicitly set Scylla's metrics listen IP

### LINKED ISSUE #622: Explicitly set metrics listen IP
We should explicitly set metrics listen IP, currently scylla defaults to `0.0.0.0` so it works, but it is about to be changed to localhost.

We should also backport it to all supported releases so the work with newer scylla.

rzetelskik:
- Explicitly set Scylla's metrics listen IP (`prometheus-address` flag) to `0.0.0.0`, because the default is going to be changed to `localhost`.

Resolves #622

### DIFF
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -211,6 +211,7 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 
 	// Listen on all interfaces so users or a service mesh can use localhost.
 	listenAddress := "0.0.0.0"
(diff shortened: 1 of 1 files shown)


zimnx (commented):
LGTM, @tnozicka is this what you had in mind? Scylla master currently still use 0.0.0.0 as default value for prometheus address, I wonder when it's going to change.

tnozicka (approved):

tnozicka:
/lgtm
thanks

I don't recall when but we have been warned it will. Using the explicit value will be safer anyways.

zimnx (approved):

PR #743, opened 2021-08-05
[v1.3] Explicitly set Scylla's metrics listen IP

rzetelskik:
backport of #741


rzetelskik:
The default `prometheus-address` value change hasn't landed yet, there's no need to backport atm.

PR #748, opened 2021-08-06
Tag Helm Charts with tagged versions

### LINKED ISSUE #746: Use tagged version for Helm Chart version
Charts published by CI are using `git describe` format for Chart Version. Instead, for tagged versions we should use tag name the same we do for container images.

zimnx:
Fixes #746

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ CODEGEN_GROUPS_VERSIONS ?="scyllaclusters/v1"
 GO_REQUIRED_MIN_VERSION ?=1.16
 
 GIT_TAG ?=$(shell git describe --long --tags --abbrev=7 --match 'v[0-9]*' || echo 'v0.0.0-unknown')
+GIT_TAG_SHORT ?=$(shell git describe --tags --abbrev=7 --match 'v[0-9]*' || echo 'v0.0.0-unknown')
 GIT_COMMIT ?=$(shell git rev-parse --short "HEAD^{commit}" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


zimnx:
Tested on fork:
- tagged build sets HELM_CHART_VERSION variable to tagged version and hence override variable in Makefile. https://github.com/zimnx/scylla-operator/runs/3256214153?check_suite_focus=true 
- builds on master and feature branches doesn't set HELM_CHART_VERSION at all, so `GIT_TAG` is used.
https://github.com/zimnx/scylla-operator/runs/3256213562?check_suite_focus=true

.github/workflows/go.yaml
```diff
@@ -259,9 +258,18 @@ jobs:
...
+
         source ./hack/lib/tag-from-gh-ref.sh
+        source ./hack/lib/semver.sh
+
         IMAGE_TAG=$( tag_from_gh_ref "${GITHUB_REF}" )
         echo "IMAGE_TAG=${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
+
+        # Helm Chart version must be in semantic format.
+        if is_semver "${IMAGE_TAG}" ; then
+          echo "HELM_CHART_VERSION=${IMAGE_TAG}" | tee -a ${GITHUB_ENV}
```
tnozicka, line 271: would it be enough just to use `git describe` without the `--long` option in the makefile?
https://github.com/scylladb/scylla-operator/blob/6439ee3ea61161855d4cf39da786238da5425495/Makefile#L56

zimnx: yeah, good idea. Changed

tnozicka (approved):
/lgtm
thanks

PR #752, opened 2021-08-11
allow overrides for helm scylla-operator replicas

saada:

Allow helm overrides for scylla operator replicas for operator and webhook server.

### DIFF
--- a/helm/scylla-operator/values.yaml
+++ b/helm/scylla-operator/values.yaml
@@ -1,3 +1,8 @@
+# Operator replicas use leader election. Setting to 1 will disable pdb creation
+# and won't be HA; creations or updates of Scylla CRs will fail during operator
+# upgrades or disruptions
+replicas: 2
+
 # Allows to customize Scylla Operator image
 image:
   repository: scylladb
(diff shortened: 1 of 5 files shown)


helm/scylla-operator/values.yaml
```diff
@@ -49,6 +52,9 @@ serviceAccount:
 ## SecurityContext holds pod-level security attributes and common container settings.
 securityContext: {}
 
+# Replicas for Webhook Server
```
tnozicka, line 55: Although it's not surprising, please explicitly add a warning comment that setting replicas to a lower value then 2 means it won't be HA and creations or updates of Scylla CRs will fail during operator upgrades or disruptions.

helm/scylla-operator/values.yaml
```diff
@@ -1,3 +1,6 @@
+# Operator replicas use leader election
+replicas: 2
```
tnozicka, line 2: Although it's not surprising, please explicitly add a warning comment that setting replicas to a lower value then 2 means it won't be HA and won't react during operator upgrades or disruptions.

tnozicka (commented):
Does this also need to handle PDBs corresponding to it when the replicas are set to 1?

helm/scylla-operator/values.yaml
saada: Thanks for your feedback! I added a comment and added a check before creating pdbs

helm/scylla-operator/values.yaml
saada: Thanks for your feedback! I added a comment and added a check before creating pdbs

zimnx (approved):
lgtm, please rebase on current master

saada:
> lgtm, please rebase on current master

done 👍🏽

zimnx (approved):

PR #753, opened 2021-08-11
Handle PRs with multiple "kind/" labels in release notes

tnozicka:
When publishing https://github.com/scylladb/scylla-operator/releases/tag/v1.4.0 the release notes wrongly assigned PR #668 as a feature and it wasn't listed under API changes. This shows that our kinds are not mutually exclusive sets and we have to prioritize it. Like in this case API changes are subset of features but we want the PR listed under API changes.

With this change, the order of sections determines the priority and the first match wins.


tnozicka:
(clean rebase)

zimnx (approved):
LGTM, thanks

Issue #497, opened 2021-03-15
Support `kubectl scale` and the /scale subresource


tnozicka:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
```
# kubectl get scyllacluster/basic
NAME    AGE
basic   3m57s

# kubectl scale scyllacluster/basic --replicas=3
Error from server (NotFound): scyllaclusters.scylla.scylladb.com "basic" not found
```

**What is use case behind this feature:**

**Additional Information:**
We need to support the scale subresource.

tnozicka:
I suppose would scale the members in every rack.

tnozicka:
@zimnx this might be a good one to look at for `ScyllaDBDatacenter`

tnozicka:
We have decided to postpone it in favour of a dedicated proposal, but for the future some context on the complexities is in https://github.com/scylladb/scylla-operator/pull/1994#discussion_r1672035663

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2470.

PR #756, opened 2021-08-12
[WIP] Scale subresource for ScyllaCluster proposal

rzetelskik:
Enhancement proposal to enable scale subresource for ScyllaCluster.

Issue link: #497

### DIFF
--- a/enhancements/proposals/scale_subresource.md
+++ b/enhancements/proposals/scale_subresource.md
@@ -1,0 +1,133 @@
+# Scale subresource for ScyllaCluster
+
+## Summary
+
+This proposal describes an enhancement of Scylla Operator enabling the scale subresource for ScyllaCluster custom resource.
+
+## Motivation
+
+When the scale subresource is enabled, the `/scale` subresource for the custom resource is exposed.
+It allows for using `kubectl scale` to scale the resource. Furthermore, it is required by [Horizontal Pod Autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) and [Vertical Pod Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) to operate on the custom resource.
+
+### Goals
+
+- Enabling the scale subresource in ScyllaCluster CustomResourceDefinition.
+- Providing a way to scale all Racks simultaneously, with each Rack's Member count set to an equal value.
+
+### Non-Goals
+
+- Providing a way to scale each Rack separately.
+- Providing support for autoscaling. Enabling the scale subresource is required, but not sufficient to support autoscaling of ScyllaClusters.
+
+## Proposal
+
+Since ScyllaCluster is a CustomResource and not a simple Kubernetes collection, to support scaling, it needs to have the scale subresource enabled.
+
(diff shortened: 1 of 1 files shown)


enhancements/proposals/scale_subresource.md
```diff
@@ -0,0 +1,133 @@
...
+type ScyllaCluster struct {
+	...
+}
+```
+
+Setting the Replicas field in ScyllaClusterSpec will enforce equal number of Scylla nodes in each Rack. The number should always match Replicas.
+To do so, Scylla Operator is extended with a mutating webhook overriding the Member counts.
+Without Replicas set, the number of Members in each Rack is determined as before, by specifying it in Rack's specification.
+
+Replicas field in ScyllaCluster's status is updated only when every Racks' readyMembers field value is equal. Replicas is set to this value then.
```
tnozicka, line 81: write it in a ways that explicitly mentions that it keeps the status keeps the previous value until all racks(statefulsets) have observedGeneration catched up, and status.replicas doesn't differ. Ideally this should be in the API comment.

enhancements/proposals/scale_subresource.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Test Plan
+
+#### Mutating webhook
+The mutating webhook propagating the Replicas value to each Rack's Member count is unit-tested.
+
+#### E2E
+##### ScyllaCluster simultaneous scaling
+- Create a ScyllaCluster with multiple Racks and `n` replicas
```
tnozicka, line 90: you should be able to verify it immediately, CREATE calls returns the object with defaulted values, you can even do a server-side dry-run to see how the object is defaulted without creating it and needing the resources.

enhancements/proposals/scale_subresource.md
```diff
@@ -0,0 +1,133 @@
...
+- Wait for the ScyllaCluster to reconcile and become fully available again
+- Verify that each Rack has `m` Members
+
+### Upgrade / Downgrade Strategy
+
+No upgrade/downgrade strategy required.
+
+### Version Skew Strategy
+
+Simultaneous scaling will not work in case of a version skew between ScyllaCluster and Scylla Operator.
```
tnozicka, line 103: Not sure what woudn't work, newer scylla-operator doesn't use the scale subresource.

enhancements/proposals/scale_subresource.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Drawbacks
+
+None.
+
+## Alternatives
+
+### Mutating webhook
+Instead of extending Scylla Operator with a mutating webhook propagating the Replicas value to Members in each Rack, we could make Members field optional and create additional validations in the admission webhook.
+Validations would enforce that Replicas and Members fields are mutually exclusive and expect either the Replicas field or every Member field to be set. This behaviour could be more transparent, albeit it would require setting the Members explicitly when unsetting Replicas.
```
tnozicka, line 122: that doesn't look viable given the `/scale` would edit only one field, it would be sentenced to fail every time.

enhancements/proposals/scale_subresource.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Implementation History
+
+- 2021-08-12 Proposal created
+
+## Drawbacks
+
+None.
+
+## Alternatives
```
tnozicka, line 118: I think it's fine to write `None` in this case :)

PR #763, opened 2021-08-18
Allow changing resources, storage, placement and repository

### LINKED ISSUE #415: At least basic properties of a rack should be editable
* Feature Request

At some point admin will want to edit e.g. cpu request/limit from 1 to 2, the answer he currently gets is
```
admission webhook "webhook.scylla.scylladb.com" denied the request: rack manager-rack: changes in resources are not currently supported
```

tnozicka:
With v1.4 and the rewrite of our reconciliation we are already able to manage updating these fields. This PR loosens the validation and add testing coverage.

Updates for requested storage capacity are reconciled up to a point where the appropriate claim template in a StatetefulSet is updated. Because StatefulSets don't manage the actual PersistentVolumeClaims (PVC), the change will only be relevant for newly created PVC, like in case of a new pod when scaling or replacing a scylla node.

Resolves #415


test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -0,0 +1,150 @@
...
+
+		framework.By("Waiting for the ScyllaCluster to redeploy")
+		waitCtx2, waitCtx2Cancel := contextForRollout(ctx, sc)
+		defer waitCtx2Cancel()
+		sc, err = waitForScyllaClusterState(waitCtx2, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, scyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Changing requested storage capacity")
+		// Requested capacity is only manifested on new PVC the old ones have to be updated manually because
+		// StatefulSets are not reconciling PVCs, only create it initially.
```
zimnx, line 104: Lets verify that old member pvc has previous storage capacity, and after scaling, new member has new storage capacity

tnozicka: actually statefulset forbid the update so we'd have to orphan delete them and recreate which isn't worth the effort now I guess

PR #764, opened 2021-08-19
Live reload webhook server certificates

### LINKED ISSUE #754: Live reload webhook server serving certificates
When the webhook server serving certificates are rotated we need to live reload them. Kubelet will already make sure those are updated on the filesystem. Ideally we'd do it similarly to kubernetes apiserver, worst case we short circuit on change.

rzetelskik:
- Add live reloading of webhook server certificates.
- Add a unit tests for live reloading of certificates

Resolves #754

### DIFF
--- a/pkg/cmd/operator/webhooks.go
+++ b/pkg/cmd/operator/webhooks.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"net"
 	"net/http"
 	"sync"
 	"time"
@@ -26,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/server/dynamiccertificates"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/klog/v2"
 )
@@ -35,12 +37,17 @@ type WebhookOptions struct {
 	Port                           int
 	InsecureGenerateLocalhostCerts bool
 
-	servingCertificate tls.Certificate
+	TLSConfig                 *tls.Config
+	dynamicCertKeyPairContent *dynamiccertificates.DynamicCertKeyPairContent
+
+	resolvedListenAddr   string
+	resolvedListenAddrCh chan struct{}
 }
 
 func NewWebhookOptions(streams genericclioptions.IOStreams) *WebhookOptions {
 	return &WebhookOptions{
-		Port: 5000,
+		Port:                 5000,
+		resolvedListenAddrCh: make(chan struct{}),
 	}
 }
 
@@ -111,8 +118,6 @@ func (o *WebhookOptions) Complete() error {
 			return err
 		}
 
-		o.servingCertificate.PrivateKey = privateKey
-
 		serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
 		serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
 		if err != nil {
@@ -138,12 +143,25 @@ func (o *WebhookOptions) Complete() error {
 			return err
 		}
 
-		o.servingCertificate.Certificate = [][]byte{derBytes}
-
+		o.TLSConfig = &tls.Config{
+			Certificates: []tls.Certificate{
+				{
+					Certificate: [][]byte{derBytes},
+					PrivateKey:  privateKey,
+				},
+			},
+		}
 	} else {
-		o.servingCertificate, err = tls.LoadX509KeyPair(o.TLSCertFile, o.TLSKeyFile)
+		o.dynamicCertKeyPairContent, err = dynamiccertificates.NewDynamicServingContentFromFiles("serving-certs", o.TLSCertFile, o.TLSKeyFile)
 		if err != nil {
-			return err
+			return fmt.Errorf("can't create DynamicServingContentFromFiles: %w", err)
+		}
+
+		o.TLSConfig = &tls.Config{
+			GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+				cert, err := tls.X509KeyPair(o.dynamicCertKeyPairContent.CurrentCertKeyContent())
+				return &cert, err
+			},
 		}
 	}
 
@@ -154,9 +172,6 @@ func (o *WebhookOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 	klog.Infof("%s version %s", cmd.Name(), version.Get())
 	cliflag.PrintFlags(cmd.Flags())
 
-	var wg sync.WaitGroup
-	defer wg.Wait()
-
 	stopCh := signals.StopChannel()
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
@@ -164,6 +179,16 @@ func (o *WebhookOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 		<-stopCh
 		cancel()
 	}()
+
+	return o.run(ctx, streams)
+}
+
+func (o *WebhookOptions) run(ctx context.Context, streams genericclioptions.IOStreams) error {
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	handler := http.NewServeMux()
 	handler.HandleFunc("/readyz", func(w http.ResponseWriter, req *http.Request) {
@@ -176,12 +201,27 @@ func (o *WebhookOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 	handler.Handle("/validate", admissionreview.NewHandler(validate))
 
 	server := http.Server{
-		Handler: handler,
-		Addr:    fmt.Sprintf(":%d", o.Port),
-		TLSConfig: &tls.Config{
-			Certificates: []tls.Certificate{o.servingCertificate},
-		},
+		Handler:   handler,
+		TLSConfig: o.TLSConfig,
 	}
+
+	if o.dynamicCertKeyPairContent != nil {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			o.dynamicCertKeyPairContent.Run(1, ctx.Done())
+		}()
+	}
+
+	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", o.Port))
+	if err != nil {
+		return fmt.Errorf("can't create listener: %w", err)
+	}
+	defer listener.Close()
+
+	o.resolvedListenAddr = listener.Addr().String()
+	// Notify anyone waiting for establishing the listen address by closing the channel.
+	close(o.resolvedListenAddrCh)
 
 	wg.Add(1)
 	go func() {
@@ -195,8 +235,8 @@ func (o *WebhookOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 		}
 	}()
 
-	klog.Infof("Starting HTTPS server on address %q.", server.Addr)
-	err := server.ListenAndServeTLS("", "")
+	klog.Infof("Starting HTTPS server on address %q.", o.resolvedListenAddr)
+	err = server.ServeTLS(listener, "", "")
 	if !errors.Is(err, http.ErrServerClosed) {
 		return err
 	}
--- a/pkg/cmd/operator/webhooks_test.go
+++ b/pkg/cmd/operator/webhooks_test.go
@@ -1,0 +1,329 @@
+// Copyright (c) 2021 ScyllaDB
+
+package operator
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"net"
+	"net/http"
+	"os"
+	"reflect"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/server/dynamiccertificates"
+)
+
+const (
+	pollInterval = 1 * time.Second
+	pollTimeout  = 3 * time.Second
+)
+
+func TestMain(m *testing.M) {
+	dynamiccertificates.FileRefreshDuration = 1 * time.Second
+	os.Exit(m.Run())
+}
+
+func TestWebhookOptionsRun(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	tt := []struct {
+		Name           string
+		WebhookOptions *WebhookOptions
+		ExpectedError  error
+	}{
+		{
+			Name: "valid webhook options",
+			WebhookOptions: func() *WebhookOptions {
+				wo := NewWebhookOptions(genericclioptions.IOStreams{})
+				wo.Port = 65535
+				wo.InsecureGenerateLocalhostCerts = true
+
+				return wo
+			}(),
+			ExpectedError: nil,
+		},
+		{
+			Name: "invalid port",
+			WebhookOptions: func() *WebhookOptions {
+				wo := NewWebhookOptions(genericclioptions.IOStreams{})
+				wo.Port = 65536
+				wo.InsecureGenerateLocalhostCerts = true
+
+				return wo
+			}(),
+			ExpectedError: fmt.Errorf("can't create listener: %w", &net.OpError{
+				Op:  "listen",
+				Net: "tcp",
+				Err: &net.AddrError{
+					Err:  "invalid port",
+					Addr: "65536",
+				},
+			}),
+		},
+	}
+
+	for i := range tt {
+		tc := tt[i]
+		t.Run(tc.Name, func(t *testing.T) {
+			t.Parallel()
+
+			err := tc.WebhookOptions.Complete()
+			if err != nil {
+				t.Fatalf("can't complete WebhookOptions: %v", err)
+			}
+
+			err = tc.WebhookOptions.run(ctx, genericclioptions.IOStreams{
+				Out:    os.Stdout,
+				ErrOut: os.Stderr,
+			})
+			if !reflect.DeepEqual(tc.ExpectedError, err) {
+				t.Errorf("expected and actual errors differ: %s", cmp.Diff(tc.ExpectedError, err))
+			}
+		})
+	}
+}
+
+func TestWebhookOptionsRunWithReload(t *testing.T) {
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	dir := t.TempDir()
+	tlsCertFile := dir + "/tls.crt"
+	tlsKeyFile := dir + "/tls.key"
+
+	cc, err := newCertificateCreator()
+	if err != nil {
+		t.Fatalf("can't create certificateCreator: %v", err)
+	}
+
+	initialCertSerialNumber := big.NewInt(1)
+	encodedInitialCert, err := cc.generateEncodedCertificate(initialCertSerialNumber)
+	if err != nil {
+		t.Fatalf("can't create new encoded certificate: %v", err)
+	}
+
+	err = createFileWithContent(tlsCertFile, encodedInitialCert.cert)
+	if err != nil {
+		t.Fatalf("can't create a file: %v", err)
+	}
+
+	err = createFileWithContent(tlsKeyFile, encodedInitialCert.privateKey)
+	if err != nil {
+		t.Fatalf("can't create a file: %v", err)
+	}
+
+	wo := NewWebhookOptions(genericclioptions.IOStreams{})
+	wo.TLSCertFile = tlsCertFile
+	wo.TLSKeyFile = tlsKeyFile
+	wo.Port = 0
+	wo.InsecureGenerateLocalhostCerts = false
+
+	err = wo.Complete()
+	if err != nil {
+		t.Fatalf("can't complete WebhookOptions: %v", err)
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		err := wo.run(ctx, genericclioptions.IOStreams{
+			Out:    os.Stdout,
+			ErrOut: os.Stderr,
+		})
+		if err != nil {
+			t.Errorf("can't run webhook server: %v", err)
+		}
+	}()
+
+	certpool := x509.NewCertPool()
+	certpool.AppendCertsFromPEM(cc.encodedCA)
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: certpool,
+			},
+		},
+	}
+
+	// Wait for the listen address to be established.
+	<-wo.resolvedListenAddrCh
+	addr := fmt.Sprintf("https://%s/readyz", wo.resolvedListenAddr)
+
+	hasExpectedCertificate := func(expectedCertificateSerialNumber *big.Int) func() (bool, error) {
+		return func() (bool, error) {
+			resp, err := client.Get(addr)
+			if err != nil {
+				t.Logf("can't make a GET request: %v", err)
+				return false, nil
+			}
+
+			if !reflect.DeepEqual(expectedCertificateSerialNumber, resp.TLS.PeerCertificates[0].SerialNumber) {
+				t.Logf("serial numbers differ: expected %s, actual %s", expectedCertificateSerialNumber, resp.TLS.PeerCertificates[0].SerialNumber)
+				return false, nil
+			}
+
+			return true, nil
+		}
+	}
+
+	err = wait.PollImmediate(pollInterval, pollTimeout, hasExpectedCertificate(initialCertSerialNumber))
+	if err != nil {
+		t.Fatalf("can't observe the initial certificate: %v", err)
+	}
+
+	updatedCertSerialNumber := big.NewInt(2)
+	encodedUpdatedCert, err := cc.generateEncodedCertificate(updatedCertSerialNumber)
+	if err != nil {
+		t.Fatalf("can't create new encoded certificate: %v", err)
+	}
+
+	err = createFileWithContent(wo.TLSCertFile, encodedUpdatedCert.cert)
+	if err != nil {
+		t.Fatalf("can't create a file: %v", err)
+	}
+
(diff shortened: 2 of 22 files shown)


pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,364 @@
...
+					TLSConfig: tls.Config{
+						GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+							cert, err := tls.X509KeyPair(dynamicCertKeyPairContent.CurrentCertKeyContent())
+							return &cert, err
+						},
+					},
+					dynamicCertKeyPairContent: dynamicCertKeyPairContent,
+				}
+			}(),
+			Callback: func(wo *WebhookOptions, t *testing.T) {
```
tnozicka, line 164: would it be better not to do table driven test for integration like scenario?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,364 @@
...
+	_, cert, privateKey, _, _, err := createCertificates(now)
+	if err != nil {
+		panic(err)
+	}
+
+	ts := []struct {
+		Name           string
+		WebhookOptions *WebhookOptions
+		ExpectedError  error
+		Callback       func(*WebhookOptions, *testing.T)
```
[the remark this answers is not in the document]
tnozicka: could this use `ExpectedWebhookOptions` pattern?

tnozicka (commented):
pls squash commit 1 with 2, and 3 with 4

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,364 @@
...
+			},
+		},
+		{
+			Name: "sets dynamicCertKeyPairContent and TLSConfig.GetCertificate",
+			WebhookOptions: func() *WebhookOptions {
+				dir := t.TempDir()
+				TLSCertFile, TLSKeyFile := dir+"/tls.crt", dir+"/tls.key"
+
+				if err = createPopulatedFile(TLSCertFile, cert); err != nil {
+					panic(err)
```
tnozicka, line 71: actually, I think you could still call t.fatal here

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,364 @@
...
+)
+
+func TestComplete(t *testing.T) {
+	now := time.Now()
+	_, cert, privateKey, _, _, err := createCertificates(now)
+	if err != nil {
+		panic(err)
+	}
+
+	ts := []struct {
```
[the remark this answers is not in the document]
rzetelskik: I'm assuming `tests`; copied from https://github.com/scylladb/scylla-operator/blob/153f6a04b15fc02d79d771e3e714c55a390aeb28/pkg/cmd/releasenotes/helpers_test.go#L10. I'm happy to change to `tt` though.

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/server/dynamiccertificates"
+)
+
+func TestComplete(t *testing.T) {
+	now := time.Now()
+	_, cert, privateKey, _, _, err := createCertificates(now)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
rzetelskik: doesn't it sort of imply that we're expecting errors, while we don't really want to test the behaviour of createCertificates?

pkg/cmd/operator/webhooks_test.go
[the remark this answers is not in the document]
tnozicka: yeah, I see `ts` as slightly unfortunate because the short variant for an item in that array would be `t` which conflicts. Same way we do the naming for `pods`->`pod`. So using `tt` and `tc` kind of avoids that implied rule while it still keeps the repeated names short.
Completely get you just took an existing one.

pkg/cmd/operator/webhooks_test.go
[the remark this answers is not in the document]
tnozicka: I don't think it does. I don't want our test binary to panic just because it can't create a file. Disk can be full, the test should fail, the others that don't need disk should continue on.

pkg/cmd/operator/webhooks_test.go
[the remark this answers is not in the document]
rzetelskik: `Certificate's serial number is generated randomly so that's an issue. DynamicCertKeyPairContent would require a custom comparator and a way to avoid concurrent reads. Additionally `GetCertificate` can't be compared at all.

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,358 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		wo := WebhookOptions{
+			TLSCertFile:                    TLSCertFile,
+			TLSKeyFile:                     TLSKeyFile,
+			Port:                           5000,
+			InsecureGenerateLocalhostCerts: false,
+			TLSConfig: tls.Config{
```
tnozicka, line 166: should this use Complete?

pkg/cmd/operator/webhooks_test.go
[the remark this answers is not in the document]
tnozicka: maybe we can split it into a table based test for those that fit and one non-table based for the live reload. On the other hand that case could be cover through the Run flow which should call Complete anyways I think. In the end we primarily care about the reload test, the rest is covered through other flows in our suite.

pkg/cmd/operator/webhooks_test.go
rzetelskik: I suppose we could, since it's turned into an integration test anyway. Are you sure we should get rid of testing Complete then (the other comment)? We don't really cover it otherwise?

tnozicka: > We don't really cover it otherwise?
We'd call it from the other unit test for Run so it will be covered, probably to a bigger extend as it will test if the completed options start the program, not just that it's non-nil. I am not against some simple and meaningful test for Complete (although not necessary).

tnozicka: (also e2e tests start the program by deploying the webhook)

rzetelskik: I've tried creating a more reasonable table for Complete but even with custom comparators I wasn't really able to test anything meaningful in a nice way, so I've removed them altogether.

pkg/cmd/operator/webhooks.go
```diff
@@ -165,6 +175,13 @@ func (o *WebhookOptions) Run(streams genericclioptions.IOStreams, commandName st
 		cancel()
 	}()
 
+	return o.run(ctx, streams)
+}
+
+func (o *WebhookOptions) run(ctx context.Context, streams genericclioptions.IOStreams) error {
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
zimnx, line 183: I would prefer explicit wg.Wait in code instead of in defer. Defer is good for cleanup, but not for logic

pkg/cmd/operator/webhooks.go
```diff
@@ -175,12 +192,18 @@ func (o *WebhookOptions) Run(streams genericclioptions.IOStreams, commandName st
 	})
 	handler.Handle("/validate", admissionreview.NewHandler(validate))
 
+	if o.dynamicCertKeyPairContent != nil {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			o.dynamicCertKeyPairContent.Run(1, ctx.Done())
```
zimnx, line 199: It's unclear what `1` means, name it in const

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+	cc, err := newCertificateCreator()
+	if err != nil {
+		t.Fatalf("can't create CertificateCreator: %v", err)
+	}
+
+	certSerialNumber := big.NewInt(1)
+	encodedCert, err := cc.newEncodedCertificate(certSerialNumber)
+	if err != nil {
+		t.Fatalf("can't create new encoded certificate: %v", err)
+	}
```
zimnx, line 52: nit: empty line after block

pkg/cmd/operator/webhooks.go
tnozicka: how would you make sure it's called in all the branches? the point of defer here is to make sure we **always** wait for the children

pkg/cmd/operator/webhooks.go
tnozicka: technically it's the name of the function argument that determines the meaning of that int, not the name of a variable you pass in so bare int has the same meaning, const is primarily used for something that needs to be set across the board or possibly changed but not to infer a "type"

pkg/cmd/operator/webhooks.go
tnozicka: (it's the same as with locks, IMO)

zimnx: Now it hangs infinitely if L224 branch is triggered. IMO it's easier to spot this kind of deadlocks when blocking paths are mentioned explictly.

pkg/cmd/operator/webhooks.go
zimnx: I cannot deduce what this 1 means without looking at the source code of third party library, nor from purpose of this library. It's clear for `WaitGroup.Add(1)`, but not for `DynamicCertKeyPairContent.Run(1, ...)`
Naming it here would make my (as a reviewer) life easier, because I would be able to read the code without going back and forth between IDE and browser/IDE tabs.

tnozicka: how do you (as a reviewer) know that someone putting `const numberOfDays` into `DynamicCertKeyPairContent.Run(numberOfDays, ...)` is doing a correct job without looking at the function or knowing that conventionally it takes `numberOfWorkers`?

pkg/cmd/operator/webhooks.go
tnozicka: We must track all our workers reliably and manual wait is a recipe for forgetting that. 
That said, we shouldn't hang - are you saying Shutdown won't cancel if `ListenAndServeTLS` fails?

We should add a cancelable context after the wg which would make sure no blocking call is left behind.

tnozicka: guess that hangs on the `<-ctx.Done()` part because it's missing the cancalable context for this function, which was present before

tnozicka: @rzetelskik this deserves a unit test - use an invalid port as a reproducer

pkg/cmd/operator/webhooks.go
zimnx: When name is given i can at least judge if it make sense. `numberOfWorkers` in Run makes me think that there're workers - additional goroutines - and this stop channel probably stops them - and I would be right.  
Where `1` or `numberOfDays` doesn't make any sense.

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+				RootCAs: certpool,
+			},
+		},
+	}
+
+	addr := fmt.Sprintf("https://localhost:%d/readyz", wo.Port)
+
+	err = wait.PollImmediate(1*time.Second, 15*time.Second, func() (bool, error) {
+		resp, err := client.Get(addr)
+		return err == nil && reflect.DeepEqual(resp.TLS.PeerCertificates[0].SerialNumber, certSerialNumber), nil
```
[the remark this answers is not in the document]
tnozicka: might worth to log SN inequality as well because all you get when this fails is `timeout waiting for condition`

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+	})
+	if err != nil {
+		t.Fatalf("can't poll immediatelly: %v", err)
```
tnozicka, line 103: `failed to observe the initial certificate: %v` (with fatal errors you don't always have to  [but can] use the regular wording as in the operator where it needs to imply it will retry)

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+	dir := t.TempDir()
+	TLSCertFile := dir + "/tls.crt"
+	TLSKeyFile := dir + "/tls.key"
+
+	cc, err := newCertificateCreator()
+	if err != nil {
+		t.Fatalf("can't create CertificateCreator: %v", err)
+	}
+
+	certSerialNumber := big.NewInt(1)
```
tnozicka, line 48: nit: initialCertSerialNumber

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+		t.Fatalf("can't create new encoded replace certificate: %v", err)
+	}
+	if err = createFileWithContent(wo.TLSCertFile, encodedReplaceCert.cert); err != nil {
+		t.Fatalf("can't create a file: %v", err)
+	}
+	if err = createFileWithContent(wo.TLSKeyFile, encodedReplaceCert.privateKey); err != nil {
+		t.Fatalf("can't create a file: %v", err)
+	}
+
+	err = wait.PollImmediate(1*time.Second, 3*time.Second, func() (bool, error) {
```
tnozicka, line 118: let's unify the wait times and share them through const

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+		t.Fatalf("can't create a file: %v", err)
+	}
+	if err = createFileWithContent(TLSKeyFile, encodedCert.privateKey); err != nil {
+		t.Fatalf("can't create a file: %v", err)
+	}
+
+	wo := WebhookOptions{
+		TLSCertFile:                    TLSCertFile,
+		TLSKeyFile:                     TLSKeyFile,
+		Port:                           5000,
```
tnozicka, line 63: the port might be already taken, we need a dynamic one

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+		return nil, err
+	}
+
+	return &CertificateCreator{
+		ca:           ca,
+		caPrivateKey: caPrivateKey,
+	}, nil
+}
+
+func (cc *CertificateCreator) getEncodedCA() ([]byte, error) {
```
tnozicka, line 163: by the name I thought this encodes the signer CA while it actually generates a new self-signed cert - could use a better name but maybe it could just live in the constructor?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+	bytes, err := x509.CreateCertificate(rand.Reader, cc.ca, cc.ca, cc.caPrivateKey.Public(), cc.caPrivateKey)
+	if err != nil {
+		return nil, err
```
tnozicka, line 166: wrap the error here, and everywhere else

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	return &encodedCertificate{
+		cert:       certEncoded,
+		privateKey: privateKeyEncoded,
+	}, nil
+}
+
+func pemEncode(cert []byte, certType string) ([]byte, error) {
+	buff := new(bytes.Buffer)
```
tnozicka, line 212: tiny nit: for consistency - let's use `buff := &bytes.Buffer{}` or `var buff bytes.Buffer`

pkg/cmd/operator/webhooks_test.go
tnozicka, line 63: a fixed port might be already taken on a machine, let's use a dynamic one with `:0` and detect which was allocated

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+	go func() {
+		defer wg.Done()
+
+		err := wo.run(ctx, genericclioptions.IOStreams{})
+		if err != nil {
+			t.Errorf("can't run webhook server: %v", err)
+		}
+	}()
+
+	encodedCa, err := cc.getEncodedCA()
```
tnozicka, line 81: rootCA

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,237 @@
...
+func (cc *CertificateCreator) getEncodedCA() ([]byte, error) {
+	bytes, err := x509.CreateCertificate(rand.Reader, cc.ca, cc.ca, cc.caPrivateKey.Public(), cc.caPrivateKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return pemEncode(bytes, "CERTIFICATE")
+}
+
+func (cc *CertificateCreator) newEncodedCertificate(serialNumber *big.Int) (*encodedCertificate, error) {
```
tnozicka, line 172: nit: generateEncodedCertificate

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			err := test.WebhookOptions.Complete()
+			if err != nil {
+				t.Fatalf("can't complete WebhookOptions: %v", err)
+			}
+
+			wg.Add(1)
+			go func() {
```
tnozicka, line 71: what if the context gets canceled before the goroutine starts? does it actually need to be running on background if there is no other code run after it?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		t.Fatalf("failed to create CertificateCreator: %v", err)
+	}
+
+	initialCertSerialNumber := big.NewInt(1)
+	encodedCert, err := cc.generateEncodedCertificate(initialCertSerialNumber)
+	if err != nil {
+		t.Fatalf("failed to create new encoded certificate: %v", err)
+	}
+	if err = createFileWithContent(tlsCertFile, encodedCert.cert); err != nil {
```
tnozicka, line 104: nit: this is pretty long, it might be clearer to initialize the variable on a dedicated line (hint: it always is)

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+		t.Fatalf("failed to complete WebhookOptions: %v", err)
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		err := wo.run(ctx, genericclioptions.IOStreams{})
+		if err != nil {
+			t.Errorf("can't run webhook server: %v", err)
```
tnozicka, line 128: fatal

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+	certpool.AppendCertsFromPEM(cc.encodedCA)
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: certpool,
+			},
+		},
+	}
+
+	wo.assignedPort.c.L.Lock()
```
tnozicka, line 142: make it a method

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+			return true, nil
+		}
+	}
+
+	err = wait.PollImmediate(pollInterval, pollTimeout, observeCertificate(initialCertSerialNumber))
+	if err != nil {
+		t.Fatalf("failed to observe the initial certificate: %v", err)
+	}
+
+	replaceCertSerialNumber := big.NewInt(2)
```
tnozicka, line 171: nit: guess the name should be `replacedCertSerialNumber` but a better naming might be `initialCert...` and `updatedCert...`

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	err = wait.PollImmediate(pollInterval, pollTimeout, observeCertificate(replaceCertSerialNumber))
+	if err != nil {
+		t.Errorf("certificate wasn't updated: %v", err)
+	}
+}
+
+type CertificateCreator struct {
+	ca           *x509.Certificate
```
tnozicka, line 190: nit: caCert

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+		t.Fatalf("failed to create a file: %v", err)
+	}
+
+	err = wait.PollImmediate(pollInterval, pollTimeout, observeCertificate(replaceCertSerialNumber))
+	if err != nil {
+		t.Errorf("certificate wasn't updated: %v", err)
+	}
+}
+
+type CertificateCreator struct {
```
tnozicka, line 189: should the class be private?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+type encodedCertificate struct {
+	cert       []byte
+	privateKey []byte
+}
+
+func newCertificateCreator() (*CertificateCreator, error) {
+	now := time.Now()
+	ca := &x509.Certificate{
```
tnozicka, line 202: nit: caCert

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,302 @@
...
+func createFileWithContent(path string, data []byte) error {
+	f, err := os.Create(path)
+	if err != nil {
+		return fmt.Errorf("can't create file: %v", err)
+	}
+	defer f.Close()
+
+	_, err = f.Write(data)
+	if err != nil {
+		return fmt.Errorf("can't write to file: %v", err)
```
tnozicka, line 298: %w (globally for errors)

pkg/cmd/operator/webhooks_test.go
rzetelskik: afaik fatal should only be called from a goroutine running the test

rzetelskik:
rebased on current master

rzetelskik:
ping @tnozicka

rzetelskik:
clean rebase

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,297 @@
...
+	pollInterval = 1 * time.Second
+	pollTimeout  = 3 * time.Second
+)
+
+func TestMain(m *testing.M) {
+	dynamiccertificates.FileRefreshDuration = 1 * time.Second
+	os.Exit(m.Run())
+}
+
+func TestRunErrors(t *testing.T) {
```
tnozicka, line 38: TestWebhookOptionsRun ?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,297 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	tt := []struct {
+		Name           string
+		WebhookOptions *WebhookOptions
+		ExpectedError  error
+	}{
+		{
```
tnozicka, line 47: wanna add a positive case as well?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,297 @@
...
+			WebhookOptions: &WebhookOptions{
+				Port:                           65536,
+				InsecureGenerateLocalhostCerts: true,
+			},
+			ExpectedError: fmt.Errorf("can't create listener: listen tcp: address 65536: invalid port"),
+		},
+	}
+
+	for i := range tt {
+		test := tt[i]
```
tnozicka, line 58:
```suggestion
		tc := tt[i]
```

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,297 @@
...
+	err = wo.Complete()
+	if err != nil {
+		t.Fatalf("failed to complete WebhookOptions: %v", err)
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		err := wo.run(ctx, genericclioptions.IOStreams{})
```
tnozicka, line 122: should we setup stdout and stderr so the errors can be seen?

rzetelskik:
@tnozicka thanks, I've addressed your comments

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,313 @@
...
+				RootCAs: certpool,
+			},
+		},
+	}
+
+	// Wait for port to be assigned.
+	<-wo.assignedPortCh
+	addr := fmt.Sprintf("https://localhost:%d/readyz", wo.assignedPort)
+
+	observeCertificate := func(expectedCertificateSerialNumber *big.Int) func() (bool, error) {
```
tnozicka, line 156: hasCertificate?

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,313 @@
...
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+
+		DNSNames: []string{"localhost"},
+	}
+
+	caPrivateKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate key: %w", err)
```
tnozicka, line 228: nit: for consistency we use "can't" pattern in errors

pkg/cmd/operator/webhooks_test.go
rzetelskik: changed to hasExpectedCertificate

pkg/cmd/operator/webhooks.go
```diff
@@ -138,25 +142,37 @@ func (o *WebhookOptions) Complete() error {
...
+
+		o.TLSConfig = &tls.Config{
+			GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+				cert, err := tls.X509KeyPair(o.dynamicCertKeyPairContent.CurrentCertKeyContent())
+				return &cert, err
+			},
 		}
 	}
 
+	o.assignedPortCh = make(chan struct{})
```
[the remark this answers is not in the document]
rzetelskik: I'm not a fan of this. It's supposed to be internal and used for testing and we don't use the constructor in tests. So it would either require using the constructor in tests, validating it or explicitly creating the channel in each struct in tests?

rzetelskik:
clean rebase

pkg/cmd/operator/webhooks_test.go
```diff
@@ -0,0 +1,330 @@
...
+	"net"
+	"net/http"
+	"os"
+	"reflect"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+
```
tnozicka, line 24: no \n

tnozicka (approved):
/lgtm
thanks!

PR #766, opened 2021-08-19
Fix backup.rateLimit type in CRD example

vavsab:
Fix backup rate limit example in [CRD page](https://operator.docs.scylladb.com/stable/scylla_cluster_crd.html) to fit its actual type.


tnozicka (approved):
/lgtm
thanks

zimnx (approved):

PR #770, opened 2021-08-31
Convert ephemeral storage to xfs on GKE

### LINKED ISSUE #760: Scylla pods fail to be migrated to new upgraded K8S nodes when local disks are used
**Describe the bug**
When we try to upgrade K8S nodes dedicated for scylla pods with configured local disks we get following errors:

```
$ kubectl get pods -A -o wide
NAMESPACE         NAME                                 READY   STATUS        RESTARTS   AGE     IP            NODE
scylla            sct-cluster-us-east1-b-us-east1-0    2/2     Running       1          57m     10.142.0.29   gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-bx8w
scylla            sct-cluster-us-east1-b-us-east1-1    2/2     Running       1          60m     10.142.0.39   gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-s0zn
scylla            sct-cluster-us-east1-b-us-east1-2    0/2     Pending       0          29s     <none>        <none>

$ kubectl describe pod sct-cluster-us-east1-b-us-east1-2 -n scylla
...
Events:
  Type     Reason            Age                From               Message
  ----     ------            ----               ----               -------
  Warning  FailedScheduling  22s (x2 over 37s)  default-scheduler  0/8 nodes are available: 1 Insufficient cpu, 1 Insufficient memory, 2 node(s) didn't have free ports for the requested pod ports, 2 node(s) were unschedulable, 3 node(s) didn't match Pod's node affinity.
  Warning  FailedScheduling  1s (x4 over 40s)   default-scheduler  0/8 nodes are available: 1 Insufficient cpu, 1 Insufficient memory, 1 node(s) had volume node affinity conflict, 1 node(s) were unschedulable, 2 node(s) didn't have free ports for the requested pod ports, 3 node(s) didn't match Pod's node affinity.
```

PVs don't get recreated as expected:
```
$ kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM                                                          STORAGECLASS       REASON   AGE
local-pv-81ab7c30                          1124Gi     RWO            Delete           Available                                                                  local-raid-disks            127m
local-pv-926c8527                          1124Gi     RWO            Delete           Bound       scylla/data-sct-cluster-us-east1-b-us-east1-1                  local-raid-disks            128m
local-pv-c0a85317                          1124Gi     RWO            Delete           Bound       scylla/data-sct-cluster-us-east1-b-us-east1-0                  local-raid-disks            124m
local-pv-ffaf0e50                          1124Gi     RWO            Delete           Released    scylla/data-sct-cluster-us-east1-b-us-east1-2                  local-raid-disks            128m
pvc-2ae4ac91-b741-4eab-b2fc-5db04d560c8b   10Gi       RWO            Delete           Bound       scylla-manager/data-scylla-manager-manager-dc-manager-rack-0   standard                    137m
pvc-47411cc4-de88-417a-a526-49ba9eead21d   10Gi       RWO            Delete           Bound       minio/minio                                                    standard                    138m
```

And, since it is not recreated, PV refers to already absent node:
```
$ kubectl describe pv local-pv-ffaf0e50
Name:              local-pv-ffaf0e50
Labels:            <none>
Annotations:       pv.kubernetes.io/bound-by-controller: yes
                   pv.kubernetes.io/provisioned-by:
                     local-volume-provisioner-gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-658c-c2b5bbdc-0bf3-477d-b595-4220eb634a7b
Finalizers:        [kubernetes.io/pv-protection]
StorageClass:      local-raid-disks
Status:            Released
Claim:             scylla/data-sct-cluster-us-east1-b-us-east1-2
Reclaim Policy:    Delete
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          1124Gi
Node Affinity:     
  Required Terms:  
    Term 0:        kubernetes.io/hostname in [gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-658c]
Message:           
Source:
    Type:  LocalVolume (a persistent volume backed by local storage on a node)
    Path:  /mnt/raid-disks/disk0
Events:    <none>
```

List of nodes:
```
$ kubectl get nodes
NAME                                                  STATUS                     ROLES    AGE    VERSION
gke-sct-test-dev-7d6d-monitoring-pool-0f3c9c15-94lw   Ready                      <none>   133m   v1.19.13-gke.700
gke-sct-test-dev-7d6daee-default-pool-28379986-f4o1   Ready                      <none>   48m    v1.20.9-gke.700
gke-sct-test-dev-7d6daee8-loader-pool-20874e70-mzkj   Ready                      <none>   53m    v1.20.9-gke.700
gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-bx8w   Ready,SchedulingDisabled   <none>   130m   v1.19.13-gke.700
gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-i5wi   Ready                      <none>   37m    v1.20.9-gke.700
gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-joyl   Ready                      <none>   41m    v1.20.9-gke.700
gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-nqwr   Ready                      <none>   130m   v1.19.13-gke.700
gke-sct-test-dev-7d6daee8-scylla-pool-0ca0312a-s0zn   Ready                      <none>   130m   v1.19.13-gke.700
```

Labels on the appropriate svc:
```
$ kubectl describe svc sct-cluster-us-east1-b-us-east1-2 -n scylla
Name:              sct-cluster-us-east1-b-us-east1-2
Namespace:         scylla
Labels:            app=scylla
                   app.kubernetes.io/managed-by=scylla-operator
                   app.kubernetes.io/name=scylla
                   scylla/cluster=sct-cluster
                   scylla/datacenter=us-east1-b
                   scylla/rack=us-east1
                   scylla/replace=10.3.243.227
...
```
As we can see, "replace" label is set.

As a summary: PDB blocks moving further than first missing PV and upgrade hangs for long time where second node scheduled for upgrade stays in "Ready,SchedulingDisabled" state. Finally, reaching out wait timeout, GKE continue upgrades and kills scylla cluster disks with nodes.

**To Reproduce**

1. Provision K8S cluster of v1.19.13-gke.700 version
2. Deploy operator v1.4.0
3. Deploy local SSDs to "scylla node pool" using following example: https://github.com/scylladb/scylla-operator/blob/v1.4/examples/gke/raid-daemonset.yaml
4. Deploy local volume provisioner using following example: https://github.com/scylladb/scylla-operator/tree/v1.4/examples/common/provisioner
5. Deploy a Scylla cluster setting "**automaticOrphanedNodeCleanup**" parameter to "**true**"
6. Upgrade gke control-plane: `$ gcloud container --project %project-name% upgrade %cluster-name% --master --cluster-version 1.20`
7. Upgrade scylla node pool nodes to new K8S version: `$ gcloud container --project %project-name% clusters upgrade %cluster-name% --node-pool scylla-pool`
8. See error

**Expected behavior**
Scylla pods and storage get recreated correctly.

**Logs**
[gke-k8s-upgrade-failure.tar.gz](https://github.com/scylladb/scylla-operator/files/6992129/gke-k8s-upgrade-failure.tar.gz)

**Environment:**
 - Platform: GKE
 - Kubernetes version: v1.19.13-gke.700 -> v1.20.9-gke.700
 - Scylla version: 4.4.4
 - Scylla-operator version: v1.4.0

**Additional context**

rzetelskik:
This pull request introduces a new DaemonSet used for formatting ephemeral storage from ext4 to xfs. The script for creating GKE clusters is also modified to take advantage of gcloud beta feature `ephemeral-storage`, which creates a RAID array out of local SSDs.

The new DaemonSet effectively replaces a DaemonSet creating a RAID0 array out of local disks. The setup would not survive a host reboot, as `kube-node-configuration.service` would detect that the local SSDs do not contain an ext4 filesystem and would later fail trying to set it up (https://github.com/kubernetes/kubernetes/blob/dd2d12f6dc0e654c15d5db57a5f9f6ba61192726/cluster/gce/gci/configure-helper.sh#L290). 

Resolves #760


examples/gke/xfs-format-daemonset.yaml
```diff
@@ -0,0 +1,156 @@
...
+                device=\$1
+                shift
+                ;;
+            esac
+          done
+
+          if [[ -n \$test && -n \$device && \$( blkid -o value -s TYPE \$device ) == "xfs" ]]; then
+            exit 0
+          fi
+          /usr/sbin/tune2fs \$args
```
tnozicka, line 65: nit: use `${variable}` globally

examples/gke/xfs-format-daemonset.yaml
```diff
@@ -0,0 +1,156 @@
...
+          mkfs.xfs -f \$raid_device
+
+          cp -a \$tmp_dir/. \$ephemeral_mountpoint/
+          rm -rf \$tmp_dir
+
+          mkdir -p \$raid_dir
+          mount --bind \$ephemeral_mountpoint \$raid_dir
+          chmod a+w \$raid_dir
+
+          iptables -P INPUT ACCEPT
```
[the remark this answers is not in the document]
rzetelskik: yup, I'm going to add comments and refactor the script for better readability later on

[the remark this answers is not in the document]
tnozicka: before you go into the refactor can you describe the high level intent in the comment? I wasn't expecting us to mess with iptables.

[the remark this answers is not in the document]
rzetelskik: configure-helper.sh adds rules to the iptables and so it fails on restart if there are any existing Chains. Another approach could be adding an `iptables` wrapper similar to the tune2fs one.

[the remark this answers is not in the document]
tnozicka: after we copy our systemd files, wouldn't we just reboot the machine to make sure it is able to start correctly before any data can be written there?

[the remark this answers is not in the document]
rzetelskik: what are the advantages of rebooting though? The script would obviously be more straightforward but other than that I don't really see why we'd want to reboot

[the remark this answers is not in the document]
tnozicka: If you reboot right after and it comes back up ok, you have a high chance that the next (involuntary) reboot is gonna work. If you just modify the units and restart the particular unit to work, a lot can go differently when the first reboot comes but at this point you already have the data there and your scylla is gonna have a downtime while you figure it out. So my point is to find out sooner rather then later. Nobody is gonna freakout if it fails initially before deploying scylla (we get time to figure it out) but when it brings down an existing cluster we are gonna get paged.

examples/gke/xfs-format-daemonset.yaml
```diff
@@ -0,0 +1,156 @@
...
+        - "chroot"
+        - "/host"
+        - "/bin/bash"
+        - "-c"
+        args:
+        - |
+          set -euExo pipefail
+
+          mkdir -p /var/xfs-format/bin/
+          cat<<EOF >/var/xfs-format/bin/tune2fs
```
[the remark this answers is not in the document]
rzetelskik: where should the scripts live? I feel like `examples` dir is not the most fortunate. Should the systemd unit files also exist in separate files?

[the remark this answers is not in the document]
tnozicka: good question :) `/hack/gke/` sounds like one option for the repository - I guess in the future we might embed them into one of our binaries (depending on the proposal for automated node setup) but we can move them to `/manifests` when that comes. `/usr/local/lib/scylla-operator/gke` would be the location in the container image.

for the other part - it should be all in dedicated files

hack/gke/xfs-format/xfs-format.sh
```diff
@@ -0,0 +1,37 @@
...
+
+tmp_dir=$( mktemp -d )
+cp -af ${ephemeral_mountpoint}/. ${tmp_dir}/
+
+umount $( grep -e '/home/kubernetes/containerized_mounter' -e '/var/lib/containerd' -e '/home/containerd' /proc/mounts | sed -E -e 's/([^ ]+) ([^ ]+) .+/\2/' | sort -u -r ) || true
+# Separate line due to existing dependencies
+umount $( grep -e "${raid_device}" /proc/mounts | sed -E -e 's/([^ ]+) ([^ ]+) .+/\2/' | sort -r ) || true
+mkfs.xfs -f ${raid_device}
+
+cp -af ${tmp_dir}/. ${ephemeral_mountpoint}/
```
tnozicka, line 30: nit: usually we go with  `-r` for symlink handling

hack/gke/xfs-format/xfs-format.sh
```diff
@@ -0,0 +1,37 @@
...
+rm -rf ${tmp_dir}
+
+mkdir -p ${ephemeral_mountpoint}/scylla
```
tnozicka, line 33: let's try to avoid `-p` if you substitute wrong path this can make a mess and still succeed, safer to create the particular dirs individually

hack/gke/xfs-format/xfs-format.sh
```diff
@@ -0,0 +1,37 @@
...
+mkdir -p ${raid_dir}
+chmod a+w ${raid_dir}
+
+reboot & disown
```
tnozicka, line 37: `exec reboot`?

hack/gke/systemd/system/kube-node-configuration.service.d/10-path.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
```
tnozicka, line 1: nit: guess we don't need the `system` dir in the path

tnozicka:
I didn't write it down in the particular places but pls use qoutes around args - like `cmd "${variable}" "$( echo subshell )"`

hack/gke/systemd/system/xfs-format.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Service converting ephemeral storage to xfs
+
+[Service]
```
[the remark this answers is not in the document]
rzetelskik: It's only invoked from the DaemonSet, it's not running on boot. Setting After/Before won't do anything here and given it is started from the Pod, kube-node-configuration and kubelet had to be active.

examples/gke/xfs-format-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+            - "-c"
+          args:
+            - |
+              set -euExo pipefail
+
+              cp -a /usr/local/lib/scylla-operator/gke/xfs-format /host/var/xfs-format
+              chmod a+x /host/var/xfs-format/xfs-format.sh
+              chmod a+x /host/var/xfs-format/bin/tune2fs
+
+              cp -a /usr/local/lib/scylla-operator/gke/systemd/. /host/etc/systemd/
```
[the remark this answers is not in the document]
rzetelskik: What if we wanted to create a similar thing for other providers? I think it should at least contain `gke` somewhere in the path, I guess we can get rid of `scylla-operator` though.

hack/gke/systemd/system/xfs-format.service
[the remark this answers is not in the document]
tnozicka: why is it a service?

examples/gke/xfs-format-daemonset.yaml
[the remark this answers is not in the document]
tnozicka: my bad, I meant the target, so `/host/usr/local/lib/systemd/system`

hack/gke/systemd/system/xfs-format.service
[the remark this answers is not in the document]
rzetelskik: Is there a better way to run `xfs-format.sh`, provided it stops kubelet right at the start?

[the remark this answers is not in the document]
tnozicka: I guess I was initially thinking `nohup` style, but the service will give a nice journald output.

hack/gke/systemd/system/xfs-format.service
```diff
@@ -0,0 +1,6 @@
...
+Type=forking
```
tnozicka, line 5: oneshot?

examples/gke/xfs-format-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+            - "/bin/bash"
+            - "-c"
+          args:
+            - |
+              set -euExo pipefail
+
+              systemctl daemon-reload
+              systemctl enable mnt-raid\\x2ddisks-disk0.mount
+              systemctl start xfs-format
+
```
tnozicka, line 61: guess you should follow the xfs_format logs so it's propagate into the DS logs

examples/gke/xfs-format-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+            - "chroot"
+            - "/host"
+            - "/bin/bash"
+            - "-c"
+          args:
+            - |
+              set -euExo pipefail
+
+              systemctl daemon-reload
+              systemctl enable mnt-raid\\x2ddisks-disk0.mount
```
tnozicka, line 59: `--now`? I guess ideally we need to wait for the mount to be actually created

rzetelskik: I don't think we care whether it's mounted before the reboot though?

hack/gke/xfs-format/xfs-format.sh
rzetelskik: I wanted to fork to avoid seeing `xfs-format.service: Main process exited, code=killed, status=15/TERM` in xfs-format's journal, I suppose its somewhat confusing. On the other hand, forking logs another start/stop of the service. Wdyt?

hack/gke/xfs-format/xfs-format.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+if [[ "$( blkid -o value -s TYPE ${raid_device} )" == "xfs" ]]; then
+  echo "Raid device's file system is already xfs."
+  exit 0
+fi
+
+systemctl stop kube-node-configuration.service
+systemctl stop kubelet.service
+/home/kubernetes/bin/crictl stopp $( /home/kubernetes/bin/crictl pods -q )
```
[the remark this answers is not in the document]
rzetelskik: that's only in root user profile; should the script switch users or is it better to keep it that way?

hack/gke/xfs-format/xfs-format.sh
```diff
@@ -0,0 +1,40 @@
...
+systemctl stop containerd.service
+
+tmp_dir=$( mktemp -d )
+cp -rf "${ephemeral_mountpoint}"/. "${tmp_dir}"/
+
+umount $( grep -e '/home/kubernetes/containerized_mounter' -e '/var/lib/containerd' -e '/home/containerd' /proc/mounts | sed -E -e 's/([^ ]+) ([^ ]+) .+/\2/' | sort -u -r ) || true
```
tnozicka, line 28: would a condition on the number of returned arguments be safer then ignoring the error?

hack/gke/xfs-format/xfs-format.sh
tnozicka: I haven't used it but looking at `systemctl show systemd-reboot.service`,`SuccessAction=reboot` might work

examples/gke/xfs-format-daemonset.yaml
tnozicka: we'd at least fail sooner if something didn't work ?

examples/gke/xfs-format-daemonset.yaml
rzetelskik: at this point kubelet is already stopped so we'll never see the logs

hack/gke/xfs-format/xfs-format.sh
rzetelskik: SuccessAction is not recognised with oneshot. I've used ExecStartPost=/usr/sbin/reboot which is only triggered if ExecStart ran successfully, so it required changing the status code to 1 if the device is already xfs and checking for that in the daemonset as well. The sigkill is logged in the journal anyway, but I guess we can live with it.
```
Sep 14 11:54:24 gke-rzetelskik-scylla-dem-scylla-pool-465fca7d-r468 xfs-format.sh[10568]: + chmod a+w /mnt/raid-disks/disk0
Sep 14 11:54:24 gke-rzetelskik-scylla-dem-scylla-pool-465fca7d-r468 systemd[1]: xfs-format.service: Control process exited, code=killed, status=15/TERM
```

hack/gke/xfs-format/xfs-format.sh
rzetelskik: I've changed it to a loop as some of the results wouldn't be mounted when they were called (I suppose they were binds) which would result in an error. So I'm checking for that first one by one.

examples/gke/xfs-format-daemonset.yaml
tnozicka: you could start it on background so the logs make it here

rzetelskik: Do you mean starting the service in the background? If so, I don't see how that helps. Kubelet is stopped right away anyway and after the reboot the container is restarted.

Makefile
```diff
@@ -493,11 +493,12 @@ test-e2e:
 	$(GO) run ./cmd/scylla-operator-tests run $(GO_TEST_E2E_EXTRA_ARGS)
 .PHONY: test-e2e
 
-test-version-script:
+test-scripts:
 	./hack/lib/tag-from-gh-ref.sh
-.PHONY: test-version-script
+	./hack/test/gke/xfs-formatter/tune2fs_test.sh
```
tnozicka, line 498: nit: maybe less dirs :)

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,77 @@
+# Daemonset that converts ephemeral-storage to xfs
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: xfs-formatter
+spec:
+  selector:
+    matchLabels:
+      name: xfs-formatter
```
tnozicka, line 9: nit: `app: xfs-formatter`

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - "/bin/bash"
+        - "-c"
+        args:
+        - |
+          set -euExo pipefail
+          shopt -s inherit_errexit
+
+          cp -r /usr/local/lib/scylla-operator/gke/xfs-formatter /host/var/lib/
+
+          if [[ ! -d /host/usr/local/lib/systemd ]]; then mkdir /host/usr/local/lib/systemd; fi
```
tnozicka, line 38: nit: make it pretty on multiple lines (applies globally)

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,77 @@
...
+          if [[ ! -d /host/usr/local/lib/systemd/system ]]; then mkdir /host/usr/local/lib/systemd/system; fi
+          cp -r /usr/local/lib/scylla-operator/gke/systemd/. /host/usr/local/lib/systemd/system
+
+          if [[ ! -d /host/usr/local/lib/systemd/system/kube-node-configuration.service.d ]]; then mkdir /host/usr/local/lib/systemd/system/kube-node-configuration.service.d; fi
+          cat<<-EOF > /host/usr/local/lib/systemd/system/kube-node-configuration.service.d/10-path.conf
+            [Service]
+            Environment="PATH=/var/lib/xfs-formatter/bin:$( chroot /host /bin/bash -c 'echo "${PATH}"' )"
+          EOF
+
+          if [[ ! -d /host/mnt/raid-disks ]]; then mkdir /host/mnt/raid-disks; fi
```
tnozicka, line 48: (last time we chatted on this I forgot that you'll also need to handle reentrancy with ifs for the mkdir, otherwise I could have lived with `-p` even with the disadvantages, sorry)

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,77 @@
...
+          if [[ ! -d /host/mnt/raid-disks/disk0 ]]; then mkdir /host/mnt/raid-disks/disk0; fi
+          chmod a+w /host/mnt/raid-disks/disk0
+
+          chroot /host /bin/bash -euo pipefail -c '
```
tnozicka, line 52: why not `-x`?

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,77 @@
...
+        readinessProbe:
+          exec:
+            command:
+              - "chroot"
+              - "/host"
+              - "/bin/bash"
+              - "-euExo"
+              - "pipefail"
+              - "-c"
+              - "shopt -s inherit_errexit; xfs_info \"$( grep -E '/mnt/raid-disks/disk0($| )' /proc/mounts | sed -E -e 's/([^ ]+) .+/\\1/' )\""
```
tnozicka, line 71: nit: multiline

hack/gke/xfs-formatter/xfs-formatter.sh
```diff
@@ -0,0 +1,35 @@
...
+
+raid_device="$( head -n 1 "${RAID_DEVICE_FILE}" )"
+
+systemctl stop kube-node-configuration.service
+systemctl stop kubelet.service
+crictl_pods="$( /home/kubernetes/bin/crictl pods -q )"
+/home/kubernetes/bin/crictl stopp ${crictl_pods}
+systemctl stop containerd.service
+
+tmp_dir=$( mktemp -d )
```
tnozicka, line 16: nit: quotes

hack/gke/xfs-formatter/xfs-formatter.sh
```diff
@@ -0,0 +1,35 @@
...
+shopt -s inherit_errexit
+
+ephemeral_mountpoint='/mnt/stateful_partition/kube-ephemeral-ssd'
+
+raid_device="$( head -n 1 "${RAID_DEVICE_FILE}" )"
+
+systemctl stop kube-node-configuration.service
+systemctl stop kubelet.service
+crictl_pods="$( /home/kubernetes/bin/crictl pods -q )"
+/home/kubernetes/bin/crictl stopp ${crictl_pods}
```
tnozicka, line 13: nit: quotes

hack/test/gke/xfs-formatter/tune2fs_test.sh
```diff
@@ -0,0 +1,54 @@
...
+shopt -s inherit_errexit
+
+message_prefix='Testing tune2fs'
+
+echo "${message_prefix}..." >&2
+
+dir=$( dirname "${0}" )
+bin=$( readlink -f "${dir}/../../../gke/xfs-formatter/bin/tune2fs" )
+if [[ ! -f "${bin}" ]]; then
+  echo "tune2fs binary does not exist at ${bin}"
```
tnozicka, line 16: `>&2`

hack/gke/systemd/xfs-formatter.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Service converting ephemeral storage to xfs
+
+[Service]
+Type=oneshot
+Environment="RAID_DEVICE_FILE=/var/lib/xfs-formatter/.raid-device"
```
tnozicka, line 6: use a tmpfs and delete action on success

tnozicka:
good job driving it this far! I had a few comments but I think this is close to being able to land.

hack/gke/xfs-formatter/xfs-formatter.sh
rzetelskik: unquoted on purpose, there's multiple pods

Makefile
rzetelskik: What would you change it to? I thought it would be nicer if the test file structure reflected `hack`.

hack/gke/xfs-formatter/xfs-formatter.sh
tnozicka: you could use the array expansion but let's keep this as podIDs are unlikely to have spaces

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,92 @@
...
+          exec:
+            command:
+              - "chroot"
+              - "/host"
+              - "/bin/bash"
+              - "-euExo"
+              - "pipefail"
+              - "-c"
+              - "shopt -s inherit_errexit"
+              - "xfs_info \"$( grep -E '/mnt/raid-disks/disk0($| )' /proc/mounts | sed -E -e 's/([^ ]+) .+/\\1/' )\""
```
tnozicka, line 86: nit: `- |` to avoid escaping quotes, ideally consistent command vs. args with https://github.com/scylladb/scylla-operator/pull/770/files#diff-c98bf62e26f5b0c21fccffc335ea032ac98cb51c11c757a32e6da6da19575bd4R31

rueian:
Hi, a question about using XFS on GCP Local SSD. According to the document:
https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance, write cache flushing should be disable when using GCP Local SSD. The document use `nobarrier` option to mount an EXT4. Should mounting XFS use a similar option?

zimnx:
> Hi, a question about using XFS on GCP Local SSD. According to the document: https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance, write cache flushing should be disable when using GCP Local SSD. The document use `nobarrier` option to mount an EXT4. Should mounting XFS use a similar option?

`nobarrier` option on XFS partition is deprecated/removed. If you would like to disable write cache flushing you can do it via
`echo "write through" > /sys/block/$device/queue/write_cache`
This gives nice performance boost, although you risk loosing 2s of data according to GCP documentation.

rueian:
> > Hi, a question about using XFS on GCP Local SSD. According to the document: https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance, write cache flushing should be disable when using GCP Local SSD. The document use `nobarrier` option to mount an EXT4. Should mounting XFS use a similar option?
> 
> `nobarrier` option on XFS partition is deprecated/removed. If you would like to disable write cache flushing you can do it via `echo "write through" > /sys/block/$device/queue/write_cache` This gives nice performance boost, although you risk loosing 2s of data according to GCP documentation.

Hi @zimnx, thanks. Does this option mean that linux will cache the write and therefore no barrier is needed?

And another question: It seems that the ephemeral storage is shared https://github.com/kubernetes/kubernetes/blob/dd2d12f6dc0e654c15d5db57a5f9f6ba61192726/cluster/gce/gci/configure-helper.sh#L499-L505. Is there a way to book a quota for scylla?

zimnx:
> Hi @zimnx, thanks. Does this option mean that linux will cache the write and therefore no barrier is needed?

It means writes will remain cached, but kernel won't issue any flushes. According to GCP documentation, caches are always flushed every 2s.
https://github.com/torvalds/linux/blob/076501ff6ba265a473689c112eda9f1f34f620b5/Documentation/block/queue-sysfs.txt#L144


> And another question: It seems that the ephemeral storage is shared https://github.com/kubernetes/kubernetes/blob/dd2d12f6dc0e654c15d5db57a5f9f6ba61192726/cluster/gce/gci/configure-helper.sh#L499-L505. Is there a way to book a quota for scylla?

It is shared, by quota you mean capacity quota or IO quota? Capacity can be set in ScyllaCluster CRD. IO is not supported on cgroups v1 k8s is using.

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,92 @@
...
+          EOF
+
+          if [[ ! -d /host/mnt/raid-disks ]]; then
+            mkdir /host/mnt/raid-disks
+          fi
+
+          if [[ ! -d /host/mnt/raid-disks/disk0 ]]; then
+            mkdir /host/mnt/raid-disks/disk0
+          fi
+          chmod a+w /host/mnt/raid-disks/disk0
```
tnozicka, line 63: why `a+w`?

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+          chroot /host /bin/bash -euxo pipefail -c '
```
tnozicka, line 65: `-O inherit_errexit`

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,92 @@
...
+          exec:
+            command:
+              - "chroot"
+              - "/host"
+              - "/bin/bash"
+              - "-euExo"
+              - "pipefail"
+              - "-c"
+              - |
+                 shopt -s inherit_errexit
```
tnozicka, line 85: to be consistent with pipefail you could do `-O inherit_errexit`

hack/gke/xfs-formatter/xfs-checker.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -euExo pipefail
+shopt -s inherit_errexit
+
+raid_device="$( grep -E "/mnt/stateful_partition/kube-ephemeral-ssd($| )" /proc/mounts | sed -E -e 's/([^ ]+) .+/\1/' )"
+
+if [[ -z "${raid_device}" ]]; then
+  if [[ -f "${RAID_DEVICE_FILE}" ]]; then
+    raid_device="$( head -n 1 "${RAID_DEVICE_FILE}" )"
+  fi
```
tnozicka, line 11: nit: add an else branch for better diagnostics

hack/gke/xfs-formatter/xfs-checker.sh
```diff
@@ -0,0 +1,24 @@
...
+  if [[ -z "${raid_device}" ]]; then
+    echo "Raid device not mounted"
+    exit 255
+  fi
+fi
+
+echo "${raid_device}" > "${RAID_DEVICE_FILE}"
```
tnozicka, line 18: should this be written only if we have autodetected the device and not when we read it from the same file?

hack/gke/xfs-formatter/xfs-formatter.sh
tnozicka: but we need to check for the case where there is no pod

hack/gke/xfs-formatter/xfs-formatter.sh
```diff
@@ -0,0 +1,35 @@
...
+systemctl stop containerd.service
+
+tmp_dir="$( mktemp -d )"
+cp -rf "${ephemeral_mountpoint}"/. "${tmp_dir}"/
+
+function safe_umount_multiple_mount_points {
+  for line in ${1}; do
+    if mountpoint -q "${line}"; then umount "${line}"; fi
```
tnozicka, line 21: nit: dedicated lines

docs/source/gke.md
```diff
@@ -129,17 +127,18 @@ kubectl create clusterrolebinding cluster-admin-binding --clusterrole cluster-ad
 
 If you don't have Helm installed, Go to the [helm docs](https://docs.helm.sh/using_helm/#installing-helm) to get the binary for your distro.
 
-#### Install RAID DaemonSet
+#### Install xfs-format DaemonSet
 
-To combine the local disks together in RAID0 arrays, we deploy a DaemonSet to do the work for us.
+To run Scylla, it is necessary to convert ephemeral storage's filesystem to xfs. Deploy the xfs-format DaemonSet to have it taken care of.
+Unfortunately, GKE is only able to provision the ephemeral storage with `ext4` filesystem while Scylla requires `xfs` filesystem. Deploying the `xfs-format` DaemonSet will format the storage as `xfs` and prevent GKE from reformatting it back to `ext4.
```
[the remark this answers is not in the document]
tnozicka: We should make it clear that this is super hacky and, although we did our best to limit the risk, the is still some in case google changes their logic.

tnozicka (commented):
pls squash the commits

hack/gke/xfs-formatter/xfs-formatter.sh
rzetelskik: does it refer to formatting or explicitly running the command on each mount point (instead of having this function)?

examples/gke/xfs-formatter-daemonset.yaml
rzetelskik: I wanted to recreate what GKE was doing in [configure-helper.sh](https://github.com/kubernetes/kubernetes/blob/c2a633e07ec6129398dfef95c9e011ea5535ded2/cluster/gce/gci/configure-helper.sh#L283), which we're interfering with. What would you suggest instead?

hack/gke/xfs-formatter/xfs-formatter.sh
tnozicka: formatting

examples/gke/xfs-formatter-daemonset.yaml
tnozicka: it's fine then

tnozicka (commented):
doc nits, otherwise it lgtm

rzetelskik:
applied suggestions and rebased on current master

tnozicka (approved):
/lgtm
@rzetelskik thanks for all the effort you've put into this!

@zimnx PTAL - when you are happy with the PR, I'd like to ask @rzetelskik to manually test before we merge. (Unless there were no changes to the code since the last time her tried.)

tnozicka:
(also noting down that @vponomaryov will try to report back with QA results before we merge)

tnozicka (changes requested):
please add hook up into update in the Makefile so the DS image matches when we branch

vponomaryov:
Pods of the newly introduced daemonset fails with the following error:
```
+ cp -r /usr/local/lib/scylla-operator/gke/xfs-formatter /host/var/lib/
+ [[ ! -d /host/usr/local/lib/systemd ]]
+ [[ ! -d /host/usr/local/lib/systemd/system ]]
+ cp -r /usr/local/lib/scylla-operator/gke/systemd/. /host/usr/local/lib/systemd/system
+ [[ ! -d /host/usr/local/lib/systemd/system/kube-node-configuration.service.d ]]
+ cat
++ chroot /host /bin/bash -c 'echo "${PATH}"'
+ [[ ! -d /host/mnt/raid-disks ]]
+ [[ ! -d /host/mnt/raid-disks/disk0 ]]
+ chmod a+w /host/mnt/raid-disks/disk0
+ chroot /host /bin/bash -euxo pipefail -O inherit_errexit -c '
    systemctl daemon-reload
    systemctl enable --now mnt-raid\\x2ddisks-disk0.mount
    systemctl start xfs-formatter.service
    journalctl -xeu xfs-formatter -f
    '
+ systemctl daemon-reload
+ systemctl enable --now 'mnt-raid\x2ddisks-disk0.mount'
+ systemctl start xfs-formatter.service
Failed to start xfs-formatter.service: Unit xfs-formatter.service is not loaded properly: Exec format error.
See system logs and 'systemctl status xfs-formatter.service' for details.
```

I suspect that the "rm" binary in the "hack/gke/systemd/xfs-formatter.service" file must have absolute path.

examples/gke/xfs-formatter-daemonset.yaml
```diff
@@ -0,0 +1,95 @@
...
+    matchLabels:
+      app: xfs-formatter
+  template:
+    metadata:
+      labels:
+        app: xfs-formatter
+    spec:
+      hostPID: true
+      nodeSelector:
+        cloud.google.com/gke-ephemeral-storage-local-ssd: "true"
```
tnozicka, line 17: You'll need to elaborate on why is this needed given the admin has already labeled nodes with the other one.

Makefile
```diff
@@ -397,6 +397,12 @@ define set-default-app-version
 	$(YQ) eval -i -P '.appVersion = "$(2)"' '$(1)'/Chart.yaml
 endef
 
+# $1 - file path
+# $2 - target image ref
+define replace-image-ref
+	sed -E -i 's,^(\s*image\s*:\s*).*,\1$(2),' '$(1)'
```
tnozicka, line 403: would YQ be less error prone?

examples/gke/xfs-formatter-daemonset.yaml
rzetelskik: I've added it after Valerii tried to deploy the DS without the beta feature enabled, so that it would be somewhat more robust, but I suppose this label could be added to a node anyway, so maybe this is redundant with the other label. Wdyt?

Makefile
rzetelskik: changed to yq

examples/gke/xfs-formatter-daemonset.yaml
rzetelskik: @tnozicka My proposed solution would be to remove the `scylla.scylladb.com/gke-ephemeral-storage-local-ssd=true` label and only leave the one created by GKE, i.e. `cloud.google.com/gke-ephemeral-storage-local-ssd: "true"`. Instead of the additional label, we could add the following toleration to the DaemonSet:
```
  tolerations:
    - key: role
      operator: Equal
      value: scylla-clusters
      effect: NoSchedule
```
which is probably a more robust way to approach it.

vponomaryov (dismissed):
Tested it. Works.

examples/gke/xfs-formatter-daemonset.yaml
tnozicka: I think it's correct to migrate over to the GKE label.

Not sure what you aim for with the toleration instead of node selector. The selector restricts "which nodes to land on" while the toleration means "allow landing on these node" (but also elsewhere).

rzetelskik: Right, that didn't make much sense. I've changed the set of nodeSelectors to two GKE provided ones:
```
      nodeSelector:
        cloud.google.com/gke-ephemeral-storage-local-ssd: "true"
        cloud.google.com/gke-nodepool: "scylla-pool"
```

docs/source/releases.md
```diff
@@ -45,10 +45,10 @@ GA images aren't build from scratch but rather promoted from an existing release
...
-| :----------------: | :---------: | :---------: | :---------: | :---------: | :---------: | :---------: | :--------: |
-| Kubernetes         | `>=1.19.10` | `>=1.19.10` | `>=1.19.10` | `>=1.19`    | `>=1.19`    | `>=1.11`    | `>=1.11`   |
-| Scylla OS          | `>=4.3`     | `>=4.3`     | `>=4.3`     | `>=4.2`     | `>=4.2`     | `>=4.0`     | `>=4.0`    |
-| Scylla Enterprise  | `>=2021.1`  | `>=2021.1`  | `>=2021.1`  | `>=2020.1`  | `>=2020.1`  | `>=2020.1`  | `>=2020.1` |
-| Scylla Manager     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`    |
-| Scylla Monitoring  | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`    |
+|                    | v1.7        | v1.6        | v1.5        | v1.4        | v1.3        | v1.2        | v1.1        | v1.0       |
+| :----------------: | :---------: | :---------: | :---------: | :---------: | :---------: | :---------: | :---------: | :--------: |
+| Kubernetes         | `>=1.20.10` | `>=1.19.10` | `>=1.19.10` | `>=1.19.10` | `>=1.19`    | `>=1.19`    | `>=1.11`    | `>=1.11`   |
```
tnozicka, line 50: why `1.20.10` and not `1.20`?

docs/source/releases.md
```diff
@@ -45,10 +45,10 @@ GA images aren't build from scratch but rather promoted from an existing release
...
+| Scylla OS          | `>=4.3`     | `>=4.3`     | `>=4.3`     | `>=4.3`     | `>=4.2`     | `>=4.2`     | `>=4.0`     | `>=4.0`    |
+| Scylla Enterprise  | `>=2021.1`  | `>=2021.1`  | `>=2021.1`  | `>=2021.1`  | `>=2020.1`  | `>=2020.1`  | `>=2020.1`  | `>=2020.1` |
+| Scylla Manager     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`     | `>=2.2`    |
+| Scylla Monitoring  | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`     | `>=1.0`    |
```
tnozicka, line 54: since 1.6 the supported versions are as follows:
```
Scylla Open Source ≥4.3
Scylla Enterprise ≥2021.1
Kubernetes ≥1.19.10
Scylla Manager ≥2.2
Scylla Monitoring Stack ≥3.0
```
https://www.scylladb.com/product/release-notes/scylla-operator-1-6/

docs/source/releases.md
rzetelskik: It's the latest master version available on GKE that runs systemd 245

docs/source/releases.md
rzetelskik: Sent a separate PR https://github.com/scylladb/scylla-operator/pull/902

tnozicka (approved):
/lgtm
thanks!

rueian:
> > Hi, a question about using XFS on GCP Local SSD. According to the document: https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance, write cache flushing should be disable when using GCP Local SSD. The document use `nobarrier` option to mount an EXT4. Should mounting XFS use a similar option?
> 
> `nobarrier` option on XFS partition is deprecated/removed. If you would like to disable write cache flushing you can do it via `echo "write through" > /sys/block/$device/queue/write_cache` This gives nice performance boost, although you risk loosing 2s of data according to GCP documentation.

Hi @rzetelskik, @zimnx, 

where should I run the echo "write through" > /sys/block/$device/queue/write_cache`?

I guess it is appropriate running it in the `xfs-checker.sh` right?

zimnx:
> Where should I run the `echo "write through" > /sys/block/$device/queue/write_cache`?
> 
> I guess it was appropriate to run it in the `xfs-checker.sh`, right?

This script is only used to prepare other Google scripts to not break when there's a XFS partition after reboot. 
Disabling the write cache is implemented in our tuning scripts, although at the moment it works only on Scylla Enterprise 2021.0+. If you're on it, you have to change the default image used for tuning scripts to Enteprise in ScyllaOperatorConfig custom resource created by default in 1.6.

PR #773, opened 2021-09-03
Fix default anti-affinity rules for scylla-operator

vponomaryov:
For the moment 'scylla-operator' pods have default pod anti-affinity
rules set to be scheduled not with "webhook-server" ones.
The same vice-versa. It makes no sense because pods of the same type
may be scheduled on the same node in such a case.
It violates the HA idea.
So, fix it by switching labels between these affinity rules making
each pod of a type be scheduled on a node where there are no pods of
such a type.

Resolves #

### DIFF
--- a/deploy/operator/50_operator.deployment.yaml
+++ b/deploy/operator/50_operator.deployment.yaml
@@ -42,7 +42,7 @@ spec:
           - podAffinityTerm:
               labelSelector:
                 matchLabels:
-                  app.kubernetes.io/instance: webhook-server
-                  app.kubernetes.io/name: webhook-server
+                  app.kubernetes.io/instance: scylla-operator
+                  app.kubernetes.io/name: scylla-operator
               topologyKey: kubernetes.io/hostname
             weight: 1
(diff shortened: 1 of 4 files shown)


tnozicka:
thanks @vponomaryov, please split the generated changes into a separate commit named `Update generated`

vponomaryov:
> thanks @vponomaryov, please split the generated changes into a separate commit named `Update generated`

Done

tnozicka:
you'll need to shuffle the files between the commits. `helm/scylla-operator/values.yaml` is the source file, the rest is generated.

vponomaryov:
> you'll need to shuffle the files between the commits. `helm/scylla-operator/values.yaml` is the source file, the rest is generated.

Done
P.S. originally I separated files based on this commit: https://github.com/scylladb/scylla-operator/commit/243296e8

tnozicka:
yeah, it's not easy to see what's generated and what isn't. in the commit that you referenced that's helm schema (`helm/scylla/values.schema.json`) which is indeed generated but not the `values.yaml` (nor the helm templates)

tnozicka (approved):
/lgtm
thanks!

zimnx (approved):

PR #774, opened 2021-09-06
Pin minikube version

tnozicka:
We should pin minikube version in the CI and bump it manually so the new version goes through CI. Currently the version is bumped through the `latest` tag at any time and if the new version doesn't work or it needs system adjustments, the CI gets borked.

3 consequent nightly failures
- https://github.com/scylladb/scylla-operator/actions/runs/1199656383
- https://github.com/scylladb/scylla-operator/actions/runs/1201866410
- https://github.com/scylladb/scylla-operator/actions/runs/1204056939

### DIFF
--- a/.github/actions/setup-minikube/action.yaml
+++ b/.github/actions/setup-minikube/action.yaml
@@ -8,7 +8,7 @@ runs:
     run: |
       set -x
       sudo apt-get install -y --no-install-recommends conntrack
-      sudo curl --fail -L https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 -o /usr/local/bin/minikube
+      sudo curl --fail -L https://storage.googleapis.com/minikube/releases/v1.22.0/minikube-linux-amd64 -o /usr/local/bin/minikube
       sudo chmod +x /usr/local/bin/minikube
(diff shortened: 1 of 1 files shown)


.github/actions/setup-minikube/action.yaml
```diff
@@ -8,7 +8,7 @@ runs:
     run: |
       set -x
       sudo apt-get install -y --no-install-recommends conntrack
-      sudo curl --fail -L https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 -o /usr/local/bin/minikube
+      sudo curl --fail -L https://storage.googleapis.com/minikube/releases/v1.22.0/minikube-linux-amd64 -o /usr/local/bin/minikube
```
zimnx, line 11: nit: version could be extracted to env variable, similar to `go_version`

zimnx (approved):
one nit. feel free to ignore and merge

.github/actions/setup-minikube/action.yaml
tnozicka: it would have to be an argument for it to be able to reuse outside this library file (`go-version` arg vs. the `go_version` variable) but given this action is dedicated for our repo and we use it with just one version for now I'd postpone that to when we actually need it

PR #780, opened 2021-09-10
Reserve time for flake retries

tnozicka:
We retry flakes on promotion jobs to prevent failures that would need to be manually retriggered. But when a job flakes it's often done by waiting for an event that doesn't happen until it hits an internal timeout, which adds up. Given the global test timeout is pretty stretched this can lead to a situation when we should retry the flake but the e2e suite is terminated by the global timeout, like in https://github.com/scylladb/scylla-operator/runs/3557449206?check_suite_focus=true#step:11:776 . 

This PR adds 10 minutes per flake attempt to the overall timeout.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -176,12 +176,14 @@ jobs:
       run: |
         set -euExo pipefail
 
+        e2e_timeout_minutes=60
         flake_attempts=0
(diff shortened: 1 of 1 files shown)


rzetelskik (approved):
lgtm, thanks

PR #793, opened 2021-09-21
Expose shard aware ports in Scylla Services

### LINKED ISSUE #750: Expose shard-aware ports

Recent Scylla versions expose additional shard-aware port 19042, drivers can use this port to directly connect to each shard.

zimnx:
Newest shard aware drivers may use this port to establish direct shard
connection to speedup initializaiton of connection pool.

Fixes #750


pkg/controller/scyllacluster/resource/resource.go
```diff
@@ -119,6 +119,14 @@ func memberServicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
 			Name: "cql-ssl",
 			Port: 9142,
 		},
+		{
+			Name: "cql-shard-aware",
+			Port: 19042,
+		},
+		{
+			Name: "cql-shard-aware-ssl",
```
tnozicka, line 127: nit: if the original port is named `cql-ssl` would it be better to use `cql-ssl-shard-aware`?

tnozicka (commented):
I wonder how practical an e2e would be

pkg/controller/scyllacluster/resource/resource.go
zimnx: done

zimnx:
> I wonder how practical an e2e would be

It's used by default and it's transparent to clients. Although I added an e2e to validate whether we are connecting to shard aware port.

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,113 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var _ = g.Describe("ScyllaCluster shardawareness", func() {
+	defer g.GinkgoRecover()
+
+	f := framework.NewFramework("scyllacluster")
+
+	g.It("connection pool initialization via shard aware ports", func() {
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,113 @@
...
+		framework.By("Waiting for the ScyllaCluster to deploy")
+		waitCtx1, waitCtx1Cancel := contextForRollout(ctx, sc)
+		defer waitCtx1Cancel()
+		sc, err = waitForScyllaClusterState(waitCtx1, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, scyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		hosts, err := getHosts(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		mut := sync.Mutex{}
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,113 @@
...
+		clusterConfig := gocql.NewCluster(hosts...)
+		clusterConfig.Dialer = DialerFunc(func(ctx context.Context, network, addr string) (net.Conn, error) {
+			sourcePort := gocql.ScyllaGetSourcePort(ctx)
+			localAddr, err := net.ResolveTCPAddr(network, fmt.Sprintf(":%d", sourcePort))
+			if err != nil {
+				return nil, err
+			}
+
+			framework.Infof("Connecting to %s using %d source port", addr, sourcePort)
+			mut.Lock()
```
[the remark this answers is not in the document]
zimnx: connection dialing shouldn't be covered by critical section, and I don't like inline lambdas just to use the defer. Limited critical section to minimum

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,113 @@
...
+		defer session.Close()
+
+		err = wait.PollImmediate(100*time.Millisecond, 5*time.Second, func() (done bool, err error) {
+			mut.Lock()
+			defer mut.Unlock()
+			return len(connections) == nrShards, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Expect(shardAwareAttempts).To(o.Equal(nrShards - 1))
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,114 @@
...
+package scyllacluster
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sync"
+	"time"
+
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
```
tnozicka, line 12: order

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,114 @@
...
+		)
+
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
+
+		sc := scyllaclusterfixture.BasicScyllaCluster.ReadOrFail()
+		sc.Spec.Datacenter.Racks[0].Members = 1
+
+		// Ensure 2 shards.
+		sc.Spec.Datacenter.Racks[0].Resources.Limits[corev1.ResourceCPU] = resource.MustParse(fmt.Sprintf("%d", nrShards))
```
tnozicka, line 45: this is harsh.

CI nodes have 2 cores exactly, so the other ~16 tests may keep you out of those CPUs for a long time (== flakes)

also if we ever run a DS by default this won't ever land on a node because of resources

tnozicka (commented):
we either need to find a way to have more shards in dev mode (currently hardcoded to cpu limits) or to test just our part - is there a shard aware port exposed - and leave testing it works to scylla itself

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
zimnx: it's not going to wait, It still has resource.requests set to low value. Limits in this case only control number of CFQ shares, doesn't affect scheduling.

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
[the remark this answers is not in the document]
tnozicka: I don't see any of these comments applied, should be `var connectionsLock sync.Mutex` or `var connectionsMutex sync.Mutex`

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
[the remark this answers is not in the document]
tnozicka: You could extract it into an object with set/get functions if you don't like lambdas, especially given this lock guards 2 variables at once so the naming convention won't be enough. Not using defer with locks is very prone to future bugs when someone edits the code in the middle. Sometimes you won't even see it in the diff on reviews.

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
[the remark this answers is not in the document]
tnozicka: did you forget to push it?

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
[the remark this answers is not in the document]
tnozicka: I don't see this addressed either

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,114 @@
...
+				o.Expect(addr).To(o.HaveSuffix(fmt.Sprintf("%d", nonShardAwarePort)))
+				continue
+			}
+			o.Expect(addr).To(o.HaveSuffix(fmt.Sprintf("%d", shardAwarePort)))
+			shardAwareAttempts++
+		}
+
+		// Control connection always use the default port to discover the shard number,
+		// and is added to the pool with shard number it randomly gets.
+		o.Expect(shardAwareAttempts).To(o.Equal(nrShards - 1))
```
tnozicka, line 106: should there be 2 shard aware attempts? this assert just 1

zimnx: No, because first connection is always to non-shard aware port to discover the number of shards. There's a comment about it above

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
zimnx: oops, now it's in right place

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,114 @@
...
+		waitCtx1, waitCtx1Cancel := utils.ContextForRollout(ctx, sc)
+		defer waitCtx1Cancel()
+		sc, err = utils.WaitForScyllaClusterState(waitCtx1, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		hosts, err := utils.GetHosts(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		connections := map[uint16]string{}
+		connectionsMut := sync.Mutex{}
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
tnozicka: What is not clear to me is that if you use the first connection to discover the number of shards, why doesn't it connect to each of them afterwards?

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -0,0 +1,113 @@
...
+		framework.By("Waiting for the ScyllaCluster to deploy")
+		waitCtx1, waitCtx1Cancel := utils.ContextForRollout(ctx, sc)
+		defer waitCtx1Cancel()
+		sc, err = utils.WaitForScyllaClusterState(waitCtx1, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		hosts, err := utils.GetHosts(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		connections := make(map[uint16]string)
```
tnozicka, line 62: you had it nicer initially as `connections := map[uint16]string{}`

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
zimnx: Because it's already established to random shard. There's no point to do termination and roundtip again if only missing ones can be established.

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
zimnx: Then we need to establish coding convention on what is nicer and not ping pong each variable instation on every PR. 
Initial approach is syntactic sugar for calling `make`: https://go.godbolt.org/z/W17z6vPex

tnozicka: We normally refer to Kube style guide although I don't see the topic there while briefly looking. While I though we have discussed it before, I'd always prefer the more readable initialization style, unless you actually want to pre-allocate the map with some value.

tnozicka (approved):
/lgtm
thanks!

PR #796, opened 2021-09-24
Allow removing empty racks from ScyllaCluster Spec

### LINKED ISSUE #287: Rack removing is not supported
**Description**
s-o throws following error, when you submit cluster with spec that has one rack less than current one:
denied the request: racks [us-east-1a-1] not found, you cannot remove racks from the spec

Enable rack removing, possible options:

- Delete empty rack (with members: 0)
- Delete non-emptu rack  
-

### LINKED ISSUE #798: Relax limitation of rack removal
* Feature Request

Currently users cannot remove racks from ScyllaClsuter due to missing automatic decommision.
Even racks with 0 members cannot be removed. In order to relax this limitation, we should allow to remove empty racks. 
Rack removal will consist of two steps until #287 is implemented:
1. Scale to 0
2. Remove rack from spec

zimnx:
Currently users cannot remove racks from ScyllaClsuter due to missing automatic decommision. Even racks having 0 members cannot be removed. 
This change relaxes this limitation and allows to remove racks in two steps:
1. Scale to 0
2. Remove rack from spec

Fixes #798

### DIFF
--- a/pkg/api/validation/cluster_validation.go
+++ b/pkg/api/validation/cluster_validation.go
@@ -134,35 +134,53 @@ func ValidateScyllaClusterRackSpec(rack scyllav1.RackSpec, rackNames sets.String
 func ValidateScyllaClusterUpdate(new, old *scyllav1.ScyllaCluster) field.ErrorList {
 	allErrs := ValidateScyllaCluster(new)
 
-	return append(allErrs, ValidateScyllaClusterSpecUpdate(&new.Spec, &old.Spec, field.NewPath("spec"))...)
+	return append(allErrs, ValidateScyllaClusterSpecUpdate(new, old, field.NewPath("spec"))...)
 }
 
-func ValidateScyllaClusterSpecUpdate(newSpec, oldSpec *scyllav1.ScyllaClusterSpec, fldPath *field.Path) field.ErrorList {
+func ValidateScyllaClusterSpecUpdate(new, old *scyllav1.ScyllaCluster, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	// Check that the datacenter name didn't change
-	if oldSpec.Datacenter.Name != newSpec.Datacenter.Name {
+	if old.Spec.Datacenter.Name != new.Spec.Datacenter.Name {
 		allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "name"), "change of datacenter name is currently not supported"))
 	}
 
 	// Check that all rack names are the same as before
 	oldRackNames, newRackNames := sets.NewString(), sets.NewString()
-	for _, rack := range oldSpec.Datacenter.Racks {
+	for _, rack := range old.Spec.Datacenter.Racks {
 		oldRackNames.Insert(rack.Name)
 	}
-	for _, rack := range newSpec.Datacenter.Racks {
+	for _, rack := range new.Spec.Datacenter.Racks {
 		newRackNames.Insert(rack.Name)
 	}
 	diff := oldRackNames.Difference(newRackNames)
-	if diff.Len() != 0 {
-		allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks"), fmt.Sprintf("racks %v not found, you cannot remove racks from the spec", diff.List())))
+	for _, rackName := range diff.List() {
+		for i, rack := range old.Spec.Datacenter.Racks {
+			if rack.Name != rackName {
+				continue
+			}
+
+			if rack.Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q can't be removed because it still has members that have to be scaled down to zero first", rackName)))
+				continue
+			}
+
+			if old.Status.Racks[rack.Name].Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q can't be removed because the members are being scaled down", rackName)))
+				continue
+			}
+
+			if !isRackStatusUpToDate(old, rack.Name) {
+				allErrs = append(allErrs, field.InternalError(fldPath.Child("datacenter", "racks").Index(i), fmt.Errorf("rack %q can't be removed because its status, that's used to determine members count, is not yet up to date with the generation of this resource; please retry later", rackName)))
+			}
+		}
 	}
 
 	rackMap := make(map[string]scyllav1.RackSpec)
-	for _, oldRack := range oldSpec.Datacenter.Racks {
+	for _, oldRack := range old.Spec.Datacenter.Racks {
 		rackMap[oldRack.Name] = oldRack
 	}
-	for i, newRack := range newSpec.Datacenter.Racks {
+	for i, newRack := range new.Spec.Datacenter.Racks {
 		oldRack, exists := rackMap[newRack.Name]
 		if !exists {
 			continue
@@ -176,4 +194,11 @@ func ValidateScyllaClusterSpecUpdate(newSpec, oldSpec *scyllav1.ScyllaClusterSpe
 	}
 
 	return allErrs
+}
+
+func isRackStatusUpToDate(sc *scyllav1.ScyllaCluster, rack string) bool {
+	return sc.Status.ObservedGeneration != nil &&
+		*sc.Status.ObservedGeneration >= sc.Generation &&
+		sc.Status.Racks[rack].Stale != nil &&
+		!*sc.Status.Racks[rack].Stale
 }
--- a/pkg/api/validation/cluster_validation_test.go
+++ b/pkg/api/validation/cluster_validation_test.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/pointer"
 )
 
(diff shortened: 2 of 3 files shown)


rzetelskik (dismissed):
lgtm, although we could open a separate issue for providing support for removing non-empty Racks in the future, before #287 is closed

zimnx:
> lgtm, although we could open a separate issue for providing support for removing non-empty Racks in the future, before #287 is closed

done, created #798 and #287 will stay until we implement non-empty removal

pkg/api/validation/cluster_validation.go
```diff
@@ -154,8 +154,17 @@ func ValidateScyllaClusterSpecUpdate(newSpec, oldSpec *scyllav1.ScyllaClusterSpe
...
-	if diff.Len() != 0 {
-		allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks"), fmt.Sprintf("racks %v not found, you cannot remove racks from the spec", diff.List())))
+	for _, rackName := range diff.List() {
+		for i, rack := range oldSpec.Datacenter.Racks {
+			if rack.Name != rackName {
+				continue
+			}
+			if rack.Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %v not found, racks having members cannot be removed, scale it down to zero members first", rackName)))
+			} else if oldStatus.Racks[rack.Name].Members != 0 {
```
tnozicka, line 164: Is it gonna panic if the rack isn't present in the status?

tnozicka (commented):
What happens in a case the status is stale, iow. `status.observedGeneration != metadata.generation` ?

pkg/api/validation/cluster_validation.go
```diff
@@ -154,8 +154,17 @@ func ValidateScyllaClusterSpecUpdate(newSpec, oldSpec *scyllav1.ScyllaClusterSpe
...
 	}
 	diff := oldRackNames.Difference(newRackNames)
-	if diff.Len() != 0 {
-		allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks"), fmt.Sprintf("racks %v not found, you cannot remove racks from the spec", diff.List())))
+	for _, rackName := range diff.List() {
+		for i, rack := range oldSpec.Datacenter.Racks {
+			if rack.Name != rackName {
+				continue
+			}
+			if rack.Members != 0 {
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation.go
zimnx: Map returns the zero value when key is missing, and status has `map[string]RackStatus` type, so zero struct value would be also fine.

tnozicka: oh, I forgot we don't use pointers there

tnozicka (commented):
I am not super excited being dependent on the observed generation as e.g. the rollout of the other rack can block the admission this way, but I can live with that. It's certainly better then not checking it at all.

pkg/api/validation/cluster_validation.go
```diff
@@ -134,35 +134,45 @@ func ValidateScyllaClusterRackSpec(rack scyllav1.RackSpec, rackNames sets.String
...
-		allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks"), fmt.Sprintf("racks %v not found, you cannot remove racks from the spec", diff.List())))
+	for _, rackName := range diff.List() {
+		for i, rack := range old.Spec.Datacenter.Racks {
+			if rack.Name != rackName {
+				continue
+			}
+
+			if rack.Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %v not found, racks having members cannot be removed, scale it down to zero members first", rackName)))
+			} else if isStatusStale(old, rack.Name) || old.Status.Racks[rack.Name].Members != 0 {
```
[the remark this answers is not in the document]
zimnx: `continue` would move to another removed rack, where i want to generate errors for those having zero members

pkg/api/validation/cluster_validation.go
```diff
@@ -177,3 +187,8 @@ func ValidateScyllaClusterSpecUpdate(newSpec, oldSpec *scyllav1.ScyllaClusterSpe
 
 	return allErrs
 }
+
+func isStatusStale(sc *scyllav1.ScyllaCluster, rack string) bool {
+	return (sc.Status.ObservedGeneration != nil && *sc.Status.ObservedGeneration != sc.Generation) ||
+		(sc.Status.Racks[rack].Stale != nil && *sc.Status.Racks[rack].Stale)
+}
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation.go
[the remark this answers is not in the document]
tnozicka: if members != 0, you can set the error and continue to the next rack, right? the case of members != 0 is mutually exclusive with checking the staleness or status

pkg/api/validation/cluster_validation.go
```diff
@@ -134,35 +134,49 @@ func ValidateScyllaClusterRackSpec(rack scyllav1.RackSpec, rackNames sets.String
...
+		for i, rack := range old.Spec.Datacenter.Racks {
+			if rack.Name != rackName {
+				continue
+			}
+			if rack.Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q not found, racks having members cannot be removed, scale it down to zero members first", rackName)))
+			} else {
+				if old.Status.Racks[rack.Name].Members != 0 {
+					allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q not found, racks having members cannot be removed, wait for decommission to complete", rackName)))
+				}
```
tnozicka, line 167: if `old.Status.Racks[rack.Name].Members != 0` you don't need to check for staleness => continue

zimnx: done

pkg/api/validation/cluster_validation.go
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation.go
```diff
@@ -134,35 +134,53 @@ func ValidateScyllaClusterRackSpec(rack scyllav1.RackSpec, rackNames sets.String
...
-	if diff.Len() != 0 {
-		allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks"), fmt.Sprintf("racks %v not found, you cannot remove racks from the spec", diff.List())))
+	for _, rackName := range diff.List() {
+		for i, rack := range old.Spec.Datacenter.Racks {
+			if rack.Name != rackName {
+				continue
+			}
+
+			if rack.Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q not found, racks having members cannot be removed, scale it down to zero members first", rackName)))
```
tnozicka, line 164:
```suggestion
				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q can't be removed because it still has members that have to be scaled down to zero first", rackName)))
```

pkg/api/validation/cluster_validation.go
```diff
@@ -134,35 +134,53 @@ func ValidateScyllaClusterRackSpec(rack scyllav1.RackSpec, rackNames sets.String
...
+				continue
+			}
+
+			if old.Status.Racks[rack.Name].Members != 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q not found, racks having members cannot be removed, wait for decommission to complete", rackName)))
```
tnozicka, line 169:
```suggestion
				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q can't be removed because the members are being scaled down", rackName)))
```

pkg/api/validation/cluster_validation.go
```diff
@@ -134,35 +134,53 @@ func ValidateScyllaClusterRackSpec(rack scyllav1.RackSpec, rackNames sets.String
...
+				continue
+			}
+
+			if !isRackStatusUpToDate(old, rack.Name) {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q not found, racks can be removed from spec only when cluster is fully reconciled", rackName)))
```
tnozicka, line 174:
```suggestion
				allErrs = append(allErrs, field.InternalError(fldPath.Child("datacenter", "racks").Index(i), fmt.Sprintf("rack %q can't be removed because its status, that's used to determine members count, is not yet up to date with the generation of this resource; please retry later", rackName)))
```

tnozicka (commented):
wording nits to make a clear message for the users, otherwise lgtm

pkg/api/validation/cluster_validation.go
zimnx: done

pkg/api/validation/cluster_validation.go
zimnx: done

pkg/api/validation/cluster_validation.go
zimnx: done

tnozicka (approved):
/lgtm
thanks

PR #799, opened 2021-10-04
Bump Scylla and Scylla Manager Agent versions

zimnx:
Our examples use Scylla which is not supported anymore and is failing
to deploy due to additional parameters that are unknown in 4.2.0.

Scylla and Scylla Manager Agent versions are bumped to latest ones.

### DIFF
--- a/examples/eks/cluster.yaml
+++ b/examples/eks/cluster.yaml
@@ -39,8 +39,8 @@ metadata:
   name: scylla-cluster
   namespace: scylla
 spec:
-  version: 4.4.2
-  agentVersion: 2.3.0
+  version: 4.4.5
+  agentVersion: 2.5.1
   cpuset: true
   network:
     hostNetworking: true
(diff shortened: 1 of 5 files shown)


tnozicka (commented):
should you bump e2e fixtures as well?

zimnx:
Fixtures are already using supported versions:
https://github.com/scylladb/scylla-operator/blob/7a58628d34932be70e851b44058071dbb2bf2475/test/e2e/set/scyllacluster/config.go#L8-L11

tnozicka:
Would it be better to match the version in the examples as only the e2e's are actually tested? z-streams for scylla are different and agent is 2.3.0 vs. 2.5.1
https://github.com/scylladb/scylla-operator/blob/de735efcbc15ef4a0b12c18d12509b1797800d87/test/e2e/fixture/scyllacluster/basic.scyllacluster.yaml#L6-L7

zimnx:
> Would it be better to match the version in the examples as only the e2e's are actually tested? z-streams for scylla are different and agent is 2.3.0 vs. 2.5.1


bumped

tnozicka (approved):
/lgtm
thanks

zimnx:
clean force push to trigger broken GH actions

tnozicka (approved):

Issue #758, opened 2021-08-13
Operator helm chart: hardcoded namespace


slamdev:
**Describe the bug**
Namespace cannot be changed for scylla-operator helm chart

**Additional context**
It was possible before https://github.com/scylladb/scylla-operator/commit/7bb3e88656684117f7332c66e1392fb1eec22424 commit.

tnozicka:
This is intentional, it's a singleton instance of the helm chart. We apologize for the inconvenience as it was possible before but now any supported deployment should be deployed to the fixed namespaces we have claimed - `scylla-operator` and `scylla-manager`. This is also needed for using the internal DNS names and future artifacts collection CLI tool the will help collecting logs for the issues.

slamdev:
quite some companies have restriction on what namespaces can be used to deploy applications. by limiting the namespace choice you might lose them as users of your operator. Or, as in my case, force ppl to maintain a fork of the chart.

btw all the operators helm charts I am aware off provide such a flexibility

tnozicka:
One option is also to make an exception for the naming, as others who have such policies did. I see a difference between an application instance and a platform extension. Operators are platform extensions (also why you need to be system:admin to install the CRD). Would you apply you naming policy to the `default`, `kube-system` or `kube-node-lease` namespaces and asked them to renameable?

Ashtonian:
+1 for making this configurable..

rngcntr:
+1

baez90:
I'd rather differentiate between CRDs being a platform extension and an operator running in any namespace that **uses** this platform extension?
I'm not aware of any other operator that enforces the namespace it will be installed to and to be honest I do not really understand why this is any better than assuming a default namespace and allowing users to override it in case they'd like to deploy it to another one.
Also the internal DNS isn't really fixed or do you also enforce users to not change the cluster DNS domain?

rngcntr:
For those who are interested, [this](https://github.com/scylladb/scylla-operator/pull/667) is the PR in which the change was introduced.

brad1121:
Is this hard coded limitation stated in the docs anywhere. If it is, its not obvious.
Wasted a couple hours triple checking for typos and the like wondering why I could not get this in any other namespace.

PR #801, opened 2021-10-04
Make manager and operator namespaces configurable

rngcntr:
According to [the discussion](https://github.com/helm/helm/issues/5465) on best practices for k8s namespaces in helm, it is preferred to use `{{ .Release.Namespace }}` instead of hard coded namespaces for helm templates.

In the `scylla` chart, this style is already used. This PR extends the usage of `{{ .Release.Namespace }}` to the charts `scylla-manager` and `scylla-operator`. This enables users to deploy custom namespaces, e.g. one namespace which hosts the entire ScyllaDB tool stack.


zimnx:
It's intentional, see discussion here: https://github.com/scylladb/scylla-operator/issues/758

PR #802, opened 2021-10-06
Docs: Update Sphinx theme to v1x

lauranovich:
Upgrades the Scylla Theme to v1.x
This change does not include any HP fixes or any other issues which will be handled with additional PRs. Please merge this FIRST.

**Description of your changes:**

**Which issue is resolved by this Pull Request:**
Resolves #

PR #804, opened 2021-10-06
adds glob pattern to TOC to prevent missing toc entry errors.

lauranovich:
FIXES https://github.com/scylladb/scylla-operator/pull/802

PR #807, opened 2021-10-10
Fix build errors in TOC and index. Add Index file to node operations directory

lauranovich:
Fixes the build errors which happened when make preview is launched.

Replaces https://github.com/scylladb/scylla-operator/pull/804

FIXES #802


dgarcia360:
I'm looking at this issue now, I'll send the fix in a separate PR.

PR #808, opened 2021-10-11
Bump Golang to 1.17.2

tnozicka:
go 1.17 had a bug for embedded data (https://github.com/golang/go/issues/47754) that prevented us from using it, but that's now fixed.


tnozicka:
yeah, given I've bumped go.mod version, even the min version is now 1.17 and given 1.17 had a bug with embedding, 1.17.1 is the minimum. 

updated, thx

PR #824, opened 2021-11-05
Build and publish Helm charts as CI artifacts

zimnx:
Having Helm charts as CI artifacts may help maintainers
to tests changes provided in Pull Requests prior merging.


tnozicka (commented):
we already have similar steps for helm-publish - is there a way we could make it in stages? (in the makefile and in CI)

tnozicka (approved):
I am fine with followups

PR #830, opened 2021-11-11
Bump Scylla to 4.5.1 and manager to 2.5.2

tnozicka:
Updating to latest OpenSource versions:
 - scylla 4.5.1
 - manager 2.5.2


zimnx:
LGTM, would you mind splitting changes into manual+autogenerated? It would be easier to spot where stuff needs to be changed in the future.

tnozicka:
ok, I can try but in this case I didn't iterate like that

zimnx (approved):

PR #794, opened 2021-09-23
Fix deploying ScyllaClusters with multiple Racks

rzetelskik:
**Description of your changes:**
This PR fixes a bug described in #755 and hence allows for deploying ScyllaClusters with multiple Racks. StatefulSets are now deployed one by one to comply with [Adding a New Node Into an Existing Scylla Cluster](https://docs.scylladb.com/operating-scylla/procedures/cluster-management/add-node-to-cluster/) procedure. Therefore no new Rack will be deployed unless all existing Racks are ready. However, members of an existing Rack can still be decommissioned, even if not all existing Racks are ready.

**Which issue is resolved by this Pull Request:**
Resolves #755 #757

PR #840, opened 2021-11-17
sidecar: fix parallel racks creation

MerlinDMC:
**Problem:**
If multiple racks are present in the initial ScyllaCluster CRD multiple StatefulSet resources will be created.
This results in multiple Pods being created in parallel.

Since the logic for the first node only triggers when there are no other Pods it will instead now create a circular dependency between the StatefulSets as seed nodes.

This changes the logic that tries to find the "first created Pod" to also include itself and will allow a uniform seed list for each rack without circular dependencies.

### DIFF
--- a/pkg/sidecar/identity/member.go
+++ b/pkg/sidecar/identity/member.go
@@ -84,9 +84,11 @@ func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (st
 		return "", fmt.Errorf("internal error: can't find any pod for this cluster, including itself")
 	}
 
+	var allPods []*corev1.Pod
 	var otherPods []*corev1.Pod
 	for i := range podList.Items {
 		pod := &podList.Items[i]
+		allPods = append(allPods, pod)
 		if pod.Name != m.Name {
 			otherPods = append(otherPods, pod)
 		}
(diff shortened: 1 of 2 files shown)


zimnx:
Problem comes from fact that multiple racks are created in parallel, which is wrong because Scylla only supports synchronous topology changes. This will be fixed in #794, which will solve your issue too.

MerlinDMC:
@zimnx understood.

However the test case for "bootstrap with first created Pod when all are down" then still seems wrong as the expected value is not the first created pod but the second.

zimnx:
Yes you're right. Looks like we are sorting `otherPods` but less function takes i'th and j'th elements from `podList.Items` when creation timestamp is compared. Good catch!

Would you mind sending a PR fixing it?

MerlinDMC:
@zimnx this PR was what you're requesting. By fixing that logic the parallel racks creation equally is being fixed.

I can resubmit this PR if that is what you're requesting?

zimnx:
You can push to the same branch, and change PR title, description and code. 
Instead creating a new slice `allPods` you should use `otherPods` inside less function passed to sort. We don't want to set seeds to itself if cluster have more than 1 node.

MerlinDMC:
It will not set the seed to itself - it will set the seed to the first created pod IF all pods are down.
Before this change it was not always setting it to the first created pod.

Given 3 pods (all Not Ready) A, B and C started 10 seconds apart of each other.

The old logic would distribute seeds like this:
| Node | Start Time | Seed Node |
| ----- | ----------- | ----------- |
| A       | 12:00:00     | B                 |
| B       | 12:00:10     | A                 |
| C       | 12:00:30     | A or B         |

This results in a circular dependency and it makes little sense as your unit test states that:
> bootstrap with first created Pod when all are down
https://github.com/scylladb/scylla-operator/blob/865f049a5ff3d6c92ebbd144691d090d9feb7088/pkg/sidecar/identity/member_test.go#L85-L91

The first created pod in this case is A and not B.

The change I made deliberately fixes that misalignment and will distribute seeds like this:
| Node | Start Time | Seed Node |
| ----- | ----------- | ----------- |
| A       | 12:00:00     | A                 |
| B       | 12:00:10     | A                 |
| C       | 12:00:30     | A or B        |

The seed for pod C is determined by the Ready state of pods A and B - but it's seed is negligible in this scenario as it would always use the old logic and gain the first Ready marked pod as a seed and if none are Ready it will use the oldest one. C would never form a circular dependency.

If I change the code to not use `allPods` the description of your test does not make sense as for pod A it will never "bootstrap with first created Pod when all are down".

Issue #755, opened 2021-08-12
ScyllaCluster Pods not coming up when deploying multiple Racks


rzetelskik:
**Describe the bug**
When trying to deploy a ScyllaCluster with multiple Racks, none of the Pods will come up.

**To Reproduce**
1. Deploy Scylla Operator and Scylla Manager
2. Deploy a ScyllaCluster (provided in artifacts) with >1 Racks

**Expected behavior**
All Pods should come up.

**Config Files**
[artifacts.tar.gz](https://github.com/scylladb/scylla-operator/files/6976058/artifacts.tar.gz)

**Logs**
```
ERROR 2021-08-12 13:19:14,191 [shard 0] rpc - client 10.102.116.70:7000: fail to connect: Connection refused
...
ERROR 2021-08-12 13:20:04,702 [shard 0] token_metadata - sorted_tokens is empty in first_token_index!
...
ERROR 2021-08-12 13:20:04,750 [shard 0] init - Startup failed: std::runtime_error (sorted_tokens is empty in first_token_index!)
2021-08-12 13:20:04,773 INFO exited: scylla (exit status 1; not expected)
```

Log dump provided in artifacts.

**Environment:**
 - Platform: minikube
 - Kubernetes version: 1.21
 - Scylla version: 4.4.0
 - Scylla-operator version: master

**Additional context**
@zimnx confirmed he has the same issue and suggested this could be due to starting all Racks at once.

rzetelskik:
I've also managed to reproduce this behaviour on GKE.
Scylla logs: [gke.tar.gz](https://github.com/scylladb/scylla-operator/files/6976790/gke.tar.gz)

tnozicka:
we need e2e coverage

rzetelskik:
I'll open a separate issue for adding multi-rack e2e then

tnozicka:
>I'll open a separate issue for adding multi-rack e2e then

Wouldn't you add it in the same PR? If it's broken now, the fix should prove it actually fixes it by adding and e2e, and you can't merge the test before the fix because it should fail.

tnozicka:
@dkropachev is QA covering this case?

dkropachev:
> @dkropachev is QA covering this case?

We recently added test case to functional suite that does the following:
1. Deploy ScyllCluster with one rack
2. Addes two nodes, one by one to another rack

No exactly the same, but close.

rzetelskik:
fixed with #843

PR #843, opened 2021-11-19
Fix deploying multiple racks

### LINKED ISSUE #757: Add e2e for deploying a ScyllaCluster with multiple Racks
* Feature Request

Add an e2e for deploying a ScyllaCluster with multiple Racks and verifying it deployed successfully

Stability

**Additional Information:**
Originated from #755

tnozicka:
We have to wait for each Scylla rack to bootstrap before creating another one.

More contained change replacing https://github.com/scylladb/scylla-operator/pull/794

Resolves #757 #755

### DIFF
--- a/.github/actions/setup-minikube/action.yaml
+++ b/.github/actions/setup-minikube/action.yaml
@@ -3,6 +3,13 @@ description: 'Installs minikube Kubernetes cluster'
 runs:
   using: "composite"
   steps:
+  - name: Configure system
+    shell: bash
+    run: |
+      set -euExo pipefail
+
+      echo 'fs.aio-max-nr = 10485760' | sudo tee /etc/sysctl.d/90-scylla.conf >/dev/null
+      sudo sysctl --system
   - name: Install minikube
     shell: bash
     run: |
--- a/pkg/controller/scyllacluster/sync_statefulsets.go
+++ b/pkg/controller/scyllacluster/sync_statefulsets.go
@@ -355,7 +355,7 @@ func (scc *Controller) pruneStatefulSets(
 }
 
 // createMissingStatefulSets creates missing StatefulSets.
-// It return true if a statefulSet was created and an error.
+// It return true if done and an error.
 func (scc *Controller) createMissingStatefulSets(
 	ctx context.Context,
 	sc *scyllav1.ScyllaCluster,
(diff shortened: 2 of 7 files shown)


.github/actions/setup-minikube/action.yaml
```diff
@@ -3,6 +3,13 @@ description: 'Installs minikube Kubernetes cluster'
 runs:
   using: "composite"
   steps:
+  - name: Configure system
+    shell: bash
+    run: |
+      set -euExo pipefail
+
+      echo 'fs.aio-max-nr = 10485760' | sudo tee /etc/sysctl.d/90-scylla.conf >/dev/null
```
zimnx, line 11: Was this number printed in scylla logs? According to https://github.com/scylladb/scylla/pull/8228/files this should be 87588. I don't mind higher number, just wondering where it's coming from.

tnozicka: This number accounts for **multiple** scylla instances running on the same node. It's just an arbitrarily high number, hopefully high enough. (Practically it's just a 10x what I had on my dev machine.)

pkg/controller/scyllacluster/sync_statefulsets.go
```diff
@@ -365,36 +365,56 @@ func (scc *Controller) createMissingStatefulSets(
...
-			updatedSts, changed, err := resourceapply.ApplyStatefulSet(ctx, scc.kubeClient.AppsV1(), scc.statefulSetLister, scc.eventRecorder, sts, false)
+			klog.V(2).InfoS("Creating missing StatefulSet", "StatefulSet", klog.KObj(req))
+			var changed bool
+			var err error
+			sts, changed, err = resourceapply.ApplyStatefulSet(ctx, scc.kubeClient.AppsV1(), scc.statefulSetLister, scc.eventRecorder, req, false)
 			if err != nil {
 				errs = append(errs, fmt.Errorf("can't create missing statefulset: %w", err))
 				continue
 			}
 			if changed {
```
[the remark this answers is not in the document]
tnozicka: it can exist, we just haven't seen it yet in the previous logic but caches are arbitrarily delayed and live updated

zimnx (approved):
LGTM, thanks

pkg/controller/scyllacluster/sync_statefulsets.go
```diff
@@ -355,7 +355,7 @@ func (scc *Controller) pruneStatefulSets(
 }
 
 // createMissingStatefulSets creates missing StatefulSets.
-// It return true if a statefulSet was created and an error.
+// It return true if done and an error.
```
rzetelskik, line 358: "It return true if done and an error." is really vague and doesn't really help understand the purpose of this function any better than by just looking at its signature.

tnozicka: yeah, I guess I could have elaborated on that more then expecting the readers to be familiar with the pattern.

Issue #844, opened 2021-11-22
Scylla Pod get restarted when node is overloaded


zimnx:
**Describe the bug**
Scylla may get restarted when it's overloaded due to timeouts on liveness and readiness probes

![Selection_004](https://user-images.githubusercontent.com/3951334/142844015-17cd9650-cc79-4461-a10b-4a931db5c907.png)

**To Reproduce**
1. Deploy Scylla cluster
2. Overload it to 100%
3. Scylla is getting restarted due to timeouts on probes

**Expected behavior**
Scylla Pods survives overload

Liveness probe doesn't need to determine address to connect to, it's already available in Prober. 
Readiness probe is getting unnecessary information, like host ID or datacenter. 
Figure out if it's possible to determine UN state using single API call instead 4.

zimnx:
AFAIR this was related to routing probes via Scylla Manager Agent container, which had low resources. We got rid of this routing long time ago, so issue shouldn't persist anymore.

rzetelskik:
@zimnx [this TODO](https://github.com/scylladb/scylla-operator/blob/d4da9dfe44e55d7dd674a3d0394eafb5f309af2f/pkg/controller/scylladbdatacenter/resource.go#L784) references this issue - should we revert it?

zimnx:
lowering them to previous values should be fine

PR #845, opened 2021-11-22
Increase Scylla probes timeout to 30s

zimnx:
Related to #844

### DIFF
--- a/pkg/controller/scyllacluster/resource/resource.go
+++ b/pkg/controller/scyllacluster/resource/resource.go
@@ -331,7 +331,10 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 							StartupProbe: &corev1.Probe{
 								// Initial delay should be big, because scylla runs benchmarks
 								// to tune the IO settings.
-								TimeoutSeconds:   int32(5),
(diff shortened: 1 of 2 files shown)


tnozicka (commented):
let's raise startup probe as well

zimnx:
> let's raise startup probe as well

done

pkg/controller/scyllacluster/resource/resource.go
```diff
@@ -342,7 +344,9 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 								},
 							},
 							LivenessProbe: &corev1.Probe{
-								TimeoutSeconds:   int32(5),
+								// TODO: temporarily increase timeout from 5s to 30s to survive node overload
```
tnozicka, line 347: nit: the TODO should say what needs to be done in the future. Like `TODO: Lower the timeout when we fix probe timeouts. We have temporarily increased the timeout not to kill nodes when under load.` 

The issue that is reverenced should have a clear action item to remove this `TODO` and lower the timeout.

tnozicka (approved):
/lgmt

tnozicka (approved):

PR #861, opened 2021-11-26
Bump Go to 1.17.2 in release script

zimnx:
Minimum Go version is 1.17.1, where release scripts use 1.16.x. Syncing it with what we use for PR builds.


tnozicka (approved):
/lgtm

PR #862, opened 2021-11-26
Parallel e2es

### LINKED ISSUE #675: Run e2e suite in parallel
Currently, we run e2e tests serially and the run takes a long time, about 45 mins at this point. We need to split the suite into "serial" (have side effects) and "parallel" tests.

Currently, the minikube cluster won't be able to handle many tests in parallel because of CPU and memory requests but we could see at least a 2x speedup. We can also advance and spin a GKE cluster (on demand) with more CPUs that could handle a lot more.

tnozicka:
This PR migrates tests to use ginkgo/v2 and to run the suite in parallel.

Resolves #675


pkg/cmd/tests/tests_run.go
```diff
@@ -164,38 +258,198 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+		if !passed {
+			return fmt.Errorf("test suite %q failed", suite)
+		}
+
+		return nil
+	}
+
+	server, err := parallel_support.NewServer(suiteConfig.ParallelTotal, reporters.NewDefaultReporter(reporterConfig, formatter.ColorableStdOut))
+	if err != nil {
+		return fmt.Errorf("can't create parallel spec server")
```
zimnx, line 334: missing %w

pkg/cmd/tests/tests_run.go
```diff
@@ -164,38 +258,198 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+	for entryIndex := range cmdEntries {
+		e := cmdEntries[entryIndex]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			klog.V(2).InfoS("Waiting for process", "ID", e.id, "Process", e.cmd.String())
+			err := e.cmd.Wait()
+			if err != nil {
+				errs = append(errs, fmt.Errorf("can't wait for command %q with args %q: %w", e.cmd.Path, e.cmd.Args, err))
```
zimnx, line 400: we have a race here, `errs` is shared and modified by all goroutines

pkg/cmd/tests/tests_run.go
```diff
@@ -164,38 +258,198 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+		break
+
+	default:
+		for _, e := range cmdEntries {
+			exitCode := e.cmd.ProcessState.ExitCode()
+			switch exitCode {
+			case 0, types.GINKGO_FOCUS_EXIT_CODE:
+				break
+			default:
+				klog.ErrorS(nil, "Process failed", "ID", e.id, "Process", e.cmd.String(), "Logs", e.out)
```
zimnx, line 425: e.out.String(), otherwhise you will get an array of bytes

pkg/cmd/tests/tests_run.go
```diff
@@ -164,38 +258,198 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+
+		args := make([]string, 0, len(commonArgs)+1)
+		args = append(args, commonArgs...)
+		args = append(args, fmt.Sprintf("--%s=%v", parallelShardFlagKey, i))
+
+		buf := &bytes.Buffer{}
+		cmd := exec.CommandContext(ctx, os.Args[0], args...)
+		cmd.Stdout = buf
+		cmd.Stderr = buf
+		cmdEntries = append(cmdEntries, &cmdEntry{
```
zimnx, line 364: what is this printing out? I thought it's test progress, but it isn't printed anywhere on happy path.

pkg/cmd/tests/tests_run.go
tnozicka: it doesn't output bytes
```
Logs="Error: unknown flag: --foo\n"
```
but it won't hurt :)

pkg/cmd/tests/tests_run.go
tnozicka: it's printed at the end only when the subprocess fails

zimnx: I meant stdout, but looks like test progress is printed by the server

zimnx (approved):
LGTM, good stuff - thanks

PR #867, opened 2021-11-30
Increase TerminationGracePeriodSeconds of Scylla pods

### LINKED ISSUE #805: A busy scyllacluster doesn't shutdown cleanly in time and is force terminated with SIGKILL
**Describe the bug**
If you update the scylla operator sidecar image string scylla operator will trigger a rolling update of the statefulset(s) that compose the scylla cluster.  This will trigger a restart of the pods in the statefulset(s).  This doesn't trigger any "upgrading status" as the scylla version has not changed.  Thus a simple SIGTERM is sent to the scylla process followed 30 seconds latter by a SIGKILL (default termination grace period).  On a busy scylla node this will likely result in a unclean shutdown. 

**To Reproduce**
1. Have a busy scylla cluster running under scylla operator
2. Update the scylla-operator image which is used as an init container in scylla
3. Watch scylla take more then 30s to shutdown after SIGTERM (termination grace period) and get SIGKILLED

**Expected behavior**
The scylla replica is drained before it is sent a SIGTERM.  This could be achieved via a prestop hook and/or potentially a longer terminationgraceperiod for the Scylla pod.

**Environment:**
 - Scylla-operator version: v1.4 (should affect master and v1.5 as well)

rzetelskik:
This PR increases TerminationGracePeriodSeconds of Scylla pods from default 30s to 900s to sync with scylla-server supervisord's timeout value introduced in https://github.com/scylladb/scylla/pull/9545. 

Resolves #805

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -396,7 +396,8 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 						PodAffinity:     placement.PodAffinity,
 						PodAntiAffinity: placement.PodAntiAffinity,
 					},
-					ImagePullSecrets: c.Spec.ImagePullSecrets,
+					ImagePullSecrets:              c.Spec.ImagePullSecrets,
+					TerminationGracePeriodSeconds: pointer.Int64Ptr(900),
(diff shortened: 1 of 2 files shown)


tnozicka (approved):
/lgtm
thanks

PR #873, opened 2021-12-03
Don't act on scylla pods that are being deleted in nodeconfigpod controller

tnozicka:
No controller is allowed to act on objects that are being deleted. We have missed that in `nodeconfigpod` controller.

@rzetelskik found out that in his cluster scylla pods couldn't be foreground deleted because the controller was always recreating the tuning CM that depends on the Pod.

@rzetelskik please pick the commit and test it in your PR.


zimnx (approved):
LGTM, thanks

PR #714, opened 2021-07-22
Add e2e test for rebootstrapping from existing PVCs

rzetelskik:
**Description of your changes:**
- Add an e2e test for rebootstrapping ScyllaCluster from existing PVCs

**Which issue is resolved by this Pull Request:**
Resolves #535

PR #882, opened 2021-12-08
Fix ownerRefs for nodeconfigpod controller

### LINKED ISSUE #881: NodeConfig ConfigMap is not orphaned/deleted on ScyllaCluster deletion
**Describe the bug**
The ConfigMaps created by NodeConfigPodController are not being deleted nor orphaned on ScyllaClusters' deletion. This is especially problematic when deleting the ScyllaCluster with propagationPolicy: foreground (`--cascade='foreground'`) as it will cause the SC deletion to hang indefinitely (see https://github.com/scylladb/scylla-operator/actions/runs/1548694108). 

**To Reproduce**
1. Deploy the operator (current master)
2. Wait for it to roll out.
3. Deploy an exemplary ScyllaCluster
4. Wait for the SC to roll out
5. Delete the SC

**Expected behavior**
The NodeConfig ConfigMap should be orphaned/deleted.
ScyllaCluster deletion should finalize with propagationPolicy: Foreground.

**Logs**
See https://github.com/scylladb/scylla-operator/actions/runs/1548694108.

**Environment:**
 - Platform: any
 - Kubernetes version: 1.22.2 
 - Scylla version: any
 - Scylla-operator version: master

tnozicka:
We have accidentally used Resource name instead of Kind in the nodeconfigpod controller. That creates invalid ownerRefs and prevents GC from executing a cascade delete because such Kind doesn't exist. This PR fixes that and adds the logic that allows us to apply a new ownerRef as long as the UID matches. (We got lucky this time.)

Resolves #881

### DIFF
--- a/pkg/resourceapply/apps_test.go
+++ b/pkg/resourceapply/apps_test.go
@@ -274,8 +274,8 @@ func TestApplyStatefulSet(t *testing.T) {
 			}(),
 			expectedSts:     nil,
 			expectedChanged: false,
-			expectedErr:     fmt.Errorf(`statefulset "default/test" isn't controlled by us`),
-			expectedEvents:  []string{`Warning UpdateStatefulSetFailed Failed to update StatefulSet default/test: statefulset "default/test" isn't controlled by us`},
+			expectedErr:     fmt.Errorf(`statefulset "default/test" isn't controlled by anyone, won't adopt it in apply`),
+			expectedEvents:  []string{`Warning UpdateStatefulSetFailed Failed to update StatefulSet default/test: statefulset "default/test" isn't controlled by anyone, won't adopt it in apply`},
 		},
 		{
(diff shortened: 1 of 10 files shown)


pkg/resourceapply/apps_test.go
```diff
@@ -274,8 +274,8 @@ func TestApplyStatefulSet(t *testing.T) {
 			}(),
```
zimnx, line 274: Please add test verifying we can override ownerRef containing different Kind in case when UID matches

zimnx:
LGTM, is this still a draft?

rzetelskik (approved):
I ran it against #714 , lgtm

zimnx (dismissed):

tnozicka:
(clean rebase)

zimnx (approved):

PR #884, opened 2021-12-10
Use direct localhost connection to Scylla in sidecar

### LINKED ISSUE #869: Probes should talk directly to scylla container
* Feature Request

Scylla probes should avoid using the manager-agent and talk directly to scylla (in the same container).

Stability. There have been multiple issues over the time:
- Timeouts from manager-agent when scylla is under load.
- Because it depends on the manager-agent container scylla will get killed even if it's working fine but the manager container is down.

zimnx:
Previously readiness and liveness probes were routed thorugh
scylla-manage-agent which has much lower resources than Scylla
and it could throtthle probes causing timeouts.
Eliminating routing through agent decouples Scylla lifetime from agent.

Fixes #869

### DIFF
--- a/pkg/controller/sidecar/sync.go
+++ b/pkg/controller/sidecar/sync.go
@@ -16,24 +16,18 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const requeueWaitDuration = 5 * time.Second
-
-func (c *Controller) getScyllaClient() (*scyllaclient.Client, error) {
-	secret, err := c.secretLister.Secrets(c.namespace).Get(c.secretName)
-	if err != nil {
-		return nil, fmt.Errorf("can't get manager agent auth secret %s/%s: %w", c.namespace, c.secretName, err)
-	}
-
-	return controllerhelpers.NewScyllaClientFromSecret(secret, []string{c.hostAddr})
-}
+const (
+	requeueWaitDuration = 5 * time.Second
+	localhost           = "localhost"
+)
 
 func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service) error {
-	scyllaClient, err := c.getScyllaClient()
+	scyllaClient, err := controllerhelpers.NewScyllaClientForLocalhost()
 	if err != nil {
 		return err
 	}
 
-	opMode, err := scyllaClient.OperationMode(ctx, c.hostAddr)
+	opMode, err := scyllaClient.OperationMode(ctx, localhost)
 	if err != nil {
 		return fmt.Errorf("can't get node operation mode: %w", err)
 	}
@@ -60,7 +54,7 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service) 
 		// Node can be in NORMAL mode while still starting up.
 		// Last thing that scylla is doing as part of startup process is brining native transport up
 		// so we check if native port is up as sign that it is not loading.
-		nativeUp, err := scyllaClient.IsNativeTransportEnabled(ctx, c.hostAddr)
+		nativeUp, err := scyllaClient.IsNativeTransportEnabled(ctx, localhost)
 		if err != nil {
 			return fmt.Errorf("can't get native transport status: %w", err)
 		}
@@ -72,11 +66,11 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service) 
 		}
 
 		// Decommission the node only if it is in normal mode and native transport is up.
-		decommissionErr := scyllaClient.Decommission(ctx, c.hostAddr)
+		decommissionErr := scyllaClient.Decommission(ctx, localhost)
 		if decommissionErr != nil {
 			// Decommission is long running task, so request fails due to the timeout in most cases.
 			// To not raise an error, when it is in progress, we check opMode.
-			opMode, err := scyllaClient.OperationMode(ctx, c.hostAddr)
+			opMode, err := scyllaClient.OperationMode(ctx, localhost)
 			if err == nil && (opMode.IsDecommissioned() || opMode.IsLeaving() || opMode.IsDecommissioning()) {
 				klog.V(2).InfoS("Decommissioning is in progress. Waiting a bit.", "Mode", opMode)
 				c.queue.AddAfter(c.key, requeueWaitDuration)
--- a/pkg/controllerhelpers/scylla.go
+++ b/pkg/controllerhelpers/scylla.go
@@ -7,7 +7,6 @@ import (
 	"github.com/scylladb/go-log"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
-	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/scyllaclient"
 	"go.uber.org/zap"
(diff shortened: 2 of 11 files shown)


tnozicka:
51 commits, looks like a wrong base

tnozicka:
looks like GH broke, changing the base to a random branch and back helped

pkg/controller/sidecar/sync.go
```diff
@@ -33,7 +31,7 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service)
 		return err
 	}
 
-	opMode, err := scyllaClient.OperationMode(ctx, c.hostAddr)
+	opMode, err := scyllaClient.OperationMode(ctx, localhost)
```
tnozicka, line 34: would it be better to use the "localhost" DNS/alias that's IP protocol independent?

pkg/controller/sidecar/sync.go
```diff
@@ -16,15 +16,13 @@ import (
...
+)
 
 func (c *Controller) getScyllaClient() (*scyllaclient.Client, error) {
-	secret, err := c.secretLister.Secrets(c.namespace).Get(c.secretName)
-	if err != nil {
-		return nil, fmt.Errorf("can't get manager agent auth secret %s/%s: %w", c.namespace, c.secretName, err)
-	}
-
-	return controllerhelpers.NewScyllaClientFromSecret(secret, []string{c.hostAddr})
+	return controllerhelpers.NewLocalhostScyllaClient()
```
tnozicka, line 25: would it be better to inline the function? it no longer needs any value from the object

pkg/controllerhelpers/scylla.go
```diff
@@ -69,28 +68,33 @@ func GetRequiredScyllaHosts(sc *scyllav1.ScyllaCluster, services map[string]*cor
 	return hosts, nil
 }
 
-func NewScyllaClientFromToken(hosts []string, authToken string) (*scyllaclient.Client, error) {
+func NewScyllaClient(cfg scyllaclient.Config) (*scyllaclient.Client, error) {
```
tnozicka, line 71: pass it by ref?

pkg/controllerhelpers/scylla.go
```diff
@@ -69,28 +68,33 @@ func GetRequiredScyllaHosts(sc *scyllav1.ScyllaCluster, services map[string]*cor
...
-	}
+func NewScyllaClientFromToken(hosts []string, authToken string) (*scyllaclient.Client, error) {
+	cfg := scyllaclient.DefaultConfig(authToken, hosts...)
+	return NewScyllaClient(cfg)
+}
+
+func NewLocalhostScyllaClient() (*scyllaclient.Client, error) {
+	cfg := scyllaclient.DefaultConfig("", "127.0.0.1")
+	cfg.Scheme = "http"
+	cfg.Port = fmt.Sprintf("%d", naming.ScyllaAPIPort)
```
tnozicka, line 91: can the port be change in the supplied config?

pkg/controllerhelpers/scylla.go
```diff
@@ -69,28 +68,33 @@ func GetRequiredScyllaHosts(sc *scyllav1.ScyllaCluster, services map[string]*cor
...
+
+	transport := scyllaclient.DefaultTransport()
+	transport.TLSClientConfig = nil
+	cfg.Transport = transport
```
tnozicka, line 95: can you write directly to cfg.Transport?

pkg/controllerhelpers/scylla.go
```diff
@@ -69,28 +68,33 @@ func GetRequiredScyllaHosts(sc *scyllav1.ScyllaCluster, services map[string]*cor
...
-	token, err := helpers.GetAgentAuthTokenFromSecret(secret)
-	if err != nil {
-		return nil, err
-	}
+func NewScyllaClientFromToken(hosts []string, authToken string) (*scyllaclient.Client, error) {
+	cfg := scyllaclient.DefaultConfig(authToken, hosts...)
+	return NewScyllaClient(cfg)
+}
+
+func NewLocalhostScyllaClient() (*scyllaclient.Client, error) {
```
tnozicka, line 88: To match the pattern and search prefix, could this be `NewScyllaClientForLocalhost`?

pkg/controllerhelpers/scylla.go
zimnx: done

pkg/controllerhelpers/scylla.go
zimnx: done

pkg/controllerhelpers/scylla.go
zimnx: done

pkg/controller/sidecar/sync.go
zimnx: done

pkg/controller/sidecar/sync.go
zimnx: done

pkg/controllerhelpers/scylla.go
zimnx: Could you elaborate? I don't get the question

pkg/scyllaclient/client.go
```diff
@@ -301,8 +301,16 @@ func (c *Client) HasSchemaAgreement(ctx context.Context) (bool, error) {
 	return len(versions) == 1, nil
 }
 
-func DefaultTransport() *http.Transport {
-	return &http.Transport{
+type TransportOption func(transport *http.Transport)
+
+func WithTLSConfig(tlsConfig *tls.Config) func(*http.Transport) {
```
tnozicka, line 306: If the field is public and can be set on the returned config, what's the benefit of option wrappers? That's usually done for private fields.

zimnx: Short extensible code. Options pattern not always modify non-exported fields, they are also used for chaning default values.
Your previous [comment](https://github.com/scylladb/scylla-operator/pull/884#discussion_r770325025) suggested writing directly to cfg.Transport which is an interface of http.RoundTripper and cannot be modified, so I added an option to achive the same with shorter block and without additional object allocation.

tnozicka: >writing directly to cfg.Transport which is an interface of http.RoundTripper and cannot be modified 

I don't feel like that argument relates to this interface as it's behind the boundary. The function returns `*http.Transport` which has the public field that needs to be set.

this looks much simpler 
```
t := DefaultTransport()
t.TLSClientConfig = myTLSConfig
```
then the lambda syntax
```
func WithTLSConfig(tlsConfig *tls.Config) func(*http.Transport) {
	return func(transport *http.Transport) {
		transport.TLSClientConfig = tlsConfig
	}
}

t := DefaultTransport(WithTLSConfig(myTLSConfig))
```
which also isn't consistent with setting other option which would lack the lambda wrapper or needed manufacturing one per each.

zimnx: > this looks much simpler
> 
> ```
> t := DefaultTransport()
> t.TLSClientConfig = myTLSConfig
> ```

suggested change was implemented in first version and received a remark to make it shorter. I'll revert to initial version then.

> which also isn't consistent with setting other option which would lack the lambda wrapper or needed manufacturing one per each.

It's a tradeoff of keeping dead code in internal library or having consistency and option to modify all possible parameters.

pkg/scyllaclient/client.go
```diff
@@ -302,7 +302,7 @@ func (c *Client) HasSchemaAgreement(ctx context.Context) (bool, error) {
 }
 
 func DefaultTransport() *http.Transport {
-	return &http.Transport{
+	t := &http.Transport{
```
tnozicka, line 305: nit: you can just do return here now

pkg/scyllaclient/client.go
tnozicka: > suggested change was implemented in first version and received a remark to make it shorter.

apologies, if we have misunderstood each other there

pkg/scyllaclient/client.go
zimnx: done

tnozicka (approved):
/lgtm
thanks!

PR #886, opened 2021-12-13
Update docs theme to 1.0.6 and add CI

tnozicka:
We need to bump vulnerable dependencies. Also the same sources must produce the same results and not randomly pick a new version that can break it.

This also bumps our docs theme to 1.0.6.


tnozicka:
hm, we need CI for this

tnozicka:
I picked up logical changes from https://github.com/scylladb/scylla-operator/pull/815

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,133 @@
+name: Docs
+
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - 'docs/**'
+    - '.github/workflows/docs.yaml'
```
zimnx, line 9: We also need to trigger it on pushes to release branches

docs/source/conf.py
```diff
@@ -169,43 +91,42 @@ def setup(sphinx):
...
-]
+# Theme options are theme-specific and customize the look and feel of a theme
+# further.  For a list of options available for each theme, see the
+# documentation.
+#
+html_theme_options = {
+    'conf_py_path': 'docs/source/',
+    'default_branch': 'master',
+    'hide_edit_this_page_button': 'false',
+    'hide_version_dropdown': ['master'],
```
zimnx, line 109: master branch contains latest release dates, so we don't want to hide it

.github/workflows/docs.yaml
tnozicka: yeah, you are right, although the design of checking random tip of the other branches is not great when all of them publish (possible races), it is what we have.

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,134 @@
...
+      with:
+        name: docs.tar.xz
+        path: ~/docs.tar.xz
+        if-no-files-found: error
+        retention-days: ${{ env.retention_days }}
+
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push' }}
```
zimnx, line 84: I won't publish backports. We either need to deploy also on pushes to release branches, or enable manual button with deploy step

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,134 @@
...
+        
+        pushd ./docs/
+        ~/.poetry/bin/poetry install
+
+    - name: Build docs
+      run: |
+        set -euExo pipefail
+        shopt -s inherit_errexit
+        
+        make -C ./docs/ multiversion SPHINXOPTS="--keep-going"
```
zimnx, line 64: multiversion builds from multiple branches specified in config. Changes in different branch shouldn't affect version we are testing here. And it doesn't build actual PR changes for PR builds.

docs/Makefile
```diff
@@ -9,70 +9,62 @@ SOURCEDIR     := source
...
-multiversion: setup
+multiversion:
 	@mkdir -p $(HOME)/.cache/pypoetry/virtualenvs
 	$(POETRY) run sphinx-multiversion $(SOURCEDIR) $(BUILDDIR)/dirhtml
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/dirhtml."
+.PHONY: multiversion
 
-.PHONY: multiversionpreview
-multiversionpreview: multiversion
```
zimnx, line 77: why is this dependency removed? It doesn't make sense to start server without building sources

.github/workflows/docs.yaml
tnozicka: yeah, I need to redo it. This checking out random tips of other branches it a nightmare for the CI.

docs/Makefile
tnozicka: you could have build the sources already. if you make it a hard dependency it forces everyone to rebuild before every run, even it nothing changed.

zimnx (approved):
LGTM, thanks

PR #887, opened 2021-12-13
Extend node optimizations with kernel parameter tuning

### LINKED ISSUE #868: Support configuring sysctls on nodes
* Feature Request

Users should be able to configure sysctls with NodeConfig. One of the use cases is setting `/proc/sys/fs/aio-max-nr`.

zimnx:
Additional kernel parameter tuning is added to our tuning suite.
It allows user to provide their own desired custom values, ability to
apply recommended settings directly from Scylla image, as well as
scaling some of the values based on number of expected Scyllas running
on the same host.

NodeConfig API was extended with KernelParameters field allowing
configuring/disabling above logic.

Fixes #868


tnozicka (commented):
I think this needs to be conditional at best. There can already be a tuning component or manual node setting that the user might not want to be overridden.

While I see that using whatever sysctls scylla sets for a dedicated VM could seems like something we should reuse, there are issues like:
- what if the user wants to bump them
- the node needs specific adjustments
- some values might need to be higher if there is more then 1 scylla process on the node
- possibly cloud specific values
- extra values

PR #888, opened 2021-12-13
Cache iotune results in persistent location

### LINKED ISSUE #146: Cache iotune results

zimnx:
Symlink in io_properties file destination pointing to persistent
location is created on startup if file doesn't exists. This allows to
cache iotune results throughout entire node lifetime, making it a single
time operation.

Fixes #146

### DIFF
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"regexp"
 	"strconv"
 	"strings"
@@ -58,20 +59,25 @@ func NewScyllaConfig(m *identity.Member, kubeClient kubernetes.Interface, scylla
 }
 
 func (s *ScyllaConfig) Setup(ctx context.Context) (*exec.Cmd, error) {
-	var err error
-	var cmd *exec.Cmd
+	klog.Info("Setting up iotune cache")
+	if err := setupIOTuneCache(); err != nil {
+		return nil, fmt.Errorf("can't setup iotune cache: %w", err)
+	}
 
 	klog.Info("Setting up scylla.yaml")
-	if err = s.setupScyllaYAML(scyllaYAMLPath, scyllaYAMLConfigMapPath); err != nil {
-		return nil, errors.Wrap(err, "failed to setup scylla.yaml")
+	if err := s.setupScyllaYAML(scyllaYAMLPath, scyllaYAMLConfigMapPath); err != nil {
+		return nil, fmt.Errorf("can't setup scylla.yaml: %w", err)
 	}
+
 	klog.Info("Setting up cassandra-rackdc.properties")
-	if err = s.setupRackDCProperties(); err != nil {
-		return nil, errors.Wrap(err, "failed to setup rackdc properties file")
+	if err := s.setupRackDCProperties(); err != nil {
+		return nil, fmt.Errorf("can't setup rackdc properties file: %w", err)
 	}
+
 	klog.Info("Setting up entrypoint script")
-	if cmd, err = s.setupEntrypoint(ctx); err != nil {
-		return nil, errors.Wrap(err, "failed to setup entrypoint")
+	cmd, err := s.setupEntrypoint(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("can't setup entrypoint: %w", err)
 	}
 
 	return cmd, nil
@@ -293,6 +299,25 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
 		klog.InfoS("suboptimal shard and cpuset config, shard count (config: 'CPU') and cpuset size should match for optimal performance",
 			"shards", shards, "cpuset", cpuSet.String())
 	}
+	return nil
+}
+
+func setupIOTuneCache() error {
+	if _, err := os.Stat(scyllaIOPropertiesPath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("can't stat file %q: %w", scyllaIOPropertiesPath, err)
+		}
+
+		cachePath := path.Join(naming.DataDir, naming.ScyllaIOPropertiesName)
+		if err := os.Symlink(cachePath, scyllaIOPropertiesPath); err != nil {
+			return fmt.Errorf("can't create symlink from %q to %q: %w", scyllaIOPropertiesPath, cachePath, err)
+		}
+
+		klog.V(2).Info("Initialized IOTune benchmark cache", "path", scyllaIOPropertiesPath, "cachePath", cachePath)
+	} else {
+		klog.V(2).Info("Found cached IOTune benchmark results", "path", scyllaIOPropertiesPath)
+	}
+
 	return nil
 }
 


tnozicka (commented):
could we create a volume mount with subpath and mount data subdir directly into that etcd path?

tnozicka:
> Unfortunately we can't mount single file if it doesn't exists in PVC, and in our case, on the first run there isn't any io_properties.yaml in data pvc.
technically, we might be able to touch it in an init container first

but generally, could we just control the filelocation from the scylla config? I suppose if we can change it to a volume, we can have a real (independent) sidecar managing it and projecting it back into a CM some day. Given our last chat on perftuning and how that can change (enhance) performance in the future releases, removing the cached tuning is gonna be a bit painful with `kubectl exec` instead of just deleting a fixed CM with label selector.

zimnx:
We can control from where Scylla is reading, but not where scylla_io_setup is writing. It's hardcoded to `/etc/scylla.d/io_properties.yaml`. Having a symlink give us control over where result is stored. 
Later we can point it to different location and manage a configmap persisting it. 
For now users will be able to either provide their own io_properties via volume/volumeMount field CRD, or retrigger it once by manually removing using `kubectl exec`.

pkg/sidecar/config/config.go
```diff
@@ -296,6 +301,15 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
 	return nil
 }
 
+func (s *ScyllaConfig) setupIOTuneCache() error {
+	if _, err := os.Stat(scyllaIOPropertiesPath); errors.Is(err, os.ErrNotExist) {
+		if err := os.Symlink(path.Join(naming.DataDir, naming.ScyllaIOPropertiesName), scyllaIOPropertiesPath); err != nil {
```
tnozicka, line 306: nit: pls split the variable assignment, the lines are already long

pkg/sidecar/config/config.go
```diff
@@ -296,6 +301,15 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
...
+			return fmt.Errorf("create iotune cache symlink: %w", err)
```
tnozicka, line 307: nit: *can't* create symlink from %q to %q: %w

pkg/sidecar/config/config.go
```diff
@@ -296,6 +301,15 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
 	return nil
 }
 
+func (s *ScyllaConfig) setupIOTuneCache() error {
```
tnozicka, line 304: I don't see the object being used, maybe it should be just a function

tnozicka (commented):
ok, let's start with the symlink

pkg/sidecar/config/config.go
zimnx: done

pkg/sidecar/config/config.go
zimnx: done

pkg/sidecar/config/config.go
zimnx: done

pkg/sidecar/config/config.go
```diff
@@ -61,6 +62,10 @@ func (s *ScyllaConfig) Setup(ctx context.Context) (*exec.Cmd, error) {
 	var err error
 	var cmd *exec.Cmd
 
+	klog.Info("Setting up iotune cache")
+	if err = setupIOTuneCache(); err != nil {
```
tnozicka, line 66: nit: did you mean to `:=`?

pkg/sidecar/config/config.go
```diff
@@ -61,6 +62,10 @@ func (s *ScyllaConfig) Setup(ctx context.Context) (*exec.Cmd, error) {
...
+		return nil, errors.Wrap(err, "failed to setup iotune cache")
```
tnozicka, line 67:
```suggestion
		return nil, fmt.Errorf("can't setup iotune cache: %w", err)
```

pkg/sidecar/config/config.go
```diff
@@ -296,6 +301,16 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
 	return nil
 }
 
+func setupIOTuneCache() error {
+	if _, err := os.Stat(scyllaIOPropertiesPath); errors.Is(err, os.ErrNotExist) {
```
tnozicka, line 305: nit: would you mind splitting it on 2 lines for better readability? (also the other)

pkg/sidecar/config/config.go
zimnx: yes, done

pkg/sidecar/config/config.go
zimnx: done

pkg/sidecar/config/config.go
zimnx: splitted, left second if using init-statement becasue it looks natural

pkg/sidecar/config/config.go
```diff
@@ -296,6 +302,22 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
 	return nil
 }
 
+func setupIOTuneCache() error {
+	_, err := os.Stat(scyllaIOPropertiesPath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	if errors.Is(err, os.ErrNotExist) {
```
tnozicka, line 311: can you reshuffle the logic so you don't check the same error twice?

zimnx: done

pkg/sidecar/config/config.go
```diff
@@ -296,6 +302,21 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
 	return nil
 }
 
+func setupIOTuneCache() error {
+	if _, err := os.Stat(scyllaIOPropertiesPath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return err
```
tnozicka, line 308:
```suggestion
			return fmt.Errorf("can't stat file %q: %w, scyllaIOPropertiesPath, err)
```

pkg/sidecar/config/config.go
```diff
@@ -296,6 +302,21 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
...
+		}
+
+		cachePath := path.Join(naming.DataDir, naming.ScyllaIOPropertiesName)
+		if err := os.Symlink(cachePath, scyllaIOPropertiesPath); err != nil {
+			return fmt.Errorf("can't create symlink from %q to %q: %w", scyllaIOPropertiesPath, cachePath, err)
+		}
+	}
```
tnozicka, line 315: in the else branch we should log that the file already exists at V2

pkg/sidecar/config/config.go
```diff
@@ -296,6 +302,21 @@ func (s *ScyllaConfig) validateCpuSet(ctx context.Context, cpusAllowed string, s
...
+func setupIOTuneCache() error {
+	if _, err := os.Stat(scyllaIOPropertiesPath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+
+		cachePath := path.Join(naming.DataDir, naming.ScyllaIOPropertiesName)
+		if err := os.Symlink(cachePath, scyllaIOPropertiesPath); err != nil {
+			return fmt.Errorf("can't create symlink from %q to %q: %w", scyllaIOPropertiesPath, cachePath, err)
+		}
```
tnozicka, line 314: we should log that we have created the symlink at V2

tnozicka (commented):
lgtm but we should add logging to help debug it when mounting volumes over it

pkg/sidecar/config/config.go
zimnx: added

pkg/sidecar/config/config.go
zimnx: added

pkg/sidecar/config/config.go
zimnx: done

PR #895, opened 2021-12-16
Fix sysctls example in docs

tnozicka:
sysctls are specified in `scyllaclusters.spec.sysctls`, not `scyllaclusters.spec.datacenter.racks`.


rzetelskik (approved):
lgtm

PR #899, opened 2021-12-16
Increase loglevel of CI e2e's

zimnx:


hack/ci-deploy.sh
```diff
@@ -40,6 +41,7 @@ cp ./examples/common/cert-manager.yaml "${deploy_dir}/"
 
 for f in $( find "${deploy_dir}"/ -type f -name '*.yaml' ); do
     sed -i -E -e "s~docker.io/scylladb/scylla-operator(:|@sha256:)[^ ]*~${OPERATOR_IMAGE_REF}~" "${f}"
+    sed -i -E -e "s~--loglevel=[0-9]+~--loglevel=${LOG_LEVEL}~" "${f}"
```
tnozicka, line 44: I didn't notice this PR at first but I think we should do it here https://github.com/scylladb/scylla-operator/pull/900/files#diff-59b0366afbddbafb181605e979731c2c5b7e104711ff2158c2dff2198cc6b882R207
it's also where we'd set the ScyllaOperatorConfig loglevel option when it's there

PR #907, opened 2021-12-30
[WIP] Automatically provision member SA

### LINKED ISSUE #450: Automatically provision a fixed SA for the member sidecar and appropriate rolebinding
* Feature Request

When reconciling ScyllaCluster object the operator should check if the specified SA has enough permissions like listing services, patching the stateful set and if not, it should set a degraded condition in scyllacluster.status.

Users should be able to identify why isn't the ScyllaCluster deploying. Given the SA has to be setup manually we should do a better job telling it to the user and a better UX then manually logging through event or the managed objects.

Alternatively we can look into automatically provisioning that SA.

tnozicka:
This will automatically setup the member service account for each ScyllaCluster. Existing ServiceAccounts will be adopted and updated if they don't match. That can also mean a rename.

Resolves #450


tnozicka:
I'll do just one SA per namespace instead per ScyllaCluster

tnozicka:
Given we already use dedicated SAs per ScyllaCluster (in helm charts and CI) it's easier to go that way. Having a global SA could possibly mean name collision which couldn't be adjusted and the migration would be easy either.

test/e2e/set/scyllacluster/scyllacluster_sa.go
```diff
@@ -0,0 +1,106 @@
...
+	g.It("should claim preexisting member ServiceAccount and RoleBinding", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
+
+		sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+
+		framework.By("Creating a ServiceAccount and a RoleBinding")
+		sa, err := f.KubeClient().CoreV1().ServiceAccounts(f.Namespace()).Create(ctx, &corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("%s-member", sc.Name),
```
zimnx, line 34: use helper function to get name

test/e2e/set/scyllacluster/scyllacluster_sa.go
```diff
@@ -0,0 +1,106 @@
...
+					APIGroup:  corev1.GroupName,
+					Kind:      rbacv1.ServiceAccountKind,
+					Namespace: sa.Namespace,
+					Name:      sa.Name,
+				},
+			},
+			RoleRef: rbacv1.RoleRef{
+				APIGroup: rbacv1.GroupName,
+				Kind:     "ClusterRole",
+				Name:     "scyllacluster-member",
```
zimnx, line 55: This could be a const in naming package

zimnx (commented):
few nits, please also remove SA and RB from examples and any mentions in documentation

pkg/controller/scyllaclustersa/controller.go
```diff
@@ -0,0 +1,337 @@
+package scyllaclustersa
```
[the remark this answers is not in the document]
tnozicka: I am open to a better name, but rbac doesn't cover it as SA is not rbac. So I justify it as "this is primarily an SA controller that sets up collateral objects for the SA to work" - shortened as SA :)

pkg/naming/names.go
```diff
@@ -167,3 +167,7 @@ func PerftuneResultName(uid string) string {
 func GetTuningConfigMapNameForPod(pod *corev1.Pod) string {
 	return fmt.Sprintf("nodeconfig-podinfo-%s", pod.UID)
 }
+
+func MemberServiceAccountNameForScyllaCluster(scName string) string {
```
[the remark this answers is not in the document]
tnozicka: unified now

pkg/controller/scyllaclustersa/resource.go
```diff
@@ -0,0 +1,49 @@
...
+				APIGroup:  corev1.GroupName,
+				Kind:      rbacv1.ServiceAccountKind,
+				Namespace: sc.Namespace,
+				Name:      saName,
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     "scyllacluster-member",
```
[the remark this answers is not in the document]
tnozicka: done

test/e2e/set/scyllacluster/scyllacluster_sa.go
tnozicka: https://github.com/scylladb/scylla-operator/pull/907#discussion_r784016893

pkg/controller/scyllaclustersa/controller.go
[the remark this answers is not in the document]
tnozicka: I wanted to write that it's similar to ScyllaCluster controller but now I am thinking whether that should be just another sync loop there...

pkg/controller/scyllacluster/sync.go
```diff
@@ -263,6 +362,18 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
...
+		errs = append(errs, fmt.Errorf("can't sync serviceaccounts: %w", err))
+		// TODO: Set degraded condition
+	}
+
+	err = scc.syncRoleBindings(ctx, sc, roleBindings)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("can't sync rolebindings: %w", err))
+		// TODO: Set degraded condition
+	}
+
```
zimnx, line 376: Shouldn't these be above statefulset? It's a dependency of it

tnozicka: I don't think these are a hard dependency but I can move them

zimnx (approved):
LGTM, thanks

tnozicka:
picked up https://github.com/scylladb/scylla-operator/pull/917 to see if it make the CI here pass

zimnx (approved):

zimnx (approved):

PR #915, opened 2022-01-17
Collect e2e container logs from all namespaces

tnozicka:
When a scylla pod is deleted (to be replaced), we lose all logs. We can do our best at least get them from CRI, if not purged already.


tnozicka:
we can't list it by namespace as the e2e namesapces get deleted, so we'd need to remap and list all containers with `crictl ps -a`

PR #917, opened 2022-01-19
Generate random names for e2e clusters to prevent cross namespace joins

tnozicka:
Random cluster names will reduce the likelihood of accidentally joining 2 clusters when an IP gets reused.

e2e flakes


zimnx (approved):
LGTM, thanks

tnozicka:
we'd have to also fix the manually provisioned SA names that take their name from the fixture, so I'll squash it into #907

PR #922, opened 2022-01-19
Create NodeConfig DaemonSet even when optimizations are disabled

zimnx:
DaemonSet wasn't created when optimizations were disabled, it forbid cleaning up
the Job resources.
Dangling resources lead to continous reruning of optimizations on every Node
restart.


zimnx:
Jobs have DS controllerRef, so there's no issue.

PR #926, opened 2022-01-24
Tune nodes immediately when tuning DaemonSet is scheduled 

zimnx:
Previously controller was waiting for Scylla Pod to be scheduled on the same Node.
This wasn't in sync with e2e logic verifying it, but unfortunately due to a bug, it wasn't even checked.

E2E test verifying it was relying on suite-scoped variable containing a list of Nodes which are subject for optimizations. But because this variable was shadowed in place where it was computed, test was using an empty default slice. 
Assert compared length of this slice with number of scheduled jobs, which "thanks" to shadowing was passing even though no jobs were created.

### DIFF
--- a/pkg/controller/nodeconfigdaemon/controller.go
+++ b/pkg/controller/nodeconfigdaemon/controller.go
@@ -156,6 +156,9 @@ func NewController(
 		DeleteFunc: snc.deleteJob,
 	})
 
+	// Start right away, Scylla might not be scheduled yet, but Node can already be tuned.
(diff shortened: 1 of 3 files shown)


tnozicka (commented):
good catch, pls squash the commits a bit

zimnx:
> good catch, pls squash the commits a bit

done, squashed one commit into another

tnozicka (approved):
/lgtm
thanks

PR #933, opened 2022-01-31
Keep custom annotations on ServiceAccounts

### LINKED ISSUE #932: no longer able to set these annotations
Hello,
A customer has updated their Scylla operator to 1.7.0. However, the new behavior of the operator, in regards to it now handling the cluster-member SAs for the other components, seems to have broken their implementation.
Previously, their Scylla cluster's SAs (including the manager's own) contained the following annotations (configured by an own chart)
`
eks.amazonaws.com/role-arn: arn:aws:iam::<aws.account.id>:role/<cluster name>-member
`
These annotations allowed access to the s3 buckets
However, now that the operator is automatically managing these SAs, it seems I am no longer able to set these annotations.
is there a way to either override this behavior at all?

zimnx:
Users may need to annotate ServiceAccount used by ScyllaCluster
to integrate with k8s provider services like IAM role management.
These shouldn't be removed by Operator during reconcilation

Fixes #932

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -12,9 +12,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/api/policy/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/utils/pointer"
 )
 
@@ -648,17 +650,45 @@ func stringOrDefault(str, def string) string {
 	return def
 }
 
-func MakeServiceAccount(sc *scyllav1.ScyllaCluster) *corev1.ServiceAccount {
+func MakeServiceAccount(sc *scyllav1.ScyllaCluster, serviceAccounts map[string]*corev1.ServiceAccount, serviceAccountLister corev1listers.ServiceAccountLister) (*corev1.ServiceAccount, error) {
+	annotations := map[string]string{}
+	name := naming.MemberServiceAccountNameForScyllaCluster(sc.Name)
+
+	existing, found := serviceAccounts[name]
+	if !found {
+		// In 1.7.0 we started managing ServiceAccounts. Because previously these were created by users
+		// there's a high chance that they don't match our selector, hence are not available in serviceAccounts.
+		// Ref: https://github.com/scylladb/scylla-operator/issues/932
+		//
+		// Cache fetch happens only once - when SA is being adopted, or it's missing.
+		// It can be removed in 1.8.0 because all SAs will be already adopted, and they will match our selector.
+		// TODO: https://github.com/scylladb/scylla-operator/issues/934
+		var err error
+		existing, err = serviceAccountLister.ServiceAccounts(sc.Namespace).Get(name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("can't get %s service account: %w", naming.ManualRef(sc.Namespace, name), err)
+		}
+	}
+
+	if existing != nil {
+		// We need to copy annotations stored in user SA because these may control integrations with k8s provider services
+		// like IAM management.
+		for k, v := range existing.Annotations {
+			annotations[k] = v
+		}
+	}
+
 	return &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      naming.MemberServiceAccountNameForScyllaCluster(sc.Name),
+			Name:      name,
 			Namespace: sc.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				*metav1.NewControllerRef(sc, controllerGVK),
 			},
-			Labels: naming.ClusterLabels(sc),
+			Labels:      naming.ClusterLabels(sc),
+			Annotations: annotations,
 		},
-	}
+	}, nil
 }
 
 func MakeRoleBinding(sc *scyllav1.ScyllaCluster) *rbacv1.RoleBinding {
--- a/pkg/resourceapply/core.go
+++ b/pkg/resourceapply/core.go
@@ -3,6 +3,7 @@ package resourceapply
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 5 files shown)


pkg/resourceapply/core.go
```diff
@@ -237,11 +238,26 @@ func ApplyServiceAccount(
 	}
 
 	requiredCopy := required.DeepCopy()
+
+	// Users may need to add custom annotations to integrate with k8s provider services.
+	// These should be filtered out before hash is computed.
+	customAnnotations := map[string]string{}
+	for key, value := range requiredCopy.Annotations {
+		if !strings.HasPrefix(key, "scylla-operator.scylladb.com") {
```
tnozicka, line 246:
```suggestion
		if !strings.HasPrefix(key, "scylla-operator.scylladb.com/") {
```

tnozicka (commented):
please also add this to the adoption e2e test

tnozicka (dismissed):
/lgtm
thanks

zimnx:
@tnozicka PTAL, e2e uncovered that in case when SA doesn't match our selector it's not available in SA list. I added a live call to fetch user object in such case.

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +651,47 @@ func stringOrDefault(str, def string) string {
...
+func MakeServiceAccount(ctx context.Context, sc *scyllav1.ScyllaCluster, serviceAccounts map[string]*corev1.ServiceAccount, coreClient corev1client.CoreV1Interface) (*corev1.ServiceAccount, error) {
+	var annotations map[string]string
+	name := naming.MemberServiceAccountNameForScyllaCluster(sc.Name)
+
+	// In 1.7.0 we started managing ServiceAccounts. Because previously these were created by users
+	// there's a high chance that they don't match our selector, hence are not available in serviceAccounts.
+	//
+	// We need to copy annotations stored in user SA because these may control integrations with k8s provider services
+	// like IAM management.
+	// Live call happens only once - when SA is being adopted. It can be removed in 1.8.0 because all SAs will be already
```
tnozicka, line 663: Why do you need a live call?
I was thinking to:
- before calling this function list the cache and extend the map with the missing SA
- set RV, so ApplyServiceAccount work correctly with the input

zimnx: replaced live call with cache fetch. No need to ammend the fetched resource since we only need to copy annotations.

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,47 @@ func stringOrDefault(str, def string) string {
 	return def
 }
 
-func MakeServiceAccount(sc *scyllav1.ScyllaCluster) *corev1.ServiceAccount {
+func MakeServiceAccount(sc *scyllav1.ScyllaCluster, serviceAccounts map[string]*corev1.ServiceAccount, serviceAccountLister corev1listers.ServiceAccountLister) (*corev1.ServiceAccount, error) {
+	var annotations map[string]string
+	name := naming.MemberServiceAccountNameForScyllaCluster(sc.Name)
+
+	// In 1.7.0 we started managing ServiceAccounts. Because previously these were created by users
```
tnozicka, line 657: Please link an issue to remove this.

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,47 @@ func stringOrDefault(str, def string) string {
...
+	// there's a high chance that they don't match our selector, hence are not available in serviceAccounts.
+	//
+	// We need to copy annotations stored in user SA because these may control integrations with k8s provider services
```
tnozicka, line 660: nit: I'd move this comment over to the block below

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,47 @@ func stringOrDefault(str, def string) string {
...
+	// like IAM management.
+	// Cache fetch happens only once - when SA is being adopted. It can be removed in 1.8.0 because all SAs will be already
```
tnozicka, line 662: technically also when the SA is really missing

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,47 @@ func stringOrDefault(str, def string) string {
...
+	// adopted, and they will match our selector.
+	existing, exists := serviceAccounts[name]
```
tnozicka, line 664:
```suggestion
	existing, found := serviceAccounts[name]
```

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,47 @@ func stringOrDefault(str, def string) string {
...
+	if !exists {
+		var err error
+		existing, err = serviceAccountLister.ServiceAccounts(sc.Namespace).Get(name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("can't get %s service account: %w", naming.ManualRef(sc.Namespace, name), err)
+		}
+		exists = !apierrors.IsNotFound(err)
```
tnozicka, line 671:
```suggestion
		exists = err == nil
```

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,47 @@ func stringOrDefault(str, def string) string {
...
+	}
+
+	if exists {
+		if existing.Annotations != nil {
+			annotations = make(map[string]string)
+		}
```
tnozicka, line 677: nit: Although this is technically correct, would it be clearer to just always create a non nil map at the beginning?

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go
zimnx: There's a difference between empty map and nil map, although it doesn't matter in this case. Added initialization to have shorter code.

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,46 @@ func stringOrDefault(str, def string) string {
...
+		// TODO: https://github.com/scylladb/scylla-operator/issues/934
+		var err error
+		existing, err = serviceAccountLister.ServiceAccounts(sc.Namespace).Get(name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("can't get %s service account: %w", naming.ManualRef(sc.Namespace, name), err)
+		}
+		found = err == nil
+	}
+
+	if found {
```
tnozicka, line 676:
```suggestion
	if existing != nil {
```

pkg/controller/scyllacluster/resource.go
```diff
@@ -648,17 +650,46 @@ func stringOrDefault(str, def string) string {
...
+		//
+		// Cache fetch happens only once - when SA is being adopted, or it's missing.
+		// It can be removed in 1.8.0 because all SAs will be already adopted, and they will match our selector.
+		// TODO: https://github.com/scylladb/scylla-operator/issues/934
+		var err error
+		existing, err = serviceAccountLister.ServiceAccounts(sc.Namespace).Get(name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("can't get %s service account: %w", naming.ManualRef(sc.Namespace, name), err)
+		}
+		found = err == nil
```
tnozicka, line 673: reusing the var seems a bit confusing because on the first look these 2 ifs could be an else block, but it's just a reused variable

pkg/controller/scyllacluster/resource.go
tnozicka: I don't see it moved though https://github.com/scylladb/scylla-operator/pull/933/files#diff-d5f374ace56ea95c01835a2bccad226009def5f02e5d88c10e261ef5ee394718R677

tnozicka (approved):
/lgtm
thanks

PR #937, opened 2022-02-03
Bring back ServiceAccount template in Scylla Helm chart

### LINKED ISSUE #936: Custom ServiceAccount annotations are ignored in Scylla Helm chart
**Describe the bug**
Scylla Helm chart ignores annotations that are supposed to be set on ServiceAccount used by ScyllaCluster Pods.

**To Reproduce**
1. Set custom annotations using `serviceAccount.annotations` in `values` file
2. Create/update ScyllaCluster via Helm
3. Annotations are not added to ServiceAccount used by ScyllaCluster Pods

**Environment:**
 - Scylla chart version: 1.7.0, 1.7.1

zimnx:
Even though Scylla Operator creates and manages Service Account
used by ScyllaClusters we have to keep it in Scylla Helm chart
to keep backward compatibility of setting custom annotations.
Operator will adopt and reconcile it anyways on creation and updates.

Fixes #936

### DIFF
--- a/helm/scylla/templates/serviceaccount.yaml
+++ b/helm/scylla/templates/serviceaccount.yaml
@@ -1,0 +1,14 @@
+{{- if (and .Values.serviceAccount.create .Values.serviceAccount.annotations) -}}
+# Even though Scylla Operator creates and manages Service Account used by ScyllaClusters
+# we have to keep it here to keep backward compatibility of setting custom annotations via helm chart.
+# Operator will adopt it anyways on creation and updates.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ printf "%s-member" (include "scylla.fullname" .) }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "scylla.labels" . | nindent 4 }}
+  annotations:
+    {{- toYaml .Values.serviceAccount.annotations | nindent 4 }}
+{{- end }}
--- a/helm/scylla/templates/_helpers.tpl
+++ b/helm/scylla/templates/_helpers.tpl
@@ -34,7 +34,10 @@ Create chart name and version as used by the chart label.
 Common labels
 */}}
 {{- define "scylla.labels" -}}
-{{ include "scylla.selectorLabels" . }}
+scylla/cluster: {{ include "scylla.fullname" . }}
+app: scylla
+app.kubernetes.io/name: scylla
+app.kubernetes.io/managed-by: scylla-operator
 {{- end }}
 
 {{/*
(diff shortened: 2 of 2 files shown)


helm/scylla/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if (and .Values.serviceAccount.create .Values.serviceAccount.annotations) -}}
+# Even though Scylla Operator creates and manages Service Account used by ScyllaClusters
+# we have to keep it here to keep backward compatibility of setting custom annotations via helm chart.
+# Operator will adopt it anyways on creation and updates.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "scylla.serviceAccountName" . }}
```
tnozicka, line 8: does this need to be fixed?

helm/scylla/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "scylla.labels" . | nindent 4 }}
```
tnozicka, line 11: this should be fixed to what the operator labels are and there should be a unit test to validate it's in sync. we want to drop the forced adoption eventually and adopt only those matching the selector

helm/scylla/templates/serviceaccount.yaml
zimnx: it was defined in _helpers. Moved it here.

helm/scylla/templates/serviceaccount.yaml
zimnx: Good point, synced labels with what is set on our end. 
Helm doesn't have unit test library, it only allows to spawn pods verifying deployment  - like checking whether proper auth is set to DB succeded. 
There're third-party plugins adding unit tests, although i haven't used any. 
Evaluation and integration seems out of scope of this PR, verified output manually.
Leaving evaluation of unit tests library as separate issue:   https://github.com/scylladb/scylla-operator/issues/938

helm/scylla/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if (and .Values.serviceAccount.create .Values.serviceAccount.annotations) -}}
```
tnozicka, line 1: do we even need the create option at this point if we are only creating it when annotations are set and the reverse doesn't make sense?

helm/scylla/templates/serviceaccount.yaml
tnozicka: I guess QA is gonna the annotation path for helm, so this is fine to address later

tnozicka (approved):

helm/scylla/templates/serviceaccount.yaml
zimnx: It doesn't make any sense but switch is already there, someone might be using it.

tnozicka: it's a dead switch though

PR #944, opened 2022-02-15
Update incorrect Docs for enabling alternator

mscheer93:
Current Docs do not mention the parameter `enabled` for enabling alternator.

Alternator API via port 8000 is only active when setting this parameter to true in values.yaml of latest scylla helm-chart.

Here you can find the corresponding code in the current values.yaml file:

https://github.com/scylladb/scylla-operator/blob/master/helm/scylla/values.yaml#L25

Resolves #

### DIFF
--- a/docs/poetry.lock
+++ b/docs/poetry.lock
@@ -7,14 +7,6 @@ optional = false
 python-versions = "*"
 
 [[package]]
-name = "argh"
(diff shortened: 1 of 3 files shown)


mscheer93:
Thanks for the hint :)

https://github.com/scylladb/scylla-operator/pull/947

PR #952, opened 2022-03-17
Expose node-exporter metrics in Scylla Pods

zimnx:
Scylla image by defualt runs node-exporter which is exposing useful
metrics our Scylla Monitoring expects. Previously they weren't exposed
on Pod level hence Prometheus wans't able to scrape them.


tnozicka (approved):
/lgtm

PR #955, opened 2022-03-18
[WIP] Add host_id annotation sync to sidecar controller

rzetelskik:


pkg/controller/sidecar/sync.go
```diff
@@ -104,6 +105,39 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service)
 	return nil
 }
 
+func (c *Controller) setHostIdLabel(ctx context.Context, svc *corev1.Service) error {
```
tnozicka, line 108: nit: setter methods don't normally fetch the value inside but take it as an argument

pkg/controller/sidecar/sync.go
```diff
@@ -104,6 +105,39 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service)
...
+		}
+
+		if len(hostid) == 0 {
+			return fmt.Errorf("received an invalid host_id")
+		}
+
+		c.hostId = &hostid
+	}
+
+	_, err := c.kubeClient.CoreV1().Services(c.namespace).Patch(
```
tnozicka, line 127: would it be cleaner to do an UPDATE here? so any of the preconditions like `deletionTimestamp != nil` holds?

pkg/controller/sidecar/sync.go
```diff
@@ -134,5 +168,13 @@ func (c *Controller) sync(ctx context.Context) error {
 		}
 	}
 
+	hostIdValue, hasHostIdAnnotation := svc.Annotations[naming.HostIdAnnotation]
+	if !hasHostIdAnnotation || (c.hostId != nil && hostIdValue != *c.hostId) {
+		err := c.setHostIdLabel(ctx, svc)
```
tnozicka, line 173: nit: I'd inline the set method but extract the `GetHostID` method that either uses cache or and API call. the condition above will be also simplified.

pkg/naming/constants.go
```diff
@@ -27,6 +27,12 @@ const (
 	LabelValueFalse = "false"
 )
 
+// Annotations used internally by sidecar controller.
+const (
+	// HostIdAnnotation annotates the host_id of a node.
+	HostIdAnnotation = "scylla.internal/host-id"
```
tnozicka, line 33:
```suggestion
	HostIdAnnotation = "internal.scylla-operator.scylladb.com/host-id"
```

pkg/scyllaclient/client.go
```diff
@@ -160,6 +160,23 @@ func (c *Client) Status(ctx context.Context, host string) (NodeStatusInfoSlice,
 	return all, nil
 }
 
+func (c *Client) LocalHostid(ctx context.Context, host string, retry bool) (string, error) {
```
tnozicka, line 163:
```suggestion
func (c *Client) LocalHostID(ctx context.Context, host string, retry bool) (string, error) {
```

tnozicka: the pattern of using verbs isn't well established here but `GetLocalHostID` would be a pretty name

pkg/controller/sidecar/sync.go
tnozicka: this one is technically a status so it is kind of allowed to be updated even with deletion timestamp set but update still seems more fitting (and easier / type safe) here

pkg/resourceapply/helpers.go
```diff
@@ -16,6 +16,23 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// ExtractManagedAnnotations extracts managed annotations from the object and returns them.
+func ExtractManagedAnnotations(obj metav1.Object) map[string]string {
```
tnozicka, line 20: isn't this actually extracting the **UN**managed annotations?

pkg/resourceapply/helpers.go
```diff
@@ -16,6 +16,23 @@ import (
...
+	annotations := obj.GetAnnotations()
+
+	customAnnotations := map[string]string{}
+	for key, value := range annotations {
+		if !strings.HasPrefix(key, "scylla-operator.scylladb.com/") {
```
tnozicka, line 25: I think this would be a good adept for a separate function (`IsManagedAnnotation`) and also making the prefix a const that is used in the annotation consts so we can be sure those are managed

tnozicka (commented):
- please an e2e
- I am having second thoughts about our label approach so skip the comments around that and I'll try ti fix that on master for you

pkg/scyllaclient/client.go
tnozicka: I'll leave the naming to @zimnx then, but I'd like at least the CamelCase

tnozicka:
#960 will give you the machinery

pkg/scyllaclient/client.go
zimnx: i'm fine with Get prefix, lets use it

pkg/controller/sidecar/controller.go
```diff
@@ -48,6 +49,8 @@ type Controller struct {
 
 	queue workqueue.RateLimitingInterface
 	key   string
+
+	hostId *string
```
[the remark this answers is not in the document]
rzetelskik: changed to atomic.Value

pkg/controller/sidecar/sync.go
```diff
@@ -134,5 +145,18 @@ func (c *Controller) sync(ctx context.Context) error {
 		}
 	}
 
+	hostId, err := c.getHostId(ctx)
```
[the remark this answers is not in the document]
rzetelskik: moved to separate method to get rid of the if-else soup

rzetelskik:
@tnozicka @zimnx thanks, I've applied your suggestions. I'll wait for #960 to land to rebase

pkg/controller/sidecar/sync.go
```diff
@@ -104,6 +104,36 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service)
...
+}
+
+func (c *Controller) syncHostIdAnnotation(ctx context.Context, svc *corev1.Service) error {
+	hostId, err := c.getHostId(ctx)
+	if err != nil {
+		return err
+	}
+
+	hostIdValue, hasHostIdAnnotation := svc.Annotations[naming.HostIdAnnotation]
+	if hasHostIdAnnotation && hostIdValue == hostId {
```
tnozicka, line 125: nit: i guess you could just compare the ids given the default value

pkg/controller/sidecar/sync.go
```diff
@@ -104,6 +104,36 @@ func (c *Controller) decommissionNode(ctx context.Context, svc *corev1.Service)
...
+		return nil
+	}
+
+	err = c.setHostIdAnnotation(ctx, svc, hostId)
```
tnozicka, line 129: nit: pls inline (this is an opinionated function that could be mistaken for setting the annotation on the Service object instead of calling the API, the name is not the best fit)

pkg/controller/sidecar/sync.go
```diff
@@ -134,5 +164,10 @@ func (c *Controller) sync(ctx context.Context) error {
 		}
 	}
 
+	err = c.syncHostIdAnnotation(ctx, svc)
```
tnozicka, line 167: maybe move it before decommission

tnozicka (commented):
few nits, please:
- wrap the errors with a context message
- cammelCase for `id` is `ID`

pkg/controller/sidecar/sync.go
rzetelskik: I could but I think this is a more idiomatic approach

pkg/controller/sidecar/controller.go
```diff
@@ -234,3 +239,28 @@ func (c *Controller) deleteService(obj interface{}) {
...
 	c.enqueue(svc)
 }
+
+func (c *Controller) getHostId(ctx context.Context) (string, error) {
+	v := c.hostId.Load()
+	if v != nil {
+		return v.(string), nil
+	}
+
+	scyllaClient, err := controllerhelpers.NewScyllaClientForLocalhost()
```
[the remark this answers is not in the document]
rzetelskik: discussed offline, switched to RWLock

pkg/naming/constants.go
```diff
@@ -50,6 +57,9 @@ const (
 	PrometheusPortAnnotation   = "prometheus.io/port"
 
 	ForceRedeploymentReasonAnnotation = "scylla-operator.scylladb.com/force-redeployment-reason"
+
+	ServiceTypeHeadless = "headless"
```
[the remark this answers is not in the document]
tnozicka:
```suggestion
	ServiceTypeHeadless = "identity"
```

pkg/scyllaclient/client.go
```diff
@@ -160,6 +160,23 @@ func (c *Client) Status(ctx context.Context, host string) (NodeStatusInfoSlice,
 	return all, nil
 }
 
+func (c *Client) GetLocalHostId(ctx context.Context, host string, retry bool) (string, error) {
+	if len(host) > 0 {
```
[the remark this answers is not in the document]
rzetelskik: The method returns the host id of the node that you're connecting with, so although it doesn't make much sense when the client is only talking to the localhost, I don't see why not when connecting to a pool of nodes, other than the naming being a little unfortunate. But then again, it follows the name of the actual endpoint https://github.com/scylladb/scylla/blob/cfe04e6437f09defce0d3acd775ac5b81b2597f4/api/api-doc/storage_service.json#L15.

rzetelskik:
thanks @tnozicka , I've addressed your comments

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
...
+	o "github.com/onsi/gomega"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("ScyllaCluster", func() {
```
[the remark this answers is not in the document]
tnozicka: tiny nit: Describe should ideally hold the feature name (related to this file)

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2022 ScyllaDB
```
tnozicka, line 1: maybe name the file just `scyllacluster_hostid.go`

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		svcs, err := f.KubeClient().CoreV1().Services(sc.Namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: utils.GetMemberServiceSelector(sc.Name).String(),
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		for _, svc := range svcs.Items {
+			hostIdValue, hasHostIdAnnotation := svc.Annotations[naming.HostIDAnnotation]
+			o.Expect(hasHostIdAnnotation).To(o.BeTrue())
```
tnozicka, line 61: remove in favor of the later

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
...
+
+			host := svc.Spec.ClusterIP
+			o.Expect(host).NotTo(o.Equal(corev1.ClusterIPNone))
+
+			hostId, err := scyllaClient.GetLocalHostId(ctx, host, false)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			o.Expect(hostIdValue).To(o.Equal(hostId))
```
tnozicka, line 69:
```suggestion
			o.Expect(svc.Annotations).To(o.HaveKeyWithValue(naming.HostIDAnnotation, hostIdValue))
```

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		for _, svc := range svcs.Items {
+			hostIdValue, hasHostIdAnnotation := svc.Annotations[naming.HostIDAnnotation]
+			o.Expect(hasHostIdAnnotation).To(o.BeTrue())
+
+			host := svc.Spec.ClusterIP
+			o.Expect(host).NotTo(o.Equal(corev1.ClusterIPNone))
+
+			hostId, err := scyllaClient.GetLocalHostId(ctx, host, false)
```
tnozicka, line 66:
```suggestion
			hostID, err := scyllaClient.GetLocalHostId(ctx, host, false)
```

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
```
tnozicka, line 67: sanity check that the hostID isn't empty

tnozicka (commented):
last batch of nits coupled to the `HaveKeyWithValue` request, rest lgtm

test/e2e/set/scyllacluster/scyllacluster_annotate_hostid.go
```diff
@@ -0,0 +1,72 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("ScyllaCluster", func() {
+	defer g.GinkgoRecover()
+
+	f := framework.NewFramework("scyllacluster")
+
+	g.It("should annotate services with hostIDs", func() {
```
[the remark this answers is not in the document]
rzetelskik: changed to "should be reflected as a Service annotation"

tnozicka (commented):
/lgtm
thanks!

pkg/controller/scyllacluster/resource.go
```diff
@@ -31,11 +31,14 @@ const (
 )
 
 func HeadlessServiceForCluster(c *scyllav1.ScyllaCluster) *corev1.Service {
+	labels := naming.ClusterLabels(c)
+	labels[naming.ScyllaServiceTypeLabel] = string(naming.ScyllaServiceTypeIdentity)
```
zimnx, line 35: would you mind splitting service type label changes to separate PR? It's worth notiifying users about the change in the release notes.

pkg/controller/sidecar/controller.go
```diff
@@ -102,6 +110,9 @@ func NewController(
 
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "scyllasidecar"),
 		key:   key,
+		hostID: hostID{
+			v: "",
+		},
```
zimnx, line 115: nit: it's a noop, zero value fits well

pkg/controller/scyllacluster/resource.go
rzetelskik: #969

rzetelskik:
Thanks @zimnx , I've extracted #969 , I'll rebase after it will have merged into master.

rzetelskik:
rebased on current master

zimnx (approved):
LGTM, thanks

PR #967, opened 2022-04-06
Fix decommissioning of drained nodes

tnozicka:
Fixes decommissioning of drained nodes and adds e2e coverage. Previously, when the node was already drained, the decommissioning would get stuck in the sidecar until a change on the corresponding service would trigger a new sync.


zimnx (approved):
lgtm, thanks

PR #976, opened 2022-04-27
Gather logs from previous iterations of restarted containers of non-e2e namespaced resources

rzetelskik:
**Description of your changes:** At the moment we are only gathering logs from previous iterations of restarted containers in e2e namespaces. This pull request introduces a modification that also gathers such logs from containers in other namespaced resources.

### DIFF
--- a/hack/ci-gather-artifacts.sh
+++ b/hack/ci-gather-artifacts.sh
@@ -64,7 +64,12 @@ for n in ${namespaces}; do
                 containers=$( kubectl -n "${n}" get "pods/${o}" --template='{{ range .spec.containers }}{{ println .name }}{{ end }}' )
                 for c in ${containers}; do
                     # Logs are best effort, a pod might not have been running yet.
-                    { kubectl -n "${n}" logs pod/"${o}" -c="${c}" > "${c}".log; } || true
+                    { kubectl -n "${n}" logs pod/"${o}" -c="${c}" > "${c}".current; } || true
+
+                    has_previous=$( kubectl -n "${n}" get "pods/${o}" --template="{{ range .status.containerStatuses }}{{ if eq .name \"${c}\" }}{{ gt .restartCount 0 | print }}{{ end }}{{ end }}" )
+                    if [[ "${has_previous}" = "true" ]]; then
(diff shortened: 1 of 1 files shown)


hack/ci-gather-artifacts.sh
```diff
@@ -64,7 +64,12 @@ for n in ${namespaces}; do
                 containers=$( kubectl -n "${n}" get "pods/${o}" --template='{{ range .spec.containers }}{{ println .name }}{{ end }}' )
                 for c in ${containers}; do
                     # Logs are best effort, a pod might not have been running yet.
-                    { kubectl -n "${n}" logs pod/"${o}" -c="${c}" > "${c}".log; } || true
+                    { kubectl -n "${n}" logs pod/"${o}" -c="${c}" > "${c}".current; } || true
+
+                    previous=$( kubectl -n "${n}" get "pods/${o}" --template="{{ range .status.containerStatuses }}{{ if eq .name \"${c}\" }}{{ gt .restartCount 0 | print }}{{ end }}{{ end }}" )
```
tnozicka, line 69:
```suggestion
                    has_previous=$( kubectl -n "${n}" get "pods/${o}" --template="{{ range .status.containerStatuses }}{{ if eq .name \"${c}\" }}{{ gt .restartCount 0 | print }}{{ end }}{{ end }}" )
```

hack/ci-gather-artifacts.sh
```diff
@@ -64,7 +64,12 @@ for n in ${namespaces}; do
...
+                    if [[ "${previous}" = true ]]; then
```
tnozicka, line 70: nit: the second value should be quoted, but would it be more defensive to check just for non empty string like with grep?

rzetelskik: previous evaluates to either true or false, why should it check for a non-empty string?

tnozicka: both are fine. the idea was to print a string only if greater then zero and then you don't have to care about the value, it's format (e.g. if a next version prints it with quotes or a capital letter) or if there are more then one match (although container names have to be unique)

tnozicka (approved):
/approve
/lgtm

PR #984, opened 2022-06-03
Use local NVMe storage class in EKS example

fee-mendes:
The EKS provisioning example defaults to provisioning i3.2xlarge machines, but cluster.yaml has no references to which storageClass to use when deploying a scyllaCluster resource.

Update the example as currently present within the GKE example (which uses the very same storageClass).

### DIFF
--- a/examples/eks/cluster.yaml
+++ b/examples/eks/cluster.yaml
@@ -26,6 +26,7 @@ spec:
     - name: a
       members: 1
       storage:
+        storageClassName: local-raid-disks
         capacity: 1800G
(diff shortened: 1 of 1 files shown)


zimnx (dismissed):
LGTM, thanks!

tnozicka (commented):
Thanks for the contribution!
Please adjust the commit message summary and you are good to go. Think of it as it should complete this sentence `If applied, this commit will <your subject line here>` (like: `Add storage class to EKS examples`)

PR #993, opened 2022-06-27
Reconcile Ingress objects for ScyllaCluster

zimnx:
ScyllaCluster CRD is extended with ExposeOptions and Domains allowing users to
configure domain, ingress class name and custom annotations.
If provided, Operator creates an Ingress object for each Scylla Node
pointing directly to it, and one additional Ingress for headless
Service allowing accessing any of available Scylla Nodes.

ExposeOptions are considered as experimental and are subject for
deprecation and incompatible changes. Using it, may require manual steps
during upgrades. User discretion is advised.


pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,44 @@ type ScyllaClusterSpec struct {
...
+}
+
+// ServiceBackendPort is the service port being referenced.
+type ServiceBackendPort struct {
+	// Name is the name of the port on the Service.
+	// This is a mutually exclusive setting with "Number".
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// Number is the numerical port number on the Service.
```
rzetelskik, line 132: Imo 'numerical' implies numerical types, not just `int32`. I'd change it to integer/integral or remove the adjective altogether, the type documents it well enough.

pkg/api/validation/cluster_validation.go
```diff
@@ -86,6 +88,50 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
+	}
+
+	if len(options.Annotations) != 0 {
+		allErrs = append(allErrs, apimachineryvalidation.ValidateAnnotations(options.Annotations, fldPath.Child("annotations"))...)
+	}
+
+	hasPortName := len(options.ServiceBackendPort.Name) > 0
+	hasPortNumber := options.ServiceBackendPort.Number != 0
+	if hasPortName && hasPortNumber {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("serviceBackendPort"), "", "cannot set both port name & port number"))
```
rzetelskik, line 122:
```suggestion
		allErrs = append(allErrs, field.Forbidden(fldPath.Child("serviceBackendPort"), "cannot set both port name & port number"))
```

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,80 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+		labels := naming.ClusterLabels(c)
+
+		switch naming.ScyllaServiceType(service.Labels[naming.ScyllaServiceTypeLabel]) {
+		case naming.ScyllaServiceTypeIdentity:
+			host = fmt.Sprintf("%s.%s", naming.ScyllaIngressAnySubdomain, c.Spec.IngressOptions.Domain)
+			labels[naming.ScyllaIngressTypeLabel] = string(naming.ScyllaIngressTypeAny)
+		case naming.ScyllaServiceTypeMember:
+			hostID, ok := service.Annotations[naming.HostIDAnnotation]
+			if !ok {
+				klog.Infof("Service %s is missing HostID annotation, waiting with Ingress object until it's available", naming.ObjRef(service), service.Labels[naming.ScyllaServiceTypeLabel])
```
rzetelskik, line 663:
```suggestion
				klog.Infof("Service %s is missing HostID annotation, postponing Ingress object creation until it's available", naming.ObjRef(service), service.Labels[naming.ScyllaServiceTypeLabel])
```

pkg/controller/scyllacluster/sync.go
```diff
@@ -336,6 +381,11 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
 		return err
 	}
 
+	ingressMap, err := scc.getIngresses(ctx, sc)
+	if err != nil {
+		return err
```
rzetelskik, line 386: nit: wrap error

pkg/controllertools/ingressmanager.go
```diff
@@ -0,0 +1,139 @@
...
+			claimedMap[ingress.Name] = ingress
+		}
+	}
+	return claimedMap, utilerrors.NewAggregate(errors)
+}
+
+func (m *IngressControllerRefManager) AdoptIngress(ingress *networkingv1.Ingress) error {
+	err := m.CanAdopt()
+	if err != nil {
+		return fmt.Errorf("can't adopt Ingress %v/%v (%v): %v", ingress.Namespace, ingress.Name, ingress.UID, err)
```
rzetelskik, line 92:
```suggestion
		return fmt.Errorf("can't adopt Ingress %v/%v (%v): %w", ingress.Namespace, ingress.Name, ingress.UID, err)
```

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,44 @@ type ScyllaClusterSpec struct {
...
+	// +optional
+	IngressOptions *IngressOptions `json:"ingressOptions,omitempty"`
+}
+
+// IngressOptions defines configuration options for Ingress objects associated with cluster nodes.
+// Experimental - subject to incompatible changes.
+type IngressOptions struct {
+	// domain specifies DNS suffix of hostname of each Scylla node.
+	// It is concatenated with node host_id to form Ingress rule pointing to particular Scylla node.
+	Domain string `json:"domain"`
```
tnozicka, line 111: domain isn't specific to ingresses (say it's also needed for standalone certs) and it also needs to be an array

pkg/api/validation/cluster_validation.go
rzetelskik: Fair. Forbidden would make sense in regard to either of the inner fields, for `serviceBackendPort` invalid seems right

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,44 @@ type ScyllaClusterSpec struct {
 	// used for pulling Scylla and Agent images.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// ingressOptions is an optional Ingress configuration options.
+	// If provided, Ingress objects are generated for each Scylla node with following options.
+	// Experimental - subject to incompatible changes.
+	// +optional
+	IngressOptions *IngressOptions `json:"ingressOptions,omitempty"`
```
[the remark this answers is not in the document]
zimnx: done

pkg/controllertools/ingressmanager.go
zimnx: done

pkg/controller/scyllacluster/sync.go
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
...
+// Experimental - subject to incompatible changes.
+type IngressOptions struct {
+	// ServiceBackendPort a port name or port number of the member service.
+	ServiceBackendPort ServiceBackendPort `json:"serviceBackendPort"`
+
+	// ingressClassName specifies Ingress class name.
+	// +optional
+	IngressClassName string `json:"ingressClassName,omitempty"`
+
+	// annotations specify custom annotations added to every Ingress object.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
...
+
+// ExposeOptions hold options related to exposing ScyllaCluster backends.
+type ExposeOptions struct {
+	// ingress is an Ingress configuration options.
+	// If provided, Ingress objects are generated for each Scylla node with following options.
+	Ingress *IngressOptions `json:"ingress,omitempty"`
+}
+
+// IngressOptions defines configuration options for Ingress objects associated with cluster nodes.
+// Experimental - subject to incompatible changes.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
...
+type IngressOptions struct {
+	// ServiceBackendPort a port name or port number of the member service.
+	ServiceBackendPort ServiceBackendPort `json:"serviceBackendPort"`
+
+	// ingressClassName specifies Ingress class name.
+	// +optional
+	IngressClassName string `json:"ingressClassName,omitempty"`
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
...
+	// ingress is an Ingress configuration options.
+	// If provided, Ingress objects are generated for each Scylla node with following options.
+	Ingress *IngressOptions `json:"ingress,omitempty"`
+}
+
+// IngressOptions defines configuration options for Ingress objects associated with cluster nodes.
+// Experimental - subject to incompatible changes.
+type IngressOptions struct {
+	// ServiceBackendPort a port name or port number of the member service.
+	ServiceBackendPort ServiceBackendPort `json:"serviceBackendPort"`
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
 	// used for pulling Scylla and Agent images.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// domains specifies a list of DNS domains this cluster is identified by.
+	// Routes and certificates to particular cluster nodes should use these domains as suffixes.
+	// +optional
+	Domains []string `json:"domains"`
+
+	// exposeOptions specifies parameters related to exposing ScyllaCluster backends.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
 	// used for pulling Scylla and Agent images.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// domains specifies a list of DNS domains this cluster is identified by.
+	// Routes and certificates to particular cluster nodes should use these domains as suffixes.
+	// +optional
+	Domains []string `json:"domains"`
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
 	// used for pulling Scylla and Agent images.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// domains specifies a list of DNS domains this cluster is identified by.
+	// Routes and certificates to particular cluster nodes should use these domains as suffixes.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
 	// used for pulling Scylla and Agent images.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// domains specifies a list of DNS domains this cluster is identified by.
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,91 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+		var hosts []string
+		labels := naming.ClusterLabels(c)
+
+		switch naming.ScyllaServiceType(service.Labels[naming.ScyllaServiceTypeLabel]) {
+		case naming.ScyllaServiceTypeIdentity:
+			for _, domain := range c.Spec.Domains {
+				hosts = append(hosts, fmt.Sprintf("%s.%s", naming.ScyllaIngressAnySubdomain, domain))
+			}
+			labels[naming.ScyllaIngressTypeLabel] = string(naming.ScyllaIngressTypeAny)
+		case naming.ScyllaServiceTypeMember:
```
tnozicka, line 662: nit: \n

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,91 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+			hostID, ok := service.Annotations[naming.HostIDAnnotation]
+			if !ok {
+				klog.Infof("Service %s is missing HostID annotation, postponing Ingress object creation until it's available", naming.ObjRef(service))
```
tnozicka, line 665: Warningf

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,91 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+				continue
+			}
+
+			for _, domain := range c.Spec.Domains {
+				hosts = append(hosts, fmt.Sprintf("%s.%s", hostID, domain))
+			}
+			labels[naming.ScyllaIngressTypeLabel] = string(naming.ScyllaIngressTypeDirect)
```
tnozicka, line 672: nit: \n

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,91 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+				},
+			},
+			Spec: networkingv1.IngressSpec{
+				IngressClassName: pointer.String(c.Spec.ExposeOptions.Ingress.IngressClassName),
+			},
+		}
+
+		portName := naming.PortNameCQLSSL
+		if c.Spec.Alternator != nil {
+			portName = naming.PortNameAlternator
```
tnozicka, line 695: does this overwrite the cql over ssl port?

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,91 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+	if c.Spec.ExposeOptions == nil || c.Spec.ExposeOptions.Ingress == nil {
+		return nil
+	}
+
+	var ingresses []*networkingv1.Ingress
+
+	pathPrefix := networkingv1.PathTypePrefix
+	for _, service := range services {
+		var hosts []string
+		labels := naming.ClusterLabels(c)
```
tnozicka, line 654: is this only being written to?

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -71,6 +71,14 @@ func TestMemberService(t *testing.T) {
 			Name: "jmx-monitoring",
 			Port: 7199,
 		},
+		{
```
tnozicka, line 74: can you add a unit test for ingress generation?

pkg/naming/constants.go
```diff
@@ -40,13 +40,23 @@ const (
 	ScyllaServiceTypeMember   ScyllaServiceType = "member"
 )
 
+type ScyllaIngressType string
+
+const (
+	ScyllaIngressTypeDirect ScyllaIngressType = "direct"
```
tnozicka, line 46:
```suggestion
	ScyllaIngressTypeNode ScyllaIngressType = "Node"
```

pkg/naming/constants.go
```diff
@@ -40,13 +40,23 @@ const (
...
+	ScyllaIngressTypeAny    ScyllaIngressType = "any"
```
tnozicka, line 47: although the `any` is the dns name, `NodeLB` or `Cluster` types might be more descriptive here. possibly `AnyNode`

pkg/naming/constants.go
```diff
@@ -40,13 +40,23 @@ const (
...
+
+	ScyllaIngressAnySubdomain = "any"
```
tnozicka, line 49: this may not need a constant e2e should inline it to enforce explicit changes to surface the impact

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+		defer cancel()
+
+		sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc.Spec.Datacenter.Racks[0].Members = 2
+		sc.Spec.Domains = []string{"private.nodes.scylladb.com", "public.nodes.scylladb.com"}
+		sc.Spec.ExposeOptions = &scyllav1.ExposeOptions{
+			Ingress: &scyllav1.IngressOptions{
+				IngressClassName: "my-ingress-controller",
+				Annotations: map[string]string{
+					"my-annotation": "my-annotation",
```
tnozicka, line 37: tiny nit: best to use distinct string to test we don't switch key and value accidentally

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+
+			serviceType := naming.ScyllaServiceType(svc.Labels[naming.ScyllaServiceTypeLabel])
+			if serviceType == naming.ScyllaServiceTypeIdentity {
+				hostPrefix = naming.ScyllaIngressAnySubdomain
+			} else if serviceType == naming.ScyllaServiceTypeMember {
+				hostID := svc.Annotations[naming.HostIDAnnotation]
+				o.Expect(hostID).ToNot(o.BeEmpty())
+
+				hostPrefix = hostID
+			}
```
tnozicka, line 83: else?

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+
+		framework.By("Verifying Ingresses")
+		ingresses, err := f.KubeClient().NetworkingV1().Ingresses(sc.Namespace).List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		services, err := f.KubeClient().CoreV1().Services(sc.Namespace).List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Expect(ingresses.Items).To(o.HaveLen(int(sc.Spec.Datacenter.Racks[0].Members) + 1))
+		for _, ingress := range ingresses.Items {
```
tnozicka, line 63: Would the assertion be more readable if you were to label select particular groups of ingresses, use o.ConsistsOf to compare the names and then assert the ingress itself (possibly sharing a common verify code in a helper function)?

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+			o.Expect(ingress.Spec.IngressClassName).ToNot(o.BeNil())
+			o.Expect(*ingress.Spec.IngressClassName).To(o.Equal(sc.Spec.ExposeOptions.Ingress.IngressClassName))
+			o.Expect(ingress.Spec.Rules).To(o.HaveLen(2))
+			o.Expect(ingress.Spec.Rules[0].Host).To(o.Equal(fmt.Sprintf("%s.private.nodes.scylladb.com", hostPrefix)))
+			o.Expect(ingress.Spec.Rules[1].Host).To(o.Equal(fmt.Sprintf("%s.public.nodes.scylladb.com", hostPrefix)))
+			for _, host := range ingress.Spec.Rules {
+				o.Expect(host.HTTP.Paths).To(o.HaveLen(1))
+				o.Expect(host.HTTP.Paths[0].Path).To(o.Equal("/"))
+				o.Expect(host.HTTP.Paths[0].PathType).ToNot(o.BeNil())
+				o.Expect(*host.HTTP.Paths[0].PathType).To(o.Equal(networkingv1.PathTypePrefix))
```
tnozicka, line 96: let's not use constants in assertions

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+				o.Expect(host.HTTP.Paths[0].Backend.Service).ToNot(o.BeNil())
+				o.Expect(host.HTTP.Paths[0].Backend.Service.Name).To(o.Equal(svc.Name))
+				o.Expect(host.HTTP.Paths[0].Backend.Service.Port.Name).To(o.Equal(naming.PortNameCQLSSL))
```
tnozicka, line 99: inline

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+		sc.Spec.Datacenter.Racks[0].Members = 2
+		sc.Spec.Domains = []string{"private.nodes.scylladb.com", "public.nodes.scylladb.com"}
+		sc.Spec.ExposeOptions = &scyllav1.ExposeOptions{
+			Ingress: &scyllav1.IngressOptions{
+				IngressClassName: "my-ingress-controller",
+				Annotations: map[string]string{
+					"my-annotation": "my-annotation",
+				},
+			},
+		}
```
tnozicka, line 40: let's enable alternator and test it as well

test/e2e/set/scyllacluster/scyllacluster_ingress.go
```diff
@@ -0,0 +1,103 @@
...
+			o.Expect(ingress.Spec.Rules[1].Host).To(o.Equal(fmt.Sprintf("%s.public.nodes.scylladb.com", hostPrefix)))
+			for _, host := range ingress.Spec.Rules {
+				o.Expect(host.HTTP.Paths).To(o.HaveLen(1))
+				o.Expect(host.HTTP.Paths[0].Path).To(o.Equal("/"))
+				o.Expect(host.HTTP.Paths[0].PathType).ToNot(o.BeNil())
+				o.Expect(*host.HTTP.Paths[0].PathType).To(o.Equal(networkingv1.PathTypePrefix))
+				o.Expect(host.HTTP.Paths[0].Backend.Service).ToNot(o.BeNil())
+				o.Expect(host.HTTP.Paths[0].Backend.Service.Name).To(o.Equal(svc.Name))
+				o.Expect(host.HTTP.Paths[0].Backend.Service.Port.Name).To(o.Equal(naming.PortNameCQLSSL))
+			}
```
tnozicka, line 100: verify alternator port as well

pkg/controller/scyllacluster/resource.go
zimnx: I don't get what you mean, could you elaborate?

pkg/api/validation/cluster_validation.go
```diff
@@ -86,6 +87,45 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
 		}
 	}
 
+	if len(spec.Domains) == 0 && spec.ExposeOptions != nil && spec.ExposeOptions.Ingress != nil {
```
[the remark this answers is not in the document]
zimnx: condition became much complex, although i left the order as it was, conjuction is commutative and it's still more natural to me.

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,40 @@ type ScyllaClusterSpec struct {
...
+	Ingress *IngressOptions `json:"ingress,omitempty"`
+}
+
+// IngressOptions defines configuration options for Ingress objects associated with cluster nodes.
+// Experimental. Do not rely on any particular behaviour controlled by this field.
+type IngressOptions struct {
+	// ingressClassName specifies Ingress class name.
+	// Experimental. Do not rely on any particular behaviour controlled by this field.
+	// +optional
+	IngressClassName string `json:"ingressClassName,omitempty"`
```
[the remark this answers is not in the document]
zimnx: That's why I called it with full name, leaving as is.

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,53 @@ type ScyllaClusterSpec struct {
...
+	// Experimental - subject to incompatible changes.
+	// +optional
+	ExposeOptions *ExposeOptions `json:"exposeOptions,omitempty"`
+}
+
+// ExposeOptions hold options related to exposing ScyllaCluster backends.
+type ExposeOptions struct {
+	// ingress is an Ingress configuration options.
+	// If provided, Ingress objects are generated for each Scylla node with following options.
+	Ingress *IngressOptions `json:"ingress,omitempty"`
```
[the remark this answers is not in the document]
zimnx: Good idea, generally Scylla doesn't recommend using both API's at the same time, but on the other hand we shouldn't enforce it. Allowed configuration of both APIs.

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,103 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+	}
+	if c.Spec.ExposeOptions.Alternator != nil && c.Spec.ExposeOptions.Alternator.Ingress != nil {
+		backends[naming.BackendAlternator] = c.Spec.ExposeOptions.Alternator.Ingress
+	}
+
+	var ingresses []*networkingv1.Ingress
+
+	pathPrefix := networkingv1.PathTypePrefix
+	for backend, ingressOptions := range backends {
+		for _, service := range services {
```
tnozicka, line 666: alternator would have just 1 backend service, right?

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -925,3 +927,255 @@ func TestStatefulSetForRack(t *testing.T) {
 		})
 	}
 }
+
+func TestMakeIngresses(t *testing.T) {
+	newBasicScyllaCluster := func() *scyllav1.ScyllaCluster {
```
tnozicka, line 932: nit: could be a var that you deep copy

pkg/naming/constants.go
```diff
@@ -132,3 +142,17 @@ const (
 const (
 	ScyllaRuntimeConfigKey string = "ScyllaRuntimeConfig"
 )
+
+const (
+	PortNameCQL              = "cql"
```
tnozicka, line 147: these seem to be limited to the desired service definition and not used elsewhere - do we want to promote them for reuse or rather inline them?

pkg/naming/constants.go
```diff
@@ -132,3 +142,17 @@ const (
...
+	PortNameCQLSSL           = "cql-ssl"
+	PortNameCQLShardAware    = "cql-shard-aware"
+	PortNameCQLSSLShardAware = "cql-ssl-shard-aware"
+	PortNameAlternator       = "alternator"
+	PortNameThrift           = "thrift"
+)
+
+const (
+	BackendCQL        = "cql"
```
tnozicka, line 156: these are also limited to a single function - is there a reason to export them?

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -0,0 +1,190 @@
...
+
+	g.It("should create ingress objects when ingress exposeOptions are provided", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
+
+		sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc.Spec.Datacenter.Racks[0].Members = 1
+		sc.Spec.Domains = []string{"private.nodes.scylladb.com", "public.nodes.scylladb.com"}
+		sc.Spec.Alternator = &scyllav1.AlternatorSpec{
+			Port:           4242,
```
tnozicka, line 36: can this default?

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,62 @@ type ScyllaClusterSpec struct {
...
+	// ingress is an Ingress configuration options.
+	// If provided, Ingress objects routing to CQL SSL port are generated for each Scylla node
+	// with following options.
+	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+	Ingress *IngressOptions `json:"ingress,omitempty"`
+}
+
+type AlternatorExposeOptions struct {
+	// ingress is an Ingress configuration options.
+	// If provided, Ingress objects routing to Alternator are generated for each Scylla node
```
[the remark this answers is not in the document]
zimnx: Alternator drivers also can load balance on the client side: https://www.scylladb.com/2021/04/13/load-balancing-in-scylla-alternator/

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,62 @@ type ScyllaClusterSpec struct {
...
+}
+
+// ExposeOptions hold options related to exposing ScyllaCluster backends.
+// At least one backend must be specified.
+// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+type ExposeOptions struct {
+	// cql specifies expose options for CQL SSL backend.
+	// +optional
+	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+	CQL *CQLExposeOptions `json:"cql,omitempty"`
```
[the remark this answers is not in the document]
zimnx: If needed we could add CQLExposeOptions.Insecure. ServerName is also required for routing binary protocol. Leaving as is, there's a mention in the field comment about SSL.

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,103 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+			var hosts []string
+			labels := naming.ClusterLabels(c)
+
+			switch naming.ScyllaServiceType(service.Labels[naming.ScyllaServiceTypeLabel]) {
+			case naming.ScyllaServiceTypeIdentity:
+				for _, domain := range c.Spec.Domains {
+					hosts = append(hosts, fmt.Sprintf("%s.%s.%s", naming.ScyllaIngressAnySubdomain, backend, domain))
```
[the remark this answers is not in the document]
zimnx: why not?

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,103 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+
+			case naming.ScyllaServiceTypeMember:
+				hostID, ok := service.Annotations[naming.HostIDAnnotation]
+				if !ok {
+					klog.Warningf("Service %q is missing HostID annotation, postponing Ingress object creation until it's available", naming.ObjRef(service))
+					continue
+				}
+
+				for _, domain := range c.Spec.Domains {
+					hosts = append(hosts, fmt.Sprintf("%s.%s.%s", hostID, backend, domain))
```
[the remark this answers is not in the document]
zimnx: as in previous comment, alternator drivers can also load balance on client side, so having per-host endpoint is beneficial

pkg/controller/scyllacluster/resource.go
zimnx: no

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: current API default value (`0`) doesn't enable alterantor :(

pkg/controller/scyllacluster/resource.go
```diff
@@ -34,6 +36,19 @@ func HeadlessServiceForCluster(c *scyllav1.ScyllaCluster) *corev1.Service {
 	labels := naming.ClusterLabels(c)
 	labels[naming.ScyllaServiceTypeLabel] = string(naming.ScyllaServiceTypeIdentity)
 
+	ports := []corev1.ServicePort{
```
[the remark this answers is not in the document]
zimnx: Do we need LB IP? Headless Service with selector returns A records pointing to Pod IPs

pkg/api/validation/cluster_validation.go
```diff
@@ -86,6 +87,66 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
 		}
 	}
 
+	for i, domain := range spec.Domains {
+		for _, msg := range apimachineryvalidation.NameIsDNSSubdomain(domain, false) {
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation.go
```diff
@@ -86,6 +87,66 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
+			allErrs = append(allErrs, field.Required(fldPath.Child("domains"), "at least one domain needs to be provided when exposing ingresses"))
+		}
+
+		if spec.ExposeOptions.Alternator != nil && spec.ExposeOptions.Alternator.Ingress != nil {
+			allErrs = append(allErrs, field.Required(fldPath.Child("domains"), "at least one domain needs to be provided when exposing ingresses"))
+		}
+	}
+
+	if spec.ExposeOptions != nil && spec.ExposeOptions.Alternator != nil && !spec.Alternator.Enabled() {
+		allErrs = append(allErrs, field.Required(fldPath.Child("alternator"), "alternator must be enabled when alternator exposing options are provided"))
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation.go
```diff
@@ -86,6 +87,66 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
+	}
+
+	return allErrs
+}
+
+func ValidateExposeOptions(options *scyllav1.ExposeOptions, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if options.CQL == nil && options.Alternator == nil {
+		allErrs = append(allErrs, field.Required(fldPath, "at least one backend must be specified when exposeOptions are provided"))
```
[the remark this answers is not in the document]
zimnx: done, added optional Disabled field

pkg/api/validation/cluster_validation.go
```diff
@@ -86,6 +87,66 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
+	}
+
+	if options.CQL != nil && options.CQL.Ingress != nil {
+		allErrs = append(allErrs, ValidateIngressOptions(options.CQL.Ingress, fldPath.Child("cql", "ingress"))...)
+	}
+
+	if options.Alternator != nil {
+		if options.Alternator.Ingress != nil {
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation_test.go
```diff
@@ -23,68 +23,149 @@ func TestValidateScyllaCluster(t *testing.T) {
...
+			},
 			expectedErrorList:   field.ErrorList{},
 			expectedErrorString: "",
 		},
 		{
 			name: "two racks with same name",
-			obj:  sameName,
+			cluster: func() *v1.ScyllaCluster {
+				cluster := validCluster.DeepCopy()
+				cluster.Spec.Datacenter.Racks = append(cluster.Spec.Datacenter.Racks, cluster.Spec.Datacenter.Racks[0])
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation_test.go
```diff
@@ -23,68 +23,149 @@ func TestValidateScyllaCluster(t *testing.T) {
...
+				cluster.Spec.ExposeOptions = &v1.ExposeOptions{
+					CQL: &v1.CQLExposeOptions{
+						Ingress: &v1.IngressOptions{},
+					},
+				}
+
+				return cluster
+			},
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeRequired, Field: "spec.domains", BadValue: "", Detail: "at least one domain needs to be provided when exposing ingresses"},
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation_test.go
```diff
@@ -23,68 +23,149 @@ func TestValidateScyllaCluster(t *testing.T) {
...
+				cluster.Spec.ExposeOptions = &v1.ExposeOptions{
+					Alternator: &v1.AlternatorExposeOptions{
+						Ingress: &v1.IngressOptions{},
+					},
+				}
+
+				return cluster
+			},
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeRequired, Field: "spec.domains", BadValue: "", Detail: "at least one domain needs to be provided when exposing ingresses"},
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/validation/cluster_validation_test.go
```diff
@@ -23,68 +23,149 @@ func TestValidateScyllaCluster(t *testing.T) {
...
-	nonUniqueManagerTaskNames.Spec.Repairs = append(nonUniqueManagerTaskNames.Spec.Repairs, v1.RepairTaskSpec{
-		SchedulerTaskSpec: v1.SchedulerTaskSpec{
-			Name: "task-name",
-		},
-	})
-
 	tests := []struct {
 		name                string
-		obj                 *v1.ScyllaCluster
+		cluster             func() *v1.ScyllaCluster
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,103 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
 	}
 }
 
+func MakeIngresses(c *scyllav1.ScyllaCluster, services map[string]*corev1.Service) []*networkingv1.Ingress {
+	// Don't create Ingresses if cluster isn't exposed.
+	if c.Spec.ExposeOptions == nil {
+		return nil
+	}
+
+	ports := map[string]string{
```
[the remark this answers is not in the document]

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +640,103 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+	if c.Spec.ExposeOptions.CQL != nil && c.Spec.ExposeOptions.CQL.Ingress != nil {
+		backends[naming.BackendCQL] = c.Spec.ExposeOptions.CQL.Ingress
+	}
+	if c.Spec.ExposeOptions.Alternator != nil && c.Spec.ExposeOptions.Alternator.Ingress != nil {
+		backends[naming.BackendAlternator] = c.Spec.ExposeOptions.Alternator.Ingress
+	}
+
+	var ingresses []*networkingv1.Ingress
+
+	pathPrefix := networkingv1.PathTypePrefix
```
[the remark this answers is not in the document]

pkg/controller/scyllacluster/resource_test.go
zimnx: done

pkg/naming/constants.go
zimnx: done, moved to unexported consts in resource.go

pkg/naming/constants.go
zimnx: done, moved to unexported consts in resource.go

zimnx:
@tnozicka PTAL

pkg/controller/scyllacluster/resource.go
tnozicka: alternator is load balanced service (unlike CQL) why would it need more then one backend service?

zimnx: As mentioned in previous comment, Alternator drivers also can load balance on the client side: https://www.scylladb.com/2021/04/13/load-balancing-in-scylla-alternator/

zimnx:
@tnozicka as discussed, support for alternator is removed.

pkg/controller/scyllacluster/resource.go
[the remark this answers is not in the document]
tnozicka: > Do we need LB IP?

it's nice to have - if you expose a LB through ingress it's nice if you can use a ServiceIP to reach the same service if you have doubts about the ingress implementation or want to use kubectl port-forward

why are these **only** on the headless service?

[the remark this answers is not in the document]
zimnx: > it's nice to have - if you expose a LB through ingress it's nice if you can use a ServiceIP to reach the same service if you have doubts about the ingress implementation or want to use kubectl port-forward

changed it to ClusterIP service

> why are these only on the headless service?

unified ports

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -925,3 +927,227 @@ func TestStatefulSetForRack(t *testing.T) {
...
+				},
+			},
+			Spec: networkingv1.IngressSpec{
+				IngressClassName: pointer.String(ingressClassName),
+				Rules:            rules,
+			},
+		}
+	}
+
+	tcs := []struct {
```
tnozicka, line 1031: nit: `tt` (aka `test table`, to keep conventions)

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -925,3 +927,227 @@ func TestStatefulSetForRack(t *testing.T) {
...
+								"my-cql-custom-annotation": "my-cql-custom-annotation-value",
+							},
+						},
+					},
+				}
+
+				return cluster
+			}(),
+			expectedIngresses: []*networkingv1.Ingress{
+				newIngress("any-cql",
```
tnozicka, line 1090: nit: these expected values are hard to asses whether defined correctly. I know it's a bit painful but I'd inline the expected values

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -925,3 +927,227 @@ func TestStatefulSetForRack(t *testing.T) {
...
+			t.Parallel()
+
+			got := MakeIngresses(tc.cluster, tc.services)
+
+			less := func(v []*networkingv1.Ingress) func(i, j int) bool {
+				return func(i, j int) bool {
+					return v[i].Name < v[j].Name
+				}
+			}
+			sort.Slice(got, less(got))
```
tnozicka, line 1145: the function must have a consistent output, if not we should fix it there

tnozicka (commented):
some comments, mostly just tests. thanks for the updates!

pkg/controller/scyllacluster/resource_test.go
zimnx: fixed

pkg/controller/scyllacluster/resource_test.go
zimnx: expanded

pkg/controller/scyllacluster/resource_test.go
zimnx: fixed

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -0,0 +1,147 @@
...
+		o.Expect(ingresses.Items).To(o.HaveLen(1))
+
+		verifyIngress(
+			&ingresses.Items[0],
+			fmt.Sprintf("%s-cql", services.Items[0].Name),
+			sc.Spec.ExposeOptions.CQL.Ingress.Annotations,
+			sc.Spec.ExposeOptions.CQL.Ingress.IngressClassName,
+			&services.Items[0],
+			[]string{
+				fmt.Sprintf("any.cql.%s", sc.Spec.DNSDomains[0]),
```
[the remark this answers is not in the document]
zimnx: fixed

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -0,0 +1,147 @@
...
+
+		framework.By("Creating a ScyllaCluster")
+		sc, err := f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for ScyllaCluster to deploy")
+		waitCtx, waitCtxCancel := utils.ContextForRollout(ctx, sc)
+		defer waitCtxCancel()
+
+		sc, err = utils.WaitForScyllaClusterState(waitCtx, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, utils.IsScyllaClusterRolledOut)
```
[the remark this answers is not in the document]
zimnx: added TODO

pkg/naming/constants.go
```diff
@@ -40,13 +40,23 @@ const (
 	ScyllaServiceTypeMember   ScyllaServiceType = "member"
 )
 
+type ScyllaIngressType string
+
+const (
+	ScyllaIngressTypeNode    ScyllaIngressType = "Node"
+	ScyllaIngressTypeAnyNode ScyllaIngressType = "AnyNode"
+
+	ScyllaIngressAnySubdomain = "any"
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +653,109 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+
+			ingresses = append(ingresses, ingress)
+		}
+	}
+
+	return ingresses
+}
+
+func isIngressEnabled(ingressOptions *scyllav1.IngressOptions) bool {
+	return ingressOptions != nil && (ingressOptions.Disabled == nil || !*ingressOptions.Disabled)
```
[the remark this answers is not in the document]

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +653,109 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+
+			default:
+				klog.Warningf("Unsupported Scylla service type %q, not creating Ingress for it", service.Labels[naming.ScyllaServiceTypeLabel])
+				continue
+			}
+
+			ingress := &networkingv1.Ingress{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        fmt.Sprintf("%s-%s", service.Name, ip.backendName),
+					Namespace:   service.Namespace,
```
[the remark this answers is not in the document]

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +653,109 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+
+			switch naming.ScyllaServiceType(service.Labels[naming.ScyllaServiceTypeLabel]) {
+			case naming.ScyllaServiceTypeIdentity:
+				for _, domain := range c.Spec.DNSDomains {
+					hosts = append(hosts, fmt.Sprintf("%s.%s.%s", naming.ScyllaIngressAnySubdomain, ip.backendName, domain))
+				}
+				labels[naming.ScyllaIngressTypeLabel] = string(naming.ScyllaIngressTypeAnyNode)
+
+			case naming.ScyllaServiceTypeMember:
+				hostID, ok := service.Annotations[naming.HostIDAnnotation]
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +653,109 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+				if !ok {
+					klog.Warningf("Service %q is missing HostID annotation, postponing Ingress object creation until it's available", naming.ObjRef(service))
```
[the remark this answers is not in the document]

pkg/controller/scyllacluster/resource.go
```diff
@@ -127,38 +132,61 @@ func memberServicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
...
 		{
-			Name: "cql",
-			Port: 9042,
+			Name: "agent-api",
+			Port: 10001,
 		},
+	}
+
+	ports = append(ports, backendPorts(cluster)...)
+	ports = append(ports, metricsPorts()...)
```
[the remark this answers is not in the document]

pkg/controller/scyllacluster/resource.go
```diff
@@ -30,10 +32,36 @@ const (
 	rootGID = 0
 )
 
+const (
+	backendCQL = "cql"
```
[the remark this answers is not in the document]

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,54 @@ type ScyllaClusterSpec struct {
...
+	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+	// +optional
+	Ingress *IngressOptions `json:"ingress,omitempty"`
+}
+
+// IngressOptions defines configuration options for Ingress objects associated with cluster nodes.
+// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+type IngressOptions struct {
+	// disabled controls if Ingress object creation is disabled.
+	// If not disabled, Ingress objects routing to backend are generated for each Scylla node.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/cluster_types.go
```diff
@@ -95,6 +95,54 @@ type ScyllaClusterSpec struct {
...
+// At least one backend must be specified.
+// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+type ExposeOptions struct {
+	// cql specifies expose options for CQL SSL backend.
+	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
+	// +optional
+	CQL *CQLExposeOptions `json:"cql,omitempty"`
+}
+
+type CQLExposeOptions struct {
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -0,0 +1,147 @@
...
+			sc.Spec.ExposeOptions.CQL.Ingress.Annotations,
+			sc.Spec.ExposeOptions.CQL.Ingress.IngressClassName,
+			&services.Items[0],
+			[]string{
+				fmt.Sprintf("%s.cql.%s", nodeHostIDs[0], sc.Spec.DNSDomains[0]),
+				fmt.Sprintf("%s.cql.%s", nodeHostIDs[0], sc.Spec.DNSDomains[1]),
+			},
+			"cql-ssl",
+		)
+	})
```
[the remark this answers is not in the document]
zimnx: Isn't that we take for granted it works? The same way we don't check whether readiness probes are sent with interval and delay we set. 
I can extend it with testing connectivity via Ingress, but maybe as a followup.

pkg/controller/scyllacluster/resource.go
```diff
@@ -625,6 +637,121 @@ func MakePodDisruptionBudget(c *scyllav1.ScyllaCluster) *v1beta1.PodDisruptionBu
...
+
+			case naming.ScyllaServiceTypeMember:
+				hostID, ok := service.Annotations[naming.HostIDAnnotation]
+				if !ok {
+					klog.V(4).Infof("Service %q is missing HostID annotation, postponing Ingress creation until it's available", naming.ObjRef(service))
+					continue
+				}
+
+				if len(hostID) == 0 {
+					klog.Warningf("Service %q has empty HostID annotation, postponing Ingress creation until it's fixed", naming.ObjRef(service))
```
tnozicka, line 683: as this is an unrecoverable error, likely caused by the issuer or a bug - we should make the user aware using an event

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -925,3 +934,415 @@ func TestStatefulSetForRack(t *testing.T) {
...
+	}
+
+	for i := range tt {
+		test := tt[i]
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			got := MakeIngresses(test.cluster, test.services)
+			if !apiequality.Semantic.DeepEqual(got, test.expectedIngresses) {
+				t.Errorf("expected and actual Ingress differ: %s", cmp.Diff(test.expectedIngresses, got))
```
tnozicka, line 1344:
```suggestion
				t.Errorf("expected and actual Ingresses differ: %s", cmp.Diff(test.expectedIngresses, got))
```

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -925,3 +934,415 @@ func TestStatefulSetForRack(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for i := range tt {
+		test := tt[i]
```
tnozicka, line 1338: nit: I see you've fixed the `tt` but this one was initially correct. Conventionally we do `tt` (test table) and `tc` (test case)

tnozicka (commented):
last round, the rest lgtm

pkg/controller/scyllacluster/resource.go
tnozicka: I guess our sidecar controller should always override it so as long as it work this would only be intermittent, so warning may be fine

zimnx: it's syncing what Scylla API is returining, so if label value doesn't match it's going to be overwritten

pkg/controller/scyllacluster/resource_test.go
zimnx: done

pkg/controller/scyllacluster/resource_test.go
zimnx: done

pkg/controller/scyllacluster/resource.go
tnozicka: If this is a nasty user playing with it, the controller shouldn't error. Also error will just make the case an infinite retry with backoff, yet the resource  won't result in a different codepath.

The option I see is keeping the warning, or sending an event. As it is supposed to be reconciled back by the sidecar, I'd lean towards the warning, sorry for the false alarm here.

zimnx: > Also error will just make the case an infinite retry with backoff, yet the resource won't result in a different codepath.

It won't be infinite, because sidecar controller doesn't use hash annotation to check whether resource match it's expectation, it will react on every service label change (user interrupting) and it will override and fix the value with what Scylla returns. 
Leaving warning would not propagate error to upper level where reconcilation status is calculated, and without one required resource, cluster should be marked as degraded. We don't mark that for other errors as well, but that's outside of scope of this PR.

tnozicka: Let's think about controllers alone, assume only the scyllacluster controller running and the sidecar one being broken. ScyllaCluster object has a broken annotation and the scyllacluster controller returns error. That means in a bit scyllacluster controller is gonna try again on exactly the same object, and again with the backoff policy... If someone fixes the annotation, you don't need the retries because you'll get requeued on changes.

zimnx: Without error, controller sync loop will succeed, properly working Ingress will be removed if it was created before interruption, and status will reflect that cluster is operational even when one of the required resources isn't created and user can't connect to all nodes in it. 

Brought back warning, even though I consider error a better choice overall here.

tnozicka: > Without error, controller sync loop will succeed, properly working Ingress will be removed if it was created before interruption

that's actually a good thing, the ingress no longer matches the desired state and is a relic of the past. whether it's still valid or not is just a guessing game - e.g. what if the hostid changed but there is none projected yet, so implicitly assuming it's still good may not be a good idea

tnozicka: and thanks ;)

tnozicka: and thanks for the update ;)

(the isolated case shows that a controller shouldn't requeue based on the state of the resource, only on external actions)

tnozicka (approved):
/approve
/lgtm

great job!

pkg/controller/scyllacluster/resource.go
zimnx: > is just a guessing game

Exactly, and controller guessed it's better to act (remove ingress) instead doing nothing because it ecountered undefined behavior. Imo it's wrong, because even when state is invalid, node might be functional and even though it cascades to application interruptions. 
Doing nothing slightly increases load on apiserver which isn't a big thing compared to possibly loosing access to precious (hopefully replicated) data.

tnozicka: well, I think it goes the other way :) if the spec doesn't say there should be a managed ingress and there is one, it shouldn't be there.

PR #995, opened 2022-06-30
Support tuning for minikube on cri-o

zimnx:
Path to cpuset file in cgroupfs is different than available ones
on minikube with cri-o. Added correct path to fallback list.


tnozicka (approved):
/approve
/lgtm

PR #1006, opened 2022-08-04
Make operator use new Scylla Manager client 

Choraden:
The mermaid client was outdated. This PR updates it to the newest version. 

- Scylla Manager swagger has been updated to this https://github.com/scylladb/scylla-manager/blob/9ad2186d6b6b922ac96ff7c08e45d4ba1b0dd31a/swagger/scylla-manager.json.
- Makefile for client generation has been created in mermaid client package. It follows Scylla Manager logic of generating clients https://github.com/scylladb/scylla-manager/blob/master/swagger/Makefile.  
- New client has been generated, unnecessary old client code removed, old script to generate client removed. 
- The operator has been adjusted to use new mermaid client.


pkg/mermaidclient/Makefile
```diff
@@ -0,0 +1,10 @@
+all: gen/mermaid
```
tnozicka, line 1: this needs to have a verify/update job in the top level makefile/CI - possibly just moved there

Choraden:
- Mermaid has been renamed to Scylla Manager 
- Rules to update and verify Scylla Manager client have been moved to top level Makefile
- Swagger has been imported to vendor files

pkg/dependencymagnet/doc.go
```diff
@@ -1,11 +1,13 @@
 // Copyright (C) 2021 ScyllaDB
 
+//go:build tools
 // +build tools
 
 // Force go mod to download and vendor code that isn't depended upon.
 package dependencymagnet
 
 import (
+	_ "github.com/go-swagger/go-swagger/cmd/swagger"
```
zimnx, line 10: This commit should be squashed with commit using it (adding make verify/update)

pkg/scyllamanagerclient/internal/client/mermaid_client.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	strfmt "github.com/go-openapi/strfmt"
 
-	"github.com/scylladb/scylla-operator/pkg/mermaidclient/internal/client/operations"
```
zimnx, line 14: This is autogenerated, move it to separate commit

Makefile
```diff
@@ -75,6 +75,10 @@ HELM_MANIFEST_CACHE_CONTROL ?=public, max-age=600
 CONTROLLER_GEN ?=$(GO) run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen --
 CRD_PATH ?= pkg/api/scylla/v1/scylla.scylladb.com_scyllaclusters.yaml
 
+SWAGGER := $(GO) run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger
+MANAGER_CLIENT_DIR := pkg/scyllamanagerclient/gen/
+MANAGER_CLIENT_SRC_DIR := pkg/scyllamanagerclient/scyllamanagerclient.json
```
zimnx, line 80: change `:=` to `?=` and remote space after it

Makefile
```diff
@@ -489,10 +493,25 @@ verify-links:
...
 .PHONY: verify-links
 
-verify: verify-gofmt verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint verify-links
+update-scylla-manager-client:
+	@rm -Rf $(MANAGER_CLIENT_DIR) && mkdir -p $(MANAGER_CLIENT_DIR)
+	$(SWAGGER) generate client -A scylla-manager -f $(MANAGER_CLIENT_SRC_DIR) -t $(MANAGER_CLIENT_DIR) &> /dev/null
+.PHONY: update-scylla-manager-client
+
+verify-scylla-manager-client: tmp_dir := $(shell mktemp -d)
+verify-scylla-manager-client: PROJECT := github.com/scylladb/scylla-operator
```
zimnx, line 502: There's already `GO_MODULE`

Makefile
```diff
@@ -489,10 +493,25 @@ verify-links:
...
+verify-scylla-manager-client: MANAGER_CLIENT_TMP_DIR := $(tmp_dir)/pkg/scyllamanagerclient/gen/
```
zimnx, line 503: isn't it `$(tmp_dir)/${MANAGER_CLIENT_DIR}`?

Makefile
```diff
@@ -489,10 +493,24 @@ verify-links:
 	fi;
 .PHONY: verify-links
 
-verify: verify-gofmt verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint verify-links
+update-scylla-manager-client:
+	@rm -Rf $(MANAGER_CLIENT_DIR) && mkdir -p $(MANAGER_CLIENT_DIR)
+	$(SWAGGER) generate client -A scylla-manager -f $(MANAGER_CLIENT_SRC_DIR) -t $(MANAGER_CLIENT_DIR) &> /dev/null
```
zimnx, line 498: nit: this use common values in two places, entire command could be extracted to macro accepting input and output dirs.

pkg/scyllamanagerclient/internal/client/mermaid_client.go
zimnx: this is not addressed

zimnx: it is, my mistake.

Makefile
Choraden: Extracted command to macro PTAL

Makefile
```diff
@@ -489,10 +493,31 @@ verify-links:
 	fi;
 .PHONY: verify-links
 
-verify: verify-gofmt verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint verify-links
+# $1 - client name
+# $2 - source file
+# $3 - client directory
+define generate-swagger-client
+	@rm -Rf $(3) && mkdir -p $(3)
+	$(SWAGGER) generate client -A $(1) -f $(2) -t $(3) &> /dev/null
```
zimnx, line 501: This is the common part, it shouldn't be parametrized

zimnx, line 501:
```suggestion
	$(SWAGGER) generate client -A scylla-manager -f $(2) -t $(3) &> /dev/null
```
This is the common part, it shouldn't be parametrized

Choraden: Right, fixed it.

Makefile
```diff
@@ -75,6 +75,10 @@ HELM_MANIFEST_CACHE_CONTROL ?=public, max-age=600
 CONTROLLER_GEN ?=$(GO) run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen --
 CRD_PATH ?= pkg/api/scylla/v1/scylla.scylladb.com_scyllaclusters.yaml
 
+SWAGGER ?=$(GO) run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger
+MANAGER_CLIENT_DIR ?=pkg/scyllamanagerclient/gen/
+MANAGER_CLIENT_SRC_DIR ?=pkg/scyllamanagerclient/scyllamanagerclient.json
```
zimnx, line 80:
```suggestion
MANAGER_CLIENT_SRC ?=pkg/scyllamanagerclient/scyllamanagerclient.json
```
It's not a directory

Makefile
```diff
@@ -75,6 +75,10 @@ HELM_MANIFEST_CACHE_CONTROL ?=public, max-age=600
 CONTROLLER_GEN ?=$(GO) run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen --
 CRD_PATH ?= pkg/api/scylla/v1/scylla.scylladb.com_scyllaclusters.yaml
 
+SWAGGER ?=$(GO) run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger
+MANAGER_CLIENT_DIR ?=pkg/scyllamanagerclient/gen/
```
zimnx, line 79:
```suggestion
MANAGER_CLIENT_GEN_DIR ?=pkg/scyllamanagerclient/gen/
```
It's not where manager client lives, it's where generated part is

Choraden:
Changed names as suggested.

zimnx (dismissed):
LGTM, thanks

tnozicka (commented):
a few comments. please also grep for `mermaid`

Makefile
```diff
@@ -489,10 +493,30 @@ verify-links:
 	fi;
 .PHONY: verify-links
 
-verify: verify-gofmt verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint verify-links
+# $1 - source file
+# $2 - client directory
+define generate-scylla-manager-client
+	@rm -Rf $(2) && mkdir -p $(2)
+	$(SWAGGER) generate client -A scylla-manager -f $(1) -t $(2) &> /dev/null
```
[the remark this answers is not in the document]
tnozicka: `SWAGGER_QUIET :=--quiet` would be an easy way to control it but I think I'd by fine with verbose output as well

Makefile
```diff
@@ -75,6 +75,10 @@ HELM_MANIFEST_CACHE_CONTROL ?=public, max-age=600
 CONTROLLER_GEN ?=$(GO) run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen --
 CRD_PATH ?= pkg/api/scylla/v1/scylla.scylladb.com_scyllaclusters.yaml
 
+SWAGGER ?=$(GO) run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger
+MANAGER_CLIENT_GEN_DIR ?=pkg/scyllamanagerclient/gen/
```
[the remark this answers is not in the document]
zimnx: consumers are using `/pkg/scyllamanagerclient/` where hand-crafted version of wrapped autogenerated client is present. Although splitting it into more packages as you suggested would be beneficial.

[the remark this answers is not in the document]
tnozicka: I though they need the types but loos like that has a manual wrapper

Makefile
```diff
@@ -489,10 +493,30 @@ verify-links:
...
+endef
+
+update-scylla-manager-client:
+	$(call generate-scylla-manager-client,${MANAGER_CLIENT_SRC},${MANAGER_CLIENT_GEN_DIR})
+.PHONY: update-scylla-manager-client
+
+verify-scylla-manager-client: tmp_dir := $(shell mktemp -d)
+verify-scylla-manager-client: MANAGER_CLIENT_TMP_DIR := $(tmp_dir)/${MANAGER_CLIENT_GEN_DIR}
+verify-scylla-manager-client:
+	cd $(tmp_dir) && $(GO) mod init $(GO_MODULE)
```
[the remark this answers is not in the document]
Choraden: Swagger target must reside inside a location in the $GOPATH/src

Makefile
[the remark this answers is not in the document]
Choraden: I did a restructure.

Choraden:
@tnozicka I've addressed your comments PTAL

tnozicka:
It's on my list for Monday, apologies for the delay. (if you we not to get a review pls ping me on Tue)

Makefile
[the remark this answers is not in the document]
tnozicka: I am not sure how this command ensures it's in the gopath but assume you are in the gopath or set the gopath for a particular command. looks at the others, generators e.g. need to be in the gopath as well

Makefile
```diff
@@ -75,6 +75,11 @@ HELM_MANIFEST_CACHE_CONTROL ?=public, max-age=600
 CONTROLLER_GEN ?=$(GO) run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen --
 CRD_PATH ?= pkg/api/scylla/v1/scylla.scylladb.com_scyllaclusters.yaml
 
+SWAGGER ?=./vendor/github.com/go-swagger/go-swagger/cmd/swagger
```
tnozicka, line 78: swagger command is executed as `$(GO) run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger` (this is a swagger cmd dir)

tnozicka:
```suggestion
SWAGGER_CMD_PKG ?=./vendor/github.com/go-swagger/go-swagger/cmd/swagger
```

Makefile
```diff
@@ -489,10 +494,31 @@ verify-links:
 	fi;
 .PHONY: verify-links
 
-verify: verify-gofmt verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-examples verify-govet verify-helm-lint verify-links
+# $1 - source file
+# $2 - client directory
+# $3 - client name
+define generate-swagger-client
+	GOPATH=$(GOPATH) $(GO) run "$(SWAGGER)" generate client --name "$(3)" --spec "$(1)" --target "$(2)"
```
tnozicka, line 501: should be simple quotes, you don't want the substitutions to evaluate bash vars and make vars are evaluated before
(applies globally)

Makefile
```diff
@@ -489,10 +494,31 @@ verify-links:
...
+
+endef
+
+update-manager-client:
+	$(call generate-swagger-client,${MANAGER_CLIENT_SPEC},${MANAGER_CLIENT_GEN_DIR},'manager')
```
tnozicka, line 506: `'manager'` ends up double quoted this way

Makefile
```diff
@@ -489,10 +494,31 @@ verify-links:
...
+.PHONY: update-manager-client
+
+verify-manager-client: tmp_dir := $(shell mktemp -d)
+verify-manager-client: MANAGER_CLIENT_TMP_DIR := "$(tmp_dir)/${MANAGER_CLIENT_GEN_DIR}"
```
tnozicka, line 510: this ends up double quoted as well

tnozicka: nit: i'd drop the extra variable as using it explicitelly makeses sure e..g you are diffing the same folder with just different prefixes

Makefile
```diff
@@ -489,10 +494,31 @@ verify-links:
...
+verify-manager-client:
+	cd $(tmp_dir) && $(GO) mod init $(GO_MODULE) && mkdir -p "${MANAGER_CLIENT_GEN_DIR}"
```
tnozicka, line 512:
```suggestion
mkdir -p "$(tmp_dir)/$(MANAGER_CLIENT_GEN_DIR)"
cd $(tmp_dir) && $(GO) mod init $(GO_MODULE)
```
`-p` is error prone, this way it's at least more safely bounded to the tmp_dir if a substitution was to fail

tnozicka: also `"${MANAGER_CLIENT_GEN_DIR}"` should be substituted in make, not bash (bracket types)

(applies globally)

Makefile
```diff
@@ -489,10 +494,31 @@ verify-links:
...
+	$(call generate-swagger-client,${MANAGER_CLIENT_SPEC},${MANAGER_CLIENT_TMP_DIR},'manager')
+	$(diff) -r "$(MANAGER_CLIENT_GEN_DIR)" "$(MANAGER_CLIENT_TMP_DIR)" || (echo 'scylla-manager-client is not up-to date. Please run `make update-scylla-manager-client` to update it.' && false)
```
tnozicka, line 514: using the dirs explicitly will make the comparison clearer

pkg/dependencymagnet/doc.go
```diff
@@ -1,11 +1,13 @@
 // Copyright (C) 2021 ScyllaDB
 
+//go:build tools
```
tnozicka, line 3: do we need both of those tags?

pkg/controller/manager/sync_action.go
```diff
@@ -10,13 +10,15 @@ import (
...
+	"github.com/scylladb/scylla-operator/pkg/managerclient"
+	"github.com/scylladb/scylla-operator/pkg/managerclient/model"
+	"github.com/scylladb/scylla-operator/pkg/managerclient/util"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/util/uuid"
 )
 
 type state struct {
-	Clusters    []*mermaidclient.Cluster
+	Clusters    []*model.Cluster
```
tnozicka, line 21: Should we keep the manual types at the same level / in the same package as the manual client. @zimnx wdyt?

pkg/managerclient/client.go
```diff
@@ -1,6 +1,6 @@
 // Copyright (C) 2017 ScyllaDB
```
tnozicka, line 1: This commit has grown pretty big. Would you mind splitting it into
- changes to the client
- manual updates for the call sites
?

Makefile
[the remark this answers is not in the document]
Choraden: As far as I know, swagger does lookups in the gopath for guessing imports an so on.
Here you can see the output when I don't init gomodule
```
make verify-manager-client
mkdir -p "/tmp/tmp.EXIla2kyk5/pkg/managerclient/gen/"
#cd /tmp/tmp.EXIla2kyk5 && go mod init github.com/scylladb/scylla-operator
GOPATH=/home/hubert/go go run './vendor/github.com/go-swagger/go-swagger/cmd/swagger' generate client --name 'manager' --spec 'pkg/managerclient/api/managerclient.json' --target '/tmp/tmp.EXIla2kyk5/pkg/managerclient/gen/'
2022/09/27 10:13:13 validating spec /home/hubert/Dev/scylla-operator/pkg/managerclient/api/managerclient.json
2022/09/27 10:13:13 preprocessing spec with option:  minimal flattening
2022/09/27 10:13:13 building a plan for generation
2022/09/27 10:13:13 generation target /tmp/tmp.EXIla2kyk5/pkg/managerclient/gen/
2022/09/27 10:13:13 target must reside inside a location in the $GOPATH/src or be a module
exit status 1
```

pkg/dependencymagnet/doc.go
Choraden: I think we don't need the old one.

pkg/mermaidclient/internal/client/mermaid_client.go
```diff
@@ -1,117 +0,0 @@
-// Code generated by go-swagger; DO NOT EDIT.
```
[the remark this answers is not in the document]
Choraden: All the files in `internal` dir were generated, thus I don't think there is anything interesting to diff.
Moreover, I didn't want to make it confusing as before
> this looks to be moved in the same commit as it's changed which is a bit confusing

I'd suggest comparing only the source files (it is in separated commit), as in both cases it is generated with the same tool.

pkg/controller/manager/sync_action.go
Choraden: I've changed the file structure so that managerclient package has all the necessary things.

Choraden:
@tnozicka ptal

tnozicka:

PR #1032, opened 2022-09-19
Unify WaitForState implementations

tnozicka:
It was very painful to add new WaitForState functions into our e2e as the code had to be duplicated and there were differences over time. This PR merges the implementation into a single generic function an unifies the interface for all types.

(This is needed for  #1021 that adds another WaitForState variant for secrets.)

### DIFF
--- a/test/e2e/utils/helpers.go
+++ b/test/e2e/utils/helpers.go
@@ -226,31 +226,31 @@ func WaitForObjectState[Object, ListObject runtime.Object](ctx context.Context, 
 	return event.Object.(Object), nil
 }
 
-func WaitForScyllaClusterState(ctx context.Context, client scyllav1client.ScyllaV1Interface, namespace string, name string, conditions ...func(*scyllav1.ScyllaCluster) (bool, error)) (*scyllav1.ScyllaCluster, error) {
-	return WaitForObjectState[*scyllav1.ScyllaCluster, *scyllav1.ScyllaClusterList](ctx, client.ScyllaClusters(namespace), name, WaitForStateOptions{}, conditions[0], conditions[1:]...)
+func WaitForScyllaClusterState(ctx context.Context, client scyllav1client.ScyllaV1Interface, namespace string, name string, options WaitForStateOptions, condition func(*scyllav1.ScyllaCluster) (bool, error), additionalConditions ...func(*scyllav1.ScyllaCluster) (bool, error)) (*scyllav1.ScyllaCluster, error) {
+	return WaitForObjectState[*scyllav1.ScyllaCluster, *scyllav1.ScyllaClusterList](ctx, client.ScyllaClusters(namespace), name, options, condition, additionalConditions...)
 }
 
-func WaitForPodState(ctx context.Context, client corev1client.PodInterface, name string, condition func(*corev1.Pod) (bool, error), options WaitForStateOptions) (*corev1.Pod, error) {
-	return WaitForObjectState[*corev1.Pod, *corev1.PodList](ctx, client, name, options, condition)
+func WaitForPodState(ctx context.Context, client corev1client.PodInterface, name string, options WaitForStateOptions, condition func(*corev1.Pod) (bool, error), additionalConditions ...func(*corev1.Pod) (bool, error)) (*corev1.Pod, error) {
+	return WaitForObjectState[*corev1.Pod, *corev1.PodList](ctx, client, name, options, condition, additionalConditions...)
 }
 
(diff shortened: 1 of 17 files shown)


test/e2e/utils/helpers.go
```diff
@@ -160,48 +161,16 @@ func IsScyllaClusterRolledOut(sc *scyllav1.ScyllaCluster) (bool, error) {
...
+	List(context.Context, metav1.ListOptions) (ListPtr, error)
+	Watch(context.Context, metav1.ListOptions) (watch.Interface, error)
 }
 
 type WaitForStateOptions struct {
 	TolerateDelete bool
 }
 
-func WaitForPodState(ctx context.Context, client corev1client.PodInterface, name string, condition func(sc *corev1.Pod) (bool, error), o WaitForStateOptions) (*corev1.Pod, error) {
+func WaitForObjectState[ObjectPtrType, ObjectListPtrType runtime.Object](ctx context.Context, client listerWatcher[ObjectListPtrType], name string, options WaitForStateOptions, condition func(obj ObjectPtrType) (bool, error), additionalConditions ...func(obj ObjectPtrType) (bool, error)) (ObjectPtrType, error) {
```
zimnx, line 173:
```suggestion
func WaitForObjectState[ObjectType, ObjectListType runtime.Object](ctx context.Context, client listerWatcher[ObjectListType], name string, options WaitForStateOptions, condition func(obj ObjectType) (bool, error), additionalConditions ...func(obj ObjectType) (bool, error)) (ObjectType, error) {
```

test/e2e/utils/helpers.go
```diff
@@ -360,77 +205,54 @@ func WaitForNodeConfigState(ctx context.Context, ncClient scyllav1alpha1client.N
 		return allDone, nil
 	}
 
-	event, err := watchtools.UntilWithSync(ctx, lw, &scyllav1alpha1.NodeConfig{}, nil, func(event watch.Event) (bool, error) {
+	event, err := watchtools.UntilWithSync(ctx, lw, reflect.New(reflect.TypeOf(*new(ObjectPtrType)).Elem()).Interface().(runtime.Object), nil, func(event watch.Event) (bool, error) {
```
zimnx, line 208:
```suggestion
	event, err := watchtools.UntilWithSync(ctx, lw, *new(ObjectPtrType), nil, func(event watch.Event) (bool, error) {
```

test/e2e/utils/helpers.go
tnozicka: It's on purpose as it isn't the ObjectType but a pointer to it (ObjectType is `corev1.Pod` but ObjectPtrType is `*corev1.Pod`)

test/e2e/utils/helpers.go
tnozicka: to many iterations, you are right that the simple one works ;)

tnozicka: fixed

test/e2e/utils/helpers.go
tnozicka: I really wanted to get the non-pointer types as arguments but without the constraints, golang didn't let me cast the interfaces later on. Good old c++ templates...

zimnx: > It's on purpose as it isn't the ObjectType but a pointer to it (ObjectType is corev1.Pod but ObjectPtrType is *corev1.Pod)

Name doesn't infer any condition on type, constraint is to implement runtime.Object, not being a pointer. It's just that `corev1.Pod` doesn't implement it, but `*corev1.Pod` does.

zimnx: `Type` seems redundant in this scope, so I would be fine with both `WaitForObjectState[Object, ObjectList runtime.Object]` and suggested one.

tnozicka: done

test/e2e/utils/helpers.go
```diff
@@ -160,48 +160,16 @@ func IsScyllaClusterRolledOut(sc *scyllav1.ScyllaCluster) (bool, error) {
...
-		default:
-			return true, fmt.Errorf("unexpected event: %#v", event)
-		}
-	})
-	if err != nil {
-		return nil, err
-	}
-
-	return event.Object.(*scyllav1.ScyllaCluster), nil
+type listerWatcher[ListPtr runtime.Object] interface {
```
zimnx, line 163:
```suggestion
type listerWatcher[ObjectList runtime.Object] interface {
```
for consistency

tnozicka: made it `ListObject`. (`ObjectList` wouldn't be a pointer as Lists are called the actual non-ptr types, not the interfaces)

zimnx (approved):
LGTM, very nice improvement, thanks!

PR #1044, opened 2022-09-23
[WIP] Multi region informers

zimnx:
Do not merge.

It's better to checkout the branch and browse code in IDE than review the diff, especially for the example usage in controller.

Proposal for multiregion client/informers/listers autogenerated code. 
It's based on existing local informers with extension layer allowing to dynamically select parciular datacenter based on a key. 
In addition controller is added which is dynamically updating/removing informers based on ScyllaCluster object keeping the credentials and DC informations. 

Please ignore nits in controllers, code is still in progress - focus on pkg/multiregionclient and usage of it.

I want to pass the review, before I start working on actual autogeneration of it for kube and crd objects.


zimnx:
No feedback received, closing in favor of upcoming PR.

PR #1048, opened 2022-09-29
Disable shadawareness E2E giving false positives

zimnx:
Both on CI and on local environments traffic between test binary and
Scylla Pods is NATed. This test give false positives
because we were checking whether driver attempted to connect to shard aware port,
but it wasn't checking whether it acutally succeeded.
It wasn't spotted because message about wrong shard assignment is
printed out by the driver to stdout and no error is reported anywhere.
Gocql doesn't expose any way of checking if shardaware ports are used
successfully.

False positive is fixed by #1038 but it requires client-node communication using 
PodIPs which we don't currenly have.


tnozicka (approved):
/approve
/lgtm

thanks!

PR #1049, opened 2022-09-29
Avoid watch cache when using UntilWithSync

tnozicka:
`UntilWithSync` is based on informers for controllers that use `LIST` call with `RV="0"` that can use watch cache on the apiserver. (Contrary to `RV=""` that avoids watch cache.) That way it can receive older events then the current state. Like when a test verifies that a ScyllaCluster has been modified, it can still receive an older version of that object and think it is immediately rolled out.

In the ListerWatcher we'll replace any list call to `RV="0"` with `RV=""` to avoid watch cache.

A lot of flakes in different forms.

### DIFF
--- a/pkg/cmd/operator/sidecar.go
+++ b/pkg/cmd/operator/sidecar.go
@@ -14,6 +14,7 @@ import (
 	sidecarcontroller "github.com/scylladb/scylla-operator/pkg/controller/sidecar"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/scylladb/scylla-operator/pkg/internalapi"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/sidecar"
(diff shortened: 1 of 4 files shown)


tnozicka:
I guess I should fix the other usages as well, although sometime the objects there are new. On it.

tnozicka:
fixed other usages as well

pkg/cmd/operator/sidecar.go
```diff
@@ -197,10 +198,10 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 	// Wait for the service that holds identity for this scylla node.
 	serviceFieldSelector := fields.OneTermEqualSelector("metadata.name", o.ServiceName)
 	serviceLW := &cache.ListWatch{
-		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+		ListFunc: helpers.UncachedListFunc(func(options metav1.ListOptions) (runtime.Object, error) {
```
zimnx, line 201: Is there any reason why skipping cache is required in sidecar? We are not patching here anything and right after waiting until change is rolled out, so i guess cache shouldn't be an issue.

tnozicka: in theory it could see any arbitrarily old object. although unlikely it could be a previous scyllacluster with different UUID

zimnx (approved):
LGTM, thanks

PR #1051, opened 2022-09-30
Allow overriding Scylla command line args

PenguinToast:

Allow `scyllaArgs` to replace existing Scylla command line arguments. This is necessary to override the hardcoded `0.0.0.0` listen address to `::` when we want to use IPv6.

Resolves #


tnozicka:
could we just always use `::` to listen to both stacks?

PenguinToast:
> could we just always use `::` to listen to both stacks?

I believe there are systems where listening on `::` doesn't also listen on `0.0.0.0`, so just listening on `::` for those would break IPv4 communication

PenguinToast:
@tnozicka let me know what you think ^

PR #1064, opened 2022-10-12
Tee test logs of e2e test runner's parallel processes to files

rzetelskik:
**Description of your changes:** Currently, in our e2e test runner, when one of the spawned processes fails before the Ginkgo reporter can fetch the intercepted output, we won't be able to access the test logs. Although the test runner prints the logs collected in a buffer of each failed process (https://github.com/scylladb/scylla-operator/blob/f33a54a2c30f55a93934da397cc94a27df338ddb/pkg/cmd/tests/tests_run.go#L428), they don't contain the intercepted test logs (see https://github.com/onsi/ginkgo/blob/2cc2219b4ec8630df2ee1b112237b4ae56da1a75/core_dsl.go#L271).

This PR adds redirection of the test logs of each process to a file. Additionally, it reformats the logs of the test runner and fixes a wait bug. This effort is supposed to facilitate tracing issues where some of the parallel processes of the E2E test runner fail (e.g. get signaled).

CI issues: https://github.com/scylladb/scylla-operator/actions/runs/3203718127/jobs/5234200661#step:3:1925, https://github.com/scylladb/scylla-operator/actions/runs/3205551747/jobs/5238266328#step:3:1983

### DIFF
--- a/pkg/cmd/tests/tests_run.go
+++ b/pkg/cmd/tests/tests_run.go
@@ -322,6 +322,29 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
 		if suiteConfig.ParallelTotal > 1 {
 			suiteConfig.ParallelHost = o.ParallelServerAddress
 			suiteConfig.ParallelProcess = o.ParallelShard
+
+			ginkgo.BeforeSuite(func() {
+				if len(o.ArtifactsDir) < 1 {
+					return
+				}
+
+				d := path.Join(o.ArtifactsDir, "e2e-parallel-processes-logs")
+				err := os.Mkdir(d, 0777)
+				if err != nil && !os.IsExist(err) {
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				}
+
+				f, err := os.Create(path.Join(d, fmt.Sprintf("%d.log", o.ParallelShard)))
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				ginkgo.GinkgoWriter.TeeTo(f)
+
+				ginkgo.DeferCleanup(func() {
+					ginkgo.GinkgoWriter.ClearTeeWriters()
+					// Ignoring the error as there's nowhere to log it.
+					_ = f.Close()
+				})
+			})
 		}
 
 		klog.InfoS("Running specs")
--- a/pkg/cmd/tests/tests_run.go
+++ b/pkg/cmd/tests/tests_run.go
@@ -379,7 +379,7 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
 		if err != nil {
 			errs = append(errs, fmt.Errorf("can't start command %q with args %v: %w", e.cmd.Path, e.cmd.Args, err))
 		}
-		klog.V(2).InfoS("Started process", "ID", e.id, "Process", e.cmd.String())
+		klog.V(2).InfoS("Started process", "ID", e.id, "Command", e.cmd.String())
 		// RegisterAlive needs to be set so ginkgo worker #1 can detect all the other workers are finished
 		// and start serial tests. It needs cmd.Wait to be called first to pick up the state.
 		server.RegisterAlive(e.id, func() bool { return e.cmd.ProcessState == nil || !e.cmd.ProcessState.Exited() })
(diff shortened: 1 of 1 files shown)


pkg/cmd/tests/tests_run.go
```diff
@@ -318,6 +318,28 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+			ginkgo.BeforeSuite(func() {
+				if len(o.ArtifactsDir) < 1 {
+					return
+				}
+
+				d := path.Join(o.ArtifactsDir, "e2e-parallel-processes-logs")
+				err := os.Mkdir(d, 0777)
+				if err != nil && !os.IsExist(err) {
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				}
```
zimnx, line 331: Shouldn't it be created by the root process? Otherwise all parallel processes will try to create it and all that didn't succedded will start failing because of exists error.

pkg/cmd/tests/tests_run.go
```diff
@@ -318,6 +318,28 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+
+				f, err := os.OpenFile(path.Join(d, fmt.Sprintf("%d.log", o.ParallelShard)), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
```
zimnx, line 333: os.Create

pkg/cmd/tests/tests_run.go
rzetelskik: Hence the check for IsExist error, so that all processes other than the first one to actually create it would just go past. It could be taken out of the `BeforeSuite` and created by the root process, I just thought this was a bit cleaner, even if not the most performant. Also that's how it's handled in the existing framework: https://github.com/scylladb/scylla-operator/blob/97c831eeb5854e58bc10500aa24c9fbb3c39998a/test/e2e/framework/framework.go#L300

zimnx: oh right, I didn't spot negation, sorry. Looks good.

pkg/cmd/tests/tests_run.go
rzetelskik: done

zimnx (approved):
LGTM, thanks

pkg/cmd/tests/tests_run.go
```diff
@@ -322,6 +322,29 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+				}
+
+				f, err := os.Create(path.Join(d, fmt.Sprintf("%d.log", o.ParallelShard)))
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				ginkgo.GinkgoWriter.TeeTo(f)
+
+				ginkgo.DeferCleanup(func() {
+					ginkgo.GinkgoWriter.ClearTeeWriters()
+					// Ignoring the error as there's nowhere to log it.
```
tnozicka, line 344: why not stderr?

pkg/cmd/tests/tests_run.go
```diff
@@ -322,6 +322,29 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+					return
+				}
+
+				d := path.Join(o.ArtifactsDir, "e2e-parallel-processes-logs")
+				err := os.Mkdir(d, 0777)
+				if err != nil && !os.IsExist(err) {
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				}
+
+				f, err := os.Create(path.Join(d, fmt.Sprintf("%d.log", o.ParallelShard)))
```
tnozicka, line 337: nit `suiteConfig.ParallelProcess` (late binding)

pkg/cmd/tests/tests_run.go
```diff
@@ -400,21 +423,22 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+			klog.V(2).InfoS("Process finished", "ID", e.id, "Command", e.cmd.String(), "ProcessState", e.cmd.ProcessState.String())
 		}()
 	}
+
+	wg.Wait()
+
 	err = apierrors.NewAggregate(errs)
 	if err != nil {
-		return err
+		klog.ErrorS(err, "Processes encountered errors")
```
tnozicka, line 439: why is it no longer returning here?

pkg/cmd/tests/tests_run.go
rzetelskik: It's intercepted so after calling `ClearTeeWriters` it won't be accessible

pkg/cmd/tests/tests_run.go
rzetelskik: I figured that wasn't intentional to return here? Otherwise the runner never reaches logging the output of failed processes: https://github.com/rzetelskik/scylla-operator/blob/11db76fb389ed12d654133b061f4c7d00a9de50e/pkg/cmd/tests/tests_run.go#L433

tnozicka: You can't ignore errors like `errors.New("exec: not started")` - those have to be manifested into our return code through errors. At the same time you are right about logs - this may be a good place to output them

rzetelskik: https://github.com/scylladb/scylla-operator/pull/1084

PR #1069, opened 2022-10-14
Adjust rate limits and timeouts for e2e tests

tnozicka:
- raises scyllacluster sync timeout
- raises rate limits for the operator when deployed to run e2e
- removes maxSyncDuration that has served it's purpose when tracing down long lived syncs but doesn't work well with the rate limitter

Some flakes.

Extracted from #1021

### DIFF
--- a/pkg/controller/scyllacluster/controller.go
+++ b/pkg/controller/scyllacluster/controller.go
@@ -44,9 +44,6 @@ import (
 
 const (
 	ControllerName = "ScyllaClusterController"
-	// maxSyncDuration enforces preemption. Do not raise the value! Controllers shouldn't actively wait,
-	// but rather use the queue.
-	maxSyncDuration = 40 * time.Second
 
 	artificialDelayForCachesToCatchUp = 10 * time.Second
 )
@@ -206,8 +203,6 @@ func (scc *Controller) processNextItem(ctx context.Context) bool {
 	}
 	defer scc.queue.Done(key)
 
-	ctx, cancel := context.WithTimeout(ctx, maxSyncDuration)
-	defer cancel()
 	err := scc.sync(ctx, key.(string))
 	// TODO: Do smarter filtering then just Reduce to handle cases like 2 conflict errors.
 	err = utilerrors.Reduce(err)
--- a/test/e2e/utils/config.go
+++ b/test/e2e/utils/config.go
@@ -5,7 +5,11 @@ package utils
 import "time"
 
 const (
-	baseRolloutTimout  = 30 * time.Second
+	// syncTimeout is the maximum time the sync loop can last. In normal circumstances this is in the order
+	// of seconds but there are special cases we need to account for. Like when the sync loop generates new keys
+	// and signs certificates it can take several seconds. When the CI creates multiple clusters in parallel on
+	// a constrained CPU, one cert can easily take over 30s.
+	syncTimeout        = 2 * time.Minute
 	imagePullTimeout   = 4 * time.Minute
 	joinClusterTimeout = 3 * time.Minute
 
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -42,6 +42,8 @@ for f in $( find "${deploy_dir}"/ -type f -name '*.yaml' ); do
     sed -i -E -e "s~docker.io/scylladb/scylla-operator(:|@sha256:)[^ ]*~${OPERATOR_IMAGE_REF}~" "${f}"
 done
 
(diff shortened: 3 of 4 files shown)


pkg/controller/scyllacluster/controller.go
```diff
@@ -44,9 +44,6 @@ import (
 
 const (
 	ControllerName = "ScyllaClusterController"
-	// maxSyncDuration enforces preemption. Do not raise the value! Controllers shouldn't actively wait,
-	// but rather use the queue.
-	maxSyncDuration = 40 * time.Second
```
zimnx, line 49: Would you mind elaborating with why this helps? 
I'm afraid removing time constraint on sync execution may lead to worker starvation in case when not time restricted request is sent to for example hanging scylla.
Setting it to higher value could prevent from starvation and maybe solve the issue you're solving.

tnozicka: The shared kube client has a rate limiter that slows downs your calls. When you hit the system hard enough, that can intentionally case high delays. Forcefully canceling the call won't help because the other loop will hit the same one (possibly in the next exponential step).

zimnx: I see. Yet this will cause the issue of worker pool exhaustion - I used wrong word in previous comment -  when running sync without any time constraint.

test/e2e/utils/config.go
```diff
@@ -5,7 +5,11 @@ package utils
 import "time"
 
 const (
-	baseRolloutTimout  = 30 * time.Second
+	// SyncTimeout is the maximum time the sync loop can last. In normal circumstances this is in the order
+	// of seconds but there are special cases we need to account for. Like when the sync loop generates new keys
+	// and signs certificates it can take several seconds. When the CI creates multiple clusters in parallel on
+	// a constrained CPU, one cert can easily take over 30s.
+	SyncTimeout        = 2 * time.Minute
```
zimnx, line 12:
```suggestion
	syncTimeout        = 2 * time.Minute
```

tnozicka: done

pkg/controller/scyllacluster/controller.go
tnozicka: If you are concerned about individual external calls taking too long (not the initial / intended purpose of what's being removed) I think that would be better handled by localized context for that particular call. Generally, it is being handled by the client timeout for any call made so it doesn't need extra context. At least for the kube client. The global timeout is not great for this because the number and types of calls can greatly differ in different execution branches or cluster sizes.

zimnx: We use two types of client in our reconcilation logic:
* scyllaclient - this one has a hardcoded timeout of 15s : https://github.com/scylladb/scylla-operator/blob/5897038e97adc5a6defd1b688dd22bc0c42402bf/pkg/scyllaclient/config.go#L56
* kube - we don't set any timeout explictly: https://github.com/scylladb/scylla-operator/blob/ecb4d96347eb22286994bf835e35195268666b3f/pkg/genericclioptions/genericclioptions.go#L64

I did a quick experiment by adding a few prints around where kube clients are initialized in options Complete():
```
	fmt.Println("proto", o.ProtoConfig.Timeout)
	fmt.Println("rest", o.RestConfig.Timeout)

	o.kubeClient, err = kubernetes.NewForConfig(o.ProtoConfig)
	if err != nil {
		return fmt.Errorf("can't build kubernetes clientset: %w", err)
	}

	fmt.Println("kube", o.kubeClient.CoreV1().RESTClient().(*rest.RESTClient).Client.Timeout)
	o.scyllaClient, err = scyllaversionedclient.NewForConfig(o.RestConfig)
	if err != nil {
		return fmt.Errorf("can't build scylla clientset: %w", err)
	}
	fmt.Println("scylla", o.scyllaClient.ScyllaV1().RESTClient().(*rest.RESTClient).Client.Timeout)
```
Both result from running it using kubeconfig and incluster are:
```
proto 0s
rest 0s
kube 0s
scylla 0s
```
So it looks like all timeouts aren't set, which means they are infinite. 
https://github.com/scylladb/scylla-operator/blob/b1cd524ea97f6dec5c61cb2315b5b716505f75d3/vendor/k8s.io/client-go/rest/config.go#L131-L132

tnozicka: In some paths they are defaulted but you are correct we should set them. I just think it's a separate / complementary effort to the change being done in this PR, as global timeout was added to avoid people doing active waiting, not to handle particular calls (where it wouldn't work well, given the number of calls differs by an execution branch and the cluster size)

Although, you are right that this is something we should look into (separately). I made https://github.com/scylladb/scylla-operator/issues/1073 to track it.

zimnx (approved):

PR #1070, opened 2022-10-14
Add Status Conditions

tnozicka:
This PR adds 3 main conditions that help users determine the ScyllaCluster status.
 - `Available` (whether all replicas are up)
 - `Progressing` (whether the controller is making changes to reach the desired state)
 - `Degraded` (whether the controller hit any issues)

A rollout is done when `Available==True`, `Progressing==False`, `Degraded==False`. This gives user an easy way to interpret the status and possibly use upstream tooling, like `kubectl wait --for=condition=`.

Conditions also allow determining the processing state and the errors manifested into degraded state without consulting operator logs (which only admins can do).

This is also needed for #1021 where it allows interpreting state for multiple controller that consume it's outputs.

### DIFF
--- a/pkg/controller/scyllacluster/conditions.go
+++ b/pkg/controller/scyllacluster/conditions.go
@@ -1,0 +1,19 @@
+package scyllacluster
+
+const (
+	serviceAccountControllerProgressingCondition = "ServiceAccountControllerProgressing"
+	serviceAccountControllerDegradedCondition    = "ServiceAccountControllerDegraded"
+	roleBindingControllerProgressingCondition    = "RoleBindingControllerProgressing"
+	roleBindingControllerDegradedCondition       = "RoleBindingControllerDegraded"
+	agentTokenControllerProgressingCondition     = "AgentTokenControllerProgressing"
+	agentTokenControllerDegradedCondition        = "AgentTokenControllerDegraded"
+	statefulSetControllerAvailableCondition      = "StatefulSetControllerAvailable"
+	statefulSetControllerProgressingCondition    = "StatefulSetControllerProgressing"
+	statefulSetControllerDegradedCondition       = "StatefulSetControllerDegraded"
+	serviceControllerProgressingCondition        = "ServiceControllerProgressing"
+	serviceControllerDegradedCondition           = "ServiceControllerDegraded"
+	pdbControllerProgressingCondition            = "PDBControllerProgressing"
+	pdbControllerDegradedCondition               = "PDBControllerDegraded"
+	ingressControllerProgressingCondition        = "IngressControllerProgressing"
+	ingressControllerDegradedCondition           = "IngressControllerDegraded"
+)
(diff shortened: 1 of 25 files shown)


test/e2e/set/scyllacluster/verify.go
```diff
@@ -83,9 +83,124 @@ func verifyPodDisruptionBudget(sc *scyllav1.ScyllaCluster, pdb *policyv1.PodDisr
...
+				condType: "PDBControllerDegraded",
+				status:   false,
+			},
+			{
+				condType: "IngressControllerProgressing",
+				status:   false,
+			},
+			{
+				condType: "IngressControllerDegraded",
+				status:   false,
```
zimnx, line 180: wouldn't it be better to use metav1.ConditionTrue/False here without need for convertion in loop below?

zimnx (commented):
two nits, lgtm once resolved.

pkg/controller/scyllacluster/conditions.go
```diff
@@ -0,0 +1,21 @@
...
+
+const (
+	serviceAccountControllerProgressingCondition = "ServiceAccountControllerProgressing"
+	serviceAccountControllerDegradedCondition    = "ServiceAccountControllerDegraded"
+	roleBindingControllerProgressingCondition    = "RoleBindingControllerProgressing"
+	roleBindingControllerDegradedCondition       = "RoleBindingControllerDegraded"
+	agentTokenControllerProgressingCondition     = "AgentTokenControllerProgressing"
+	agentTokenControllerDegradedCondition        = "AgentTokenControllerDegraded"
+	certControllerProgressingCondition           = "CertControllerProgressing"
+	certControllerDegradedCondition              = "CertControllerDegraded"
```
zimnx, line 11: nit: these should belong to other PR

pkg/controllerhelpers/core.go
```diff
@@ -31,6 +37,137 @@ func IsPodReady(pod *corev1.Pod) bool {
...
+
+	for _, c := range conditions {
+		reasons = append(reasons, fmt.Sprintf("%s_%s", c.Type, c.Reason))
+
+		for _, line := range strings.Split(c.Message, "\n") {
+			messages = append(messages, fmt.Sprintf("%s: %s", c.Type, line))
+		}
+	}
+
+	// TODO: Strip aggregated reasons and messages to fit into API limits.
```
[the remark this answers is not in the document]
tnozicka: https://github.com/scylladb/scylla-operator/issues/1075

test/e2e/set/scyllacluster/verify.go
tnozicka: done

pkg/controller/scyllacluster/conditions.go
tnozicka: done

zimnx (approved):
LGTM, thanks

PR #1074, opened 2022-10-19
[WIP] Create scylla connection bundle

tnozicka:
TODO

Resolves #


tnozicka:
it creates a bundle, I haven't tried using it as I don't have a cluster at hand but if anyone tries, let me know :)

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -0,0 +1,306 @@
...
+	secrets map[string]*corev1.Secret,
+	configMaps map[string]*corev1.ConfigMap,
+) (*corev1.Secret, error) {
+	clientCertSecretName := naming.GetScyllaClusterLocalUserAdminCertName(sc.Name)
+	clientCertSecret, found := secrets[clientCertSecretName]
+	if !found {
+		return nil, fmt.Errorf("secret %q doesn't exist or is not own by this object", naming.ManualRef(sc.Namespace, clientCertSecretName))
+	}
+
+	clientCertsBytes, clientKeyBytes, err := okubecrypto.GetCertKeyDataFromSecret(clientCertSecret)
```
zimnx, line 41: it's the other way around, GetCertKey returns Key and Cert. Maybe name should be adjusted to the order of return values. 
Otherwise I tested the generated bundle and it works.

tnozicka: thanks

pkg/scylla/api/cqlclient/v1alpha1/types_connectionconfig.go
```diff
@@ -0,0 +1,124 @@
...
+	// Server is the address of the Scylla cluster. All requests are sent to this address.
+	// Example: https://hostname:port
+	Server string `json:"server"`
+
+	// nodeDomain is the domain suffix that is concatenated with the host_id of the node driver wants to connect to.
+	// The concatenated string is then set as TLS server name for this TLS connection. This value is only used
+	// for setting the ServerName in TLS ClientHello, never resolved or used as a server address.
+	// Example: host_id.<nodeDomain>
+	// Client uses this nodeDomain as bare TLS server name when reaching out to the load balanced (discovery) endpoint.
+	ServiceDomain string `json:"nodeDomain"`
```
zimnx, line 25: I like `NodeDomain` better and it's in sync with json representation.  It's in fact domain pointing to scylla nodes, either any or particular one via subdomain.

zimnx:
My Ingress e2e passes with this bundle.

pkg/scylla/api/cqlclient/v1alpha1/types_connectionconfig.go
tnozicka: I am not too happy about that name as it seems only like a subset of what it is used for:
- `TLS.ServerName :uuid-1.cql.sc-name.sc-namespace` (`cql.sc-name.sc-namespace` is the node domain)
- `TLS:ServerName: cql.sc-name.sc-namespace` (`cql.sc-name.sc-namespace` doesn't seem like a domain used for nodes but rather a service DNS name) - it used to be different when we had `any.` subdomain there.
but I'll keep my reasons for v1alpha1 and use the wire name here.

zimnx (approved):
LGTM, thanks

PR #1085, opened 2022-11-01
Don't use "any" prefix for LB CQL service

tnozicka:
The CQL load balanced service (`any.cql.nodeDomain`) doesn't logically belong at the same level as the `host_id` subdomains. Although highly unlikely, they could possibly conflict. Using `cql.nodeDomain` will avoid conflicts and put it on a different level. It will allow us to drop `tlsServerName` field from CQLClientConfig (aka "bundle") as it can be now covered by only one field, `nodeDomain` - to be eventually renamed to `clusterDomain`, `serviceDomain` or something more descriptive.

In the end, sending all traffic to `any.cql.nodeDomain` (with ServerName set to `host_id.cql.nodeDomain`) feels a bit weird. This gets fixed by using the service like `server: cql.nodeDomain` where all traffic goes to.

Default destination
```
server: cql.<sc_name>.<sc_namespace>.<router_domain>:9142
nodeDomain: cql.<sc_name>.<sc_namespace>.<router_domain>
```
Explicit destination
```
server: 172.16.42.42:9142
nodeDomain: cql.<sc_name>.<sc_namespace>.<router_domain>
```

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -715,7 +715,7 @@ func MakeIngresses(c *scyllav1.ScyllaCluster, services map[string]*corev1.Servic
 	type params struct {
 		ingressNameSuffix     string
 		portName              string
-		identitySubdomainFunc func(string) string
+		protocolSubdomainFunc func(string) string
 		memberSubdomainFunc   func(string, string) string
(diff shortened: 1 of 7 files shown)


zimnx (approved):
LGTM, thanks

zimnx (approved):

Issue #1031, opened 2022-09-19
Flake: Scylla Manager integration [It] should discover cluster and sync tasks


tnozicka:
https://github.com/scylladb/scylla-operator/actions/runs/3067025626/jobs/4954445399#step:12:441
https://github.com/scylladb/scylla-operator/suites/8323775595/artifacts/365955898
```
• [FAILED] [462.877 seconds]
Scylla Manager integration
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:18
  [It] should discover cluster and sync tasks
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:23

  Begin Captured StdOut/StdErr Output >>
    I0916 12:29:55.912346     159 framework/framework.go:249] "Namespace removed." Namespace="e2e-test-scyllacluster-7t8gg-mxsmm"
  << End Captured StdOut/StdErr Output

  Begin Captured GinkgoWriter Output >>
    [BeforeEach] Scylla Manager integration
      github.com/scylladb/scylla-operator/test/e2e/framework/framework.go:57
    STEP: Creating a new namespace 09/16/22 12:22:13.035
    Sep 16 12:22:13.121: INFO: Created namespace "e2e-test-scyllacluster-7t8gg-mxsmm".
    STEP: Waiting for ServiceAccount "e2e-user" in namespace "e2e-test-scyllacluster-7t8gg-mxsmm". 09/16/22 12:22:13.216
    STEP: Waiting for default ServiceAccount in namespace "e2e-test-scyllacluster-7t8gg-mxsmm". 09/16/22 12:22:15.18
    [It] should discover cluster and sync tasks
      github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:23
    STEP: Creating a ScyllaCluster 09/16/22 12:22:15.184
    STEP: Waiting for the ScyllaCluster to deploy 09/16/22 12:22:15.199
    Sep 16 12:24:25.070: INFO: ScyllaCluster e2e-test-scyllacluster-7t8gg-mxsmm/basic-sxvd8 (RV=3430) is rolled out
    STEP: Inserting data with RF=1 09/16/22 12:24:30.151
    STEP: Verifying the ScyllaCluster 09/16/22 12:24:30.729
[...]

PR #1087, opened 2022-11-04
Make sure manager controller reacts to SA token changes

Choraden:
This PR adds secret handlers to manager controller, so that it requeues when a sa token is created/updated. That will fix the controller getting into an extremely long backoff, which was caused by syncing token error.

Resolves https://github.com/scylladb/scylla-operator/issues/1031


tnozicka (approved):
/approve
/lgtm
thanks!

tnozicka (approved):

PR #1091, opened 2022-11-04
Use annotations to mark node replacement address instead of labels

PenguinToast:

In IPv6 clusters, node addresses contain `:`. This is not valid in K8S label values, so the operator is unable to replace a node in IPv6 clusters.

To fix this, instead of storing the replace address in a label on the service, store it in an annotation - K8S annotations do not have this restriction on their values.

This was tested by deploying the changes to an IPv6 cluster and verifying that node replacement works as intended.

Resolves #


tnozicka:

PR #1098, opened 2022-11-16
Fix repository paths so it can be vendored

tnozicka:
Currently the repository can't be vendored because special character in a path that's not even a go subpackage. This PR fixes it and adds a test to prevent it from happening again.

```
go mod tidy && go mod vendor
go: finding module for package github.com/scylladb/scylla-operator/pkg/scheme
go: downloading github.com/scylladb/scylla-operator v1.7.4
github.com/tnozicka/go-sni-proxy/cmd/sni-proxy imports
	github.com/scylladb/scylla-operator/pkg/scheme: create zip: hack/gke/systemd/mnt-raid\x2ddisks-disk0.mount: malformed file path "hack/gke/systemd/mnt-raid\\x2ddisks-disk0.mount": invalid char '\\'
```

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -73,7 +73,7 @@ jobs:
       run: make verify-deps --warn-undefined-variables
 
   verify-vendorability:
-    name: Verify scylla-operator repo can be vendored by someone else
+    name: Verify vendorability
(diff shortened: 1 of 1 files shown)


zimnx (approved):
LGTM, thanks!

PR #1099, opened 2022-11-16
Use explicitly created SA token in E2E test suite

rzetelskik:
In k8s 1.24, secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount (see [release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#no-really-you-must-read-this-before-you-upgrade)). This PR removes this dependency from our E2E test suite.

Resolves: running e2e test suite in k8s >= 1.24.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -24,13 +24,14 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
-	watchtools "k8s.io/client-go/tools/watch"
 	"k8s.io/klog/v2"
 )
 
 const (
-	ServiceAccountName        = "e2e-user"
-	serviceAccountWaitTimeout = 1 * time.Minute
+	ServiceAccountName                   = "e2e-user"
+	ServiceAccountTokenSecretName        = "e2e-user-token"
+	serviceAccountWaitTimeout            = 1 * time.Minute
+	serviceAccountTokenSecretWaitTimeout = 1 * time.Minute
 )
 
 type Framework struct {
@@ -193,39 +194,35 @@ func (f *Framework) setupNamespace(ctx context.Context) {
 	}, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	// Wait for user ServiceAccount.
-	By(fmt.Sprintf("Waiting for ServiceAccount %q in namespace %q.", userSA.Name, userSA.Namespace))
-	ctxUserSa, ctxUserSaCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountWaitTimeout)
-	defer ctxUserSaCancel()
-	userSA, err = WaitForServiceAccount(ctxUserSa, f.KubeAdminClient().CoreV1(), userSA.Namespace, userSA.Name)
+	// Create a service account token Secret for the user ServiceAccount.
+	userSATokenSecret, err := f.KubeAdminClient().CoreV1().Secrets(ns.Name).Create(ctx, &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: ServiceAccountTokenSecretName,
+			Annotations: map[string]string{
+				corev1.ServiceAccountNameKey: userSA.Name,
+			},
+		},
+		Type: corev1.SecretTypeServiceAccountToken,
+	}, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	// Create a restricted client using the default SA.
-	var token []byte
-	for _, secretName := range userSA.Secrets {
-		secret, err := f.KubeAdminClient().CoreV1().Secrets(userSA.Namespace).Get(ctx, secretName.Name, metav1.GetOptions{})
-		o.Expect(err).NotTo(o.HaveOccurred())
+	By("Waiting for service account token Secret %q in namespace %q.", userSATokenSecret.Name, userSATokenSecret.Namespace)
+	ctxUserSATokenSecret, ctxUserSATokenSecretCancel := context.WithTimeout(ctx, serviceAccountTokenSecretWaitTimeout)
+	defer ctxUserSATokenSecretCancel()
+	userSATokenSecret, err = WaitForServiceAccountTokenSecret(ctxUserSATokenSecret, f.KubeAdminClient().CoreV1(), userSATokenSecret.Namespace, userSATokenSecret.Name)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(userSATokenSecret.Data).To(o.HaveKey(corev1.ServiceAccountTokenKey))
 
-		if secret.Type == corev1.SecretTypeServiceAccountToken {
-			name := secret.Annotations[corev1.ServiceAccountNameKey]
-			uid := secret.Annotations[corev1.ServiceAccountUIDKey]
-			if name == userSA.Name && uid == string(userSA.UID) {
-				t, found := secret.Data[corev1.ServiceAccountTokenKey]
-				if found {
-					token = t
-					break
-				}
-			}
-		}
-	}
-	o.Expect(token).NotTo(o.HaveLen(0))
+	token := userSATokenSecret.Data[corev1.ServiceAccountTokenKey]
+	o.Expect(token).NotTo(o.BeEmpty())
 
+	// Create a restricted client using the user SA.
 	f.clientConfig = restclient.AnonymousClientConfig(f.AdminClientConfig())
 	f.clientConfig.BearerToken = string(token)
 
 	// Wait for default ServiceAccount.
-	By(fmt.Sprintf("Waiting for default ServiceAccount in namespace %q.", ns.Name))
-	ctxSa, ctxSaCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountWaitTimeout)
+	By("Waiting for default ServiceAccount in namespace %q.", ns.Name)
+	ctxSa, ctxSaCancel := context.WithTimeout(ctx, serviceAccountWaitTimeout)
 	defer ctxSaCancel()
 	_, err = WaitForServiceAccount(ctxSa, f.KubeAdminClient().CoreV1(), ns.Namespace, "default")
 	o.Expect(err).NotTo(o.HaveOccurred())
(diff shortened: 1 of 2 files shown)


rzetelskik:
Alternatively, instead of creating a TokenRequest, we could also create a non-expiring Secret API object as described here: https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets.

test/e2e/framework/framework.go
```diff
@@ -193,35 +195,18 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
 
-	// Wait for user ServiceAccount.
-	By(fmt.Sprintf("Waiting for ServiceAccount %q in namespace %q.", userSA.Name, userSA.Namespace))
-	ctxUserSa, ctxUserSaCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountWaitTimeout)
-	defer ctxUserSaCancel()
-	userSA, err = WaitForServiceAccount(ctxUserSa, f.KubeAdminClient().CoreV1(), userSA.Namespace, userSA.Name)
+	// Send a TokenRequest for user service account.
+	userTokenRequest, err := f.KubeAdminClient().CoreV1().ServiceAccounts(ns.Name).CreateToken(ctx, userSA.Name, &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{
+			ExpirationSeconds: pointer.Int64(int64(24 * time.Hour / time.Second)),
```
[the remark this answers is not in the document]
rzetelskik: What do you mean by that it's dangerous? Duration is an int64 nanosecond count, so there's nothing wrong with dividing it. It's not a valid Duration after that of course, but that's why the int64 cast. `func (d *Duration) Seconds() float64` does the same thing partially, except it produces a float64 and then you truncate it anyway.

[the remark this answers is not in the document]
zimnx: > It's not a valid Duration after that of course

That's the reason why it's dangerous, you get time.Duration which needs special handling.
There're tools in a library to do it right and devs should use them. These functions were introduced later than initial implementation, probably because people were doing it your way. I recall I debugged simillar code so now I'm extra careful about time.Duration division.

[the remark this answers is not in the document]
rzetelskik: Well, I agree that it's more bug proof to use the lib function. I've sent a fix.

zimnx (dismissed):

test/e2e/framework/framework.go
```diff
@@ -193,35 +196,18 @@ func (f *Framework) setupNamespace(ctx context.Context) {
 	}, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	// Wait for user ServiceAccount.
-	By(fmt.Sprintf("Waiting for ServiceAccount %q in namespace %q.", userSA.Name, userSA.Namespace))
-	ctxUserSa, ctxUserSaCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountWaitTimeout)
-	defer ctxUserSaCancel()
-	userSA, err = WaitForServiceAccount(ctxUserSa, f.KubeAdminClient().CoreV1(), userSA.Namespace, userSA.Name)
+	// Send a TokenRequest for user service account.
+	userTokenRequest, err := f.KubeAdminClient().CoreV1().ServiceAccounts(ns.Name).CreateToken(ctx, userSA.Name, &authenticationv1.TokenRequest{
```
tnozicka, line 200: I don't think TokenRequest is a good fit here. This is user account (long lived) and practically if I want to debug with the same token as CI I explicitly want it to be shared and stored in the Secret. Please create the token secret manually.

tnozicka (changes requested):

test/e2e/framework/framework.go
```diff
@@ -29,8 +29,10 @@ import (
 )
 
 const (
-	ServiceAccountName        = "e2e-user"
-	serviceAccountWaitTimeout = 1 * time.Minute
+	ServiceAccountName                   = "e2e-user"
+	ServiceAccountTokenSecretName        = "e2e-user-token-secret"
```
tnozicka, line 33: let's not suffix the type to the name, it's already a Secret

test/e2e/framework/framework.go
```diff
@@ -200,26 +202,28 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
-		}
-	}
-	o.Expect(token).NotTo(o.HaveLen(0))
+	By(fmt.Sprintf("Waiting for service account token Secret %q in namespace %q.", userSATokenSecret.Name, userSATokenSecret.Namespace))
+	ctxUserSATokenSecret, ctxUserSATokenSecretCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountTokenSecretWaitTimeout)
+	defer ctxUserSATokenSecretCancel()
+	userSATokenSecret, err = WaitForServiceAccountTokenSecret(ctxUserSATokenSecret, f.KubeAdminClient().CoreV1(), userSATokenSecret.Namespace, userSATokenSecret.Name)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	token := userSATokenSecret.Data[corev1.ServiceAccountTokenKey]
```
tnozicka, line 223: nit: you should check the map has the key

test/e2e/framework/util.go
```diff
@@ -65,6 +64,44 @@ func WaitForServiceAccount(ctx context.Context, c corev1client.CoreV1Interface,
...
+	event, err := watchtools.UntilWithSync(
+		ctx,
+		lw,
+		&corev1.Secret{},
+		nil,
+		func(e watch.Event) (bool, error) {
+			switch t := e.Type; t {
+			case watch.Added, watch.Modified:
+				s := e.Object.(*corev1.Secret)
+				return len(s.Data[corev1.ServiceAccountTokenKey]) > 0, nil
```
tnozicka, line 88: this could panic

test/e2e/framework/framework.go
rzetelskik: Looking up a non-existent key will return a zero value, in this case an empty array, and I'm asserting non-emptiness in the next line, so would that really make any difference? Unless you just mean it for the sake of explicitness?

test/e2e/framework/util.go
rzetelskik: Would there be any other reason than the interface conversion? I'm looking at existing code (https://github.com/scylladb/scylla-operator/blob/7e9ea2c21ee76299f5342419c37513a70eab556c/test/e2e/framework/util.go#L50) almost identical to this and I'm wondering whether this was missed before or if I'm missing something now?

tnozicka: the cast is fine but maps can be nil

test/e2e/framework/framework.go
tnozicka: If the next line fails and you are debugging an e2e test there are 2 possible ways how it could have failed. It either didn't have the key or the value is really empty. And you start wondering which it is while you could have just known if the previous line used `o.HaveKey()`

test/e2e/framework/util.go
rzetelskik: Fetching from nil map would still return a zero value though?

tnozicka: you are right, i though it will panic but that's only for writes

test/e2e/framework/framework.go
```diff
@@ -200,26 +202,29 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
-				t, found := secret.Data[corev1.ServiceAccountTokenKey]
-				if found {
-					token = t
-					break
-				}
-			}
-		}
-	}
-	o.Expect(token).NotTo(o.HaveLen(0))
+	By(fmt.Sprintf("Waiting for service account token Secret %q in namespace %q.", userSATokenSecret.Name, userSATokenSecret.Namespace))
```
zimnx, line 217: By() already takes a format string, there's no need for fmt.Sprintf()

test/e2e/framework/framework.go
```diff
@@ -200,26 +202,29 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
+	ctxUserSATokenSecret, ctxUserSATokenSecretCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountTokenSecretWaitTimeout)
```
zimnx, line 218: nit: context.WithTimeout, you don't need OptionalTimeout since it's hardcoded

test/e2e/framework/framework.go
```diff
@@ -200,26 +202,29 @@ func (f *Framework) setupNamespace(ctx context.Context) {
 	userSA, err = WaitForServiceAccount(ctxUserSa, f.KubeAdminClient().CoreV1(), userSA.Namespace, userSA.Name)
```
zimnx, line 202: Previously we waited until token was added in SA.Secrets, now we wait until secret is populated, so this watch is no longer needed.

zimnx (approved):
lgtm, thanks

test/e2e/framework/framework.go
```diff
@@ -193,39 +194,35 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
+		Type: corev1.SecretTypeServiceAccountToken,
+	}, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	// Create a restricted client using the default SA.
-	var token []byte
-	for _, secretName := range userSA.Secrets {
-		secret, err := f.KubeAdminClient().CoreV1().Secrets(userSA.Namespace).Get(ctx, secretName.Name, metav1.GetOptions{})
-		o.Expect(err).NotTo(o.HaveOccurred())
+	By("Waiting for service account token Secret %q in namespace %q.", userSATokenSecret.Name, userSATokenSecret.Namespace)
```
tnozicka, line 209: for consistency, the other `By` steps don't end with `.`

test/e2e/framework/framework.go
```diff
@@ -193,39 +194,35 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
+	o.Expect(token).NotTo(o.BeEmpty())
 
+	// Create a restricted client using the user SA.
 	f.clientConfig = restclient.AnonymousClientConfig(f.AdminClientConfig())
 	f.clientConfig.BearerToken = string(token)
 
 	// Wait for default ServiceAccount.
-	By(fmt.Sprintf("Waiting for default ServiceAccount in namespace %q.", ns.Name))
-	ctxSa, ctxSaCancel := watchtools.ContextWithOptionalTimeout(ctx, serviceAccountWaitTimeout)
+	By("Waiting for default ServiceAccount in namespace %q.", ns.Name)
```
tnozicka, line 224: and here

tnozicka (commented):
last nit, rest lgtm

test/e2e/framework/framework.go
tnozicka: seeing the diff after the comment this wasn't changed, so let's skip it

Issue #769, opened 2021-08-26
Scylla doesn't yet listen for CQL connections even though it's ready


zimnx:
```
I0826 13:26:41.857697       1 framework/framework.go:244] "Namespace removed." Namespace="e2e-test-scyllacluster-g87j8-cwrjq"

• Failure [182.680 seconds]
ScyllaCluster authentication
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_auth.go:24
  agent requires authentication [It]
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_auth.go:29

  Unexpected error:
      <*fmt.wrapError | 0xc0004c2940>: {
          msg: "can't select data: gocql: no hosts availa
ble in the pool",
          err: {
              s: "gocql: no hosts available in the pool",
          },
      }
      can't select data: gocql: no hosts available in the pool
  occurred

  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/verify.go:119

  Full Stack Trace
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.verifyScyllaCluster(0x1cef4e8, 0xc00007d920, 0x1d12c18, 0xc0008d5600, 0xc000433900, 0xc0000a41e0)
  	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/verify.go:119 +0x12f8
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.glob..func1.1()
  	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_auth.go:161 +0x2359
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc0002c8c00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
  	github.com/onsi/ginkgo@v1.15.0/internal/leafnodes/runner.go:113 +0xa3
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).run(0xc0002c8c00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
  	github.com/onsi/ginkgo@v1.15.0/internal/leafnodes/runner.go:64 +0x15c
[...]

PR #1108, opened 2022-11-30
Declare readiness only if the node and its peers consider at most one node not-UN

### LINKED ISSUE #1077: During rollout restart operator doesn't wait until previously restarted pod becomes part of a Scylla cluster
**Describe the bug**
If we change the number of cores for Scylla (it will trigger rollout pod restart) then we start facing following racy condition:
First pod gets restarted and if it gets into a cluster with some unexpected delay then operator doesn't wait for it and restarts second pod.
In a cluster of 3 nodes it makes cluster be inoperational not having quorum for several minutes.
It happens on GKE and doesn't happen on EKS because in the latter one case a restarted pod gets into a cluster much faster.

Below are the logs from a run on GKE cluster.
`pod-2` (the first restarted) logs:
```
2022-10-27 18:40:55,710 INFO waiting for scylla to stop
INFO  2022-10-27 18:40:55,710 [shard 0] init - Signal received; shutting down
...
INFO  2022-10-27 18:40:57,897 [shard 0] gossip - Gossip is now stopped
...
INFO  2022-10-27 18:40:59,866 [shard 0] init - Scylla version 5.1.0~rc3-0.20221009.9deeeb4db1cd shutdown complete.
2022-10-27 18:40:59,940 INFO stopped: scylla (exit status 0)
...
I1027 18:41:03.305900       1 operator/sidecar.go:438] "Scylla process finished"
rpc error: code = NotFound desc = an error occurred when try to find container "ef77bed95aa5838282bc5d55420a2718d6b111912fc2954dc645c35c7ce87d3f": not foundI1027 18:41:11.371885       1 operator/sidecar.go:147] sidecar version "v1.8.0-alpha.0-133-g97c831e"
...
I1027 18:41:12.363766       1 operator/sidecar.go:360] "Starting scylla"
...
INFO  2022-10-27 18:41:15,764 [shard 0] database - Resharded 7kB for system.compaction_history in 0.54 seconds, 13kB/s
...
INFO  2022-10-27 18:41:18,583 [shard 0] init - loading system_schema sstables
...
INFO  2022-10-27 18:41:19,311 [shard 0] init - setting up system keyspace
...
INFO  2022-10-27 18:41:21,558 [shard 0] database - Resharded 235MB for keyspace1.standard1 in 2.02 seconds, 116MB/s
...
INFO  2022-10-27 18:41:21,621 [shard 0] storage_service - entering STARTING mode
...
INFO  2022-10-27 18:41:21,700 [shard 0] storage_service - Starting up server gossip
...
E1027 18:41:24.931854       1 sidecar/probes.go:110] "readyz probe: can't get scylla native transport" err="giving up after 1 attempts: agent [HTTP 404] Not found" Service="scylla/sct-cluster-us-east1-b-us-east1-2" Node="10.80.11.2"
...
E1027 18:41:34.932431       1 sidecar/probes.go:110] "readyz probe: can't get scylla native transport" err="giving up after 1 attempts: agent [HTTP 404] Not found" Service="scylla/sct-cluster-us-east1-b-us-east1-2" Node="10.80.11.2"
...
INFO  2022-10-27 18:41:44,256 [shard 0] storage_service - entering JOINING mode
...
INFO  2022-10-27 18:41:44,828 [shard 0] storage_service - Node 10.80.11.2 state jump to normal
E1027 18:41:44.931320       1 sidecar/probes.go:110] "readyz probe: can't get scylla native transport" err="giving up after 1 attempts: agent [HTTP 404] Not found" Service="scylla/sct-cluster-us-east1-b-us-east1-2" Node="10.80.11.2"
INFO  2022-10-27 18:41:45,035 [shard 0] storage_service - entering NORMAL mode
...
INFO  2022-10-27 18:41:57,051 [shard 0] init - Scylla version 5.1.0~rc3-0.20221009.9deeeb4db1cd initialization completed.
WARN  2022-10-27 18:42:45,043 [shard 0] cdc - Could not update CDC description table with generation (2022/10/27 18:24:22, 69265668-5186-4a22-a2b1-6b6bad5a0f55): exceptions::unavailable_exception (Cannot achieve consistency level for cl QUORUM. Requires 2, alive 1). Will try again.
WARN  2022-10-27 18:43:14,796 [shard 0] gossip - Skip marking node 10.80.4.107 with status = shutdown as UP
INFO  2022-10-27 18:43:14,796 [shard 0] gossip - InetAddress 10.80.14.183 is now UP, status = NORMAL
INFO  2022-10-27 18:43:14,798 [shard 0] storage_service - Node 10.80.14.183 state jump to normal
INFO  2022-10-27 18:43:14,801 [shard 0] storage_service - Node 10.80.4.107 state jump to normal
INFO  2022-10-27 18:43:14,950 [shard 0] gossip - InetAddress 10.80.12.113 is now DOWN, status = LEFT
INFO  2022-10-27 18:43:14,952 [shard 0] gossip - Node 10.80.12.113 will be removed from gossip at [2022-10-30 18:31:53]: (expire = 1667154713831435293, now = 1666896194952348069, diff = 258518 seconds)
...
WARN  2022-10-27 18:43:36,752 [shard 0] gossip - Fail to send EchoMessage to 10.80.4.107: seastar::rpc::closed_error (connection is closed)
INFO  2022-10-27 18:43:36,892 [shard 0] gossip - InetAddress 10.80.4.107 is now UP, status = NORMAL
```

`pod-1` logs (second restarted):
```
...
INFO  2022-10-27 18:32:55,041 [shard 0] gossip - 60000 ms elapsed, 10.80.12.113 gossip quarantine over
INFO  2022-10-27 18:40:55,897 [shard 0] gossip - Got shutdown message from 10.80.11.2, received_generation=1666893961, local_generation=1666893961
INFO  2022-10-27 18:40:55,898 [shard 0] gossip - InetAddress 10.80.11.2 is now DOWN, status = shutdown
2022-10-27 18:42:05,276 INFO waiting for scylla to stop
INFO  2022-10-27 18:42:05,291 [shard 0] compaction_manager - Asked to stop
...
```

So, from the logs above we can see that the time when `pod-1` (second restarted) got `shutdown` message the first one
was not ready yet. The appeared `quorum` error from CDC is the proof for it. 

**To Reproduce**
1. Deploy operator
2. Deploy 3-pod Scylla cluster on GKE
3. Run some constant load
4. Change the number of CPUs in the `scyllacluster` spec
5. Wait for the restart of the second pod
6. See quorum-errors on the loader side. 

**Expected behavior**
Operator should wait for the previous pod return back to a Scylla cluster before restarting next one.

**Logs**
K8S logs: https://cloudius-jenkins-test.s3.amazonaws.com/0512c157-c4d7-4d3f-9b44-2bcce9d34de9/20221027_191640/kubernetes-0512c157.tar.gz
DB logs: https://cloudius-jenkins-test.s3.amazonaws.com/0512c157-c4d7-4d3f-9b44-2bcce9d34de9/20221027_191640/db-cluster-0512c157.tar.gz

**Environment:**
 - Platform: GKE
 - Kubernetes version: 1.22
 - Scylla version: 5.0.5 , 5.1.0-rc3
 - Scylla-operator version: e.g.: v1.8.0-alpha.0-133-g97c831e | v1.7.4

rzetelskik:
Currently our readiness probe only checks whether the node considers itself UN. We've hit issues related to it many times. By the time a node declares readiness, the gossip should've settled, although nothing stops it from settling with a split brain. For example, it's quite easy to hit the following sequence:

1. Entire 3 node cluster up: `{1: {1: UN, 2: UN, 3:UN}, 2: {1: UN, 2: UN, 3: UN}, 3: {1: UN, 2: UN, 3: UN}}`
2. Restart node 3 and wait for it to come back up and declare readiness.
3. Node 3 is up but cluster has a split-brain: `{1: {1: UN, 2: UN, 3:DN}, 2: {1: UN, 2: UN, 3: DN}, 3: {1: DN, 2: DN, 3: UN}}`
4. Restart node 2: `{1: {1: UN, 2: DN, 3:DN}, 3: {1: DN, 2: DN, 3: UN}}`

At this point you'll get errors from queries with quorum consistency regardless of which node you hit.

So we've discussed different approaches to how we could modify the readiness probe or the operator to get closer to what the documentation advises, i.e. all nodes UP before performing a topology change.
Essentially we want our readiness probes' semantics to no longer just be "I'm ready to serve traffic" but "I'm ready for a topology change". However we weren't able to just declare readiness when, and only when, a node considers all of its peers UN, since that caused some issues with rollouts.

This PR implements an approach suggested by @tnozicka: each node declares readiness when it and its peers consider at most one (and the same) node not-UN.

Resolves #1077

### DIFF
--- a/pkg/cmd/operator/sidecar.go
+++ b/pkg/cmd/operator/sidecar.go
@@ -391,6 +391,7 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 
 		http.HandleFunc(naming.LivenessProbePath, prober.Healthz)
 		http.HandleFunc(naming.ReadinessProbePath, prober.Readyz)
+		http.HandleFunc(naming.InternalNodeStatusesPath, prober.InternalNodeStatuses)
 
 		err := server.ListenAndServe()
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -148,6 +148,10 @@ func servicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
 			Port: 9100,
 		},
 		{
+			Name: "prober",
+			Port: naming.ProbePort,
+		},
+		{
 			Name: portNameCQL,
 			Port: 9042,
 		},
--- a/pkg/controller/scyllacluster/resource_test.go
+++ b/pkg/controller/scyllacluster/resource_test.go
@@ -92,6 +92,10 @@ func TestMemberService(t *testing.T) {
 			Port: 9100,
 		},
 		{
+			Name: "prober",
+			Port: 8080,
+		},
+		{
 			Name: "cql",
 			Port: 9042,
 		},
--- a/pkg/naming/constants.go
+++ b/pkg/naming/constants.go
@@ -102,10 +102,11 @@ const (
 
 	DataDir = "/var/lib/scylla"
 
-	ReadinessProbePath = "/readyz"
-	LivenessProbePath  = "/healthz"
-	ProbePort          = 8080
-	ScyllaAPIPort      = 10000
+	ReadinessProbePath       = "/readyz"
+	LivenessProbePath        = "/healthz"
+	InternalNodeStatusesPath = "/internal_node_statuses"
+	ProbePort                = 8080
+	ScyllaAPIPort            = 10000
 
 	OperatorEnvVarPrefix = "SCYLLA_OPERATOR_"
 )
--- a/pkg/sidecar/probes.go
+++ b/pkg/sidecar/probes.go
@@ -2,12 +2,18 @@ package sidecar
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
+	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/scyllaclient"
+	apierrors "k8s.io/apimachinery/pkg/util/errors"
 	corev1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 )
@@ -59,6 +65,30 @@ func (p *Prober) getNodeAddress() (string, error) {
 	return controllerhelpers.GetScyllaIPFromService(svc)
 }
 
+func GetInternalNodeStatuses(ctx context.Context, addr string) (scyllaclient.NodeStatusInfoSlice, error) {
+	statuses := scyllaclient.NodeStatusInfoSlice{}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, addr, nil)
+	if err != nil {
+		return statuses, fmt.Errorf("can't create request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	client := http.DefaultClient
+	res, err := client.Do(req)
+	if err != nil {
+		return statuses, fmt.Errorf("can't send a request: %w", err)
+	}
+	defer res.Body.Close()
+
+	err = json.NewDecoder(res.Body).Decode(&statuses)
+	if err != nil {
+		return statuses, fmt.Errorf("can't decode json: %w", err)
+	}
+
+	return statuses, nil
+}
+
 func (p *Prober) Readyz(w http.ResponseWriter, req *http.Request) {
 	ctx, ctxCancel := context.WithTimeout(req.Context(), p.timeout)
 	defer ctxCancel()
@@ -100,27 +130,105 @@ func (p *Prober) Readyz(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	klog.V(4).InfoS("readyz probe: node statuses", "Statuses", nodeStatuses)
+
+	// Check for self UN and isNativeTransportEnabled first.
+	selfStatusFound := false
 	for _, s := range nodeStatuses {
-		klog.V(4).InfoS("readyz probe: node state", "Node", s.Addr, "Status", s.Status, "State", s.State)
+		if s.Addr != nodeAddress {
+			continue
+		}
 
-		if s.Addr == nodeAddress && s.IsUN() {
-			transportEnabled, err := scyllaClient.IsNativeTransportEnabled(ctx, localhost)
-			if err != nil {
-				w.WriteHeader(http.StatusServiceUnavailable)
-				klog.ErrorS(err, "readyz probe: can't get scylla native transport", "Service", p.serviceRef(), "Node", s.Addr)
-				return
-			}
+		selfStatusFound = true
 
-			klog.V(4).InfoS("readyz probe: node state", "Node", s.Addr, "NativeTransportEnabled", transportEnabled)
-			if transportEnabled {
-				w.WriteHeader(http.StatusOK)
-				return
+		if !s.IsUN() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			klog.V(2).InfoS("readyz probe: node is not UN", "Service", p.serviceRef(), "Node", s.Addr)
+			return
+		}
+
+		transportEnabled, err := scyllaClient.IsNativeTransportEnabled(ctx, localhost)
+		if err != nil {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			klog.ErrorS(err, "readyz probe: can't get scylla native transport", "Service", p.serviceRef(), "Node", s.Addr)
+			return
+		}
+
+		if !transportEnabled {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			klog.V(2).InfoS("readyz probe: scylla native transport is not enabled", "Service", p.serviceRef(), "Node", s.Addr)
+			return
+		}
+	}
+
+	if !selfStatusFound {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		klog.V(2).InfoS("readyz probe: node's own status is missing", "Service", p.serviceRef())
+		return
+	}
+
+	receivedStatuses := make(map[string]scyllaclient.NodeStatusInfoSlice)
+	receivedStatuses[nodeAddress] = nodeStatuses
+
+	notUNMap := make(map[string]bool)
+	for _, s := range nodeStatuses {
+		if !s.IsUN() {
+			notUNMap[s.Addr] = true
+		}
+	}
+
+	if len(notUNMap) > 1 {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		klog.V(2).InfoS("readyz probe: node considers more than one peer not UN", "Service", p.serviceRef(), "Statuses", nodeStatuses)
+		return
+	}
+
+	errs := make([]error, 0)
+	for _, s := range nodeStatuses {
+		if s.Addr == nodeAddress {
+			continue
+		}
+
+		if notUNMap[s.Addr] {
+			continue
+		}
+
+		statusURL := url.URL{
+			Scheme: "http",
+			Host:   net.JoinHostPort(s.Addr, strconv.Itoa(naming.ProbePort)),
+			Path:   naming.InternalNodeStatusesPath,
+		}
+		statuses, err := GetInternalNodeStatuses(ctx, statusURL.String())
+		if err != nil {
+			errs = append(errs, fmt.Errorf("can't get statuses from node %s: %w", s.Addr, err))
+			break
+		}
+
+		klog.V(4).InfoS("readyz probe: received statuses", "Node", s.Addr, "Statuses", statuses)
+		receivedStatuses[s.Addr] = statuses
+
+		for _, rs := range statuses {
+			if !rs.IsUN() {
+				notUNMap[rs.Addr] = true
 			}
 		}
 	}
 
-	klog.V(2).InfoS("readyz probe: node is not ready", "Service", p.serviceRef())
-	w.WriteHeader(http.StatusServiceUnavailable)
+	err = apierrors.NewAggregate(errs)
+	if err != nil {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		klog.ErrorS(err, "readyz probe: can't gather statuses from peers", "Service", p.serviceRef(), "Err", err)
+		return
+	}
+
+	if len(notUNMap) > 1 {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		klog.V(2).InfoS("readyz probe: more than one node considered not UN by peers", "Service", p.serviceRef(), "Statuses", receivedStatuses)
+		return
+	}
+
+	klog.V(2).InfoS("readyz probe: node is ready", "Service", p.serviceRef())
+	w.WriteHeader(http.StatusOK)
 }
 
 func (p *Prober) Healthz(w http.ResponseWriter, req *http.Request) {
@@ -157,4 +265,34 @@ func (p *Prober) Healthz(w http.ResponseWriter, req *http.Request) {
 	}
 
 	w.WriteHeader(http.StatusOK)
+}
+
+func (p *Prober) InternalNodeStatuses(w http.ResponseWriter, req *http.Request) {
+	ctx, ctxCancel := context.WithTimeout(req.Context(), p.timeout)
+	defer ctxCancel()
+
+	scyllaClient, err := controllerhelpers.NewScyllaClientForLocalhost()
+	if err != nil {
+		klog.ErrorS(err, "internal_node_statuses: can't get scylla client", "Service", p.serviceRef())
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	defer scyllaClient.Close()
+
+	nodeStatuses, err := scyllaClient.Status(ctx, localhost)
+	if err != nil {
+		klog.ErrorS(err, "internal_node_statuses: can't get scylla node status", "Service", p.serviceRef())
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	payload, err := json.Marshal(nodeStatuses)
+	if err != nil {
+		klog.ErrorS(err, "internal_node_statuses: can't marshall node statuses", "Service", p.serviceRef())
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(payload)
 }


rzetelskik:
Generally we should also try moving away from the current implementation of [`scyllaClient.Status`](https://github.com/scylladb/scylla-operator/blob/7be1034e465e5de55c2f867e31cef41989db7446/pkg/scyllaclient/client.go#L106) which makes 5 calls to Scylla API and use [`/failure_detector/endpoints`](https://github.com/scylladb/scylladb/blob/c5121cf27365abf5989ed17078f2edff5ac59c1b/api/api-doc/failure_detector.json#L65) instead. Looks like it would require some ugly parsing but it would also let us resolve https://github.com/scylladb/scylla-operator/issues/769 (as suggested here https://github.com/scylladb/scylladb/issues/8275#issuecomment-953426583). @tnozicka lmk what you think and if it should land separately or in this PR.

mykaul:
@tgrabiec - let's update with today's meeting discussion on what the right way to do it is.
Thanks.

vladzcloudius:
@rzetelskik FYI: `UN` check is not good enough because what you really care is `UN` and node listening on a CQL port. And AFAIR our bootup code allows the later only after the former.

Hence if you only wait for a gossip state (which `UN` is about) before restarting the following node you allow a race when 2 nodes will be unreachable to clients.

The solution we use in our Ansible Role is probing the CQL (TCP) port instead of the gossip state. And implementing this is trivial in Ansible which I suspect should also be the case in K8S.

@asias Please, correct me if I'm missing something in the above.

tnozicka:
I guess we need to check CQL in addition to UN but we need UN, most importantly from the other nodes, not itself. After talking with Avi we have a go ahead to implement the N^2 algorithm where node considers itself ready, only if a quorum of nodes considers it as ready. As quorum depends on a current state of keyspaces and we don't have a way to enforce minimal quorum, quorum is always N-1 for us.

tnozicka:
(e.g. all the nodes can serve CQL and not be able to talk to each other, serving only C_ONE while we need to have C_QUORUM)

tnozicka:
> @tnozicka lmk what you think and if it should land separately or in this PR.
- separately
- whether to do it now or plan it depends on a time estimate - for now, I can live with N==5 for something that grows O(N^2)

vladzcloudius:
> (e.g. all the nodes can serve CQL and not be able to talk to each other, serving only C_ONE while we need to have C_QUORUM)

This should not possibly happen - a node will have to at least be able to contact seeders which are not itself. 
And what if cluster IS partitioned? You won't allow a rolling restart?
What if you cluster is messed up and multiple nodes crashed and to bring them back you need to execute a RR?

The bottom line: it does make sense to make sure that the restarting node is UN from other nodes perspective however in certain cases this should not be a hard requirement for a RR. 

And, still, you need to check CQL port *in addition* to UN check.

vladzcloudius:
> I guess we need to check CQL in addition to UN but we need UN, most importantly from the other nodes, not itself. After talking with Avi we have a go ahead to implement the N^2 algorithm where node considers itself ready, only if a quorum of nodes considers it as ready. As quorum depends on a current state of keyspaces and we don't have a way to enforce minimal quorum, quorum is always N-1 for us.

The more I think of what you wrote above the more I think it's simply wrong to implement this on the Operator level, @tnozicka.

Let's start the last thing you wrote - you are going to not allow next node restart till ALL nodes (except for this node that is) consider this node as UN.

This is definitely going to backfire in cases I mentioned above. 
So, it's already problematic.

But let's take a step back for a moment and recall what is the problem you want to avoid by this PR according to the opening message: a "split brain" situation in order to allow QUORUM queries. Why QUORUM? Why not ALL, or LOCAL_QUORUM, or LOCAL_ONE, or EACH_QUORUM? I think this distinction is... strange because there are too many assumptions:
* For RF<3 even a single node DOWN is not going to allow QUORUM.
* Who said the user cares about the CL=QUORUM? 

A more understandable distinction is that we want to ensure that there is not more than a single node DOWN during a rolling restart. Why? Because this is the best we can do. And it has nothing to do with a particular CL.

So it all comes down to restarting the next node only after the current node is fully operational - including joining the ring.

But isn't this ("joining the ring" part) the problem Raft is meant to resolve? - And, surprise, it IS!
And we know that this is not trivial. So, what makes you think you will be able to resolve it (in it's completeness!) on the Operator level?

You may THINK you did but I've just shown you before a few examples which you probably missed and which are not going to be covered by you "N-1" idea.

I think this is impossible unless you give up on high availability or implement Raft on the Operator level and I really hope this is not what you want to do.

Clearly this has to be resolved on the scylla level (and I believe it's already done since according to @kostja Raft is fully operational) and the nodes should not move to `UN` state till they haven't properly joined the cluster.

And I believe this is the direction, right, Kostja?

If "yes" then no need to re-invent the ~~Raft~~ wheel and you'd better follow my original proposal (probe the TCP port) which is going to simplify your solution significantly.

rzetelskik:
@vladzcloudius thanks for the input, I think you've raised quite a few valid points. Let me try and address some of them.

> FYI: UN check is not good enough because what you really care is UN and node listening on a CQL port. And AFAIR our bootup code allows the later only after the former.

That's true and in the current implementation of the readiness probe we are checking for `is_native_transport_running`. It's racy, it doesn't guarantee that CQL server is up and we should move away from it - we have an issue for it and I've linked it in one of my previous comments (issue: https://github.com/scylladb/scylla-operator/issues/769, approach suggested by Asias: https://github.com/scylladb/scylladb/issues/8275#issuecomment-953426583). However, getting an OK from `is_native_transport_running` means that the same prerequisites were met as when you're probing the CQL port - specifically, what we care about here - the gossip has settled.

> This should not possibly happen - a node will have to at least be able to contact seeders which are not itself.

Which doesn't prevent it from the scenario mentioned above, does it?

> But let's take a step back for a moment and recall what is the problem you want to avoid by this PR according to the opening message: a "split brain" situation in order to allow QUORUM queries. Why QUORUM? 

I believe the confusion is coming from the issue that this PR originated from - https://github.com/scylladb/scylla-operator/issues/1077. 

What you've written right below
> A more understandable distinction is that we want to ensure that there is not more than a single node DOWN during a rolling restart. Why? Because this is the best we can do. And it has nothing to do with a particular CL.
So it all comes down to restarting the next node only after the current node is fully operational - including joining the ring.

is what this PR is trying to achieve, i.e. for our readiness probes to only allow for one non-UN node in the entire cluster, but according to all the nodes. So the readiness probe of each node is going to state whether the entire cluster is ready for a topology change, according to this node.

vladzcloudius:
> That's true and in the current implementation of the readiness probe we are checking for is_native_transport_running. It's racy, it doesn't guarantee that CQL server is up and we should move away from it - we have an issue for it and I've linked it in one of my previous comments (issue: https://github.com/scylladb/scylla-operator/issues/769, approach suggested by Asias: https://github.com/scylladb/scylladb/issues/8275#issuecomment-953426583). 

@rzetelskik This is not supposed to be racy and scylla is not supposed to listen to the CQL port before the node considers itself as UN. If that's not the case I think this is supposed to be a bug in scylla and this is where it's supposed to be fixed - not here.

> However, getting an OK from is_native_transport_running means that the same prerequisites were met as when you're probing the CQL port - specifically, what we care about here - the gossip has settled.

Again, the CQL port is not supposed to become available until gossip has settled and Operator (or an Ansible Role for the matter) is not supposed to be the place to ensure that - scylla is. 

> 

>>    A more understandable distinction is that we want to ensure that there is not more than a single node DOWN during a rolling restart. Why? Because this is the best we can do. And it has nothing to do with a particular CL.
    So it all comes down to restarting the next node only after the current node is fully operational - including joining the ring.

> is what this PR is trying to achieve, i.e. for our readiness probes to only allow for one non-UN node in the entire cluster, but according to all the nodes. So the readiness probe of each node is going to state whether the entire cluster is ready for a topology change, according to this node.

Yes and on the way it breaks the HA - see my previous comments on the matter.

>> But isn't this ("joining the ring" part) the problem Raft is meant to resolve? - And, surprise, it IS!

> I believe we're all aware of that. The thing is this issue breaks HA during updates, upgrades, scaling etc. And afaik 5.1 does not yet handle topology changes with raft. I'd personally consider these changes only a temporary workaround until the support for topology changes is released.

My point was that you are not going to be able to resolve this issue on the Operator level and as a result you are always going to either break the HA or break some other corner case - therefore you shouldn't even try.

Requiring all nodes to be up and not allowing any partitioning during a rolling restart is a too harsh of a requirement. See more examples why it is terrible here: https://github.com/scylladb/scylla-operator/pull/1108#issuecomment-1335444473

In short my points were: 
1) Don't try to solve the scylla problem on the Operator level.   
2) Use the correct probing criteria and checking for `UN` is a wrong one because it's not the last thing that happens in the initialization path (see @asias comment you referenced above).

tnozicka:
I'll try to explain a bit more but this thread has branched a bit so if you feel a particular point was missed I am happy to get back to it. Also we have talked about readiness probe a lot this week in the operator team with an intention to unstuck the fix / adjust it's direction so even some of our previous views written here might have changed.

We have went through various approaches and reasoning but what we ended up with is that, for the purposes of the scylla operator and k8s, the readiness needs to mean "I am up and I am part of the quorum (aka N/2+1)". This primarily comes from a need to tolerate voluntary disruptions like rolling restart of Kubernetes nodes where we don't have any control except using a PDB, contrary to say a topology or any other change when we can instrument an action. During a Kubernetes node upgrade / rolling restart we need to maintain the ability for users to use scylladb with at least consistency quorum. The only way to do that is using a PDB (currently max 1 unavailable for RF=3) and PDBs are bound to readiness - so a pod may become ready, only if it is able to serve CL_QUORUM or the next one could go down sooner and we'd break the invariant.

So the /readyz algorithm will be something like:
1) check you are UN
2) check what everyone else thinks about you
3) if SUM(UN) > N/2 => ready
with raft:
1) ask the local node to ask the leader whether you are part of the quorum

For voluntary topology changes we won't use readiness but rather instrument a hook that checks that all the nodes in the cluster are UN - if it were to be done before raft it would mean "everyone thinks about everyone that they are UN".

Regarding a rolling restart, that gets blocked in some cases even today. E.g. StatefulSets won't trigger an update when there are broken nodes, so this PR won't change that. Then obviously if a quorum of nodes goes down with the new readiness approach, all the rest will go unready as well. The question that seems separate from this PR would be whether a rolling restart is required as an unstuck procedure - there are differences like some of the configs being auto reloaded, failed containers auto restart and there are other options like `kubectl delete pods` with cluster selector (it's not rolling but those pods are probably not working either).

vladzcloudius:
> I'll try to explain a bit more but this thread has branched a bit so if you feel a particular point was missed I am happy to get back to it. Also we have talked about readiness probe a lot this week in the operator team with an intention to unstuck the fix / adjust it's direction so even some of our previous views written here might have changed.
> 
> We have went through various approaches and reasoning but what we ended up with is that, for the purposes of the scylla operator and k8s, the readiness needs to mean "I am up and I am part of the quorum (aka N/2+1)". This primarily comes from a need to tolerate voluntary disruptions like rolling restart of Kubernetes nodes where we don't have any control except using a PDB, contrary to say a topology or any other change when we can instrument an action. During a Kubernetes node upgrade / rolling restart we need to maintain the ability for users to use scylladb with at least consistency quorum. The only way to do that is using a PDB (currently max 1 unavailable for RF=3) and PDBs are bound to readiness - so a pod may become ready, only if it is able to serve CL_QUORUM or the next one could go down sooner and we'd break the invariant.
> 
> So the /readyz algorithm will be something like:
> 
>     1. check you are UN
> 
>     2. check what everyone else thinks about you
> 
>     3. if SUM(UN) > N/2 => ready
>        with raft:
> 
>     4. ask the local node to ask the leader whether you are part of the quorum
> 
> 
> For voluntary topology changes we won't use readiness but rather instrument a hook that checks that all the nodes in the cluster are UN - if it were to be done before raft it would mean "everyone thinks about everyone that they are UN".
> 
> Regarding a rolling restart, that gets blocked in some cases even today. E.g. StatefulSets won't trigger an update when there are broken nodes, so this PR won't change that. Then obviously if a quorum of nodes goes down with the new readiness approach, all the rest will go unready as well. The question that seems separate from this PR would be whether a rolling restart is required as an unstuck procedure - there are differences like some of the configs being auto reloaded, failed containers auto restart and there are other options like `kubectl delete pods` with cluster selector (it's not rolling but those pods are probably not working either).

@tnozicka You seem to have ignored a lot of what I wrote before. Here (https://github.com/scylladb/scylla-operator/pull/1108#issuecomment-1335444473) in particular.

This "readiness" check is not supposed to be used in a RR of a healthy cluster according to a description (if it is - please, update the description).

And as such it's supposed to work and not cause "dead-locks" in any possible situation.

Please, explain to me how will you be able to start 7 out of 9 Scylla PODs if they crash using the algorithm you've described?

tnozicka:
> @tnozicka You seem to have ignored a lot of what I wrote before. Here (https://github.com/scylladb/scylla-operator/pull/1108#issuecomment-1335444473) in particular.
This "readiness" check is not supposed to be used in a RR of a healthy cluster according to a description (if it is - please, update the description).

I've tried to take a bit of a fresh look, for the record it wasn't meant to discard anyone's input, but it was hard to reply to all given the conversation has branched and the input (algorithm / our approach) has changed as well. 

The PR title and descriptions are outdated and before we reach a consensus I'd prefer we focus on the proposed solution described in https://github.com/scylladb/scylla-operator/pull/1108#issuecomment-1464181691. When we have that consensus let's update the PR but it felt better we don't oscillate the PR it self but I don't feet that strong. (e.g. PR description / title updates retrigger the CI every time [about 4h of compute time].)

> And as such it's supposed to work and not cause "dead-locks" in any possible situation.

I don't think it does. It may block k8s node upgrades, for rolling updates we still have the control. Even today the "single node down" blocks rolling update / restart. It doesn't mean there is a "dead lock", as I've mentioned in https://github.com/scylladb/scylla-operator/pull/1108#issuecomment-1464181691 there are other ways to deal with it.

> Please, explain to me how will you be able to start 7 out of 9 Scylla PODs if they crash using the algorithm you've described?

The 9 pods are started and restarted (on failure) independently and readiness is not used as a precondition to join the cluster.

Also I've been thinking a bit more about the algorithm I described over the weekend and while I still think it would work if nodes were what determines the quorum, with scylladb that's actually the keyspace. With many keyspaces scattered over the nodes this may become challenging. I guess in the worst case the keyspaces can be distributed over the nodes in a way that you can only ever take down 1 node even with RF>=5. So I think we need to come up with something better. I am also happy to invite you for the next arch session we have on the topic, if you'd be interested.

vladzcloudius:
> The 9 pods are started and restarted (on failure) independently and readiness is not used as a precondition to join the cluster.

I wasn't talking about bootstrapping, @tnozicka.
I was referring the situation when 7 out 9 PODs crashed. 
Then you fixed  the reason for the crash (including the nodes that didn't crash) and now what to execute a RR procedure.
You algorithms would not allow that.

> Also I've been thinking a bit more about the algorithm I described over the weekend and while I still think it would work if nodes were what determines the quorum, with scylladb that's actually the keyspace. With many keyspaces scattered over the nodes this may become challenging. I guess in the worst case the keyspaces can be distributed over the nodes in a way that you can only ever take down 1 node even with RF>=5. So I think we need to come up with something better. I am also happy to invite you for the next arch session we have on the topic, if you'd be interested.

Indeed. 
On top of that - "QUORUM" requirement in the context of CQL availability is synthetic (not generic). You, guys, invented this requirement based on not clear what and now work hard trying to achieve that. 

CQL availability is completely defined by CLs that are used in a specific use case.
And QUORUM is neither the only one nor the mainly used one.

tnozicka:
> I wasn't talking about bootstrapping, @tnozicka.
I was referring the situation when 7 out 9 PODs crashed.
Then you fixed the reason for the crash (including the nodes that didn't crash) and now what to execute a RR procedure.
You algorithms would not allow that.

I wasn't talking about bootstrapping either. As I've mentioned in the comment before, rolling restart may not be the only way out and it isn't possible even today. So you'd say `kubectl delete` the DN pods and kube would recreate them. It also technically doesn't have to depend on the readiness - for voluntary operations we have options beyond readiness probes. 

> You, guys, invented this requirement based on not clear what and now work hard trying to achieve that.

I think the ability to sustain C_QUORUM is an important invariant we have to keep so the apps keep working correctly. I wouldn't say we have invented it, recall this mentioned by people higher in the chain then me. Making updates not disruptive is important for the users as well, I believe that's what scylla cloud and our docs try to achieve too. It's just in the presence of all the keyspace combination, the intersection of those internal quorum is `N-1` instead of `N/2-1`, which is what the docs say, except is doesn't say how to reliably make sure the new node is UN (+ considered UN by the others).

vladzcloudius:
> I think the ability to sustain C_QUORUM is an important invariant we have to keep so the apps keep working correctly.

Which part of CQL protocol defines this "important invariant"? 

> Making updates not disruptive is important for the users as well, I believe that's what scylla cloud and our docs try to achieve too. 

Updates with CL=ONE, LOCAL_ONE, LOCAL_QUORUM in a multi-DC config, TWO do not require that strong guarantee you are trying to achieve. So, again, could you clarify what you are talking about exactly and which docs you refer? AFAIK none of our docs specify that "only CL=QUORUM is going to guarantee <whatever>".

> I wasn't talking about bootstrapping either. 

"join the cluster" means bootstrapping in Cassandra/Scylla lingo. ;)

> As I've mentioned in the comment before, rolling restart may not be the only way out and it isn't possible even today. So you'd say kubectl delete the DN pods and kube would recreate them. It also technically doesn't have to depend on the readiness.

How come? How will you define that the node has started successfully? You have to have a corresponding condition. 

>> You, guys, invented this requirement based on not clear what and now work hard trying to achieve that.
>  I wouldn't say we have invented it, recall this mentioned by people higher in the chain then me.

Are you suggesting that "people higher in the chain then me" can not be mistaken? ;)
In any case - it's you who's representing a specific point of view claiming some not-very-obvious assumptions are "invariants". 

In any case, this conversation began going in circles.
I suggest we have a call and discuss this if you want. 

My point stands as it was before: checking a UN on a node that just started and then checking a CQL port is a procedure that is more than enough. It's a DB responsibility to ensure uninterrupted service when a single node is being restarted. When it starts listening to CQL denotes a node's readyness to serve CQL requests. If they are going to fail - it's a DB (Scylla) bug and you can't and should not fix it on the Operator level.

Any other complication you are trying to introduce are (1) not obvious/baseless and (2) not supposed to be implemented on the Operator level.

tnozicka:
> Updates with CL=ONE, LOCAL_ONE, LOCAL_QUORUM in a multi-DC config, TWO do not require that strong guarantee you are trying to achieve. So, again, could you clarify what you are talking about exactly and which docs you refer? AFAIK none of our docs specify that "only CL=QUORUM is going to guarantee whatever".

Apologies for the confusion, as we only do single DC with the operator I was technically talking about CL=LOCAL_QUORUM. Would you agree that's an invariant we should maintain?

> Which part of CQL protocol defines this "important invariant"?
Not the protocol but saying your cluster is HA with RF>=3 kind of implies it from user's point of view.

> How come? How will you define that the node has started successfully? You have to have a corresponding condition.
We can have other conditions / status. The only think that is strictly bound to readiness is PDBs. While we are technically able not to use readiness for it we'd try to avoid it, if possible.



>> You, guys, invented this requirement based on not clear what and now work hard trying to achieve that.
I wouldn't say we have invented it, recall this mentioned by people higher in the chain then me.

> Are you suggesting that "people higher in the chain then me" can not be mistaken? ;)
In any case - it's you who's representing a specific point of view claiming some not-very-obvious assumptions are "invariants". Own it! ;)

I don't think that's what I said. I've said that people beyond my team shared that view. Any and all of us can be mistaken ;)

> In any case, this conversation began going in circles.
I suggest we have a call and discuss this if you want.
My point stands as it was before: checking a UN on a node that just started and then checking a CQL port is a procedure that is more than enough. It's a DB responsibility to ensure uninterrupted service when a single node is being restarted. When it starts listening to CQL denotes a node's readyness to serve CQL requests. If they are going to fail - it's a DB (Scylla) bug and you can't and should not fix it on the Operator level.

I am happy to have the call, but I feel like we'd need an external opinion on whether a rolling restart should keep CL=LOCAL_QUORUM working all the time. That would be a good starting point to talk about the right way to solve it.

In the meantime, I've also filed https://github.com/scylladb/scylladb/issues/13156 to have a discussion on the procedure that's independent of scylla-operator, e.g. scylla cloud is affected in the same way.

rzetelskik:
Closing as stale until there's any progress in https://github.com/scylladb/scylladb/issues/13156

PR #1109, opened 2022-12-01
Allow to apply our manifests using server-side apply

zimnx:
Some of the injected fields were set explicitly making server-side apply to conflict.


tnozicka (approved):
/approve
/lgtm

thanks!

PR #1111, opened 2022-12-05
Add ScyllaDBMonitoring

tnozicka:
**Description of your changes:**
This PR introduces experimental `scyllav1alpha1.ScyllaDBMonitoring` API object that can setup and manage the monitoring stack for Scylla - both Prometheus and Grafana with our dashboard(s). Long overdue but finally here!

Prometheus requires mTLS to be able to access it but doesn't enforce it yet. That said, anyone using it should setup mTLS properly and be aware this could be fixed at any point.

Grafana is by default using *Basic Auth* so it's easier to access it by users through UI. Other Auth options may come in the future. 

Requires:
- [x] #1125
- [x] #1127 
- [x] #1138 
- [x] #1143
- [x] #1146
- [ ] Make sure the new e2e runs in both parallel suites (right before we merge)

**Which issue is resolved by this Pull Request:**
Resolves #1094

PR #1123, opened 2022-12-20
Move scylla validation into its group folder

tnozicka:
Validation belongs to its group. We need to fix this historical misplacement for  #1111


zimnx (approved):
LGTM, thanks

PR #1125, opened 2022-12-20
[WIP] Use generic handlers

tnozicka:
We have a lot of duplicated code that is growing with every new controller and handler. This PR add generic handlers that allow keeping the logic in one place. This PR is switching the easy 90% we'll switch the leftovers when we have more bandwidth.

Needed  by #1111


zimnx:
Suggestions after going through it: https://github.com/tnozicka/scylla-operator/pull/18
LGTM when squashed

tnozicka:
@zimnx thanks for your fixes!

squashed and rebased

zimnx (dismissed):
LGTM, thanks

tnozicka:
(clean rebase)

PR #1126, opened 2022-12-22
Fix flakyness of replace node e2e

zimnx:
Full quorum check is sensitive to seeing nodes not being part of the cluster, when such node is spotted it immediately returns an error.
In replace node e2e, it could happen that cluster become ready before replaced node was acknowledged by the others.
In this case, full quorum check returned an error, because it saw a IP not bound to any cluster Service.

To fix it, test waits until cluster is rolled out, and replaced node is seen by the other one, before the full quorum check is validated.

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_replace.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_replace.go
@@ -86,6 +86,23 @@ var _ = g.Describe("ScyllaCluster replace", func() {
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
 
+		framework.By("Waiting for the other nodes to acknowledge the replace")
+
+		client, _, err := utils.GetScyllaClient(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		replacedNodeService, err := f.KubeClient().CoreV1().Services(sc.Namespace).Get(ctx, utils.GetNodeName(sc, 0), metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


tnozicka (commented):
would the readiness check @rzetelskik is building help here?

zimnx:
> would the readiness check @rzetelskik is building help here?

AFAIR it allows for one DN, we have 2 nodes in this e2e, so it wouldn't help.

rzetelskik:
> AFAIR it allows for one DN, we have 2 nodes in this e2e, so it wouldn't help.

Yes, it wouldn't help in this case. After it's been merged you could just increase the number of nodes in this test to 3 though.

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -86,6 +86,25 @@ var _ = g.Describe("ScyllaCluster replace", func() {
...
+		otherNodeService, err := f.KubeClient().CoreV1().Services(sc.Namespace).Get(ctx, utils.GetNodeName(sc, 1), metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func(g o.Gomega) {
+			status, err := client.Status(ctx, otherNodeService.Spec.ClusterIP)
+			g.Expect(err).NotTo(o.HaveOccurred())
+			g.Expect(status.LiveHosts()).To(o.ContainElement(replacedNodeService.Spec.ClusterIP))
+		}).WithPolling(time.Second).WithTimeout(5 * time.Minute).Should(o.Succeed())
+
+		framework.By("Replace has been acknowledged by the other node")
```
tnozicka, line 106: this doesn't read well. it should be either a separate framework.info but usually the "next" message will imply waiting was finished

zimnx: removed

tnozicka (approved):
/lgtm
/approve

tnozicka (approved):

PR #1127, opened 2022-12-22
Use generic controller ref manager

tnozicka:
We have been duplication code for every type that needed a controller ref manager. This PR uses generics to unify the implementation.

Needed by #1111


pkg/controllerhelpers/sync.go
```diff
@@ -0,0 +1,97 @@
...
+	ctx context.Context,
+	controller metav1.Object,
+	controllerGVK schema.GroupVersionKind,
+	selector labels.Selector,
+	filterFunc func(T) bool,
+	filterObjectsSelector labels.Selector,
+	control ControlleeManagerGetObjectsInterface[CT, T],
+) (map[string]T, error) {
+	// List all objects to find even those that no longer match our selector.
+	// They will be orphaned in ClaimObjects().
```
zimnx, line 70: This comment explains why label.Everything() is hardcoded in above function, it shouldn't be here.

pkg/controllerhelpers/sync.go
```diff
@@ -0,0 +1,97 @@
...
+	control ControlleeManagerGetObjectsInterface[CT, T],
+) (map[string]T, error) {
+	return GetObjectsWithFilter[CT, T](
+		ctx,
+		controller,
+		controllerGVK,
+		selector,
+		func(T) bool {
+			return true
+		},
```
zimnx, line 54: nit: GetObjectsWithFileter is only used here. I would remove this predicate from function signature together with filtering capabilities if it's basically a dead code.

pkg/controllertools/refmanager.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	crm.CanAdoptFunc = crm.canAdopt
+
+	return crm
+}
+
+func (m *ControllerRefManager[T]) canAdopt() error {
+	// If any adoptions are attempted, we should first recheck for deletion with
+	// an uncached quorum read sometime after listing Secrets.
```
zimnx, line 87:
```suggestion
	// an uncached read.
```

pkg/controllerhelpers/sync.go
tnozicka: i think it still applies, `filterObjectsSelector != selector`

pkg/controllerhelpers/sync.go
tnozicka: I think I added it because it was needed by some handlers that I haven't switched in the end because it proved to be complex. Probably for a case of a distributed / sharded controller.

tnozicka: i guess I can remove it and bring it back if we need it

pkg/controllerhelpers/sync.go
tnozicka: removed

pkg/controllerhelpers/sync.go
tnozicka: removed

pkg/controllertools/refmanager.go
tnozicka: done

pkg/controller/nodeconfig/sync.go
```diff
@@ -8,19 +8,20 @@ import (
...
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 )
 
+type CT = *scyllav1alpha1.NodeConfig
```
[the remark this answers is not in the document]
tnozicka: done

zimnx (approved):
lgtm, great improvement, thanks

Issue #1129, opened 2022-12-27
Missing cpuset setting in docs example for guaranteed QoS


soyacz:
**Describe the bug**
In https://operator.docs.scylladb.com/stable/performance.html is mentioned to use static CPU manager policy. To make use of it `cpuset` setting should be enabled. The mentioned example misses that part.

zimnx:
These two aren't related. 
CPU Manager Policy is kubelet configuration option, where `cpuset` in ScyllaCluster is a flag controlling additional validation logic. Not having `cpuset` enabled, doesn't prevent Pod from getting Guaranteed QoS class.

mykaul:
> These two aren't related. CPU Manager Policy is kubelet configuration option, where `cpuset` in ScyllaCluster is a flag controlling additional validation logic. Not having `cpuset` enabled, doesn't prevent Pod from getting Guaranteed QoS class.

Yes, but we do need the CPU manager, to control alignment of nodes to cores, no?
```
The following policy options exist for the static CPUManager policy:

full-pcpus-only (beta, visible by default) (1.22 or higher)
distribute-cpus-across-numa (alpha, hidden by default) (1.23 or higher)
align-by-socket (alpha, hidden by default) (1.25 or higher)
```

We probably need the 1st and 3rd?

soyacz:
According to docs:
```
cpuset: Optional field. If it’s true, then the operator will start Scylla with cpu pinning for maximum performance
```
This setting is not mentioned in performance docs.

Anyway, isn't `cpuset` flag needed to make full benefit of static CPU Manager Policy?

zimnx:
> Yes, but we do need the CPU manager, to control alignment of nodes to cores, no?

We do, but I don't see how `cpuset` is related. It's not a prerequisite for having Guaranteed QoS class, which is required for getting exclusive CPU.  

> We probably need the 1st and 3rd?

Yes.


> According to docs:
> 
> ```
> cpuset: Optional field. If it’s true, then the operator will start Scylla with cpu pinning for maximum performance
> ```
> 
> This setting is not mentioned in performance docs.

This isn't accurate, afaik cpu is pinned even without this setting.

> Anyway, isn't cpuset flag needed to make full benefit of static CPU Manager Policy?

No, Guaranteed QoS class is required.

soyacz:
I also did some training regarding scylla-operator, in Scylla University quiz cpu pinning is explicitly mentioned: 
[...]

PR #1130, opened 2022-12-27
CPU manager policy for QoS example

soyacz:
Setting the CPU manager policy to static seems to be required for guaranteed QoS and example misses that setting.

This commit adds this setting to the example.
Resolves https://github.com/scylladb/scylla-operator/issues/1129


soyacz:
wrong base branch

PR #1131, opened 2022-12-27
Fix NodeConfigPod controller to exclusively enqueue and sync Scylla Pods

### LINKED ISSUE #1119: Scylla is patching unrelated config maps
**Describe the bug**

Scylla is patching some of our config maps, removing the `controllerRef` field. These config maps are unrelated to scylla.

It seems to happen when config map additions are observed in `nodeconfigpod/controller.go`. The owners are enqueued, but having inspected the code, it seems that there is no check that this is a scylla pod (like what happens in `addPod`). Eventually the pod is being synced and the `controllerRef` on the config map is removed.

I have a change that also performs the check from `addPod` in `enqueueOwner`, I can make a PR if that's the desired solution.

**To Reproduce**
Adding a config map with a `controllerRef` and the controller pod running in the same namespace should cause this problem.

**Logs**
```
nodeconfigpod/controller.go:265] "Observed addition of ConfigMap" ConfigMap="default/<unrelated-cm>"
nodeconfigpod/controller.go:177] "Enqueuing owner" ConfigMap="default/<unrelated-cm>" Pod="default/<unrelated-pod>"
nodeconfigpod/controller.go:139] "Enqueuing Pod" Pod="default/<unrelated-pod>"
nodeconfigpod/sync.go:74] "Started syncing Pod" Pod="default/<unrelated-pod>" startTime="2022-12-12 09:31:35.922543176 +0000 UTC m=+522.945119537"
controllertools/configmapmanager.go:106] "Patching ConfigMap to remove its controllerRef" ConfigMap="default/<unrelated-cm>" ControllerRef="v1/Pod:<unrelated-pod>"
nodeconfigpod/sync.go:76] "Finished syncing Pod" Pod="default/<unrelated-pod>" duration="16.360041ms"
```

**Environment:**
 - Scylla version: 5.1.0
 - Scylla-operator version: e.g.: 1.7.4

rzetelskik:
As described in https://github.com/scylladb/scylla-operator/issues/1119, currently the NodeConfigPod controller removes ownerReferences from *any* ConfigMap that has a controllerRef to *any* Pod. This is undesired as the controller shouldn't affect any resources that the operator doesn't manage. This PR adds a condition verifying whether the owner of the ConfigMap is a Scylla Pod.

Resolves #1119

### DIFF
--- a/pkg/controller/nodeconfigpod/controller.go
+++ b/pkg/controller/nodeconfigpod/controller.go
@@ -165,7 +165,7 @@ func (ncpc *Controller) enqueueAllScyllaPodsOnNode(depth int, obj kubeinterfaces
 
 	klog.V(4).InfoSDepth(depth, "Enqueuing all pods on Node", "Pods", len(pods), "Node", klog.KObj(node))
 	for _, pod := range pods {
-		ncpc.handlers.EnqueueWithDepth(depth+1, pod, op)
+		ncpc.handlers.Enqueue(depth+1, pod, op)
 	}
 
 	return
@@ -212,7 +212,7 @@ func (ncpc *Controller) enqueueScyllaPod(depth int, obj kubeinterfaces.ObjectInt
 		return
 	}
 
-	ncpc.handlers.EnqueueWithDepth(depth+1, pod, op)
+	ncpc.handlers.Enqueue(depth+1, pod, op)
 }
 
 func (ncpc *Controller) addPod(obj interface{}) {
--- a/pkg/controller/nodeconfigpod/controller.go
+++ b/pkg/controller/nodeconfigpod/controller.go
@@ -147,6 +147,10 @@ func NewController(
 	return ncpc, nil
 }
 
+func (ncpc *Controller) enqueueScyllaPodFunc() controllerhelpers.EnqueueFuncType {
+	return ncpc.handlers.EnqueueWithFilterFunc(controllerhelpers.IsScyllaPod)
+}
+
 func (ncpc *Controller) enqueueAllScyllaPodsOnNode(depth int, obj kubeinterfaces.ObjectInterface, op controllerhelpers.HandlerOperationType) {
 	node := obj.(*corev1.Node)
 
@@ -199,26 +203,10 @@ func (ncpc *Controller) enqueueAllScyllaPodsForNodeConfig(depth int, obj kubeint
 	}
 }
 
-func (ncpc *Controller) enqueueScyllaPod(depth int, obj kubeinterfaces.ObjectInterface, op controllerhelpers.HandlerOperationType) {
-	pod := obj.(*corev1.Pod)
-
-	// TODO: extract and use a better label, verify the container
-	if pod.Labels == nil {
-		return
-	}
-
-	_, isScyllaPod := pod.Labels[naming.ClusterNameLabel]
-	if !isScyllaPod {
-		return
-	}
-
-	ncpc.handlers.Enqueue(depth+1, pod, op)
-}
-
 func (ncpc *Controller) addPod(obj interface{}) {
 	ncpc.handlers.HandleAdd(
 		obj.(*corev1.Pod),
-		ncpc.enqueueScyllaPod,
+		ncpc.enqueueScyllaPodFunc(),
 	)
 }
 
@@ -226,7 +214,7 @@ func (ncpc *Controller) updatePod(old, cur interface{}) {
 	ncpc.handlers.HandleUpdate(
 		old.(*corev1.Pod),
 		cur.(*corev1.Pod),
-		ncpc.enqueueScyllaPod,
+		ncpc.enqueueScyllaPodFunc(),
 		nil,
 	)
 }
@@ -234,7 +222,7 @@ func (ncpc *Controller) updatePod(old, cur interface{}) {
 func (ncpc *Controller) addConfigMap(obj interface{}) {
 	ncpc.handlers.HandleAdd(
 		obj.(*corev1.ConfigMap),
-		ncpc.handlers.EnqueueOwner,
+		ncpc.handlers.EnqueueOwnerFunc(ncpc.enqueueScyllaPodFunc()),
 	)
 }
 
@@ -242,7 +230,7 @@ func (ncpc *Controller) updateConfigMap(old, cur interface{}) {
 	ncpc.handlers.HandleUpdate(
 		old.(*corev1.ConfigMap),
 		cur.(*corev1.ConfigMap),
-		ncpc.handlers.EnqueueOwner,
+		ncpc.handlers.EnqueueOwnerFunc(ncpc.enqueueScyllaPodFunc()),
 		ncpc.deleteConfigMap,
 	)
 }
@@ -250,7 +238,7 @@ func (ncpc *Controller) updateConfigMap(old, cur interface{}) {
 func (ncpc *Controller) deleteConfigMap(obj interface{}) {
 	ncpc.handlers.HandleDelete(
 		obj,
-		ncpc.handlers.EnqueueOwner,
+		ncpc.handlers.EnqueueOwnerFunc(ncpc.enqueueScyllaPodFunc()),
 	)
 }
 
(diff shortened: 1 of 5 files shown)


zimnx (approved):
lgtm, thanks

pkg/controllerhelpers/handlers.go
```diff
@@ -89,29 +89,41 @@ func (h *Handlers[T]) EnqueueAll(depth int, untypedObj kubeinterfaces.ObjectInte
...
+			return
+		}
+
+		if owner.GetUID() != controllerRef.UID {
+			utilruntime.HandleError(err)
+			return
+		}
+
+		if precondition(owner) {
+			klog.V(4).InfoSDepth(depth, "Enqueuing owner", append([]any{"OwnerGVK", h.gvk, "OwnerRef", klog.KObj(owner), "OwnerUID", owner.GetUID()}, getObjectLogContext(obj, nil)...)...)
```
tnozicka, line 115: nit: would it be better to make the function extendable and use  `'!precondition` pattern?

pkg/controllerhelpers/handlers.go
```diff
@@ -89,29 +89,41 @@ func (h *Handlers[T]) EnqueueAll(depth int, untypedObj kubeinterfaces.ObjectInte
 	}
 }
 
-func (h *Handlers[QT]) EnqueueOwner(depth int, obj kubeinterfaces.ObjectInterface, operation HandlerOperationType) {
-	controllerRef := metav1.GetControllerOf(obj)
-	if controllerRef == nil {
-		return
-	}
+func (h *Handlers[QT]) EnqueueOwnerIf(precondition func(QT) bool) EnqueueFuncType {
+	return func(depth int, obj kubeinterfaces.ObjectInterface, operation HandlerOperationType) {
```
tnozicka, line 93: this one could be a named function

pkg/controllerhelpers/handlers.go
```diff
@@ -89,29 +89,41 @@ func (h *Handlers[T]) EnqueueAll(depth int, untypedObj kubeinterfaces.ObjectInte
 	}
 }
 
-func (h *Handlers[QT]) EnqueueOwner(depth int, obj kubeinterfaces.ObjectInterface, operation HandlerOperationType) {
-	controllerRef := metav1.GetControllerOf(obj)
-	if controllerRef == nil {
-		return
-	}
+func (h *Handlers[QT]) EnqueueOwnerIf(precondition func(QT) bool) EnqueueFuncType {
```
tnozicka, line 92: We should distinguish the function name from those that do that do the work as this one returns a function - those usually have `Func` suffix. How about `EnqueOwnerWithConditionFunc`

pkg/controllerhelpers/handlers.go
```diff
@@ -89,29 +89,41 @@ func (h *Handlers[T]) EnqueueAll(depth int, untypedObj kubeinterfaces.ObjectInte
...
+			utilruntime.HandleError(err)
+			return
+		}
+
+		if owner.GetUID() != controllerRef.UID {
+			utilruntime.HandleError(err)
+			return
+		}
+
+		if precondition(owner) {
```
tnozicka, line 114: actually similarly to
https://github.com/scylladb/scylla-operator/blob/c5c6c6ffabcacf504e7f39263da377a755f276a1/pkg/controller/nodeconfigpod/controller.go#L202-L216
would it be more generic to customize the enqueue function?

tnozicka (commented):
I think we should also add the check to sync too. Handlers are more about optimization, which is good as well, but there should be a single point of enforcement.

pkg/controllerhelpers/handlers.go
rzetelskik: It would, and that's what I wanted to do initially, but with `enqueueOwner` you'd have to duplicate the entire logic of extracting the owner.

tnozicka: i meant to plumb `enqueue` func, not `enqueueOwner` although that would need to pass it as well

pkg/controllerhelpers/handlers.go
rzetelskik: How? It uses a closure to pass the precondition. Unless you just meant assigning the anonymous func to a variable and returning it?
```
a := func(...) {...}
return a
```

pkg/controller/nodeconfigpod/controller.go
```diff
@@ -231,26 +225,36 @@ func (ncpc *Controller) updatePod(old, cur interface{}) {
 	)
 }
 
+func isScyllaPod(pod *corev1.Pod) bool {
+	// TODO: extract and use a better label, verify the container
```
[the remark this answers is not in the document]
rzetelskik: that's a leftover so I wasn't exactly sure what the author meant here :stuck_out_tongue:
I removed "extract and" from the TODO comment

zimnx (dismissed):

pkg/controller/nodeconfigpod/sync.go
```diff
@@ -39,6 +39,11 @@ func (ncpc *Controller) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("can't list pods: %w", err)
 	}
 
+	_, hasScyllaLabel := pod.Labels[naming.ClusterNameLabel]
```
tnozicka, line 42: can labels be nil? you've also made a function handling that int this PR

pkg/controller/nodeconfigpod/sync.go
```diff
@@ -39,6 +39,11 @@ func (ncpc *Controller) sync(ctx context.Context, key string) error {
...
+	if !hasScyllaLabel {
+		return nil
```
tnozicka, line 44: this should log a warning given we also pre-filter in the handlers unless there is some good path how it can land here in which case I'd still log it at least at info 4

pkg/controllerhelpers/handlers.go
tnozicka: nvm, it reads some of the object data as well, so it would have too many args to be a free standing function

pkg/controllerhelpers/handlers.go
tnozicka: I wanted to avoid duplication a version of the function in every variant (enqueue*, enqueueOwner*), see
https://github.com/tnozicka/scylla-operator/commit/803da0be8c47cd7f66fffed1edb2cf5d4c0ca568

rzetelskik: I see, thanks. Would you get rid of the `WithFilter` func though? I left it in as it allows for adding custom enqueue funcs without duplicating code.

pkg/controller/nodeconfigpod/sync.go
rzetelskik: good point, I missed that. Moved func to `naming` and used it here as well.

pkg/naming/labels.go
```diff
@@ -81,3 +82,17 @@ func mergeLabels(l1, l2 map[string]string) map[string]string {
 	}
 	return res
 }
+
+func IsScyllaPod(pod *corev1.Pod) bool {
```
tnozicka, line 86: i think this would be better fit for `/pkg/controllerhelpers`

pkg/controllerhelpers/handlers.go
tnozicka: The default rule is we don't leave the functions that are not used and bring them when needed.

I don't think it brings simplification because in this case, where the filtering isn't done in the middle, a composition can easily replace it, if not be even more readable:
```
func myEnqueue() {
 // filter
 // call enqueue
}
```
the withFilter could actually be modeled this way as well

rzetelskik: You're right, it's a bit cleaner to compose it this way. Updated. I hope we're on the same page now and we can merge it?

tnozicka: > I don't think it brings simplification because in this case, where the filtering isn't done in the middle, a composition can easily replace it

My point was that the with filter isn't needed at all
```
func (ncpc *Controller) enqueueScyllaPod(depth int, obj kubeinterfaces.ObjectInterface, op controllerhelpers.HandlerOperationType) {
	if !controllerhelpers.IsScyllaPod(pod.(*corev1.Pod)) {
		return
	}

	ncpc.handlers.Enqueue(depth+1, obj, op)
}
```

rzetelskik: It's not *needed*, but it abstracts away the depth increment nicely

tnozicka (dismissed):
/approve
/lgtm

rzetelskik:
rebased on current master, @zimnx please approve

zimnx (approved):

PR #1139, opened 2023-01-09
helm: allow setting custom CA when not using CertManager

glothriel:
If helm chart user disables CertManager integration, setting custom CA bundle is required.

Otherwise the whole feature doesn't make sense, because you won't be able to deploy resources, the validation during `kubectl apply` will fail with:

```
Error: Internal error occurred: failed calling webhook "webhook.scylla.scylladb.com": could not get REST client: unable to load root certificates: unable to parse bytes as PEM block

```

### DIFF
--- a/helm/scylla-operator/values.yaml
+++ b/helm/scylla-operator/values.yaml
@@ -41,6 +41,10 @@ webhook:
   # Name of a secret containing custom certificate
   # If not set and createSelfSignedCertificate is true, a name is generated using fullname
   certificateSecretName: ""
+  # If createSelfSignedCertificate=false, you need to supply your own certificate and also
+  # provide CA for the API server, so that it accepts your certificate. The value holds base64-encoded
+  # CA in PEM format.
+  caBundle: ""
 
 serviceAccount:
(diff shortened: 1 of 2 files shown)


glothriel:
I believe, the `required` part may be questionable. If you also feel so, let me know, I'm going to make it optional.

glothriel:
@rzetelskik did you guys notice this PR?

rzetelskik:
^ @zimnx ptal

helm/scylla-operator/values.yaml
```diff
@@ -41,6 +41,10 @@ webhook:
   # Name of a secret containing custom certificate
   # If not set and createSelfSignedCertificate is true, a name is generated using fullname
   certificateSecretName: ""
+  # If createSelfSignedCertificate=false, you need to supply your own certificate and also
```
zimnx, line 44:
```suggestion
  # If createSelfSignedCertificate is disabled, you need to supply your own certificate and also
```

glothriel:
Closing for now so it's not hanging open, I'll get back to it once i have some time

PR #1146, opened 2023-01-17
Fix NodeConfigDaemonController working on objects from a different shard

tnozicka:
When switching to generic controllerRef in #1127, somewhere between the rebases and splitting pieces into dedicated PRs from monitoring PR, a piece of code was lost that was ensuring the old behaviour for NodeConfigDaemonController.
https://github.com/scylladb/scylla-operator/pull/1127/files#diff-db044f9316b6b26534f5e4845170e8105167aea9e0b26f3b2338528b45620297L49-L56

This resulted in the different shards fighting over objects and going trough adopt/release loop. To manifest, it needs multiple kubernetes nodes running scylladb with tuning enabled. Unfortunately our presubmits have only a single node. This would have likely been caught by QA later on but we should still look into better presubmits eventually.

Thanks @zimnx for helping me track this down and figuring the missing piece.

```
E0117 09:58:14.017329       1 nodeconfigdaemon/controller.go:182] syncing key 'key' failed: can't sync jobs: can't create job scylla-operator-node-tuning/perftune-containers-e34847fa-8b52-49e7-b385-11d4772b529f: batch/v1, Kind=Job "scylla-operator-node-tuning/perftune-containers-e34847fa-8b52-49e7-b385-11d4772b529f" isn't controlled by us
I0117 09:58:14.017349       1 nodeconfigdaemon/sync.go:83] "Started sync" startTime="2023-01-17 09:58:14.017341953 +0000 UTC m=+49826.480182456"
I0117 09:58:14.017434       1 controllertools/refmanager.go:150] "Removing controllerRef" GVK="batch/v1, Kind=Job" Object="scylla-operator-node-tuning/perftune-node-3b938f41-3203-4975-ac6a-c043bfa8b846" ControllerRef="apps/v1/DaemonSet:scylladb"
I0117 09:58:14.017605       1 record/event.go:294] "Event occurred" object="scylla-operator-node-tuning/perftune-containers-e34847fa-8b52-49e7-b385-11d4772b529f" fieldPath="" kind="Job" apiVersion="batch/v1" type="Warning" reason="UpdateJobFailed" message="Failed to update Job scylla-operator-node-tuning/perftune-containers-e34847fa-8b52-49e7-b385-11d4772b529f: batch/v1, Kind=Job \"scylla-operator-node-tuning/perftune-containers-e34847fa-8b52-49e7-b385-11d4772b529f\" isn't controlled by us"
```

### DIFF
--- a/pkg/controllerhelpers/sync.go
+++ b/pkg/controllerhelpers/sync.go
@@ -37,18 +37,26 @@ func (f ControlleeManagerGetObjectsFuncs[CT, T]) PatchObject(ctx context.Context
 
 var _ ControlleeManagerGetObjectsInterface[kubeinterfaces.ObjectInterface, kubeinterfaces.ObjectInterface] = ControlleeManagerGetObjectsFuncs[kubeinterfaces.ObjectInterface, kubeinterfaces.ObjectInterface]{}
 
-func GetObjects[CT, T kubeinterfaces.ObjectInterface](
+func GetObjectsWithFilter[CT, T kubeinterfaces.ObjectInterface](
 	ctx context.Context,
 	controller metav1.Object,
 	controllerGVK schema.GroupVersionKind,
 	selector labels.Selector,
+	filterFunc func(T) bool,
 	control ControlleeManagerGetObjectsInterface[CT, T],
 ) (map[string]T, error) {
 	// List all objects to find even those that no longer match our selector.
 	// They will be orphaned in ClaimObjects().
-	objects, err := control.ListObjects(labels.Everything())
+	allObjects, err := control.ListObjects(labels.Everything())
 	if err != nil {
 		return nil, err
+	}
+
+	var objects []T
+	for i := range allObjects {
+		if filterFunc(allObjects[i]) {
+			objects = append(objects, allObjects[i])
+		}
 	}
 
 	return controllertools.NewControllerRefManager[T](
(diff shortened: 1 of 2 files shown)


pkg/controllerhelpers/sync.go
```diff
@@ -37,20 +37,28 @@ func (f ControlleeManagerGetObjectsFuncs[CT, T]) PatchObject(ctx context.Context
...
 	// They will be orphaned in ClaimObjects().
-	objects, err := control.ListObjects(labels.Everything())
+	allObjects, err := control.ListObjects(labels.Everything())
 	if err != nil {
 		return nil, err
 	}
 
+	var objects []T
+	for i := range allObjects {
+		if !filterFunc(allObjects[i]) {
```
zimnx, line 57: Shouldn't it append objects matching the filter? Doing the opposite is usually called filtering **out**.
Say if you have a set of numbers, if you apply isEven filter, then as a result you get even numbers. 
In this controller we want to select every job matching a selector, and running on the same host. Now we are selecting these Jobs that are matching the selector, then filter is checking whether it's not running on the node, and thanks to double negation, we get jobs running on the same node.

tnozicka: it should have ;)

tnozicka: i even think the previous version that got dropped did this. i guess that what happens while you have to code over while having lunch :D

zimnx (approved):
LGTM, thanks!

PR #1148, opened 2023-01-17
Fix support for immutable fields in Job Apply

zimnx:
Recent change adding generic apply function (#1138), introduced a regression in Apply function for batchv1.Jobs. Immutable fields weren't verified and delete/create logic wasn't triggered. Our NodeConfig controllers are changing immutable Pod Template when a new ScyllaDB Pod lands on a Node.


pkg/resourceapply/batch.go
```diff
@@ -16,7 +17,26 @@ func ApplyJobWithControl(
...
+		options,
+		nil,
+		func(required *batchv1.Job, existing *batchv1.Job) string {
+			if !equality.Semantic.DeepEqual(existing.Spec.Template, required.Spec.Template) {
+				return "spec.template is immutable"
+			}
+			if !equality.Semantic.DeepEqual(existing.Spec.Completions, required.Spec.Completions) {
+				return "spec.completions is immutable"
+			}
+			if required.Spec.ManualSelector != nil && *required.Spec.ManualSelector && !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
```
tnozicka, line 34: why the manual selector? I don't see that in
https://github.com/kubernetes/kubernetes/blob/fc69084bf1fd32e55256795a6db7a74d75679d8b/pkg/apis/batch/validation/validation.go#L377

tnozicka:
can we catch this in an e2e?

pkg/resourceapply/batch.go
zimnx: aligned with kube admission.

zimnx:
> can we catch this in an e2e?

possibly, it requires 2 ScyllaClusters with Guaranteed QoS, I hope we can fit those on our CI.

tnozicka:
> possibly, it requires 2 ScyllaClusters with Guaranteed QoS, I hope we can fit those on our CI.

Can you file an issue so we don't forget to extend the CI when we have the capacity in the future? (we don't have the resources today)

fyi, @vponomaryov to make sure this is at least covered by QA (maybe it already is)

tnozicka (approved):
/approve
/lgtm

thanks!

vponomaryov:
@zimnx @tnozicka 

Please, provide steps for reproducing so we could correctly cover the use case in our automation.

zimnx:
@vponomaryov Scenario is described here: https://github.com/scylladb/scylla-operator/issues/1149 i guess you have it covered in multitenant scenario

PR #1158, opened 2023-01-30
Switch cgroup manager from systemd to cgroupfs

zimnx:


tnozicka (approved):
let's try if it stabilizes the CI, there isn't much room to make it worse :)

tnozicka (approved):
/approve
/lgtm

PR #1159, opened 2023-02-01
Make preleases even for alpha version to be easy to consume

tnozicka:
When we tag an alpha we should do a release as well as its easier to share. I think we have initially avoided them because the alpha.0 is right after branching, and in theory could be limited in the regex, but I think consistency is better.


zimnx (approved):
makes sense, thanks

PR #1161, opened 2023-02-06
[WIP] Add (optional) extended dashboards in ScyllaDBMonitoring and drop the unmanaged setup

### LINKED ISSUE #1155: Add (optional) extended dashboards in ScyllaDBMonitoring and drop the unmanaged setup
Having introduced the new API, setup and management for monitoring of our clusters in https://github.com/scylladb/scylla-operator/pull/1111, we should update the monitoring section of our documentation with new deployment instructions and remove the outdated manifests from example files.

rzetelskik:
**Description of your changes:** WIP

Resolves #1155


assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   selector:
     {{- .endpointsSelector | toYAML | nindent 4 }}
-  jobLabel: scylla/cluster
+  jobLabel: app.kubernetes.io/name
```
tnozicka, line 8: this needs to stay - this is the label from the service so in the end, if you name your scyllacluster scylla, it will make sure there is a job=scylla label used by some dashboards

pkg/api/scylla/v1alpha1/types_monitoring.go
```diff
@@ -162,6 +162,16 @@ type ScyllaDBMonitoringSpec struct {
 
 	// components hold additional config for the monitoring components in use.
 	Components *Components `json:"components"`
+
+	// type determines what platform service model the monitoring is deployed for. Defaults to Platform. Valid options
+	// are Platform and SaaS.
+	// "Platform" means the monitoring is deployed in the ScyllaDB cluster owner's platform. A full monitoring setup
+	// designed for cluster administrators is deployed.
+	// "SaaS" means the monitoring is deployed for a SaaS offering. A simplified monitoring setup is deployed.
+	// +kubebuilder:default:="Platform"
```
tnozicka, line 171: sounds like breaking change

pkg/api/scylla/v1alpha1/types_monitoring.go
```diff
@@ -162,6 +162,16 @@ type ScyllaDBMonitoringSpec struct {
...
+	// +kubebuilder:validation:Enum="";"SaaS";"Platform"
```
tnozicka, line 172: why ""?

tnozicka (commented):
a few comments while scanning the approach

assets/monitoring/prometheus/v1/scylladbmanager.servicemonitor.yaml
```diff
@@ -4,14 +4,14 @@ metadata:
   name: scylla-manager-service-monitor
   namespace: scylla-manager
 spec:
-  jobLabel: "app.kubernetes.io/name"
+  jobLabel: app.kubernetes.io/name
   selector:
     matchLabels:
```
tnozicka, line 9: i think this would need to be configurable

assets/monitoring/prometheus/v1/scylladbmanager.servicemonitor.yaml
```diff
@@ -4,14 +4,14 @@ metadata:
...
       app.kubernetes.io/name: scylla-manager
```
tnozicka, line 10: i think this would need to be configurable

rzetelskik: Sure, that was only a WIP. Dropping the manager support until we have an automated manager setup as agreed.

assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml
rzetelskik: All of scylla-monitoring dashboards , including the serverless one, have "job" label hardcoded to either `scylla` or `scylla-manager`. This change breaks multiple metrics. What do you propose?

pkg/api/scylla/v1alpha1/types_monitoring.go
rzetelskik: Would you rather have no default then and make it required or use "SaaS" as default?

pkg/api/scylla/v1alpha1/types_monitoring.go
rzetelskik: I thought it was necessary to pass validation with implicit value. Removed, thanks.

assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml
rzetelskik: Discussed internally. Temporarily fixed job labels in metrics used by dashboards. This is supposed to be fixed in upstream dashboards in long term.

pkg/api/scylla/v1alpha1/types_monitoring.go
rzetelskik: @tnozicka ping

tnozicka: in v1alpha1 it should behave the same. with "alpha" you can break when introducing v1alpha2

it needs to be set in a way that it keeps behaving the same way (if you have deployed it already it shouldn't change your dashboards type when updating the operator+crd), which would be default=SaaS, at least for v1alpha1

rzetelskik: Right. I changed the default to `SaaS`, but explicitly set `Platform` in out examples. I believe it would make sense to change the default to `Platform` with the next API version though.

tnozicka: I feel the same, my point was purely about API compatibility guarantees

pkg/api/scylla/v1alpha1/types_monitoring.go
```diff
@@ -162,6 +162,16 @@ type ScyllaDBMonitoringSpec struct {
...
 	// components hold additional config for the monitoring components in use.
 	Components *Components `json:"components"`
+
+	// type determines what platform service model the monitoring is deployed for. Defaults to SaaS. Valid options
+	// are Platform and SaaS.
+	// "Platform" means the monitoring is deployed in the ScyllaDB cluster owner's platform. A full monitoring setup
+	// designed for cluster administrators is deployed.
+	// "SaaS" means the monitoring is deployed for a SaaS offering. A simplified monitoring setup is deployed.
+	// +kubebuilder:default:="SaaS"
+	// +kubebuilder:validation:Enum="SaaS";"Platform"
```
tnozicka, line 172:
```suggestion
	// type determines the deployment type of the monitoring setup. Can be "Platform" or "Saas". 
	// "Platform" sets up a detailed ScyllaDB monitoring, including the infrastructure.
	// "Saas" deploys monitoring setup focusing on the ScyllaDB service.
	// Defaults to "SaaS".
	// +kubebuilder:validation:Enum="SaaS";"Platform"
	// +kubebuilder:default:="SaaS"
```

tnozicka: (i think it's better to first introduce the values and then say what the default is)

pkg/controller/scylladbmonitoring/sync_grafana.go
```diff
@@ -140,9 +140,18 @@ func makeGrafanaConfigs(sm *scyllav1alpha1.ScyllaDBMonitoring) (*corev1.ConfigMa
...
+	case naming.ScyllaDBMonitoringTypePlatform:
+		return grafanav1alpha1assets.GrafanaPlatformDashboardsConfigMapTemplate.RenderObject(map[string]any{
+			"scyllaDBMonitoringName": sm.Name,
+		})
+	case naming.ScyllaDBMonitoringTypeSaaS:
+		return grafanav1alpha1assets.GrafanaSaaSDashboardsConfigMapTemplate.RenderObject(map[string]any{
+			"scyllaDBMonitoringName": sm.Name,
+		})
+	default:
+		return nil, "", fmt.Errorf("unkown type: %s", sm.Spec.Type)
```
tnozicka, line 153: nit: `switch t:=`
also `unknown monitoring type`

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+```
+
+For details, refer to the below command:
+```console
+$ kubectl explain scylladbmonitorings.scylla.scylladb.com/v1alpha1
+```
+
+## Deploy managed monitoring
+
+**Note**: as of v1.8, ScyllaDBMonitoring is in alpha state and is susceptible to backwards incompatible changes.
```
tnozicka, line 44:
```suggestion
**Note**: as of v1.8, ScyllaDBMonitoring is experimental. The API is currently at v1alpha1 version and may change in the future versions.
```

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+
+### Requirements
+
+Before you can set up your ScyllaDB Monitoring, you must install the Scylla Operator in your Kubernetes cluster
```
tnozicka, line 48:
```suggestion
Before you can set up your ScyllaDB Monitoring, you have to have Scylla Operator already installed in your Kubernetes cluster
```

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+```
+
+## Deploy managed monitoring
+
+**Note**: as of v1.8, ScyllaDBMonitoring is in alpha state and is susceptible to backwards incompatible changes.
+
+### Requirements
+
+Before you can set up your ScyllaDB Monitoring, you must install the Scylla Operator in your Kubernetes cluster 
+and have an instance of ScyllaCluster running. For details see:
```
tnozicka, line 49: i don't think that's technically true

tnozicka: You can find more information on how to deploy Scylla Operator here:

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+* [Deploying Scylla on a Kubernetes Cluster](generic.md)
+* [Deploying Scylla stack using Helm Charts](helm.md)
+
+The managed monitoring setup also requires HAProxy Ingress and Prometheus Operator deployed in your Kubernetes cluster.
```
tnozicka, line 53: it's not required by the monitoring itself, it's just that the example uses it

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+You can deploy them using the provided third party examples.
+
+#### Deploy Prometheus Operator
+Deploy Prometheus Operator using kubectl:
```
tnozicka, line 57: if your cluster already has a prometheus operator deployed you should skip this step

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+```console
+$ kubectl -n prometheus-operator apply --server-side -f examples/third-party/prometheus-operator
+```
+
+##### Verify Prometheus Operator status
+```console
+$ kubectl -n prometheus-operator get deployments.apps/prometheus-operator
```
tnozicka, line 64: kubectl rollout status

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+NAME                  READY   UP-TO-DATE   AVAILABLE   AGE
+prometheus-operator   1/1     1            1           5s
+```
+
+#### Deploy HAProxy Ingress
+Deploy HAProxy Ingress using kubectl:
```
tnozicka, line 70: same skip here

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+```console
+$ kubectl -n haproxy-ingress apply --server-side -f examples/third-party/haproxy-ingress
+```
+
+##### Verify HAProxy Ingress status
+```console
+$ kubectl -n haproxy-ingress get deployments.apps/haproxy-ingress
```
tnozicka, line 77: rollout status

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+NAME              READY   UP-TO-DATE   AVAILABLE   AGE
+haproxy-ingress   1/1     1            1           5s
+```
+
+Before you begin, update the `endpointSelector` in `examples/monitoring/v1alpha1/scylladbmonitoring.yaml` with a label 
+matching your ScyllaCluster instance.
```
tnozicka, line 83: instance name.

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+
+### Deploy ScyllaDBMonitoring
+
+Deploy the monitoring setup using kubectl:
+```console
+$ kubectl apply --server-side -f examples/monitoring/v1alpha1/scylladbmonitoring.yaml
```
tnozicka, line 89:
```suggestion
$ kubectl apply --server-side -f ./examples/monitoring/v1alpha1/scylladbmonitoring.yaml
```

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+```
+
+The Operator should pick up the object's creation and create all the necessary resources for you.
```
tnozicka, line 92:
```suggestion
Scylla Operator will notice the new ScyllaDBMonitoring object and create and manage all the necessary resources for you.
```

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+
+#### Verify ScyllaDBMonitoring status
+```console
+$ kubectl get ScyllaDBMonitoring example
```
tnozicka, line 96: wait for conditions

docs/source/monitoring.md
```diff
@@ -1,76 +1,130 @@
...
+
+#### Verify Grafana admin credentials secret
+Among other resources, the Operator creates a secret containing the admin credentials for you Grafana deployment.
+```console
+$ kubectl get secrets/example-grafana-admin-credentials
+NAME                                TYPE     DATA   AGE
+example-grafana-admin-credentials   Opaque   2      68s
+```
+
+### Access Grafana locally
```
tnozicka, line 123: I think this should be a variant when you don't want to use/install haproxy, or expose it to the world

examples/monitoring/v1alpha1/scylladbmonitoring.yaml
```diff
@@ -26,4 +26,3 @@ spec:
             - test-grafana.test.svc.cluster.local
             annotations:
               haproxy-ingress.github.io/ssl-passthrough: "true"
-
```
tnozicka, line 28: why?

pkg/api/scylla/v1alpha1/types_monitoring.go
rzetelskik: It sort of copied it from core Service object, but I don't mind either.

docs/source/monitoring.md
rzetelskik: I've put a message above saying you can skip these steps if you already have them deployed in your cluster.

docs/source/monitoring.md
rzetelskik: Do you suggest removing it or adding it as an alternative?

docs/source/monitoring.md
rzetelskik: There are multiple so I had to use a workaround

docs/source/monitoring.md
tnozicka: I think we'd primarily describe a way to use the ingress, this would be an alternative, if needed

pkg/api/scylla/v1alpha1/types_monitoring.go
```diff
@@ -162,6 +162,15 @@ type ScyllaDBMonitoringSpec struct {
 
 	// components hold additional config for the monitoring components in use.
 	Components *Components `json:"components"`
+
+	// type determines the platform type of the monitoring setup. Can be "Platform" or "SaaS".
+	// "Platform" sets up a detailed ScyllaDB monitoring, including the infrastructure.
+	// "Saas" deploys monitoring setup focusing on the ScyllaDB service.
+	// Defaults to "SaaS".
+	// +kubebuilder:validation:Enum="SaaS";"Platform"
```
zimnx, line 170: Since enumeration values are part of the API, shouldn't these be declared as consts in api package instead in `naming`?

pkg/api/scylla/v1alpha1/types_monitoring.go
```diff
@@ -162,6 +162,15 @@ type ScyllaDBMonitoringSpec struct {
...
+	// +kubebuilder:default:="SaaS"
+	// +optional
+	Type string `json:"type,omitempty"`
```
zimnx, line 173: ScyllaDBMonitoringType instead of string

pkg/api/scylla/v1alpha1/types_monitoring.go
rzetelskik: good point, changed

docs/source/monitoring.md
rzetelskik: Also asked on https://github.com/kubernetes/kubernetes/issues/95759 if I can give fixing this a try

rzetelskik:
@zimnx @tnozicka thanks, I believe I've addressed all of your comments

zimnx (dismissed):
lgtm, thanks

tnozicka:
> Note: we decided not to include ScyllaDB Manager dashboards until we start managing ScyllaDB Manager deployments with Operator.

Given the plan for manager is changing, I'd use something more general, like:
`Note: We have decided not to include ScyllaDB Manager dashboards yet until we figure out the how we'll integrate manager functionality in a better way.`

docs/source/monitoring.md
```diff
@@ -1,76 +1,148 @@
...
+
+#### Verify Grafana admin credentials secret
+Among other resources, the Operator creates a secret containing the admin credentials for you Grafana deployment.
+```console
+$ kubectl get secrets/example-grafana-admin-credentials
+NAME                                TYPE     DATA   AGE
+example-grafana-admin-credentials   Opaque   2      68s
+```
+
+### Access Grafana
```
tnozicka, line 123: I feel that this article is too focused on how to hack it with minikube and someone trying to set this up on a real cluster / domain would be confused.

I think the generic steps are:
- create ScyllaDBMonitoring with a domain (`test-grafana.test.svc.cluster.local` is fine)
- [real domain / cname to ingress controller] curl
- [fake domain / port-forward / not a cname to ingress controller] curl + `--resolve`

pkg/api/scylla/v1alpha1/types_monitoring.go
```diff
@@ -170,7 +177,7 @@ type ScyllaDBMonitoringSpec struct {
 	// +kubebuilder:validation:Enum="SaaS";"Platform"
 	// +kubebuilder:default:="SaaS"
 	// +optional
-	Type string `json:"type,omitempty"`
+	Type ScyllaDBMonitoringType `json:"type,omitempty"`
```
tnozicka, line 180: ok, it's here but there needs to be just one commit  touching this enum

pkg/naming/constants.go
```diff
@@ -148,10 +148,3 @@ type ProtocolDNSLabel string
 const (
 	CQLProtocolDNSLabel = "cql"
 )
-
-type ScyllaDBMonitoringType string
```
tnozicka: you need to clean your commits

tnozicka (changes requested):

pkg/api/scylla/v1alpha1/types_monitoring.go
tnozicka: now i see it has both the new and old changes which needs cleanup

assets/monitoring/grafana/v1alpha1/dashboards-saas.cm.yaml
```diff
@@ -4602,15 +4602,15 @@ data:
                             ]
                         },
                         "datasource": "prometheus",
-                        "definition": "query_result(count(up{job=\"scylla\"}) by (dc))",
+                        "definition": "query_result(count(up{job=~\"$cluster|$^\"}) by (dc))",
```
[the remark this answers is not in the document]
rzetelskik: Sent https://github.com/scylladb/scylla-operator/pull/1199

docs/source/monitoring.md
```diff
@@ -1,76 +1,148 @@
...
+deployment "example-grafana" successfully rolled out
+```
+
+#### Verify Grafana admin credentials secret
+Among other resources, the Operator creates a secret containing the admin credentials for you Grafana deployment.
+```console
+$ kubectl get secrets/example-grafana-admin-credentials
+NAME                                TYPE     DATA   AGE
+example-grafana-admin-credentials   Opaque   2      68s
+```
```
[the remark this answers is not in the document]
rzetelskik: Yeah it doesn't really make sense, I left it there to demonstrate how to access the admin credentials. Changed.

docs/source/monitoring.md
```diff
@@ -1,76 +1,142 @@
...
+$ kubectl -n haproxy-ingress apply --server-side -f ./examples/third-party/haproxy-ingress
+```
+
+##### Verify HAProxy Ingress status
+```console
+$ kubectl -n haproxy-ingress rollout status --timeout=5m deployments.apps/haproxy-ingress
+deployment "haproxy-ingress" successfully rolled out
+```
+
+Before you begin, update the `endpointSelector` in `examples/monitoring/v1alpha1/scylladbmonitoring.yaml` with a label
```
vponomaryov, line 81: The code uses plural form of the first word -> `endpointsSelector`.

docs/source/monitoring.md
rzetelskik: > - [real domain / cname to ingress controller] curl
> - [fake domain / port-forward / not a cname to ingress controller] curl + --resolve

I don't understand these lines. Could you elaborate on what these options in the brackets mean? Do you mean that there should be two sections demonstrating all of these options, verifying them with either curl or curl --resolve? Are you also suggesting not using minikube to demonstrate it?

tnozicka: I think that could be a joint section saying if your scylldbmonitoring domain isn't resolvable (to the ingress controller IP accessible from outside of the cluster), then you have to add the --resolve flag with ingress controller IP or use `/etc/hosts` alias. The ingress controller IP (accessible from outside of the cluster) can be constructed in a few ways - as service externalIP, nodeIP +  service nodePort, or 127.0.0.1+kubectl port-forward to ingress controller svc, ...

docs/source/monitoring.md
```diff
@@ -1,76 +1,142 @@
...
+```
+
+Scylla Operator will notice the new ScyllaDBMonitoring object, and it will reconcile all necessary resources.
+
+#### Verify ScyllaDBMonitoring status
+```console
+$ kubectl wait --for='condition=Available=True' scylladbmonitorings.scylla.scylladb.com/example
+scylladbmonitoring.scylla.scylladb.com/example condition met
+
+$ kubectl wait --for='condition=Progressing=False' scylladbmonitorings.scylla.scylladb.com/example
```
tnozicka, line 98: with this raciness, you should check progressing first and available last. it can be still available for a while when it starts progressing and when it rolls out you won't check it again

rzetelskik: good point, updated

docs/source/monitoring.md
```diff
@@ -1,76 +1,234 @@
...
+HTTP/1.1 200 OK
+Cache-Control: no-cache
+Content-Type: text/html; charset=UTF-8
+Expires: -1
+Pragma: no-cache
+X-Content-Type-Options: nosniff
+X-Frame-Options: deny
+X-Xss-Protection: 1; mode=block
+Date: Fri, 07 Apr 2023 14:32:09 GMT
+```
```
tnozicka: I've tried to simplify this and also some of the templates. Please verify the instructions.

### Accessing Grafana

For accessing Grafana service from outside the Kubernetes cluster we'd recommend using an Ingress but there are many other ways. What matters when using the Ingress is to direct your packets to the ingress controller service/pods and have the correct TLS SNI field set by the caller when reaching out the service, so it is routed properly, and it can successfully validate the grafana serving certificate. Obviously this is easier when you are using a real DNS domain that resolves to your Ingress controller IP address but most of the client/tools allow to set the SNI field manually.

#### Prerequisites

For accessing Grafana, you first need to collect the serving CA and the credentials.

```console
$ GRAFANA_SERVING_CERT="$( kubectl -n scylla get secret/example-grafana-serving-ca --template '{{ index .data "tls.crt" }}' | base64 -d )"
$ GRAFANA_USER="$( kubectl -n scylla get secret/example-grafana-admin-credentials --template '{{ index .data "username" }}' | base64 -d )"
$ GRAFANA_PASSWORD="$( kubectl -n scylla get secret/example-grafana-admin-credentials --template '{{ index .data "password" }}' | base64 -d )"
```

#### Connecting through Ingress using a resolvable domain

In case of production cluster someone has already set up the Ingress controller and appropriate DNS records. Often there is a generic wildcard record like `*.app.mydomain` already pointing to the ingress controller external IP. For custom service domains, these are usually a CNAME pointing the Ingress controller A record.

Note: The ScyllaDBMonitoring example creates an Ingress object with `test-grafana.test.svc.cluster.local` DNS domain - you need to adjust this to your domain. Here it is represented as `example-grafana.apps.mydomain`.

Note: To test a resolvable domain from your machine without creating DNS records, you can adjust `/etc/hosts` or similar.

```console
$ curl --fail -s -o /dev/null -w '%{http_code}' -L --cacert <( echo "${GRAFANA_SERVING_CERT}" ) "https://example-grafana.apps.mydomain" --user "${GRAFANA_USER}:${GRAFANA_PASSWORD}"
200
```

#### Connecting through Ingress using an unresolvable domain

To connect to an ingress without a resolvable domain you first need to find out your Ingress controller IP that can be resolved externally. Again there are more ways to do it, not limited to the few examples shown here.

Unless stated otherwise, we will assume your ingress is running on port 443.
```console
$ INGRESS_PORT=443
```

##### Variants
###### Ingress ExternalIP

When you are running in a real cluster there is usually a cloud LB or a bare metal alternative providing you with an externally reachable IP.

```console
$ INGRESS_IP="$( kubectl -n=haproxy-ingress get service/haproxy-ingress --template='{{ ( index .loadBalancer.ingress 0 ).ip }}' )"
```

###### Ingress NodePort

NodePort is a bit of a less convenient but it's available in dev clusters as well.

```console
# FIXME: the template needs to prefer ExternalIP, if set. 
$ INGRESS_IP="$( kubectl get nodes --template='{{ range ( index .items 0 ).status.addresses }}{{ if eq .type "InternalIP" }}{{ .address }}{{ end }}{{ end }}' )"
$ INGRESS_PORT="$( kubectl -n=haproxy-ingress get services/haproxy-ingress --template='{{ range .spec.ports }}{{ if eq .port 443 }}{{ .nodePort }}{{ end }}{{ end }}' )"
```

###### Ingress Port Forwarding

Port forwarding may be useful, if your nodes are not reachable through a public IP, nor your Ingress, or in general for development purposes. 

```console
$ kubectl -n=haproxy-ingress port-forward services/haproxy-ingress 5000:443 >/dev/null &
$ INGRESS_IP="localhost"
$ INGRESS_PORT="5000"
```

##### Connection

```console
$ curl --fail -s -o /dev/null -w '%{http_code}' -L --cacert <( echo "${GRAFANA_SERVING_CERT}" ) "https://test-grafana.test.svc.cluster.local"  --resolve "test-grafana.test.svc.cluster.local:${INGRESS_PORT}:${INGRESS_IP}" --user "${GRAFANA_USER}:${GRAFANA_PASSWORD}"
200
```

pkg/controller/scylladbmonitoring/sync_grafana.go
```diff
@@ -140,9 +140,18 @@ func makeGrafanaConfigs(sm *scyllav1alpha1.ScyllaDBMonitoring) (*corev1.ConfigMa
 }
 
 func makeGrafanaDashboards(sm *scyllav1alpha1.ScyllaDBMonitoring) (*corev1.ConfigMap, string, error) {
-	return grafanav1alpha1assets.GrafanaDashboardsConfigMapTemplate.RenderObject(map[string]any{
-		"scyllaDBMonitoringName": sm.Name,
-	})
+	switch t := *sm.Spec.Type; t {
```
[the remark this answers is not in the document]
rzetelskik: It has a default value, so it should never happen. I've added a sanity check though.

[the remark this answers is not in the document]
rzetelskik: Good point! I assumed it would never happen because of the default value, but I haven't thought of the version skew.

assets/monitoring/grafana/v1alpha1/registry.go
```diff
@@ -36,9 +36,13 @@ var (
 	grafanaProvisioningConfigMapTemplateString string
 	GrafanaProvisioningConfigMapTemplate       = ParseObjectTemplateOrDie[*corev1.ConfigMap]("grafana-provisioning-cm", grafanaProvisioningConfigMapTemplateString)
 
-	//go:embed "dashboards.cm.yaml"
-	grafanaDashboardsConfigMapTemplateString string
-	GrafanaDashboardsConfigMapTemplate       = ParseObjectTemplateOrDie[*corev1.ConfigMap]("grafana-dashboard-cm", grafanaDashboardsConfigMapTemplateString)
```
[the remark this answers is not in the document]
rzetelskik: It doesn't get pruned but it get updated, yes.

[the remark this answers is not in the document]
tnozicka: oh, right, the name is still the same, thx

docs/source/monitoring.md
rzetelskik: I guess we had a different perspective on how this should look like but this looks good to me - I'll vet the commands and fix the typos and other errors.

rzetelskik:
@tnozicka thanks for the review. As discussed internally I've dropped the `port-forward` part and will add it once I've investigated the issue. I've changed the wording of the last section here and there and fixed some errors so you might want to double check.

pkg/controller/scylladbmonitoring/sync_grafana.go
```diff
@@ -140,9 +140,24 @@ func makeGrafanaConfigs(sm *scyllav1alpha1.ScyllaDBMonitoring) (*corev1.ConfigMa
 }
 
 func makeGrafanaDashboards(sm *scyllav1alpha1.ScyllaDBMonitoring) (*corev1.ConfigMap, string, error) {
-	return grafanav1alpha1assets.GrafanaDashboardsConfigMapTemplate.RenderObject(map[string]any{
-		"scyllaDBMonitoringName": sm.Name,
-	})
+	if sm.Spec.Type == nil {
+		// This should have a default value, but it can be nil due to a version skew.
+		scyllaDBMonitoringTypeSAAS := scyllav1alpha1.ScyllaDBMonitoringTypeSAAS
+		sm.Spec.Type = &scyllaDBMonitoringTypeSAAS
```
tnozicka, line 146: this will break the informers, you can't write to the shared cache. pls make a variable for the type to switch on

docs/source/monitoring.md
```diff
@@ -1,76 +1,180 @@
...
+### Accessing Grafana
 
-   First you need to download Scylla Monitoring, which contains Grafana dashboards and custom Prometheus rules.
-   You can do this by running the following command:
-   ```
-   mkdir scylla-monitoring
-   curl -L https://github.com/scylladb/scylla-monitoring/tarball/branch-4.0 | tar -xzf - -C scylla-monitoring --strip-components=1
-   ```
+For accessing Grafana service from outside the Kubernetes cluster we recommend using an Ingress, although there are many other ways to do so.
+When using Ingress, what matters is to direct your packets to the ingress controller Service/Pods and have the correct TLS SNI field set by the caller when reaching out to the service, so it is routed properly, and it can successfully validate the grafana serving certificate.
```
tnozicka, line 120:
```suggestion
When using Ingress, what matters is to direct your packets to the ingress controller Service/Pods and have the correct TLS SNI field set by the caller when reaching out to the service, so it is routed properly, and your client can successfully validate the grafana serving certificate.
```

docs/source/monitoring.md
```diff
@@ -1,76 +1,180 @@
...
+```
 
-   You can find it on `http://127.0.0.1:3000` and login with the credentials `admin`:`admin`.
\ No newline at end of file
+##### Ingress NodePort
+
+NodePort is slightly less convenient, but it's available in development clusters as well.
+
+```console
+$ INGRESS_IP="$( kubectl get nodes --template='{{ $internal_ip := "" }}{{ $external_ip := "" }}{{ range ( index .items 0 ).status.addresses }}{{ if eq .type "InternalIP" }}{{ $internal_ip = .address }}{{ else if eq .type "ExternalIP" }}{{ $external_ip = .address }}{{ end }}{{ end }}{{ if $external_ip }}{{ $external_ip }}{{ else }}{{ $internal_ip }}{{ end }}' )"
```
tnozicka, line 171: Shouldn't we prefer the `ExternalIP` first and `InternalIP` as a fallback as it work only if you are in the same network like with your VM but not for say a cloud? The other way around both should work with a fallback.

tnozicka (changes requested):

docs/source/monitoring.md
tnozicka: you are right, I've misread while scrolling

pkg/controller/scylladbmonitoring/sync_grafana.go
rzetelskik: Fixed. Btw should I add a test for `makeGrafanaDashboards`?

pkg/controller/scylladbmonitoring/sync_grafana.go
```diff
@@ -140,9 +140,25 @@ func makeGrafanaConfigs(sm *scyllav1alpha1.ScyllaDBMonitoring) (*corev1.ConfigMa
...
-		"scyllaDBMonitoringName": sm.Name,
-	})
+	smType := sm.Spec.Type
+	if smType == nil {
+		// This should have a default value, but it can be nil due to a version skew.
+		scyllaDBMonitoringTypeSAAS := scyllav1alpha1.ScyllaDBMonitoringTypeSAAS
+		smType = &scyllaDBMonitoringTypeSAAS
+	}
+
+	switch t := *smType; t {
```
tnozicka, line 150: there is no need for this to be a pointer
```suggestion
	t := scyllav1alpha1.ScyllaDBMonitoringTypeSAAS
	if sm.Spec.Type != nil {
	    t = *sm.Spec.Type
	}
	
	
	switch t {
```

pkg/controller/scylladbmonitoring/sync_grafana.go
tnozicka: I don't mind wither way but I'd not hold this over it

rzetelskik: Let's go ahead then

tnozicka (approved):
/approve
/lgtm
thanks!

mykaul:
> /approve /lgtm thanks!

Hallelujah!

PR #1164, opened 2023-02-09
Fix serial job artifact names

zimnx:
Names of artifacts from serial jobs were swapped.
Serial job produced 'serial-alpha' artifact, and Serial with alpha flags produced 'serial'.


tnozicka (approved):
/approve
/lgtm

thanks

PR #1165, opened 2023-02-09
Add support for tuning EKS 1.24 using containerd CRI

zimnx:
Path to cpuset in cgroup is different on 1.24 EKS using containerd than 1.23 using docker. Our tuning machinary wasn't able to get information about Scylla Pod cpuset and hence calculate IRQ cpumask.


tnozicka (approved):
/approve
/lgtm

thanks!

tnozicka (approved):

PR #1168, opened 2023-02-14
Fix kubelet cgroup slices in CI setup

tnozicka:
This bring back the systemd driver as switching back to cgroups in #1158 didn't help and it is recommended to run systemd driver on systems running with systemd.

Hopefully a lot of flaked / the borked CI

### DIFF
--- a/.github/actions/setup-kubernetes/action.yaml
+++ b/.github/actions/setup-kubernetes/action.yaml
@@ -54,7 +54,7 @@ runs:
       cat << EOF | sudo tee /etc/crio/crio.conf.d/02-cgroup-manager.conf
       [crio.runtime]
       conmon_cgroup = "pod"
-      cgroup_manager = "cgroupfs"
+      cgroup_manager = "systemd"
       EOF
       
       sudo systemctl enable --now crio
@@ -110,10 +110,21 @@ runs:
       kind: InitConfiguration
       nodeRegistration:
         criSocket: "unix:///var/run/crio/crio.sock"
+        kubeletExtraArgs:
+          runtime-cgroups: /system.slice/crio.service
       ---
-      kind: KubeletConfiguration
       apiVersion: kubelet.config.k8s.io/v1beta1
-      cgroupDriver: cgroupfs
+      kind: KubeletConfiguration
+      cgroupDriver: systemd
+      cgroupRoot: /
+      systemCgroups: /system.slice
+      kubeletCgroups: /system.slice/kubelet.service
+      enforceNodeAllocatable:
+      - pods
+      systemReservedCgroup: /system.slice
+      kubeReservedCgroup: /system.slice/kubelet.service
+      systemReserved:
+        ephemeral-storage: 1Gi
       EOF
       
       sudo kubeadm init --config=/root/kubeadm-config.yaml


.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,34 @@ runs:
       kind: InitConfiguration
       nodeRegistration:
         criSocket: "unix:///var/run/crio/crio.sock"
+        kubeletExtraArgs:
+          container-runtime: remote
```
rzetelskik, line 114: Isn't this set by default? As it's the only accepted value now?

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,34 @@ runs:
...
       apiVersion: kubelet.config.k8s.io/v1beta1
-      cgroupDriver: cgroupfs
+      kind: KubeletConfiguration
+      cgroupDriver: systemd
+      cgroupRoot: /
+      systemCgroups: /system.slice
+      kubeletCgroups: /system.slice/kubelet.service
+      enforceNodeAllocatable:
+      - pods
+      #- kube-reserved # try again when we bump system to get cgroupsv2
```
rzetelskik, line 125: Do you think you could remove the commented out lines and create a tracking issue instead? With a note saying why you couldn't do it here?

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,34 @@ runs:
...
+      #- system-reserved # try again when we bump system to get cgroupsv2
+      systemReservedCgroup: /system.slice
+      kubeReservedCgroup: /system.slice/kubelet.service
+      systemReserved:
+      #  cpu: "200m" # try again when we bump system to get cgroupsv2
+        ephemeral-storage: 1Gi
+      #kubeReserved:
+      #  cpu: "200m" # try again when we bump system to get cgroupsv2
+      #  ephemeral-storage: "1Gi"
+      serializeImagePulls: false
```
rzetelskik, line 135: Should it be introduced in this PR?

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,34 @@ runs:
...
+      containerLogMaxSize: 50Mi
+      #cpuManagerPolicy: static # try again when we bump system to get cgroupsv2
+      maxPods: 50
```
rzetelskik, line 138: Are these just experimental values? Or is there some specific reasoning behind them? If so, could you add it to the PR description?

rzetelskik (commented):
I'm not qualified to say whether the config is correct altogether, but from looking at the docs/reference it seems sane to me. Left a few comments.

.github/actions/setup-kubernetes/action.yaml
tnozicka: more of a failsafe, but I can drop it if you want - it's the default currently

.github/actions/setup-kubernetes/action.yaml
tnozicka: I'll try to restructure it into multiple PRs, hopefully the next one being able to be based on top of ubuntu 22.04 bump and see where we stand

.github/actions/setup-kubernetes/action.yaml
tnozicka: I don't want to do a PR per line in this case, but I'll try a minimal change for the slices and bring the rest as tunning PR

.github/actions/setup-kubernetes/action.yaml
tnozicka: a baseline to have a starting point

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,22 @@ runs:
...
+      cgroupRoot: /
+      systemCgroups: /system.slice
+      kubeletCgroups: /system.slice/kubelet.service
+      enforceNodeAllocatable:
+      - pods
+      - system-reserved
+      systemReservedCgroup: /system.slice
+      kubeReservedCgroup: /system.slice/kubelet.service
+      systemReserved:
+        ephemeral-storage: 1Gi
```
zimnx, line 127: https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
> systemReserved is a set of ResourceName=ResourceQuantity (e.g. cpu=200m,memory=150G) pairs that describe resources reserved for non-kubernetes components. **Currently only cpu and memory are supported**. See http://kubernetes.io/docs/user-guide/compute-resources for more detail. Default: nil

Are you sure storage is supported? It is mentioned in kubeReserved that it is, but not in systemReserved

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,22 @@ runs:
...
       EOF
```
zimnx, line 128: add `hairpinMode: "hairpin-veth"`, Scylla requires it for TRUNCATE queries

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,22 @@ runs:
...
+      kind: KubeletConfiguration
+      cgroupDriver: systemd
+      cgroupRoot: /
+      systemCgroups: /system.slice
+      kubeletCgroups: /system.slice/kubelet.service
+      enforceNodeAllocatable:
+      - pods
+      - system-reserved
+      systemReservedCgroup: /system.slice
+      kubeReservedCgroup: /system.slice/kubelet.service
```
zimnx, line 125: EKS isolates kubelet from system, any reason why it's different here?
```
  "systemReservedCgroup": "/system",
  "kubeReservedCgroup": "/runtime"
```

.github/actions/setup-kubernetes/action.yaml
```diff
@@ -110,10 +110,22 @@ runs:
...
       apiVersion: kubelet.config.k8s.io/v1beta1
-      cgroupDriver: cgroupfs
+      kind: KubeletConfiguration
+      cgroupDriver: systemd
+      cgroupRoot: /
+      systemCgroups: /system.slice
+      kubeletCgroups: /system.slice/kubelet.service
+      enforceNodeAllocatable:
+      - pods
+      - system-reserved
```
rzetelskik, line 124: this was commented out before, did you add it on purpose?

tnozicka: yep, trying it out on its own still a fail because GH CI is missing a cgroup path that is present on my brand new 20.04 ubuntu VM :shrug: (going back)

.github/actions/setup-kubernetes/action.yaml
tnozicka: my uderstanding is that this describes where the cgroup for the systemd unit is, not where you want it to be

.github/actions/setup-kubernetes/action.yaml
tnozicka: thx, pls save it for the other "tune PR" as I was asked to keep this only for the slices fix

.github/actions/setup-kubernetes/action.yaml
tnozicka: I've checked and OKD uses it  as well and the code looks generic enough to propagate to node status https://github.com/kubernetes/kubernetes/blob/64af1adaceba4db8d0efdb91453bce7073973771/pkg/kubelet/nodestatus/setters.go#L344-L355

seems like a stale comment

zimnx: Looks like `system-reserved` entry is required in `enforceNodeAllocatable` to reserve it. From kubelet source:
```
if nc.EnforceNodeAllocatable.Has(kubetypes.SystemReservedEnforcementKey) {
		klog.V(2).InfoS("Enforcing system reserved on cgroup", "cgroupName", nc.SystemReservedCgroupName, "limits", nc.SystemReserved)
		if err := enforceExistingCgroup(cm.cgroupManager, cm.cgroupManager.CgroupName(nc.SystemReservedCgroupName), nc.SystemReserved); err != nil {
			message := fmt.Sprintf("Failed to enforce System Reserved Cgroup Limits on %q: %v", nc.SystemReservedCgroupName, err)
			cm.recorder.Event(nodeRef, v1.EventTypeWarning, events.FailedNodeAllocatableEnforcement, message)
			return fmt.Errorf(message)
		}
		cm.recorder.Eventf(nodeRef, v1.EventTypeNormal, events.SuccessfulNodeAllocatableEnforcement, "Updated limits on system reserved cgroup %v", nc.SystemReservedCgroupName)
	}
```

.github/actions/setup-kubernetes/action.yaml
tnozicka: (I half forgot whether this failed on its own yesterday with just the storage, already reverted to what was commented out before)

.github/actions/setup-kubernetes/action.yaml
tnozicka: Isn't available ephemeral storage tracked on the node, not in a cgroup?

zimnx: yep, looks like it's

.github/actions/setup-kubernetes/action.yaml
zimnx: i can't tell from skimming through the code, but their example in CLI parameter description would suggest the opposite. I guess it's fine to keep them in the same slice.

zimnx (approved):

.github/actions/setup-kubernetes/action.yaml
tnozicka: included in https://github.com/scylladb/scylla-operator/pull/1169

PR #1173, opened 2023-02-23
Fix GH action setting up Kubernetes

zimnx:
Two issues were found with setup-kubernetes action when Kube 1.25 was used:
* kubelet exec probes wasn't working - due to old conmon
* control plane nodes couldn't be untained - master taint is removed in 1.25 and kubectl isn't idempotent


tnozicka (approved):
/approve
/lgtm

thanks

PR #1174, opened 2023-02-24
Inrease maxPods value in kubelet config in CI

rzetelskik:
The `maxPods` value in kubelet config in our CI recently set in https://github.com/scylladb/scylla-operator/pull/1169 proved to be insufficient - see e.g. https://github.com/scylladb/scylla-operator/actions/runs/4255280318/jobs/7402754729#step:3:2191:
```
Feb 23 17:54:25.061: INFO: At 2023-02-23 17:51:15 +0000 UTC - event for basic-22qhv-us-east-1-us-east-1a-0: {default-scheduler } FailedScheduling: 0/1 nodes are available: 1 Too many pods. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.
 ```
This PR sets the value to the default 110.

Resolves CI issues


zimnx (dismissed):

tnozicka (approved):
it's pretty dense already for 2CPUs and I assume we'll hit that anyways just later on with the parallelism we have today but I don't mind trying.  Did the pod stay pending for long enough to fail a test?

tnozicka:
Thinking about it a bit more, for CI purposes, we don't really ever want to see `Too many pods`. I'd actually bump it much higher, like 1024 and leave it purely managed by the actually resources. We want as many SC in parallel because they spend most of their time waiting for gossip to init.

rzetelskik:
> it's pretty dense already for 2CPUs and I assume we'll hit that anyways just later on with the parallelism we have today but I don't mind trying. Did the pod stay pending for long enough to fail a test?

Yes, multiple tests failed on `STEP: Waiting for the ScyllaCluster to rollout`. In an unfortunate interlace tests seem to starve for Pods. Bare in mind that happened on a branch with a monitoring stack deployed (~10 pods), but that means we're already on thin ice with the Pod limit.

> Thinking about it a bit more, for CI purposes, we don't really ever want to see `Too many pods`. I'd actually bump it much higher, like 1024 and leave it purely managed by the actually resources. We want as many SC in parallel because they spend most of their time waiting for gossip to init.

Agreed.

tnozicka:
yeah, I think we are close. It's fine to fail with the other errors, adjust default default resources so it waits in pending, but I don't think we should fail with this error. Thanks for bumping it.

tnozicka (approved):
/lgtm
/approve

rzetelskik:
https://github.com/scylladb/scylla-operator/actions/runs/4295142958
image

I suppose resource starvation? I'll try with lower parallelism on current master.

tnozicka (approved):

PR #1182, opened 2023-03-07
Tune CI and adjust timeouts

tnozicka:
- Fixes a permission issue on `sysctl` dump
- Collects `dmesg`, `df`, `mount` and `lsblk`
- Enforces e2e timeout above the container layer in case something in between would get stuck or the program timeout didn't work
- Frees disk space on GH runners (SDK we don't need)
- Moves PVs from `/mnt` (very small) to `/`
- Defaults test timeout is now `60m` so it can fit even with flakes under the global `90m` GH actions timeout
- Reduces audit size and resource consumption by the kube-apiserver

### DIFF
--- a/.github/actions/run-e2e/action.yaml
+++ b/.github/actions/run-e2e/action.yaml
@@ -121,7 +121,7 @@ runs:
       group="$( id -g )"
       ingress_address="$( kubectl -n haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' )"
       
-      sudo podman run --user="${user}:${group}" --rm \
+      sudo timeout "$(( ${e2e_timeout_minutes} + 5 ))m" podman run --user="${user}:${group}" --rm \
       --entrypoint=/usr/bin/scylla-operator-tests \
       -v="${ARTIFACTS_DIR}:${ARTIFACTS_DIR}:rw" \
       -v="${HOME}/.kube/config:/kubeconfig:ro" -e='KUBECONFIG=/kubeconfig' \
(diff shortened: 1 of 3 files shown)


.github/actions/run-e2e/action.yaml
```diff
@@ -162,6 +162,7 @@ runs:
       sudo crictl images > crictl.images
       sudo podman stats -a --no-stream --no-reset > podman.stats
       journalctl -u kubelet > kubelet.log
+      journalctl -u kube-proxy > kube-proxy.log
```
rzetelskik, line 165: Aren't we already collecting it from a mirroring Pod?

rzetelskik:
One of the commit messages has a typo: [Setup out PVs in CI on / which has more storage available](https://github.com/scylladb/scylla-operator/pull/1182/commits/bbbd871d0674bf37b49173abcaf0aefa10d86a15) (s/out/our)

.github/actions/run-e2e/action.yaml
tnozicka: what's `a mirroring Pod`?

tnozicka: you are right, we already have the logs from the static pod. I'll drop this one.

tnozicka: dropped

rzetelskik: https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/ it's actually a mirror* Pod, but anyway a reflection of a static Pod in apiserver

zimnx (approved):
lgtm, thanks

PR #1183, opened 2023-03-07
Update dependencies

tnozicka:
Updates all dependencies.


zimnx:
Please verify v1 CRI is available on EKS. Afaik containerd is used there. According to [1] it's supported since containerd 1.6.0+, 1,5.8+

[1] - https://containerd.io/releases/#kubernetes-support

rzetelskik:
nit: typo in one of the commit messages [Remove deprecates SlowSpecThreshold in tests](https://github.com/scylladb/scylla-operator/pull/1183/commits/e83ac774906c7fed0aef67f94673baf2465b8107) (s/deprecates/deprecated)

tnozicka:
eks < 1.24 doesn't use containerd

eks 1.24 has v1 support 
```
/usr/bin/containerd --version
containerd github.com/containerd/containerd 1.6.6 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
```

looking at kubelet code, contrary to the docs I see the v1alpha2 fallback only for <=1.25 so v1.26 would likely need v1 so we probably get broken one way or the other. So we either need to move forward to v1 or get stuck in 1.25 and implement the fallback
https://github.com/kubernetes/kubernetes/blob/007d93ad08428c335382028cbdcb21a9d1b270ab/pkg/kubelet/cri/remote/remote_runtime.go#L141-L145

regarding our compatibility metrics we don't claim any deps on cri, only kube so that may be harder to express - do we just go with >=1.24 ?

tnozicka:
In the end I've aded the CRI API to our support metrics so we don't have to bundle this to the kube version which isn't really connected to it.

PR #1195, opened 2023-03-27
Fix sysctl settings in kubernetes setup Action

zimnx:
Couple of 'tee' weren't appending to a file but they were overwriting it preventing previous settings to be applied.


tnozicka (approved):
/approve
/lgtm
thanks

PR #1199, opened 2023-04-04
Fix "job" labels in Grafana dashboards' queries

rzetelskik:
Currently there is a bug breaking our Grafana dashboards introduced in https://github.com/scylladb/scylla-operator/pull/1111. The "job" label in some of the PromQL queries is hardcoded to "scylla", while the ServiceMonitor manifests set it to the value of "scylla/cluster" labels of the identity Service.
https://github.com/scylladb/scylla-operator/blob/ee48da7fa9235eb8167f569c321123740c57252f/assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml#L8

For reference see https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitor.

This PR fixes it by setting it to a `cluster` label defined in the ServiceMonitor: https://github.com/scylladb/scylla-operator/blob/ee48da7fa9235eb8167f569c321123740c57252f/assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml#L25. For reference see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#service and https://prometheus.io/docs/prometheus/latest/configuration/configuration/#service.

### DIFF
--- a/assets/monitoring/grafana/v1alpha1/dashboards.cm.yaml
+++ b/assets/monitoring/grafana/v1alpha1/dashboards.cm.yaml
@@ -4602,7 +4602,7 @@ data:
                             ]
                         },
                         "datasource": "prometheus",
-                        "definition": "query_result(count(up{job=\"scylla\"}) by (dc))",
+                        "definition": "query_result(count(up{job=~\"$cluster|$^\"}) by (dc))",
                         "error": null,
                         "hide": 2,
                         "includeAll": true,
@@ -4610,7 +4610,7 @@ data:
                         "name": "count_dc",
                         "options": [],
                         "query": {
-                            "query": "query_result(count(up{job=\"scylla\"}) by (dc))",
+                            "query": "query_result(count(up{job=~\"$cluster|$^\"}) by (dc))",
                             "refId": "StandardVariableQuery"
                         },
                         "refresh": 2,


assets/monitoring/grafana/v1alpha1/dashboards.cm.yaml
```diff
@@ -4602,15 +4602,15 @@ data:
                             ]
                         },
                         "datasource": "prometheus",
-                        "definition": "query_result(count(up{job=\"scylla\"}) by (dc))",
+                        "definition": "query_result(count(up{job=~\"$cluster|$^\"}) by (dc))",
```
tnozicka, line 4605: can you pls explain the syntax beyond `job=~'${cluster}'` to me? in particular why the "or" (`|`) and what `$^` means here?

rzetelskik: I shamelessly copied it over from the other queries - and I suppose you did the same, as it originates from scylla-monitoring dashboards. I actually tried to find something on it in the docs with no success. This essentially substitutes the variables and then matches a regex, so for a cluster named `simple-cluster` you end up matching e.g. `simple-cluster|$^`, but I don't see how that's any different than matching just `simple-cluster`, with or without the regex. @amnonh would you be able to explain? Is that some grafana/prom query magic?

tnozicka: I am fine deferring that one and wait for @amnonh's input. Please file an issue to track this and we can merge this at least.

tnozicka (approved):
/approve
/lgtm

(better then what we have now an we'll track the regex reasoning in a follow up issue)

vponomaryov (approved):

assets/monitoring/grafana/v1alpha1/dashboards.cm.yaml
rzetelskik: Opened https://github.com/scylladb/scylla-operator/issues/1209

PR #1200, opened 2023-04-04
Workaround mkfs.xfs min size limits in newer versions

tnozicka:
```
mkfs.xfs since v5.19.0-rc1 [1] refuses to create filesystems < 300 MB.
```
https://lore.kernel.org/all/Yv4i0gWiHTkfWB5m@yuki/T/

We'll set the extra variables (`TEST_DIR=1 TEST_DEV=1 QA_CHECK_FS=1`) to allow us still create small testing filesystems.

### DIFF
--- a/hack/test/gke/xfs-formatter/tune2fs_test.sh
+++ b/hack/test/gke/xfs-formatter/tune2fs_test.sh
@@ -26,8 +26,9 @@ trap 'err_handler "${LINENO}"' ERR
 tmp_file=$( mktemp )
 
 function setup_fs {
-  dd if=/dev/zero of="${tmp_file}" bs=4096 count=4096 
-  mkfs -t "${1}" "${tmp_file}" 
+  dd if=/dev/zero of="${tmp_file}" bs=4096 count=4096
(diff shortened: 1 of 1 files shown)


zimnx (approved):

hack/test/gke/xfs-formatter/tune2fs_test.sh
```diff
@@ -26,8 +26,9 @@ trap 'err_handler "${LINENO}"' ERR
 tmp_file=$( mktemp )
 
 function setup_fs {
-  dd if=/dev/zero of="${tmp_file}" bs=4096 count=4096 
-  mkfs -t "${1}" "${tmp_file}" 
+  dd if=/dev/zero of="${tmp_file}" bs=4096 count=4096
+  # Enable small filesystems with these 3 extra vars - https://lore.kernel.org/all/Yv4i0gWiHTkfWB5m@yuki/T/
+  TEST_DIR=1 TEST_DEV=1 QA_CHECK_FS=1 mkfs -t "${1}" "${tmp_file}"
```
mykaul, line 31: you should use -K for mkfs.xfs - since there's no need to try and discard anything.

tnozicka: possibly, but I don't think the speed up would be measurable on a device of this size, right?

PR #1204, opened 2023-04-04
Enforce cpu limits for scylla-operator in CI

tnozicka:
Scylla-operator has peak CPU usage when generating certs. On our small CI node 2 cores for everything this may be the last drop so we should limit the peaks.

Given this will slow down cert generation we may decide to wait for #1202 but let's see if or how bad it actually is.

Some flakes hopefully.


tnozicka:
haven't debugged anything in particular but I thought this is a good step to take no matter what

PR #1205, opened 2023-04-04
Install Prometheus ClusterRoles in Helm deployments

rzetelskik:
Prometheus ClusterRole manifests introduced in #1111 should be generated from templates in `helm/scylla-operator` like all the other deploy manifests. At the moment, although present in `deploy/operator/`, the files are missing from Helm templates, which additionally results in missing permissions in case the Operator is deployed using Helm.


rzetelskik:
@tnozicka did you leave commented out lines in the manifests on purpose or should we remove them? https://github.com/scylladb/scylla-operator/blob/0910f208cd8817b9d48e82983b31c095135f2a9f/deploy/operator/00_scylladbmonitoring_prometheus_clusterrole_def.yaml#L26

helm/scylla-operator/templates/scylladbmonitoring_prometheus_clusterrole_def.yaml
```diff
@@ -0,0 +1,37 @@
...
+kind: ClusterRole
+metadata:
+  name: scylladb:aggregate-to-scylladb-monitoring-prometheus
+  labels:
+    rbac.operator.scylladb.com/aggregate-to-scylladb-monitoring-prometheus: "true"
+rules:
+- apiGroups:
+  - ""
+  resources:
+#  - nodes
```
zimnx, line 11: lets remove commented out code. If these would be required, e2e would catch it.

Makefile
```diff
@@ -373,6 +373,8 @@ define generate-operator-manifests
 	mv '$(3)'/scylla-operator/templates/edit_clusterrole.yaml '$(2)'/00_scyllacluster_clusterrole_edit.yaml
 	mv '$(3)'/scylla-operator/templates/scyllacluster_member_clusterrole.yaml '$(2)'/00_scyllacluster_member_clusterrole.yaml
 	mv '$(3)'/scylla-operator/templates/scyllacluster_member_clusterrole_def.yaml '$(2)'/00_scyllacluster_member_clusterrole_def.yaml
+	mv '$(3)'/scylla-operator/templates/scylladbmonitoring_prometheus_clusterrole.yaml '$(2)'/00_scylladbmonitoring_prometheus_clusterrole.yaml
```
zimnx, line 376: shouldn't this be projected into separate commit with autogenerated stuff? If it's already there, it should be removed in this one.

helm/scylla-operator/templates/scylladbmonitoring_prometheus_clusterrole_def.yaml
rzetelskik: done

Makefile
rzetelskik: It was already there and I wasn't sure how to go about it. Done.

vponomaryov (approved):

tnozicka (commented):
dropping the commented lines is the right course of action, we intend to keep those minimal

/approve

tnozicka:
this wasn't yet release so I wonder whether a bug label is fitting, given it makes it to the release notes. The title sounds like cleanup. Keeping the bug label would be fine with me but you should reword the title in that case, like `Install Prometheus ClusterRoles in Helm deployments` (or something like that)

rzetelskik:
@tnozicka thanks, changed the title

zimnx (approved):

vponomaryov:
@rzetelskik @tnozicka Shouldn't also the commit topic be changed the same way as the PR description?

rzetelskik:
> @rzetelskik @tnozicka Shouldn't also the commit topic be changed the same way as the PR description?

Well, yes, but it already merged. :wink: Not too big of an issue since the concern was the release notes, which are generated from PR titles and labels.

PR #1227, opened 2023-05-02
Use Dynamic Local Volume Provisioner

zimnx:
This brings integration with scylladb/k8s-local-volume-provisioner.
In CI, instead using statically provisioned PVs, ScyllaCluster volumes are
created on demand with capacity restrictions.

Because ScyllaCluster storage cannot grow over the limit now, e2e basic cluster
storage needed to be extended, because nodes weren't boostraping due to
commitlog taking almost entire space (~98MB).


.github/actions/setup-local-volume-provisioner/action.yaml
```diff
@@ -1,37 +1,32 @@
...
-    required: true
 runs:
   using: "composite"
   steps:
-  - name: Prepare host directories
+
+  - name: Checkout Dynamic Local Volume Provisioner
+    uses: actions/checkout@v3
+    with:
+      repository: scylladb/k8s-local-volume-provisioner
```
tnozicka, line 10: checking out master will break at some point - it will make an update without going trough CI and will bork this repo. Can we copy the deploy files and consumed the image as any other project?

We need to replace the images with sha refs in those files anyways.

test/e2e/fixture/scylla/basic.scyllacluster.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     - name: us-east-1a
       members: 1
       storage:
-        capacity: 100Mi
+        capacity: 200Mi
```
tnozicka, line 15: can you elaborate on why did this need an adjustment?

.github/actions/setup-local-volume-provisioner/action.yaml
```diff
@@ -1,37 +1,32 @@
...
+  - name: Create XFS volume
     shell: bash
     run: |
       set -euExo pipefail
       shopt -s inherit_errexit
       
-      sudo mkdir /mnt/persistent-volumes
-      # /mnt has limited space in GH Actions, use /
-      sudo mkdir /persistent-volumes
-      sudo mount --bind /persistent-volumes /mnt/persistent-volumes
```
tnozicka, line 22: shouldn't this stay?

test/e2e/fixture/scylla/basic.scyllacluster.yaml
zimnx: i mentioned it in commit and PR description, commit log grows to 98MB when second node joins. I saw out of space error for 200Mi PVC, so I ended up leaving 300Mi.

.github/actions/setup-local-volume-provisioner/action.yaml
zimnx: done, not on `/` but `/var` holds the disk image which is on the same device.

.github/actions/setup-local-volume-provisioner/manifests/local-csi-driver/10_csidriver.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: storage.k8s.io/v1
```
rzetelskik, line 1: Do you think these manifests could be moved to replace `examples/common/provisioner` instead? We could use them in eks/gke examples with some tweaks.

rzetelskik (approved):
one comment, other than that lgtm

.github/actions/setup-local-volume-provisioner/manifests/local-csi-driver/10_csidriver.yaml
zimnx: current examples are still using `examples/common/provisioner`. I wouldn't mind moving these there when existing examples are converted to the dynamic provisioner. But as a part of separate PR.

.github/actions/setup-local-volume-provisioner/action.yaml
```diff
@@ -4,34 +4,52 @@ inputs:
...
     shell: bash
     working-directory: ${{ inputs.repositoryPath }}
     run: |
       set -euExo pipefail
       shopt -s inherit_errexit
       
-      kubectl -n local-volume-provisioner apply --server-side -f ./.github/actions/setup-local-volume-provisioner/manifests/
-      kubectl -n local-volume-provisioner rollout status --timeout=5m daemonset.apps/local-volume-provisioner
+      kubectl apply --server-side -f ./.github/actions/setup-local-volume-provisioner/manifests/storageclass_xfs.yaml
+      kubectl patch storageclass scylladb-local-xfs -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
```
tnozicka, line 55: the storage class manifest is dedicated to CI setup - any reason not to have it there directly?

test/e2e/fixture/scylla/basic.scyllacluster.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     - name: us-east-1a
       members: 1
       storage:
-        capacity: 100Mi
+        capacity: 300Mi
```
tnozicka, line 15: just for the record, I recall there were some implications around xfs min size - can you comment on the reason for the bump?

.github/actions/setup-local-volume-provisioner/action.yaml
zimnx: These manifests are copy pasted from upstream, this way they won't need any changes when they are updated.

test/e2e/fixture/scylla/basic.scyllacluster.yaml
zimnx: XFS min size is related to entire volume quota can be even a single block.

.github/actions/setup-local-volume-provisioner/action.yaml
zimnx: moved inline anyway.

test/e2e/fixture/scylla/basic.scyllacluster.yaml
zimnx: >can you comment on the reason for the bump? 

It's part of PR description

PR #1230, opened 2023-05-04
Fix command go-template in monitoring docs

rzetelskik:
One of the commands in the Monitoring section of the documentation has a broken go-template resulting in the following error:
```
error: error executing template "{{ ( index .loadBalancer.ingress 0 ).ip }}": template: output:1:5: executing "output" at <index .loadBalancer.ingress 0>: error calling index: index of untyped nil
```
It must have gone unnoticed when applying review remarks to https://github.com/scylladb/scylla-operator/pull/1161. This PR fixes the template to read the correct field.


tnozicka (approved):
/approve
/lgtm

PR #1234, opened 2023-05-09
Refactor release notes publishing into a reusable GitHub Action 

zimnx:
With current approach, the code for generating and publishing release notes
would need to be copied and pasted across multiple repositories,
leading to maintenance issues and duplication of effort.
To address this, I have extracted the release notes functionality into
a standalone GitHub Action, which can be easily reused across different projects.
The new composite action takes parameters that allow it to be customized
for each repository's specific needs, while still using the same underlying release note
generation tool.

### DIFF
--- a/.github/actions/release-notes/action.yaml
+++ b/.github/actions/release-notes/action.yaml
@@ -1,0 +1,122 @@
+name: "Generate and publish release notes"
+description: 'Generate and publish release notes'
+inputs:
+  githubRepository:
+    description: "Name of repository which release notes should be generated and published in 'owner/repository' format"
+    required: true
+  githubRef:
+    description: "The fully-formed ref of the branch or tag for publishing a release"
+    required: true
+  githubToken:
+    description: "Token with access to pull permissions to scylladb/scylla-operator and release permissions to repository which release is published"
+    required: true
+  goVersion:
+    description: "Version of Go"
+    required: true
+  genReleaseNotesVersionRef:
+    description: "Version ref of gen-release-notes"
+    required: true
+runs:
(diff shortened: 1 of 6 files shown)


.github/actions/release-notes/action.yaml
```diff
@@ -0,0 +1,123 @@
...
+    required: true
+  githubToken:
+    description: "Token with access to pull permissions to scylladb/scylla-operator and release permissions to repository which release is published"
+    required: true
+  goVersion:
+    description: "Version of Go"
+    required: true
+  genReleaseNotesVersionRef:
+    description: "Version ref of gen-release-notes"
+    default: master
```
tnozicka, line 18: can we make this required instead? I'd like to avoid called accidentally using a rolling ref that breaks CI

.github/actions/release-notes/action.yaml
```diff
@@ -0,0 +1,123 @@
...
+runs:
+  using: "composite"
+  steps:
+
```
tnozicka, line 22: nit: 
```suggestion
```

tnozicka (approved):
/Approve
/lgtm

I think long term we should move the tool to a separate repo and publish an image with the tool instead but this looks good for now

.github/actions/release-notes/action.yaml
zimnx: done

.github/actions/release-notes/action.yaml
zimnx: done

tnozicka (approved):

PR #1247, opened 2023-05-12
Use ScyllaDB 5.1.9 and ScyllaDB Manager 3.1.0 as default versions in Helm chart, examples and e2e

zimnx:
Examples, e2e and the default version used by ScyllaDB and ScyllaDB Manager Helm charts was changed to 5.1.9 for ScyllaDB and 3.1.0 for ScyllaDB Manager Server and Agent.


tnozicka (approved):
/approve
/lgtm

Issue #1245, opened 2023-05-12
An e2e test from a suite was not run (occasionally)


tnozicka:
**Describe the bug**
#1238 [has missed to run some tests](https://github.com/scylladb/scylla-operator/actions/runs/4929383274/jobs/8820672795#step:3:2862) from the set `Test e2e parallel` that should have been ran. We found out because it managed to break master as a result.
```
Ran 17 of 24 Specs in 1849.933 seconds
SUCCESS! -- 17 Passed | 0 Failed | 0 Pending | 7 Skipped
```
while the current master (or the broken PRs result in 
```
Ran 19 of 24 Specs in 2536.036 seconds
FAIL! -- 18 Passed | 1 Failed | 0 Pending | 5 Skipped
```
The test that got broken is
```
ScyllaCluster replace [It] should replace a node
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_replace.go:26
```
and there is no notion of a skip in the job logs either.

this may be connected to #1176.

**Additional context**
I'd propose to add a test running several iterations of a dry run with each suite and compare it with an expected set or something in those lines.

Issue #1176, opened 2023-02-28
Test runner duplicates spec runs while omitting other ones when running with parallelism


rzetelskik:
**Describe the bug**
https://github.com/scylladb/scylla-operator/actions/runs/4255280318/jobs/7402754729 ran one spec twice in one suite (see https://github.com/scylladb/scylla-operator/actions/runs/4255280318/jobs/7402754729#step:3:2103 and https://github.com/scylladb/scylla-operator/actions/runs/4255280318/jobs/7402754729#step:3:2157 right after) while omitting some other tests, e.g. `ScyllaCluster should setup and maintain up to date TLS certificates` (`scyllacluster_tls.go`).

Since our parallel server is copied from `onsi/gingko` I've looked through the existing issues and release notes since 2.4.0 which we're using and haven't found anything related. I also went through the code changes in the parallel server since it's been copied and again haven't found anything relevant either. Looking at other obvious possible causes - all processes are started with a unique `parallel-shard`.

rzetelskik:
Completed in #1256

PR #1253, opened 2023-05-16
Extend test runner with a report verifier checking if every Spec was ran exactly once

rzetelskik:
**Description of your changes:** Recently we've hit https://github.com/scylladb/scylla-operator/issues/1245 which managed to bork master. The issue is most likely caused by https://github.com/scylladb/scylla-operator/issues/1176. Until the root cause is fixed, this PR mitigates it by checking the generated report and verifying that all specs were assigned to exactly one parallel process.

Mitigates https://github.com/scylladb/scylla-operator/issues/1176 temporarily to unbork the CI.

### DIFF
--- a/pkg/cmd/tests/tests_run.go
+++ b/pkg/cmd/tests/tests_run.go
@@ -309,6 +309,63 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
 
 	gomega.RegisterFailHandler(ginkgo.Fail)
 
+	// Temporary workaround for https://github.com/scylladb/scylla-operator/issues/1176.
+	// Run a check after suite has completed to verify whether all specs were run by a single process.
+	// TODO: remove once #1176 has been fixed.
+	ginkgo.ReportAfterSuite("Verify report", func(r ginkgo.Report) {
+		expectedSpecs := []string{
+			"Node Setup should make RAID0, format it to XFS, and mount at desired location out of one loop device",
+			"Node Setup should make RAID0, format it to XFS, and mount at desired location out of three loop devices",
+			"NodeConfig Optimizations should correctly project state for each scylla pod",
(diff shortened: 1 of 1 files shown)


rzetelskik:
@tnozicka please let me know if the hardcoded list of expected specs is acceptable for a temporary mitigation or if it should be gathered in a CI action

pkg/cmd/tests/tests_run.go
```diff
@@ -309,6 +309,63 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
 
 	gomega.RegisterFailHandler(ginkgo.Fail)
 
+	// Temporary workaround for https://github.com/scylladb/scylla-operator/issues/1176.
+	// Run a check after suite has completed to verify whether all specs were run by a single process.
+	// TODO: remove once #1176 has been fixed.
+	ginkgo.ReportAfterSuite("Verify report", func(r ginkgo.Report) {
+		expectedSpecs := []string{
```
tnozicka, line 316: how does this work for different suites or for feature flags?

rzetelskik: Every suite, regardless of the feature flags, builds the same spec tree - they just have a different status in the final report. The only difference is the status of the nodes - different subsets are ran/skipped. Only in case of an invalid parallel run the specs are missing from the final report, as multiple processes get the same subset of specs.

tnozicka: didn't the #1245 had the same number of overall test just 2 more were skipped instead of being run?

rzetelskik: There was a higher number of skipped specs than usual but that's coincidental - the duplicated tests that were selected instead of the missed ones happened to be the ones being skipped.

zimnx (approved):

Issue #1233, opened 2023-05-09
Flake: ScyllaCluster upgrades should deploy and update [It] with 3 member(s) and 1 rack(s) from 5.0.4 to 5.0.5


rzetelskik:
https://github.com/scylladb/scylla-operator/actions/runs/4885577292/jobs/8729441531?pr=1230
[e2e-artifacts-parallel-alpha.tar.lz4.zip](https://github.com/scylladb/scylla-operator/files/11404414/e2e-artifacts-parallel-alpha.tar.lz4.zip)
[logs_13091.zip](https://github.com/scylladb/scylla-operator/files/11404415/logs_13091.zip)

```
[FAILED] [962.724 seconds]
ScyllaCluster upgrades should deploy and update [It] with 3 member(s) and 1 rack(s) from 5.0.4 to 5.0.5
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_upgrades.go:102

  Timeline >>
  STEP: Creating a new namespace @ 05/05/23 05:26:05.682
  May  5 05:26:05.832: INFO: Created namespace "e2e-test-scyllacluster-8gf6p-78bwk".
  STEP: Waiting for service account token Secret "e2e-user-token" in namespace "e2e-test-scyllacluster-8gf6p-78bwk". @ 05/05/23 05:26:05.976
  STEP: Waiting for default ServiceAccount in namespace "e2e-test-scyllacluster-8gf6p-78bwk". @ 05/05/23 05:26:06.063
  STEP: Creating a ScyllaCluster @ 05/05/23 05:26:06.123
  STEP: Waiting for the ScyllaCluster to rollout (RV=2049) @ 05/05/23 05:26:06.198
  May  5 05:38:04.291: INFO: ScyllaCluster e2e-test-scyllacluster-8gf6p-78bwk/basic-zgms6 (RV=8835) is rolled out
  STEP: Verifying the ScyllaCluster @ 05/05/23 05:38:09.292
  May  5 05:38:09.316: INFO: Found 3 pvc(s) in namespace "e2e-test-scyllacluster-8gf6p-78bwk"
  May  5 05:38:09.317: INFO: Found 3 pvc(s) for ScyllaCluster "e2e-test-scyllacluster-8gf6p-78bwk/basic-zgms6"
  STEP: Waiting for the ScyllaCluster to reach consistency ALL @ 05/05/23 05:38:09.333
[...]

PR #1255, opened 2023-05-17
Increase webhook cpu

zgrim:
Increase webhook cpu limit from 10m to 50m to address the starvation seen in #1233 

Addresses #1233


tnozicka (changes requested):
why do you think changing this requests will make a difference in the overall cpu division scheme?

zgrim:
from studying the logs with @zimnx it seems like the most plausible explanation, that there wasn't enough cpu, so he suggested this increase, which should lower the chances of this flake in the future

PR #1256, opened 2023-05-19
Fix seed propagation in test runner

### LINKED ISSUE #1245: An e2e test from a suite was not run (occasionally)

rzetelskik:
**Description of your changes:** Unless a `random-seed` flag was provided explicitly to `scylla-operator-tests`, the default RandomSeed was not being propagated to child processes while running tests in parallel. This resulted in the workers generating the seeds themselves and hence a possibility of running with different seeds. Since ginkgo uses seed for ordering the specs (https://github.com/scylladb/scylla-operator/blob/a2cdd29417b536f798dc5b954f5b073c2bad69ec/vendor/github.com/onsi/ginkgo/v2/internal/ordering.go#L65), workers were being assigned overlapping subsets of specs, and in turn some specs were not being ran altogether.
This PR fixes the bug in test runner by always propagating the seed to its child processes. It also removes some unnecessary assignments and updates the internal Ginkgo files copied over from Ginkgo repository.

Resolves #1245 #1176

### DIFF
--- a/pkg/cmd/tests/tests_run.go
+++ b/pkg/cmd/tests/tests_run.go
@@ -20,6 +20,7 @@ import (
 	gomegaformat "github.com/onsi/gomega/format"
 	"github.com/scylladb/scylla-operator/pkg/cmdutil"
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/scylladb/scylla-operator/pkg/signals"
 	ginkgotest "github.com/scylladb/scylla-operator/pkg/test/ginkgo"
 	"github.com/scylladb/scylla-operator/pkg/thirdparty/github.com/onsi/ginkgo/v2/exposedinternal/parallel_support"
@@ -38,6 +39,7 @@ import (
 const (
 	parallelShardFlagKey            = "parallel-shard"
 	parallelServerAddressFlagKey    = "parallel-server-address"
+	randomSeedFlagKey               = "random-seed"
 	ginkgoOutputInterceptorModeNone = "none"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/cmd/tests/tests_run.go
```diff
@@ -412,16 +413,25 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+		if strings.HasPrefix(commonArgs[i], fmt.Sprintf("--%s", randomSeedFlagKey)) {
+			foundRandomSeedArg = true
+			break
+		}
+	}
+	// Random seed has to be propagated to child processes.
+	if !foundRandomSeedArg {
+		commonArgs = append(commonArgs, fmt.Sprintf("--%s=%v", randomSeedFlagKey, suiteConfig.RandomSeed))
+	}
+
```
zimnx, line 427: nit: a little bit shorter
```suggestion
// Propagate seed to the childs.
	if !helpers.Contains(commonArgs, func(arg string) bool {
		return strings.HasPrefix(arg, fmt.Sprintf("--%s", randomSeedFlagKey))
	}) {
		commonArgs = append(commonArgs, fmt.Sprintf("--%s=%v", randomSeedFlagKey, suiteConfig.RandomSeed))
	}
```

zimnx (dismissed):
one optional nit

zimnx (dismissed):

tnozicka (approved):
Great job finding out the cause of our most painful CI issue!

/approve
/lgtm

PR #1264, opened 2023-05-28
updating chart readme + adding optional XFS storage class definition

viggin543:
- operator helm charet: now can create an xfs storage class that can be used by the scylla cluster
- updating helm README.md

K8S now supports XFS file system StorageClass and it can be created by the operator chart if enabled

// values.yaml
```
storageClass:
  create: true
  type: standard
  provisioner: ''
  volumeBindingMode: 'WaitForFirstConsumer'
  allowVolumeExpansion: true
```

k8s 1.24 also supports volume expansion that allows increasing disk size of existing PVC dynamically. The storage class the operator creates enables this flag by default.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -25,10 +25,11 @@ You can then run `helm search repo scylla-operator` to see the charts.
 
 ```console
 # Install Scylla Operator
-$ helm install [NAME] scylla-operator/scylla-operator --create-namespace --namespace scylla-operator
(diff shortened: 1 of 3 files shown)


zimnx:
I don't think `scylla` chart is a good place for cluster wide resource such as StorageClass.

viggin543:
agree.
But it can be convenient if optional

PR #1269, opened 2023-06-06
Replace xfs-formatter with NodeConfig in GKE examples

rzetelskik:
**Description of your changes:** In https://github.com/scylladb/scylla-operator/pull/770, the `xfs-formatter` DaemonSet has been introduced as a workaround to https://github.com/scylladb/scylla-operator/issues/760. We used `--ephemeral-storage` beta feature which provided an ext4 filesystem on RAID0 array. Because of that, and the fact that the filesystem would always be recreated on reboot, we had to use the DaemonSet as a nasty hack to create and maintain an xfs filesystem instead.
Fortunately, starting with 1.25.3-gke.1800, a GKE node pool can be configured with `--local-nvme-ssd-block` to provide local SSDs for [raw block access](https://cloud.google.com/kubernetes-engine/docs/concepts/local-ssd#block-access). Because of that, we are able to drop the hacky workaround in favour of the NodeConfig.

Resolves: N/A


examples/gke/gke.sh
```diff
@@ -204,9 +194,20 @@ kubectl -n cert-manager rollout status --timeout=5m deployment.apps/cert-manager
...
 
+kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com
 kubectl wait --for condition=established crd/scyllaclusters.scylla.scylladb.com
 wait-for-object-creation scylla-operator deployment.apps/scylla-operator
 kubectl -n scylla-operator rollout status --timeout=5m deployment.apps/scylla-operator
+kubectl -n scylla-operator rollout status --timeout=5m deployment.apps/webhook-server
+
+echo "Configuring nodes..."
+kubectl apply -f nodeconfig-alpha.yaml
+wait-for-object-creation default nodeconfig.scylla.scylladb.com/cluster
```
tnozicka, line 205: isn't this what `kubectl apply` has already done?

examples/gke/gke.sh
```diff
@@ -204,9 +194,20 @@ kubectl -n cert-manager rollout status --timeout=5m deployment.apps/cert-manager
...
+
+# Install local volume provisioner
+echo "Installing local volume provisioner..."
+helm install local-provisioner ../common/provisioner
```
tnozicka, line 209: let's not mix helm into it

tnozicka: might be good to wait for it to get ready

examples/gke/nodeconfig-alpha.yaml
```diff
@@ -0,0 +1,29 @@
...
+      mountPoint: /mnt/raid-disks/local
+      unsupportedOptions:
+      - prjquota
+    raids:
+    - name: nvmes
+      type: RAID0
+      RAID0:
+        devices:
+          modelRegex: nvme_card
+          nameRegex: ^/dev/nvme\d+n\d+$
```
tnozicka, line 21: Why not use the stable names `/dev/disk/by-uuid/google-local-ssds-nvme-block/local-ssd-GENERATED_UUID1`? 
https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd#raw-block

examples/gke/gke.sh
rzetelskik: > let's not mix helm into it

It's not a change

examples/gke/gke.sh
rzetelskik: I thought it only means that the request was processed by the server? Even the docs explicitly mention waiting for the endpoint to be created https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition.

tnozicka: this is a CR, not a CRD. when a CR or any object like Pod is created, it is committed to etcd.
(CRD registration is asynchronous and waits for with `kubectl wait --for condition=established`. `wait-for-object-creation` is usually used to check objects created by other controllers / asynchronously.)

examples/gke/nodeconfig-alpha.yaml
zimnx: Current code only matches names from lsblk output. In theory it could be extended with matching files/symlinks from filesystem in backward compatible manner.

examples/gke/gke.sh
rzetelskik: Sorry, I misread and thought it referred to `kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com` a couple lines above. Removed.

examples/gke/nodeconfig-alpha.yaml
rzetelskik: @tnozicka your call then - I can send a PR extending the matching in node setup controller

tnozicka: I think matching the official paths would be good but if we don't have it now, I don't want this PR to take the hit. thx

tnozicka (commented):
/approve
/lgtm

tnozicka (approved):

PR #1272, opened 2023-06-09
Use Dynamic Local Volume Provisioner in EKS and GKE examples

rzetelskik:
**Description of your changes:** Since k8s-local-volume-provisioner was released and is now being used in our CI, we should move our GKE/EKS examples to it as well. This PR does this and also removes verification of helm dependency, as with removal of local static provisioner, it's no longer needed.

**Requires**: https://github.com/scylladb/scylla-operator/pull/1269


zimnx (approved):
LGTM, thanks!

PR #1275, opened 2023-06-09
Fix release notes generation

zimnx:
In case when release notes Action was executed with scylladb/scylla-operator repository gen-release-notes binary was built in the same directory where target repository was checked out. By default checkout action cleans the directory, which removed the binary resulting in missing command.

To overcome this, binary is moved to other PATH directory which won't be cleared out.


rzetelskik (approved):
Lgtm, thanks

PR #1278, opened 2023-06-13
Fix NodeConfig DaemonSet prunning

zimnx:
Wrong function was passed to controllerhelpers.Prune resulting in sync failing due to failure in deletion.


tnozicka (approved):
/approve
/lgtm

thanks

PR #1283, opened 2023-06-16
Support scaling changes during ongoing scaling

### LINKED ISSUE #1188: Adjusting the scale doesn't take the shortest path
* Feature Request

The operator should stop scaling up if a new desired state has less members.

Say we have a scylla cluster with `3` nodes and want to adjust members to `5` but accidentally put `55` there. The controller will create `55` services which is fine but when we correct ourselves and change the members to `5`, the controller will first scale up to `55` nodes and only then goes back to `5`. In the worst case your infra won't be big enough to scale to `55` instances and you'll get stuck.

**Additional Information:**
We are creating the services ahead of time which is fine but we need to make sure we stop scaling up as soon as we can and somehow differentiate the services that never had a bootstrapped node (which may be hard).

zimnx:
Decommissioning label reconcilation was moved from StatefulSet controller into Service controller.
StatefulSet controller can now react dynamically to scaling events occuring while scaling operations are still ongoing.

Fixes #1188

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_scaling.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_scaling.go
@@ -4,25 +4,265 @@ package scyllacluster
 
 import (
 	"context"
+	"sync/atomic"
+	"time"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	"github.com/scylladb/scylla-operator/pkg/controller/scyllacluster"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/scyllaclient"
 	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/scylladb/scylla-operator/test/e2e/utils"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 )
 
 var _ = g.Describe("ScyllaCluster", func() {
 	defer g.GinkgoRecover()
 
 	f := framework.NewFramework("scyllacluster")
+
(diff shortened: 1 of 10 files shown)


test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -4,26 +4,266 @@ package scyllacluster
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 )
 
 var _ = g.Describe("ScyllaCluster", func() {
 	defer g.GinkgoRecover()
 
 	f := framework.NewFramework("scyllacluster")
 
+	g.It("should react on scaling up in the middle of scaling down", func() {
```
rzetelskik, line 34:
```suggestion
	g.It("should react to scaling up in the middle of scaling down", func() {
```

test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -4,26 +4,266 @@ package scyllacluster
...
+		framework.By("Waiting for the ScyllaCluster to rollout (RV=%s)", sc.ResourceVersion)
+		sc, err = utils.WaitForScyllaClusterState(waitCtx3, f.ScyllaClient().ScyllaV1(), sc.Namespace, sc.Name, utils.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f.KubeClient(), sc)
+
+		o.Expect(secondNodeWasLeaving.Load()).To(o.BeFalse())
+	})
+
+	g.It("should react on scaling down in the middle of scaling up", func() {
```
rzetelskik, line 143:
```suggestion
	g.It("should react to scaling down in the middle of scaling up", func() {
```

rzetelskik (changes requested):
Overall well done, I like how much more readable the scaling code is now. I left some comments though, especially regarding the edge case of scaling down and up immediatelly.

test/e2e/utils/helpers.go
```diff
@@ -566,3 +566,38 @@ func PodIsRunning(pod *corev1.Pod) (bool, error) {
...
+func NodeIndexToIP(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster, rack scyllav1.RackSpec, idx int) (string, error) {
+	svcName := naming.MemberServiceName(rack, sc, idx)
+
+	s, err := client.Services(sc.Namespace).Get(ctx, svcName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	if s.Spec.Type != corev1.ServiceTypeClusterIP {
+		return "", fmt.Errorf("service %s/%s is of type %q instead of %q", s.Namespace, s.Name, s.Spec.Type, corev1.ServiceTypeClusterIP)
```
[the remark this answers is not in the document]
tnozicka: kobj is to be used with klog, there is an equivalent function in naming package for everything else

PR #1297, opened 2023-07-13
Identify nodes by their Host ID instead of IP address in readiness probes

zimnx:
IP address may be ephemeral, where Host ID is constant across node lifetime.
It also solves an issue when IPs returned from API have different representation (IPv6), causing readiness probe to fail because node status cannot be found.

### DIFF
--- a/pkg/sidecar/probes.go
+++ b/pkg/sidecar/probes.go
@@ -93,9 +93,9 @@ func (p *Prober) Readyz(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	nodeAddress, err := p.getNodeAddress()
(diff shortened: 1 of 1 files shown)


tnozicka (commented):
I like the change! What's the minimum ScyllaDB version to return this?

zimnx:
> What's the minimum ScyllaDB version to return this?

Probably very old one, status is built from return values from multiple rpc calls, first `host-id <--> ip` is collected, then `ip -> status` is mapped onto.
Note `Status` function wasn't modified, so it was always available.

tnozicka (approved):
ok, let's do this. thanks!

/approve
/lgtm

PR #1300, opened 2023-07-24
Enhancement proposal for running cleanup after scaling

zimnx:
The Operator supports both horizontal and vertical scaling, but the procedure isn’t in sync with ScyllaDB documentation,
because the cleanup part is not implemented. It’s important because upon scaling, the stored on node disk might become
stale taking up unnecessary space. Operator should support running the cleanup to keep disk space as low as possible and ensure clusters are stable and reliable over time.

Details can be found in actual enhancement.

### DIFF
--- a/enhancements/proposals/1207-cleanup-after-scaling/README.md
+++ b/enhancements/proposals/1207-cleanup-after-scaling/README.md
@@ -1,0 +1,152 @@
+# Run node cleanup after scaling the cluster
+
+## Summary
+
+The Operator supports both horizontal and vertical scaling, but the procedure isn’t in sync with ScyllaDB documentation,
+because the cleanup part is not implemented. It’s important because upon scaling, the stored data on node disk might
+become stale taking up unnecessary space or even cause data to be resurrected. Operator should support running the
+cleanup to
+keep disk space as low as possible and ensure clusters are stable and reliable over time.
+
+## Motivation
+
+The motivation behind running node cleanup after scaling a cluster is to avoid the accumulation of unnecessary data on
+the node disks, and prevent from data resurrection. When nodes are added or removed from the cluster, they gain or lose
+some tokens, which can result in files stored on the node disks still containing data associated with lost tokens. Over
+time, this can lead to a build-up
+of unnecessary data and cause disk space issues. Data resurrection happens when tombstones delete data on other nodes,
+and the tombstone is eventually purged, leaving behind neither the data nor the tombstone. When cluster is scaled down
+and token ownership is moved back to the
+original node, that might still have the data, because the tombstone that deleted it was purged, the data gets
+resurrected.
+By running node cleanup after scaling, these files can be cleared, freeing up disk space and prevent from data
+resurrection.
+
+### Goals
+
+* Running a node cleanup procedure after horizontally scaling the cluster.
+* Running a node cleanup procedure after scaling is completed, on all nodes but the last one added.
+
+### Non-Goals
+
+* Running a node cleanup during off-peak hours.
+* Running a node cleanup on all necessary cases.
+
+## Proposal
+
+I propose to extend the sidecar with a new controller that will project the latest cleaned up and existing hash of
+tokens for each node as an annotation in the member service.
+In addition, a new controller in Operator will be responsible for managing Jobs that will execute a cleanup on nodes
+that require it. The trigger for the Job creation will be a mismatch between the current and latest hash.
+
+### Risks and Mitigations
+
+#### Cluster bootstrap time is increased
+
+When a cluster is created for the first time, tokens are reshuffled when new nodes are joining sequentially. This will
+cause the token hash mismatch and will require a cleanup once all nodes join the cluster. This will cause cluster
+bootstrap time to be increased. But since the time window is short, nodes shouldn’t have much data stored so cleanup
+won’t take long.
+
+#### Cleanup is run when not necessary
+
+This design doesn’t take into account whether a node received a token or lost it, it only detects ring changes and
+reacts with a cleanup trigger upon change. When a node is decommissioned, tokens are redistributed and nodes getting
+them don't require a cleanup since there’s no stale data on their disks associated with these new tokens. Operator
+will trigger the cleanup anyway.
+ScyllaDB plans to automate cleanup and cover all the cases, until then we will run unnecessary cleanup in some cases.
+
+#### Cleanup is not run when necessary
+
+When keyspace RF is decreased, nodes no longer need to keep extraneous copies of the data, cleanup could free the disks.
+Approach designed here doesn't detect this case because the token ring is not changed.
+When decommission fails in the middle the token ring won’t be changed, but the tokens might be rebuilt on the nodes
+requiring a cleanup. Because this isn’t visible externally, the proposed solution won’t trigger a cleanup.
+When replication strategy of keyspace is changed, token ownership of secondary replicas is changed but the ring stays
+intact,
+hence cleanup won't be triggered.
+
+ScyllaDB plans to automate cleanup and cover all the cases, until then we won’t cover these cases.
+
+## Design Details
+
+Scylla-operator will be extended with a controller responsible for running a cleanup procedure when needed.
+
+### Cleanup trigger
+
+The root cause of the cleanup requirement is that during cluster changes tokens are moved between nodes.
+There are two alternatives of detecting when this happened:
+
+1. ScyllaCluster topology changes - new rack is added/removed, number of members changes, or additional CPUs are
+   added/removed.
+2. Check tokens in Scylla API, detect when it changed.
+
+The first one is edge driven, there are multiple cases when tokens might move or not, for example if cpu resource change
+won’t be big enough to allocate additional core the cleanup is not needed. For hybrid clusters or multi datacenter
+clusters, topology change in the non-managed part should cause the cleanup of nodes in the managed part which is
+impossible to detect. It’s also harder to implement since StatefulSets, which changes need to be detected, doesn't allow
+for per node distinction, the state needs to be carried to other resources (for example member Service) which brings
+unnecessary coupling.
+
+The second one is checking the actual root cause, and seems more bulletproof.
+
+### Implementation
+
+Two new annotations in member services will be reconciled. Controller in the sidecar will
+reconcile `internal.scylla-operator.scylladb.com/current-token-ring-hash`
+and Operator will reconcile `internal.scylla-operator.scylladb.com/last-cleaned-up-token-ring-hash`.
+
+The `current-token-ring-hash` one will always reflect the hashed output of Scylla API `/storage_service/tokens` endpoint
+from a given
+node. The `last-cleaned-up-token-ring-hash` will be initially empty and set after a cleanup job finishes
+to `current-token-ring-hash` value that was present at the time the cleanup job was created.
+
+The output of tokens endpoint is json formatted sorted array of strings (representing numbers), the annotation will hold
+SHA512
+hash of it as a value.
+
+The cleanup call is synchronous, client only triggers the operation, and all the resource intensive operations are done
+by Scylla. The duration of the operation is tied to the size of the entire dataset and disk performance. Cleanup on
+nodes
+that don't require it, also impacts the disk IO.
+To prevent the controller sync loop from being too long, the cleanup controller in Operator will be responsible for
+managing cleanup Jobs. The preconditions for Job for given node are:
+
+* `current-token-ring-hash` and `last-cleaned-up-token-ring-hash` token ring hash annotations values are different.
+* Observed generation is equal to ScyllaCluster generation.
+* Topology changes are done and upgrade is done - ScyllaCluster Status contains Condition `StatefulSetControllerProgressing=False`.
+* All nodes are UN - ScyllaCluster Status contains condition `Available=True` and `Degraded=False` conditions.
+
+The Job requires access to Scylla API to initiate the cleanup. Because Scylla API is bound to localhost, it means we
+have to go through Scylla Manager Agent which is able to forward the traffic to Scylla. It uses token based
+authentication, which means the controller will have to mount the Secret into the Job as a volume.
+The alternative is to extend the Operator sidecar with request forwarding capabilities, but since we already have this
+capability there’s no need for duplication.
+
+### Status
+
+Cleanup status will be visible as condition in `ScyllaCluster.status.conditions` via `CleanupControllerProgressing` and
+`CleanupControllerDegraded` and will influence the status of aggregated `Progressing` and `Degraded` ScyllaCluster
+conditions.
+
+## Test Plan
+
+E2E tests testing whether correct annotations are projected into services, whether Jobs are created for all nodes
+requiring a cleanup.
+
+## Upgrade / Downgrade Strategy
+
+Regular Operator upgrade/downgrade strategy.
+In case of a downgrade, new annotation will persist in Services, cleanup Jobs won’t be removed until ScyllaCluster
+exists.
+
+## Version Skew Strategy
+
+In case when an old Operator would be a leader, it would keep existing hash annotations without triggering cleanup. In
+case a new Operator is a leader, cleanup would be triggered when all conditions are matched.
+
+## Drawbacks
+
+As one of non-goals is lack of running cleanup during off-peak hours, and because cleanup is run in some cases when it's
+not needed users may see increased latency and decreased
+throughput after scaling the cluster for a period of running cleanup Jobs.


mykaul:
cc @bhalevy  - can you review the suggestion?

bhalevy:
Since this PR only contains the document for the proposal not the enhacement itself, how about renaming the PR subject to be the same as the patch: `Enhancement proposal for running cleanup after scaling`?

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
+# Run node cleanup after scaling the cluster
+
+## Summary
+
+The Operator supports both horizontal and vertical scaling, but the procedure isn’t in sync with ScyllaDB documentation,
+because the cleanup part is not implemented. It’s important because upon scaling, the stored on node disk might become
```
bhalevy, line 6: nit: stored data?

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+stale taking up unnecessary space. Operator should support running the cleanup to keep disk space as low as possible and
+ensure clusters are stable and reliable over time.
+
+## Motivation
+
+The motivation behind running node cleanup after scaling a cluster is to avoid the accumulation of unnecessary data on
+the node disks. When nodes are added or removed from the cluster, they gain or lose some tokens, which can result in
+files stored on the node disks still containing data associated with lost tokens. Over time, this can lead to a build-up
+of unnecessary data and cause disk space issues. By running node cleanup after scaling, these files can be cleared,
+freeing up disk space.
```
bhalevy, line 16: See above.  It's more than just cleaning up disk space.

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Goals
+
+* Running a node cleanup procedure after horizontally scaling the cluster.
+* Running a node cleanup procedure after scaling is completed, on all nodes but the last one added.
+
+### Non-Goals
+
+Running node cleanup during off-peak hours.
```
bhalevy, line 25: We can add: Running node cleanup after vertical scaling.
(Since it is not needed in this case)

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Proposal
+
+I propose to extend the sidecar with a new controller that will project the latest cleaned up and existing hash of
+tokens for each node as an annotation in the member service.
+In addition, a new controller in Operator will be responsible for managing Jobs that will execute a cleanup on nodes
+that require it. The trigger for the Job creation will be a mismatch between the current and latest hash. The controller
+will ensure that there will be only one cleanup Job running at the same time to prevent extraneous load on the cluster
```
bhalevy, line 33: nit: s/Job/job/

Should be one cleanup job per node I guess.
But we can (and should) run cleanup on all nodes in parallel.

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+This design doesn’t take into account whether a node received a token or lost it, it only detects ring changes and
+reacts with a cleanup trigger upon change. When a node is decommissioned, tokens are redistributed and nodes getting
+them doesn’t require a cleanup since there’s no stale data on their disks associated with these new tokens. Operator
+will trigger the cleanup anyway.
+ScyllaDB plans to automate cleanup and cover all the cases, until then we will run unnecessary cleanup.
+
+#### Cleanup is not run when necessary
+
+When keyspace RF is decreased, nodes no longer need to keep extraneous copies of the data, cleanup could free the disks.
+Approach designed here doesn't detect this case because the token ring is not changed.
```
bhalevy, line 56: Similarly, cleanup is needed after changing replication strategy, e.g. from Simple to NetworkTopology as token ownership of secondary replica will change.

By the way, speaking of automation in this respect - when increasing RF, repair need to run in order to build the additional replicas.  This is probably not automated as well IIUC.

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+The second one is checking the actual root cause, and seems more bullet proof.
+
+### Implementation
+
+Two new annotations in member services will be reconciled. Controller in the sidecar will
+reconcile `internal.scylla-operator.scylladb.com/current-token-ring-hash`
+and Operator will reconcile `internal.scylla-operator.scylladb.com/last-cleaned-up-token-ring-hash`.
+
+The `current` one will always reflect the hashed output of Scylla API `/storage_service/tokens` endpoint from a given
+node. The `last` will be either equal to `current` if it’s not yet present, or will be updated to current only when
```
tnozicka, line 90:
```suggestion
node. The `last-cleaned-up-token-ring-hash` will be initially empty and set after a clean up job finishes to `current` value that was present at the time the cleanup job was created.
```

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+cleanup
+Job for a given node is `Completed`.
+
+The output of tokens endpoint is an json sorted array of strings representing numbers, the annotation will keep SHA512
```
tnozicka, line 94:
```suggestion
The output of tokens endpoint is json formatted sorted array of strings (representing numbers), the annotation will hold SHA512
```

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+hash of it as a value.
+
+The cleanup call is synchronous, client only triggers the operation, and all the resource intensive operations are done
+by Scylla. The duration of the operation is tied to the size of the entire dataset and disk performance. Cleanup on node
```
tnozicka, line 98:
```suggestion
by Scylla. The duration of the operation is tied to the size of the entire dataset and disk performance. Cleanup on nodes
```

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+that don't require it, also impacts the disk IO.
+To prevent the controller sync loop from being too long, the cleanup controller in Operator will be responsible for
+managing cleanup Jobs. The preconditions for Job for given node are:
+
+* `current` and `last` token ring hash annotations values are different
+* All previous nodes have their `current` and `last` token ring hash annotations values equal - the order is rack order
+  in ScyllaCluster.spec and Pod ordinal - to prevent two cleanup Jobs running concurrently.
```
tnozicka, line 105: should be parallel

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+* StatefulSet controller is not progressing - topology changes are done.
```
tnozicka, line 106: should this be `Available=true,Progressing=False,Degraded=False`?

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+* All nodes in ScyllaCluster are UN - every rack status readyMembers matches the number of members in spec.
```
tnozicka, line 107: how will the controller check this without a client? through readiness state (which is already in SC conditions)?

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+
+The Job requires access to Scylla API to initiate the cleanup. Because Scylla API is bound to localhost, it means we
+have to go through Scylla Manager Agent which is able to forward the traffic to Scylla. It uses token based
+authentication, which means the controller will have to mount the Secret keeping it inside the Job as a volume.
```
tnozicka, line 111:
```suggestion
authentication, which means the controller will have to mount the Secret into the Job as a volume.
```

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+### Status
+
+Cleanup status will be visible as condition in `ScyllaCluster.status.conditions` via `CleanupControllerProgressing` and
+`CleanupControllerDegraded` and will influence the status of aggregated `Progressing` and `Degraded` ScyllaCluster
+conditions.
+
+## Test Plan
+
+E2E tests testing whether correct annotations are projected into services, whether Jobs are created for all nodes
+requiring a cleanup and whether there’s at most one Job running per cluster.
```
tnozicka, line 124: parallel

tnozicka:
> Also, what is this file for: enhancements/proposals/dir_placeholder.delete_me_with_first_proposal?

git doesn't track folders, so this was used as a placeholder for that directory until we have the first proposal here which should remove it.

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
+# Run node cleanup after scaling the cluster
+
+## Summary
+
+The Operator supports both horizontal and vertical scaling, but the procedure isn’t in sync with ScyllaDB documentation,
+because the cleanup part is not implemented. It’s important because upon scaling, the stored on node disk might become
+stale taking up unnecessary space. Operator should support running the cleanup to keep disk space as low as possible and
+ensure clusters are stable and reliable over time.
```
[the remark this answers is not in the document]
zimnx: Added mention about the data resurrection to Motivation paragraph

enhancements/proposals/cleanup_after_scaling.md
zimnx: Added mention about the data resurrection to Motivation paragraph

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: fixed

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+When a cluster is created for the first time, tokens are reshuffled when new nodes are joining sequentially. This will
+cause the token hash mismatch and will require a cleanup once all nodes join the cluster. This will cause cluster
+bootstrap time to be increased. But since the time window is short, nodes shouldn’t have much data stored so cleanup
+won’t take long.
+
+#### Cleanup is run when not necessary
+
+This design doesn’t take into account whether a node received a token or lost it, it only detects ring changes and
+reacts with a cleanup trigger upon change. When a node is decommissioned, tokens are redistributed and nodes getting
+them doesn’t require a cleanup since there’s no stale data on their disks associated with these new tokens. Operator
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: Added mention about it.

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,140 @@
...
+will trigger the cleanup anyway.
+ScyllaDB plans to automate cleanup and cover all the cases, until then we will run unnecessary cleanup.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
zimnx: No, because aggregated conditions is also computed from JobController conditions. Added what is checked.

enhancements/proposals/cleanup_after_scaling.md
zimnx: Added mention about how it's checked

enhancements/proposals/cleanup_after_scaling.md
zimnx: done

enhancements/proposals/cleanup_after_scaling.md
tnozicka: progressing==False doesn't mean all the nodes are UN, that's why I've suggested the other conditions as well.
Technically, this should also check generation for the status it sees.

There is also the question of whether we'd do a live call in case the informers are stale.

enhancements/proposals/cleanup_after_scaling.md
tnozicka: progressing==False doesn't mean all the nodes are UN, that's why I've suggested the other conditions as well.
Technically, this should also check generation for the status it sees.

There is also the question of whether we'd do a live call in case the informers are stale.

zimnx: `StatefulSetControllerProgressing=False`  means all required statefulsets are fully rolled out which means all reports UN.
https://github.com/scylladb/scylla-operator/blob/30f6e46908f00e9e493392b7570ad2dac0bb872e/pkg/controller/scyllacluster/sync_statefulsets.go#L597-L617
In addition it means upgrade is finished which is also thing we we would like to wait before cleanup.

Good point about generation, i'll cover it it implementation.

tnozicka: > StatefulSetControllerProgressing=False means all required statefulsets are fully rolled out which means all reports UN.

I don't think that's what `Progressing` means. The controller currently sets `Progressing` condition because the algorithm "artificially" waits for a rollout internally but the implication goes only one way - if the controller waits for a rollout it sets the progressing condition to manifest it to external observers. It doesn't imply that `Progressing` condition means all nodes are UN.

`Progressing` - Is the controller waiting for something to happen before making progress?
`Available` - Is enough pods up/available? (Say for an HA service that can be 2 out of 3)
`Degraded` - Are some pods down? (Object can be both available and degrade if it's an HA service)

That's why our e2e also wait for those conditions and not just `Progressing`.

https://github.com/scylladb/scylla-operator/blob/9518217ba718d30af4343d85959c90b43a47c32a/test/e2e/utils/helpers.go#L148-L159

In general and as an example, say, in a fully rolled out cluster, a container gets suddenly restarted, that should only flip `Degraded` (possibly `Available` as well), not `Progressing` as the controller isn't taking any action or waiting for something before making a reconciliation action (unless there is one based on that).

zimnx: It means it's in the code at given revision, not what it is supposed to mean. Described expected behavior doesn't reflect what it is currently in the code.

But I agree relying on side effect isn't a good idea. Updated the proposal.

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,153 @@
...
+nodes
+that don't require it, also impacts the disk IO.
+To prevent the controller sync loop from being too long, the cleanup controller in Operator will be responsible for
+managing cleanup Jobs. The preconditions for Job for given node are:
+
+* `current-token-ring-hash` and `last-cleaned-up-token-ring-hash` token ring hash annotations values are different.
+* Observed generation is equal to ScyllaCluster generation.
+* Topology changes are done - ScyllaCluster Status contains Condition `StatefulSetControllerProgressing=False`.
+* Upgrade is not running - Upgrade status in ScyllaCluster status is not set.
+* All nodes are UN - rack status ReadyMembers is equal to member count in ScyllaCluster rack spec.
```
tnozicka, line 119: can we phrase this as checking `Degraded` and `Available` conditions that already aggregate node statuses?

enhancements/proposals/cleanup_after_scaling.md
```diff
@@ -0,0 +1,153 @@
...
+by Scylla. The duration of the operation is tied to the size of the entire dataset and disk performance. Cleanup on
+nodes
+that don't require it, also impacts the disk IO.
+To prevent the controller sync loop from being too long, the cleanup controller in Operator will be responsible for
+managing cleanup Jobs. The preconditions for Job for given node are:
+
+* `current-token-ring-hash` and `last-cleaned-up-token-ring-hash` token ring hash annotations values are different.
+* Observed generation is equal to ScyllaCluster generation.
+* Topology changes are done - ScyllaCluster Status contains Condition `StatefulSetControllerProgressing=False`.
+* Upgrade is not running - Upgrade status in ScyllaCluster status is not set.
```
tnozicka, line 118: should that be manifested in the progressing condition?

zimnx: it's subcondition of `StatefulSetControllerProgressing=False` which I check for topology changes, so i merged these into checking just single one.

rzetelskik:
I think that instead of adding a file with the proposal directly to the top directory, we should adopt a similar naming scheme to the original KEP repository, i.e. a subdirectory prefixed with a tracking issue number and a readme inside of it. I believe it provides better readability and ordering, as well as a reference to the origin of the proposal. See [What is the number at the beginning of the KEP name?](https://github.com/kubernetes/enhancements/blob/master/keps/README.md#what-is-the-number-at-the-beginning-of-the-kep-name).

zimnx:
> I think that instead of adding a file with the proposal directly to the top directory, we should adopt a similar naming scheme to the original KEP repository, i.e. a subdirectory prefixed with a tracking issue number and a readme inside of it. I believe it provides better readability, as well as a reference to the origin of the proposal. See [What is the number at the beginning of the KEP name?](https://github.com/kubernetes/enhancements/blob/master/keps/README.md#what-is-the-number-at-the-beginning-of-the-kep-name).

I agree, in addition having a separate directory allows for external file injection like images, yamls etc. Moved it.

tnozicka (approved):
/approve
/lgtm

thanks

PR #1306, opened 2023-07-31
Add enhancement proposal for controlling external seeds

### LINKED ISSUE #1304: Create enhancmet proposal for controlling external seeds on ScyllaClusters
We need a proposal describing ScyllaCluster API extentions to expose a way to specify seeds for connecting to other datacenters. It should explore how it works in ScyllaDB, if the flag(s) we can be idempotent / reentrant and what changes are needed for the operator.

This proposal should be limited to connecting 2+ ScyllaCluster together manually using an API primitive on our existing ScyllaCluster CR. We may later build on top of that with a new CR but this is the necessary building block that unblocks a way to set the datacenters manually.

While https://github.com/scylladb/scylla-operator/issues/1270 is ongoing it shouldn't limit trying this out as technically you should be able to connect 2+ datacenters (ScyllaCluster) within the same Kubernetes cluster. (You don't get the extra production availability as you are limited to a single region but technically that should work fine.) 

Maybe we can play with `ScyllaCluster.spec.scyllaArgs` to set the seed before we have a dedicated API.

rzetelskik:
Currently, ScyllaDB Operator does not provide any support, neither automated, nor manual, for controlling external seeds propagated to ScyllaDB cluster nodes.
Controlling the provision of external seeds is a prerequisite for supporting both manual and automated setup of multi datacenter ScyllaDB clusters, a feature long requested by many of its users, and as such is a vital step on our roadmap.

This PR proposes an extension to the existing ScyllaCluster API and corresponding changes to the ScyllaDB operator and sidecar.

Resolves #1304


enhancements/proposals/external_seeds.md
```diff
@@ -0,0 +1,145 @@
...
+    ...
+	
+	// externalSeeds specifies the external seeds to propagate to ScyllaDB binary on startup.
+	// +optional
+	ExternalSeeds []string `json:"externalSeeds,omitempty"`
+}
+```
+
+The provided `externalSeeds` are passed to the sidecar through a new `--external-seeds` flag. StatefulSet reconciliation in ScyllaDB operator is modified accordingly.
+If any external seeds are provided, the sidecar overrides the locally calculated seeds with the provided external seeds.
```
zimnx, line 79: How would Scylla behave when one of the seeds is invalid, and other one is correct? I wonder if we should completely rely on user input, or rather we should add a local node as seed as a failover.

enhancements/proposals/external_seeds.md
```diff
@@ -0,0 +1,145 @@
...
+In addition to that, the document also proposes the procedure for compiling the list of seeds propagated to ScyllaDB processes when external seeds are provided.
+
+### User Stories
+
+#### As a ScyllaDB cluster administrator:
+- I want to control the list of external seeds provided to ScyllaDB nodes.
+- I want to create a multi datacenter ScyllaDB by manually providing the external seeds to chosen datacenters using the ScyllaCluster CR.
+- I want to add a datacenter to existing ScyllaDB datacenters by manually providing the external seeds to chosen datacenters using the ScyllaCluster CR.
+
+### Notes/Constraints/Caveats [Optional]
```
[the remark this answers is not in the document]
rzetelskik: It's not empty though? There are notes on procedures and current roles of seeds in ScyllaDB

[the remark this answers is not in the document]
zimnx: Right, bug in my brain parser.

enhancements/proposals/external_seeds.md
```diff
@@ -0,0 +1,145 @@
...
+[Create a ScyllaDB Cluster - Multi Data Centers (DC)](https://opensource.docs.scylladb.com/branch-5.3/operating-scylla/procedures/cluster-management/create-cluster-multidc.html#procedure) procedure in ScyllaDB documentation explicitly states to use IP address of the first node, *and only the first node*, as seed for all nodes in all datacenters.
+This has proven to be incorrect in two ways: 
+1. Domain name addresses are also accepted as seeds. 
+2. There is no special significance to the first node of the first DC - it is just a simplification used for the purpose of minimising the risk of creating multiple clusters instead of a single, multi DC cluster.
+
+The specified constrains hold - the datacenters need a matching cluster name and distinct datacenter and rack names.
+
+### Risks and Mitigations
+
+The proposed enhancement does not carry any risks for the functionality or stability of the operator.
```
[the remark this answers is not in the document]
rzetelskik: Changes in topology won't be possible either way in case of an unstable cluster (not all nodes UN). But there are cases where providing additional seeds from local DC is going to make a difference, e.g. the case we've discussed offline today - where all nodes are DN for some reason and they fall back to using seeds https://github.com/scylladb/scylladb/blob/6c66030b7bbcf3ef5d962a0f4ca48d46582ca3a2/gms/gossiper.cc#L947. Overall appending a seed from local DC seems to improve resiliency in some cases - I'll amend the proposal

enhancements/proposals/external_seeds.md
```diff
@@ -0,0 +1,145 @@
...
+type ScyllaClusterSpec struct {
+    ...
+	
+	// externalSeeds specifies the external seeds to propagate to ScyllaDB binary on startup.
+	// +optional
+	ExternalSeeds []string `json:"externalSeeds,omitempty"`
+}
+```
+
+The provided `externalSeeds` are passed to the sidecar through a new `--external-seeds` flag. StatefulSet reconciliation in ScyllaDB operator is modified accordingly.
```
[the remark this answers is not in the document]
rzetelskik: https://github.com/scylladb/scylladb/blob/6c66030b7bbcf3ef5d962a0f4ca48d46582ca3a2/gms/gossiper.cc#L947 Scylla falls back to using initial contact points in case it can't reach any other nodes, so I believe providing the option to update externalSeeds seems like a failsafe which we might not want to get rid of. Wdyt? Also I'll add this information to the proposal.

enhancements/proposals/external_seeds.md
rzetelskik: If it's an unresolvable or invalid address, scylla process is going to fail either way: https://github.com/scylladb/scylladb/blob/39ca07c49b257d37856164ea5b845f7b98224b98/init.cc#L37.
I assume you meant a case where a node can't communicate over this provided external address, but can communicate with nodes from its DC. In that case, assuming the entire clusters is UN, using only the external seeds could block operations, you're right. As long as we use only the external seeds for the first joining node in a DC, we should be safe from accidentally creating separate clusters, which was my concern. I'll amend the proposal.

rzetelskik: > If it's an unresolvable or invalid address, scylla process is going to fail either way: https://github.com/scylladb/scylladb/blob/39ca07c49b257d37856164ea5b845f7b98224b98/init.cc#L37.

Also logged an issue for that: https://github.com/scylladb/scylladb/issues/14945

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Drawbacks
+
+### Potential misconceptions related to multi datacenter setup
+One of the non-goals of this proposal is automating the creation of multi datacenter ScyllaDB clusters. Extending the ScyllaCluster API with an option of controlling the external seeds provided to ScyllaDB nodes might create a misconception that ScyllaDB operator automates the set operations related to a multi datacenter setup, which would normally be handled in a single datacenter cluster. Such operations might involve e.g. decommissioning the nodes of a datacenter before deleting the ScyllaCluster resource.
+
+## Alternatives
+
+### Seed overriding in sidecar
+As ana alternative to the proposed algorithm compiling the seeds provided to ScyllaDB nodes subsequent to the first in the DC, overriding any seeds with just the external seeds has been considered.
```
zimnx, line 151:
```suggestion
As an alternative to the proposed algorithm compiling the seeds provided to ScyllaDB nodes subsequent to the first in the DC, overriding any seeds with just the external seeds has been considered.
```

zimnx (dismissed):
LGTM, thanks!

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: done

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+Other than that, seeds do not have any additional significance past bootstrap.
+
+#### Multi DC setup procedure
+
+[Create a ScyllaDB Cluster - Multi Data Centers (DC)](https://opensource.docs.scylladb.com/branch-5.3/operating-scylla/procedures/cluster-management/create-cluster-multidc.html#procedure) procedure in ScyllaDB documentation explicitly states to use IP address of the first node, *and only the first node*, as seed for all nodes in all datacenters.
+This has proven to be incorrect in two ways: 
+1. Domain name addresses are also accepted as seeds. 
+2. There is no special significance to the first node of the first DC - it is just a simplification used for the purpose of minimising the risk of creating multiple clusters instead of a single, multi DC cluster.
+
+The specified constrains hold - the datacenters need a matching cluster name and distinct datacenter and rack names.
```
tnozicka, line 55: Can you clarify the sentence so it's clear whether rack names have to be unique across DC, or just withing a DC?

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+The following extension to the existing ScyllaCluster CR API is proposed:
+```go
+// ScyllaClusterSpec defines the desired state of Cluster.
+type ScyllaClusterSpec struct {
+    ...
+	
+	// externalSeeds specifies the external seeds to propagate to ScyllaDB binary on startup as "seeds".
+	// See https://opensource.docs.scylladb.com/stable/kb/seed-nodes.html.
+	// For the first running node in a datacenter, the external seeds are the only seeds provided.
+	// For the remaining nodes, one other node from their datacenter is appended to the external seeds.
```
tnozicka, line 84: nit: I'd not put his level of detail into the API.
```suggestion
```

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+The provided `externalSeeds` are passed to the sidecar through a new `--external-seeds` flag. StatefulSet reconciliation in ScyllaDB operator is modified accordingly.
+The sidecar's strategy for choosing seeds to provide to ScyllaDB processes is modified to follow the below pseudocode:
+```go
+func chooseSeeds(externalSeeds []string, seedFromLocalDC string) []string {
+  if is first node in DC {
+    if externalSeeds not empty {
+      return externalSeeds
+    }
+    return seedFromLocalDC
```
tnozicka, line 98: if you are the first node, what's the seed here?

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+The below examples assume the two datacenters are to be deployed in one Kubernetes cluster.
+
+### Creating a multi datacenter ScyllaDB cluster
+To create a multi datacenter ScyllaDB cluster, an end user would first deploy a ScyllaCluster. For the purpose of this document, let's assume name `simple-cluster`, namespace `scylla1`, datacenter `us-east-1` and a single rack `us-east-1a`.
+To create the second datacenter, the end user would deploy a second ScyllaCluster with the same name, in a different namespace, with a distinct datacenter and rack names, and `.spec.externalSeeds` set to a list of IP address or domain names under which one or more nodes of the first datacenter are reachable from the second datacenter. 
+ScyllaDB operator will propagate the declared `externalSeeds` through a sidecar flag, and the sidecar will take it into account when configuring the seeds provided to ScyllaDB nodes of the second datacenter. The two datacenters will then form a single multi datacenter cluster.
+
+### Adding a new datacenter to an existing ScyllaDB cluster
+To add a new datacenter to an existing ScyllaCluster, the same procedure as above is followed.
```
tnozicka, line 117: Can we explicitly mention this working for connecting 2 existing and rolled out ScyllaClusters?

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Test Plan
+
+To test the proposed API extension, unit tests verifying that the external seeds are propagated to the sidecar will be added.
+Additionally, E2E tests should verify whether a single, multidatacenter ScyllaDB cluster is formed when additional datacenters with external seeds pointing to existing datacenter are created.
```
tnozicka, line 122: should we test connecting 2 existing cluster as well?

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Upgrade / Downgrade Strategy
+
+Regular Operator upgrade/downgrade strategy.
+
+### Version Skew Strategy
+
+#### Operator and sidecar version skew
+
+In case of a new operator and old sidecar, sidecar containers will fail due to an unknown flag.
```
tnozicka, line 132: I struggle to see the case on upgrade. Even if the operator version is changed it's still the old sidecar with old flags (flags go step in step with sidecar image)

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+In case of an old operator and new sidecar, the external seeds won't be propagated to sidecar.
+
+#### Operator and ScyllaCluster CR version skew
+In case of a new operator and old ScyllaCluster CR, the default, empty slice will be used for external seeds.
+In case of old operator and new ScyllaCluster CR, external seeds will not be supported.
```
tnozicka, line 137: > and old ScyllaCluster CR,

these cases shouldn't be about the presence of a CR(D) but rather about the value stored there (the CRD may not be downgraded):
- old operator and `externalSeeds` are set or null (after downgrade) - operator will ignore them and replaces the deployment to unconfigure them
- new operator seeing `externalSeeds` as null (after an upgrade) - no action taken

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: Connecting two bootstrapped DCs isn't guaranteed with just a rollout. Since both DCs are already bootstrapped, patching one of them with external seeds will only result in forming a multi DC cluster if at least one of the nodes falls back to using initial contact points, i.e. only if it is not able to talk with any of the other nodes from its own DC, as described in `Current role of seeds in ScyllaDB` section of this proposal. All of that despite the fact that patching an existing DC with external seeds results in following the procedure described in [this page of ScyllaDB docs]( https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/add-dc-to-existing-dc.html). It's probably outdated, since I don't see how just restarting the nodes of a bootstrapped cluster one by one would always result in forming a multi DC cluster. I don't have any ideas for covering this case easily, but I'm actually wondering whether this should even be expected to be covered?

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: see the above comment

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+This document proposes extending the existing ScyllaCluster CR with a new field in its Specification which allows the users to control the external seeds provided to ScyllaDB processes.
+The changes required for ScyllaDB Operator and the sidecar to support the enhancement are also described.
+In addition to that, the document also proposes the procedure for compiling the list of seeds propagated to ScyllaDB processes when external seeds are provided.
+
+### User Stories
+
+#### As a ScyllaDB cluster administrator:
+- I want to control the list of external seeds provided to ScyllaDB nodes.
+- I want to create a multi datacenter ScyllaDB by manually providing the external seeds to chosen datacenters using the ScyllaCluster CR.
+- I want to add a datacenter to existing ScyllaDB datacenters by manually providing the external seeds to chosen datacenters using the ScyllaCluster CR.
```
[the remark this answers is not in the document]
rzetelskik: Can you elaborate? I'm not sure I understand what you mean by that. Removing the external seeds? Once the seeds have been provided for bootstrap, removing them from the seed list past bootstrap won't do anything in terms of topology. I guess I could mention that in risks, if anywhere?

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: added a comment about getting seeds from own DC

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: done

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,156 @@
...
+## Proposal
+
+This document proposes extending the existing ScyllaCluster CR with a new field in its Specification which allows the users to control the external seeds provided to ScyllaDB processes.
+The changes required for ScyllaDB Operator and the sidecar to support the enhancement are also described.
+In addition to that, the document also proposes the procedure for compiling the list of seeds propagated to ScyllaDB processes when external seeds are provided.
+
+### User Stories
+
+#### As a ScyllaDB cluster administrator:
+- I want to control the list of external seeds provided to ScyllaDB nodes.
```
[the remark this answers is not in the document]
rzetelskik: done

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: You're right. Removed this section.

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: done

enhancements/proposals/1304-external-seeds/README.md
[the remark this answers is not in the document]
tnozicka: I meant a complement to `I want to add a datacenter` - `I want to remove a datacenter`. It's fine to list as a non-goal or to say, you remove the seed and manually remove the DC from ScyllaDB

enhancements/proposals/1304-external-seeds/README.md
tnozicka: ok, let's list it as a non-goal then and be aware that the the DC can be join only when being created.

enhancements/proposals/1304-external-seeds/README.md
tnozicka: who is the seedFromLocalDC if you are the first node (and the only node) in that DC?

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,147 @@
...
+    return seedFromLocalDC
+  }
+
+  return append(externalSeeds, seedFromLocalDC)
+}
+```
+The strategy for getting a seed from local DC (`seedFromLocalDC`) is not modified, i.e. for the first running node in the DC, its own address (`broadcast_address`) is selected, otherwise an address of one other running node from its own DC is selected.
+
+It is important that the external seeds, and only the external seeds, are provided as seeds for the first running node in a datacenter, if they are at all provided by the end user. If external seeds were only appended to the node's own IP, it could result in creating split clusters in case of lack of connectivity between the datacenters.
+For the other nodes, a seed from their own DC is appended to the external seeds for increased resiliency.
```
tnozicka, line 104: can other nodes default to it's own IP when the rest is unavailable resulting in the same situation as described for the first node?

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+The above changes allow for configuring multi datacenter ScyllaDB clusters by manually providing external seeds.
+The below examples assume the two datacenters are to be deployed in one Kubernetes cluster.
+
+### Creating a multi datacenter ScyllaDB cluster
+To create a multi datacenter ScyllaDB cluster, an end user would first deploy a ScyllaCluster. For the purpose of this document, let's assume name `simple-cluster`, namespace `scylla1`, datacenter `us-east-1` and a single rack `us-east-1a`.
```
tnozicka, line 110: nit: use a `-` for better clarity of the namespace name

enhancements/proposals/1304-external-seeds/README.md
```diff
@@ -0,0 +1,147 @@
...
+To create the second datacenter, the end user would deploy a second ScyllaCluster with the same name, in a different namespace, with a distinct datacenter and rack names, and `.spec.externalSeeds` set to a list of IP address or domain names under which one or more nodes of the first datacenter are reachable from the second datacenter.
```
tnozicka, line 111: do the rack names have to be distinct as well? while this sentence says so, [L54](https://github.com/scylladb/scylla-operator/pull/1306/files#diff-b726dfb1c3c35fbb0f4714d31b9bf2b57b225df6b69cb1b8458836a89ad56f38R54) says otherwise

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: No, as described there the seeds they are provided on bootstrap are external seeds + an address of one other running node from their DC. The first one in the DC won't be able to get up if there is no connectivity between them and external seeds. For the the other nodes in the DC, they will learn the topology from the other node in their DC even if there's no connectivity to the external seeds. So it's impossible that they form a split cluster at this point.

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: removed the names altogether as they are not used later in the description

enhancements/proposals/1304-external-seeds/README.md
rzetelskik: no, amended

enhancements/proposals/1304-external-seeds/README.md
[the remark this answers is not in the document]
rzetelskik: done

enhancements/proposals/1304-external-seeds/README.md
tnozicka: Scenario:
- DC A is already bootstrapped
- dc-b-node-0 starts and bootstraps
- dc-b-node-1 pods is created but before it starts, dc-b-node-0 gets evicted
- dc-b-node-1 has no external seeds (as for this proposal) and hits
https://github.com/scylladb/scylla-operator/blob/d8e7d3c7580bbdd613ca673b115137a70c0f9294/pkg/sidecar/identity/member.go#L74-L77
which uses it's own Service IP as a seed instead of the first node or the other DC - witting it down it seems that might be an issue in the single DC case as well

rzetelskik: > dc-b-node-1 has no external seeds (as for this proposal)

I don't see this situation happening. There's two possibilities for the seeds if external seeds were provided to a DC:
- *only* external seeds for first running node in DC
- external seeds + one seed from own DC for other nodes

So in this situation, dc-b-node-1 would be provided with only the external seeds and should join the cluster correctly. 

As for the single DC, yes, it looks like it might be an issue, but it's not introduced by this proposal.

Issue #1201, opened 2023-04-04
docs: update theme 1.5


dgarcia360:
Related issue https://github.com/scylladb/sphinx-scylladb-theme/issues/310

ScyllaDB Sphinx Theme 1.5 is now released 🥳 You can read more about all notable changes [here](https://sphinx-theme.scylladb.com/master/upgrade/CHANGELOG.html#may-2023).

## Upgrade guide

For most repositories, we are sending pull-requests with new updates. Unfortunately, we cannot do this for this project since the [docs workflows](https://github.com/scylladb/sphinx-scylladb-theme/tree/master/.github/workflows) shared between all projects to build the docs [were highly modified](https://github.com/scylladb/scylla-operator/tree/master/.github/workflows).

> **TIP:** If you modify these workflows in your repository, you will need to maintain the changes. So instead, we recommend you to open an issue / pull-request in the [sphinx-scylladb-theme repository](https://github.com/scylladb/sphinx-scylladb-theme) so that all projects can benefit from the improvements you made.

Here you have resources to help you upgrade the docs toolchain to the latest version:

**1.2. to 1.3**
- [Migrating from 1.2 to 1.3](https://sphinx-theme.scylladb.com/stable/upgrade/1-2-to-1-3.html)
- [Sample pull-request 1.2 to 1.3](https://github.com/scylladb/scylla-monitoring/pull/1778)

**1.3 to 1.4**

- [Migrating from 1.3 to 1.4](https://sphinx-theme.scylladb.com/stable/upgrade/1-3-to-1-4.html)
- [Sample pull-request 1.3 to 1.4](https://github.com/scylladb/care-pet/pull/124/files)

**1.4 to 1.5**

- [Migrating from 1.4 to 1.5](https://sphinx-theme.scylladb.com/master/upgrade/1-4-to-1-5.html)
- [Sample pull-request 1.4 to 1.5](https://github.com/scylladb/care-pet/pull/139/files)

Do not hesitate to contact me (@dgarcia360 in Slack) if you face any issues during the process.

dgarcia360:
We've released a new version. Updated the initial issue with the 1.4 to 1.5 migration guide.

annastuchlik:
@tnozicka Could you review and merge this PR?
The Operator documentation is behind the rest of the docs, as it has the Sphinx theme version 1.3.5. The later releases have different fixes that should be applied to all project docs.

annastuchlik:
@tnozicka ping

PR #1308, opened 2023-08-03
Update docs theme version to 1.5

rzetelskik:
**Description of your changes:** This PR updates the docs theme version to 1.5 as per https://sphinx-theme.scylladb.com/stable/upgrade/1-3-to-1-4.html and https://sphinx-theme.scylladb.com/stable/upgrade/1-4-to-1-5.html.
Underscores in source files are also replaced with hyphens due to a warning being thrown by Sphinx, causing the build to fail.

Resolves https://github.com/scylladb/scylla-operator/issues/1201


annastuchlik:
@rzetelskik The new version of the theme triggers a warning if an underscore is encountered in a file name. The following files should be renamed to use a hyphen instead of the underscore:
- known_issues
- nodeoperations/automatic_cleanup
- nodeoperations/maintenance_mode
- nodeoperations/replace_node
- nodeoperations/scylla_upgrade
- scylla_cluster_crd

rzetelskik:
@annastuchlik yup, I've updated the PR just before your comment. Should be fine now.

docs/Makefile
```diff
@@ -18,6 +18,7 @@ all: dirhtml
 
 clean:
 	rm -rf '$(BUILDDIR)'/*
+	rm -f poetry.lock
```
tnozicka, line 21: I don't think this belongs here. `clean` is meant to clean build files which can always be regenerated to the same output, `poetry.log` actually holds information (snapshot of all the moving references in a certain time), same as go.sum.

tnozicka (commented):
modulo Makefile, this lgtm

docs/Makefile
rzetelskik: I followed the [upgrade guide](https://sphinx-theme.scylladb.com/stable/upgrade/1-3-to-1-4.html), but personally I agree, it sort of defeats the purpose of even having the `.lock` file. @annastuchlik why is that considered a "breaking change" in the upgrade guide?

annastuchlik:
Let me look into it with @dgarcia360 before I approve or suggest an update. Thanks.

docs/Makefile
tnozicka: > I followed the upgrade guide
We don't follow their Makefile, nor CI for differences on opinions on what reproducibility means or make should look like. Totally fine to differ in `clean` target.

Having to remove `.lock` probably suggest something wrong is being done somewhere else but even so it should be one shot, not a part of regular cleanup.

rzetelskik: I agree. I removed the Makefile change.

annastuchlik:
I'm unable to build the docs. It looks like more updates are required to the Makefile.

tnozicka (approved):
/lgtm

I'll give @annastuchlik a while to see if it works for her

rzetelskik:
> I'm unable to build the docs. It looks like more updates are required to the Makefile. 

@annastuchlik can you elaborate? What did you try? What do you mean by being unable to build the docs?

annastuchlik:
I get the following error:

```
Traceback (most recent call last):
  File "C:\Users\annstu\.poetry\bin\poetry", line 17, in <module>
    from poetry.console import main
ModuleNotFoundError: No module named 'poetry.console'
make: *** [Makefile:24: preview] Error 1
```

rzetelskik:
> I get the following error:
> 
> ```
> Traceback (most recent call last):
>   File "C:\Users\annstu\.poetry\bin\poetry", line 17, in <module>
>     from poetry.console import main
> ModuleNotFoundError: No module named 'poetry.console'
> make: *** [Makefile:24: preview] Error 1
> ```

@annastuchlik please try the following:
```
podman run -it --rm -v="$( pwd ):/go/$( go list -m )" --workdir="/go/$( go list -m )/docs" -p 5500:5500 ubuntu:20.04 bash -c 'apt-get update && apt-get install -y curl python3 python3-distutils make git && ./hack/install-poetry.sh && $HOME/.poetry/bin/poetry install && make dirhtml SPHINXOPTS="--keep-going" && make redirects && make preview'
```
It's what our workflow does essentially.

dgarcia360:
Hi @annastuchlik, @tnozicka, and @rzetelskik,

While these changes will build the docs as expected for prod, building the docs locally differs from how we do it in other projects.  I'm sharing some proposed modifications to the Makefile:

1. Modify the `POETRY `variable to ensure compatibility for Windows:

    ```
    POETRY        = poetry
    
    # For Windows users
    ifeq ($(OS),Windows_NT)
        POETRY = $(APPDATA)\Python\Scripts\poetry
    endif
    ```
    
2. Include the  `setup` step before executing commands to generate documentation. For example:

    ```
    # Preview commands
    .PHONY: preview
    preview: setup
	    $(POETRY) run sphinx-autobuild -b dirhtml $(ALLSPHINXOPTS) $(BUILDDIR)/dirhtml --port 5500
    ```
  
    This ensures that the latest version of the dependencies are installed before building the docs.
  
     If this suggestions don't align with your preferences, I'd like to propose an alternative. To provide more context, I'm sharing a README from another project that outlines their build process, which is different from the one provided by the Sphinx Theme:https://github.com/scylladb/cpp-driver/blob/dacab2ced2fa84d798723d13ecb8c8f658f28d30/README-dev.md
  
     You could create a similar "README-dev.md" within the docs directory, explaining the considerations needed to run the docs locally. For example: 
  
    ```md
    # Building the docs
    
    This project uses [Sphinx Theme](https://sphinx-theme.scylladb.com/) , but we have modified the base project to adapt it to our workflows.
    
    ## Local preview
    
    To build the documentation locally, run `make setup` first.
    
    Then, you can build and preview the documentation by following the steps outlined in the [Quickstart guide](https://sphinx-theme.scylladb.com/stable/getting-started/quickstart.html).
    
    ## Custom workflows 
    
    (optional) Explain other  customization. 
    ```

rzetelskik:
@dgarcia360 thanks! 
Ad 2. I suppose we could add dirhtml/multiversion and redirects as prerequisites for preview and multiversionpreview.
I won't affect our workflow and if it'll align our Makefile with other projects, then I don't see why not. Same goes for Windows compatibility.

Having said that, it should go in a separate PR. @tnozicka any objections?

rzetelskik:
@tnozicka @annastuchlik @dgarcia360 please see https://github.com/scylladb/scylla-operator/pull/1323

tnozicka:
/lgtm

With readme shipping in #1323 are there any concerns about the content of this PR or can we land this bump?

rzetelskik:
> With readme shipping in https://github.com/scylladb/scylla-operator/pull/1323 are there any concerns about the content of this PR or can we land this bump?

I don't think they interfere with each other so I believe they can land in any order.

annastuchlik:
@rzetelskik @tnozicka Thanks!

PR #1321, opened 2023-08-10
Expose external seeds in ScyllaCluster

### LINKED ISSUE #1318: Expose external seeds on ScyllaClusters
We need to expose external seeds based on proposal from #1304

rzetelskik:
Currently, ScyllaDB Operator does not provide any support, neither automated, nor manual, for controlling external seeds propagated to ScyllaDB cluster nodes.
Controlling the provision of external seeds is a prerequisite for supporting both manual and automated setup of multi datacenter ScyllaDB clusters, a feature long requested by many of its users, and as such is a vital step on our roadmap.

This PR adds an option to expose external seeds in ScyllaClusters and extends the operator with necessary changes. It also adjusts our E2E framework and datainserter to support multi datacenter deployments and adds an E2E covering a multidatacenter deployment using external seeds.

The implementation is aligned with the proposal https://github.com/scylladb/scylla-operator/pull/1306.

Resolves #1318 

**Prerequisites**:
- [x] https://github.com/scylladb/scylla-operator/pull/1306

### DIFF
--- a/pkg/api/scylla/v1/types_cluster.go
+++ b/pkg/api/scylla/v1/types_cluster.go
@@ -106,6 +106,9 @@ type ScyllaClusterSpec struct {
 	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
 	// +optional
 	ExposeOptions *ExposeOptions `json:"exposeOptions,omitempty"`
+
+	// externalSeeds specifies the external seeds to propagate to ScyllaDB binary on startup as "seeds" parameter of seed-provider.
+	ExternalSeeds []string `json:"externalSeeds,omitempty"`
 }
 
 // ExposeOptions hold options related to exposing ScyllaCluster backends.
--- a/pkg/controller/scyllacluster/resource_test.go
+++ b/pkg/controller/scyllacluster/resource_test.go
@@ -929,6 +929,8 @@ func TestStatefulSetForRack(t *testing.T) {
 		}
 	}
 
+	const scyllaContainerIndex = 0
+
 	tt := []struct {
 		name                string
 		rack                scyllav1.RackSpec
@@ -1081,6 +1083,24 @@ func TestStatefulSetForRack(t *testing.T) {
 			expectedStatefulSet: func() *appsv1.StatefulSet {
 				sts := newBasicStatefulSet()
 				sts.Spec.Template.Annotations[naming.ForceRedeploymentReasonAnnotation] = "reason"
+				return sts
+			}(),
+			expectedError: nil,
+		},
+		{
+			name: "new StatefulSet with non-empty externalSeeds in scylla container",
+			rack: newBasicRack(),
+			scyllaCluster: func() *scyllav1.ScyllaCluster {
+				sc := newBasicScyllaCluster()
+				sc.Spec.ExternalSeeds = []string{"10.0.1.1", "10.0.1.2", "10.0.1.3"}
+				return sc
+			}(),
+			existingStatefulSet: nil,
+			expectedStatefulSet: func() *appsv1.StatefulSet {
+				sts := newBasicStatefulSet()
+
+				sts.Spec.Template.Spec.Containers[scyllaContainerIndex].Command = append(sts.Spec.Template.Spec.Containers[scyllaContainerIndex].Command, "--external-seeds=10.0.1.1,10.0.1.2,10.0.1.3")
+
 				return sts
 			}(),
 			expectedError: nil,
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -64,9 +64,10 @@ type ScyllaConfig struct {
 	scyllaRackDCPropertiesPath          string
 	scyllaRackDCPropertiesConfigMapPath string
 	cpuCount                            int
+	externalSeeds                       []string
 }
 
-func NewScyllaConfig(m *identity.Member, kubeClient kubernetes.Interface, scyllaClient scyllaversionedclient.Interface, cpuCount int) *ScyllaConfig {
+func NewScyllaConfig(m *identity.Member, kubeClient kubernetes.Interface, scyllaClient scyllaversionedclient.Interface, cpuCount int, externalSeeds []string) *ScyllaConfig {
 	return &ScyllaConfig{
 		member:                              m,
 		kubeClient:                          kubeClient,
@@ -74,6 +75,7 @@ func NewScyllaConfig(m *identity.Member, kubeClient kubernetes.Interface, scylla
 		scyllaRackDCPropertiesPath:          scyllaRackDCPropertiesPath,
 		scyllaRackDCPropertiesConfigMapPath: scyllaRackDCPropertiesConfigMapPath,
 		cpuCount:                            cpuCount,
+		externalSeeds:                       externalSeeds,
 	}
 }
 
@@ -204,10 +206,9 @@ func appendScyllaArguments(ctx context.Context, s *ScyllaConfig, scyllaArgs stri
 
 func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 	m := s.member
-	// Get seeds
-	seed, err := m.GetSeed(ctx, s.kubeClient.CoreV1())
+	seeds, err := m.GetSeeds(ctx, s.kubeClient.CoreV1(), s.externalSeeds)
 	if err != nil {
-		return nil, errors.Wrap(err, "error getting seeds")
+		return nil, fmt.Errorf("can't get seeds: %w", err)
 	}
 
 	// Check if we need to run in developer mode
@@ -232,7 +233,7 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 		"listen-address":        &listenAddress,
 		"broadcast-address":     &m.StaticIP,
 		"broadcast-rpc-address": &m.StaticIP,
-		"seeds":                 pointer.StringPtr(seed),
+		"seeds":                 pointer.String(strings.Join(seeds, ",")),
 		"developer-mode":        &devMode,
 		"overprovisioned":       &overprovisioned,
 		"smp":                   pointer.StringPtr(strconv.Itoa(s.cpuCount)),
--- a/pkg/sidecar/identity/member.go
+++ b/pkg/sidecar/identity/member.go
@@ -48,7 +48,7 @@ func NewMemberFromObjects(service *corev1.Service, pod *corev1.Pod) *Member {
 	}
 }
 
-func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (string, error) {
+func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, externalSeeds []string) ([]string, error) {
 	clusterLabels := naming.ScyllaLabels()
 	clusterLabels[naming.ClusterNameLabel] = m.Cluster
 
@@ -56,11 +56,11 @@ func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (st
 		LabelSelector: labels.SelectorFromSet(clusterLabels).String(),
 	})
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
 	if len(podList.Items) == 0 {
-		return "", fmt.Errorf("internal error: can't find any pod for this cluster, including itself")
+		return nil, fmt.Errorf("internal error: can't find any pod for this cluster, including itself")
 	}
 
 	var otherPods []*corev1.Pod
@@ -73,7 +73,11 @@ func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (st
 
 	if len(otherPods) == 0 {
 		// We are the only one, assuming first bootstrap.
-		return m.StaticIP, nil
+		if len(externalSeeds) > 0 {
+			return externalSeeds, nil
+		}
+
+		return []string{m.StaticIP}, nil
 	}
 
 	sort.Slice(otherPods, func(i, j int) bool {
@@ -88,8 +92,12 @@ func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (st
 
 	svc, err := coreClient.Services(m.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
-	return svc.Spec.ClusterIP, nil
+	res := make([]string, 0, len(externalSeeds)+1)
+	res = append(res, externalSeeds...)
+	res = append(res, svc.Spec.ClusterIP)
+
+	return res, nil
 }
--- a/test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
@@ -1,0 +1,97 @@
+// Copyright (c) 2023 ScyllaDB
+
+package scyllacluster
+
+import (
+	"context"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+var _ = g.Describe("MultiDC cluster", func() {
+	defer g.GinkgoRecover()
+
+	f1 := framework.NewFramework("scyllacluster")
+	f2 := framework.NewFramework("scyllacluster")
+
+	g.It("should form when external seeds are provided to ScyllaClusters", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
+
+		sc1 := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc1.Name = "basic-cluster"
+		sc1.Spec.Datacenter.Name = "us-east-1"
+		sc1.Spec.Datacenter.Racks[0].Name = "us-east-1a"
+		sc1.Spec.Datacenter.Racks[0].Members = 3
+
+		framework.By("Creating first ScyllaCluster")
+		sc1, err := f1.ScyllaClient().ScyllaV1().ScyllaClusters(f1.Namespace()).Create(ctx, sc1, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the first ScyllaCluster to rollout (RV=%s)", sc1.ResourceVersion)
+		waitCtx1, waitCtx1Cancel := utils.ContextForRollout(ctx, sc1)
+		defer waitCtx1Cancel()
+		sc1, err = utils.WaitForScyllaClusterState(waitCtx1, f1.ScyllaClient().ScyllaV1(), sc1.Namespace, sc1.Name, utils.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f1.KubeClient(), sc1)
+		hosts1 := getScyllaHostsAndWaitForFullQuorum(ctx, f1.KubeClient().CoreV1(), sc1)
+		di1 := insertAndVerifyCQLData(ctx, hosts1)
+		defer di1.Close()
+
+		sc2 := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc2.Name = "basic-cluster"
+		sc2.Spec.Datacenter.Name = "us-east-2"
+		sc2.Spec.Datacenter.Racks[0].Name = "us-east-2a"
+		sc2.Spec.Datacenter.Racks[0].Members = 3
+		sc2.Spec.ExternalSeeds = []string{
+			naming.CrossNamespaceServiceNameForCluster(sc1),
+		}
+
+		framework.By("Creating second ScyllaCluster")
+		sc2, err = f2.ScyllaClient().ScyllaV1().ScyllaClusters(f2.Namespace()).Create(ctx, sc2, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the second ScyllaCluster to rollout (RV=%s)", sc2.ResourceVersion)
+		waitCtx2, waitCtx2Cancel := utils.ContextForRollout(ctx, sc2)
+		defer waitCtx2Cancel()
+		sc2, err = utils.WaitForScyllaClusterState(waitCtx2, f2.ScyllaClient().ScyllaV1(), sc2.Namespace, sc2.Name, utils.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f2.KubeClient(), sc2)
+
+		framework.By("Verifying a multi datacenter cluster was formed with the first ScyllaCluster")
+		dcClientMap := make(map[string]corev1client.CoreV1Interface, 2)
+		dcClientMap[sc1.Spec.Datacenter.Name] = f1.KubeClient().CoreV1()
+		dcClientMap[sc2.Spec.Datacenter.Name] = f2.KubeClient().CoreV1()
+		hostsByDC := getScyllaHostsByDCAndWaitForFullQuorum(ctx, dcClientMap, []*scyllav1.ScyllaCluster{sc1, sc2})
+		o.Expect(hostsByDC[sc1.Spec.Datacenter.Name]).To(o.ConsistOf(hosts1))
+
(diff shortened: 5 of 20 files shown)


tnozicka:
(I'd like to avoid having the same comments twice but please ping me here when the proposal lands and this incorporates the changes to align with it.)

tnozicka (commented):
(I'd like to avoid having the same comments twice but please ping me here when the proposal lands and this incorporates the changes to align with it.)

rzetelskik:
/hold
Waiting for prerequisites to merge

rzetelskik:
@tnozicka I aligned the PR according to the enhancement proposal that merged yesterday, it's ready for review.

test/e2e/framework/framework.go
```diff
@@ -174,7 +186,62 @@ func (f *Framework) setupNamespace(ctx context.Context) {
...
+	By("Creating a new secondary namespace")
+	o.Expect(f.namespace).NotTo(o.BeNil())
+
+	generateChildName := func() string {
+		return names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", f.namespace.Name))
+	}
+	withOwnerReferences := WithOwnerReferences([]metav1.OwnerReference{
+		{
+			APIVersion: "v1",
+			Kind:       "namespace",
```
zimnx, line 203: `Namespace` is the correct Kind

test/e2e/framework/framework.go
```diff
@@ -186,7 +253,9 @@ func (f *Framework) setupNamespace(ctx context.Context) {
 	}, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	// Grant it edit permission in this namespace.
+	f.userSA = userSA
+
+	// Grant userSA edit permission in parent namespace.
```
zimnx, line 258: nit: lets call these relations consistently, here you use "parent" in other "primary".

test/e2e/framework/framework.go
```diff
@@ -296,19 +361,22 @@ func (f *Framework) afterEach() {
 			return
 		}
 
-		f.deleteNamespace(ctx, f.namespace)
+		f.deletePrimaryNamespace(ctx)
```
zimnx, line 364: wouldn't it be cleaner to delete every secondary namespace first, and then primary? Here you're letting GC to clean up, which may end up being cleared after test finishes. If something is off, we would succeed the test, but secondary ns would be dangling.

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
```diff
@@ -0,0 +1,80 @@
...
+		waitCtx2, waitCtx2Cancel := utils.ContextForRollout(ctx, sc2)
+		defer waitCtx2Cancel()
+		sc2, err = utils.WaitForScyllaClusterState(waitCtx2, f.ScyllaClient().ScyllaV1(), sc2.Namespace, sc2.Name, utils.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f.KubeClient(), sc2)
+
+		framework.By("Verifying a multi datacenter cluster was formed with the first ScyllaCluster")
+		hostsByDC := getScyllaHostsByDCAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc1, sc2)
+		o.Expect(hostsByDC[sc1.Spec.Datacenter.Name]).To(o.ConsistOf(hosts1))
```
zimnx, line 72: missing check for second dc hosts

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
```diff
@@ -0,0 +1,80 @@
...
+
+		di2 := insertAndVerifyCQLDataByDC(ctx, hostsByDC)
+		defer di2.Close()
+
+		framework.By("Verifying data of datacenter %q", sc1.Spec.Datacenter.Name)
+		verifyCQLData(ctx, di1)
```
zimnx, line 78: validate whether cluster status pulled from scyllaclient consist of all expected hosts and their datacenter information matches names from both SCs

test/e2e/set/scyllacluster/verify.go
```diff
@@ -259,11 +260,19 @@ func verifyScyllaCluster(ctx context.Context, kubeClient kubernetes.Interface, s
...
-	framework.By("Waiting for the ScyllaCluster to reach consistency ALL")
-	hosts, err := utils.GetScyllaHostsAndWaitForFullQuorum(ctx, client, sc)
+func getScyllaHostsAndWaitForFullQuorum(ctx context.Context, client corev1client.CoreV1Interface, scs ...*scyllav1.ScyllaCluster) []string {
+	hosts := getScyllaHostsByDCAndWaitForFullQuorum(ctx, client, scs...)
+	return helpers.Flatten(helpers.GetMapValues(hosts))
+}
+
+func getScyllaHostsByDCAndWaitForFullQuorum(ctx context.Context, client corev1client.CoreV1Interface, scs ...*scyllav1.ScyllaCluster) map[string][]string {
+	framework.By("Waiting for the ScyllaCluster(s) to reach consistency ALL")
+	hosts, err := utils.GetScyllaHostsByDCAndWaitForFullQuorum(ctx, client, scs...)
```
zimnx, line 270:
```suggestion
	dcHosts, err := utils.GetScyllaHostsByDCAndWaitForFullQuorum(ctx, client, scs...)
```

test/e2e/utils/datainserter.go
```diff
@@ -32,6 +33,11 @@ type TestData struct {
 }
 
 func NewDataInserter(hosts []string) (*DataInserter, error) {
+	// Instead of specifying hosts for the provided datacenter, use 'replication_factor' as a single key to specify a default RF.
+	return NewByDCDataInserter(map[string][]string{"replication_factor": hosts})
+}
+
+func NewByDCDataInserter(hosts map[string][]string) (*DataInserter, error) {
```
zimnx, line 40:
```suggestion
func NewMultiDCDataInserter(dcHosts map[string][]string) (*DataInserter, error) {
```

test/e2e/utils/datainserter.go
```diff
@@ -43,14 +49,19 @@ func NewDataInserter(hosts []string) (*DataInserter, error) {
 		data = append(data, &TestData{Id: i, Data: utilrand.String(32)})
 	}
 
+	replicationFactor := make(map[string]int, 0)
+	for k, v := range hosts {
```
zimnx, line 53:
```suggestion
	for dc, hosts := range hosts {
```

test/e2e/utils/datainserter.go
```diff
@@ -43,14 +49,19 @@ func NewDataInserter(hosts []string) (*DataInserter, error) {
 		data = append(data, &TestData{Id: i, Data: utilrand.String(32)})
 	}
 
+	replicationFactor := make(map[string]int, 0)
```
zimnx, line 52:
```suggestion
	replicationFactor := make(map[string]int, len(dcHosts))
```

test/e2e/utils/datainserter.go
```diff
@@ -83,11 +94,17 @@ func (di *DataInserter) SetClientEndpoints(hosts []string) error {
 }
 
 func (di *DataInserter) Insert() error {
-	framework.Infof("Creating keyspace %q with RF %d", di.keyspace, di.replicationFactor)
+	ss := make([]string, 0)
```
zimnx, line 97:
```suggestion
	ss := make([]string, len(di.replicationFactor))
```

test/e2e/utils/datainserter.go
```diff
@@ -83,11 +94,17 @@ func (di *DataInserter) SetClientEndpoints(hosts []string) error {
...
+	for k, v := range di.replicationFactor {
```
zimnx, line 98:
```suggestion
	for dc, hosts := range di.replicationFactor {
```

zimnx (commented):
first pass

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
rzetelskik: it doesn't really make sense to test it though, as this is the structure that would provide them? Unless you mean a check for the number of hosts - then it's in the called function

test/e2e/framework/framework.go
rzetelskik: Are you suggesting removing the owner references altogether or just adding the explicit deletion on top of that?

pkg/sidecar/config/config_test.go
```diff
@@ -389,3 +389,52 @@ client_encryption_options:
...
+			name:           "externalSeeds and localSeed are chosen if there are externalSeeds and localSeed is different then memberStaticIP",
+			memberStaticIP: "10.0.0.1",
+			localSeed:      "10.0.0.2",
+			externalSeeds:  []string{"10.0.1.1", "10.0.1.2"},
+			expected:       []string{"10.0.1.1", "10.0.1.2", "10.0.0.2"},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/framework/framework.go
rzetelskik: done

test/e2e/framework/framework.go
rzetelskik: done

test/e2e/set/scyllacluster/verify.go
rzetelskik: done

test/e2e/utils/datainserter.go
rzetelskik: done

test/e2e/utils/datainserter.go
rzetelskik: done

test/e2e/utils/datainserter.go
rzetelskik: do you mean `ss := make([]string, 0, len(di.replicationFactor))` for capacity?

test/e2e/utils/datainserter.go
rzetelskik: done, although replaced hosts with rf

test/e2e/utils/datainserter.go
rzetelskik: done

test/e2e/framework/framework.go
rzetelskik: Added explicit deletion

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
rzetelskik: > validate whether cluster status pulled from scyllaclient consist of all expected hosts

this is already done in `getScyllaHostsByDCAndWaitForFullQuorum`

> their datacenter information matches names from both SCs

added

rzetelskik:
@zimnx thanks, I've addressed all your comments

pkg/sidecar/config/config.go
```diff
@@ -204,12 +206,13 @@ func appendScyllaArguments(ctx context.Context, s *ScyllaConfig, scyllaArgs stri
 
 func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 	m := s.member
-	// Get seeds
-	seed, err := m.GetSeed(ctx, s.kubeClient.CoreV1())
+	localSeed, err := m.GetSeed(ctx, s.kubeClient.CoreV1())
```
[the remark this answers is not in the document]
zimnx: There's just one parameter where they end up so there's no distinction. I bet in a month no one will remember such details, and logic of choosing seeds isn't that complex. Code should be written to be easily maintainable in future.
In my view, it's error prone to have two separate functions that are spread across two packages, where one of them is public, second one isn't.

test/e2e/utils/datainserter.go
zimnx: yes, in general I meant preallocation to avoid unnecessary copies.

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
zimnx: so what's the point of this check if you're checking just one out of two DCs?

test/e2e/framework/framework.go
zimnx: owner references are good, we need those. I suggest explicit remove, to clean everything we create before we call it success. Relying on GC is good when test it interrupted.

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
tnozicka: I'd prefer avoid any uncertainty. We should assert `len(hostsByDC)==2` and the content of data for each of those 2 keys.

test/e2e/set/scyllacluster/verify.go
```diff
@@ -279,16 +288,26 @@ func verifyCQLData(ctx context.Context, di *utils.DataInserter) {
...
+	return insertAndVerifyCQLDataFn(utils.NewDataInserter)(ctx, hosts)
+}
 
-	err = di.Insert()
-	o.Expect(err).NotTo(o.HaveOccurred())
+func insertAndVerifyCQLDataByDC(ctx context.Context, hosts map[string][]string) *utils.DataInserter {
+	return insertAndVerifyCQLDataFn(utils.NewMultiDCDataInserter)(ctx, hosts)
+}
+
+func insertAndVerifyCQLDataFn[T any](dataInserterFn func(hosts T) (*utils.DataInserter, error)) func(context.Context, T) *utils.DataInserter {
```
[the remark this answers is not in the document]
tnozicka: yeah, given the arguments are just a subset of each other, there should be an easy way to share a single static implementation.

test/e2e/utils/datainserter.go
```diff
@@ -32,6 +33,11 @@ type TestData struct {
 }
 
 func NewDataInserter(hosts []string) (*DataInserter, error) {
+	// Instead of specifying hosts for the provided datacenter, use 'replication_factor' as a single key to specify a default RF.
+	return NewMultiDCDataInserter(map[string][]string{"replication_factor": hosts})
```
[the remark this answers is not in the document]
rzetelskik: That's why I've put the comment there, it's somewhat hacky but avoids modifying the existing tests

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
rzetelskik: > so what's the point of this check if you're checking just one out of two DCs?

It's basically comparing the hosts of the first DC returned by scylla client before and after forming the multi DC cluster.
There is no other source of hosts of the second DC other than that returned structure, so there is nothing to compare it to.

> We should assert len(hostsByDC)==2

The number of hosts is already verified in `getScyllaHostsByDCAndWaitForFullQuorum` for every SC.

> the content of data for each of those 2 keys.

ditto, any check I do here is essentially going to compare the hosts coming from the same source of truth.

pkg/sidecar/config/config.go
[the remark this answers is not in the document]
rzetelskik: Ok, I don't feel strongly about any of the two options. I felt that from the object-oriented point of view it's a bit weird for the member to implement a policy of choosing seeds to provide to itself, but I agree that it's generally more coherent and less error prone to do it that way.

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -106,6 +106,10 @@ type ScyllaClusterSpec struct {
 	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
 	// +optional
 	ExposeOptions *ExposeOptions `json:"exposeOptions,omitempty"`
+
+	// externalSeeds specifies the external seeds to propagate to ScyllaDB binary on startup as "seeds" parameter of seed-provider.
+	// See https://opensource.docs.scylladb.com/stable/kb/seed-nodes.html.
```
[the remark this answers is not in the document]
rzetelskik: I suppose if we don't provide the reference to the actual page then there's no point in saying something as obvious as consult the docs - after all we'd have to say that for most of the API fields. I removed the reference altogether.

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -1085,6 +1085,28 @@ func TestStatefulSetForRack(t *testing.T) {
...
+				sc := newBasicScyllaCluster()
+				sc.Spec.ExternalSeeds = []string{"10.0.1.1", "10.0.1.2", "10.0.1.3"}
+				return sc
+			}(),
+			existingStatefulSet: nil,
+			expectedStatefulSet: func() *appsv1.StatefulSet {
+				sts := newBasicStatefulSet()
+
+				for i := range sts.Spec.Template.Spec.Containers {
+					if sts.Spec.Template.Spec.Containers[i].Name == naming.ScyllaContainerName {
```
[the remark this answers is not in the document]
rzetelskik: What would you suggest? I'd rather avoid explicitly setting `sts.Spec.Template.Spec.Containers` field, since that adds a lot of bloat. Would a `panic` on not found scylla container be fine?

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
```diff
@@ -0,0 +1,93 @@
...
+
+	g.It("should form when external seeds are provided to ScyllaClusters", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
+
+		sc1 := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc1.Name = "basic-cluster"
+		sc1.Spec.Datacenter.Name = "us-east-1"
+		sc1.Spec.Datacenter.Racks[0].Name = "us-east-1a"
+		sc1.Spec.Datacenter.Racks[0].Members = 2
```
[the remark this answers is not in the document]
rzetelskik: I'm not sure what exactly you're referring to other than just maintaining HA, I believe that'd only break tests related to topology changes. This only adds the nodes so it requires all nodes UN anyway, but I think it won't hurt to have to proper SCs anyway. Changed.

test/e2e/set/scyllacluster/verify.go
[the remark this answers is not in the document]
rzetelskik: agreed and changed, thanks

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
```diff
@@ -0,0 +1,93 @@
...
+	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("MultiDC cluster", func() {
+	defer g.GinkgoRecover()
+
+	f := framework.NewFramework("scyllacluster")
```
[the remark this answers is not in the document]
rzetelskik: I did it this way for ease of debugging, but if you think the separation is more important here then yes, it's definitely possible and easier implementation-wise.

test/e2e/utils/datainserter.go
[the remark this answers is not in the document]
rzetelskik: Alternatively, the `NewMultiDCDataInserter` could take a flat slice of hosts as a first argument and `map[string]int` with replication factor as a second. A bit redundant and leaking the implementation, but I don't see any other options atm. Wdyt?

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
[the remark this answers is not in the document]
rzetelskik: It will require some tweaks to verifying functions though, like passing pairs of resources and clients from different namespaces

test/e2e/set/scyllacluster/verify.go
```diff
@@ -259,13 +260,21 @@ func verifyScyllaCluster(ctx context.Context, kubeClient kubernetes.Interface, s
...
+
+func getScyllaHostsByDCAndWaitForFullQuorum(ctx context.Context, client corev1client.CoreV1Interface, scs ...*scyllav1.ScyllaCluster) map[string][]string {
+	framework.By("Waiting for the ScyllaCluster(s) to reach consistency ALL")
+	dcHosts, err := utils.GetScyllaHostsByDCAndWaitForFullQuorum(ctx, client, scs...)
 	o.Expect(err).NotTo(o.HaveOccurred())
-	o.Expect(hosts).To(o.HaveLen(int(utils.GetMemberCount(sc))))
 
-	return hosts
+	for _, sc := range scs {
+		o.Expect(dcHosts[sc.Spec.Datacenter.Name]).To(o.HaveLen(int(utils.GetMemberCount(sc))))
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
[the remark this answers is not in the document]
tnozicka: If you mean the helpers that already take map of host by DCs / SCs, it wouldn't seem unreasonable to put dedicated clients there as well. In the end, you'd need 2 clients for real 2 DCs in different kube clusters - this also makes sure we are better setup for a richer CI infra in the future.

test/e2e/utils/datainserter.go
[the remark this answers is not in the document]
tnozicka: Just don't call it `replication_factor` when it means a DC and the value is "hosts". (you later use the value to compute replication factor at some place but that shouldn't be the name of the key. In the other func, the key in that map means a datacenter, so let's put in a dummy key for a DC instead, like `_datacenter` or something and let the comment say, that the key doesn't matter as it's ignored by the func.)

pkg/controller/scyllacluster/resource_test.go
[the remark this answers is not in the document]
rzetelskik: done, hardcoded

test/e2e/utils/datainserter.go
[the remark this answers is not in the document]
rzetelskik: The keys are used for specifying the replication_factor though. You can either specify RF per DC `"DC1": 3, "DC2": 6` (same for single DC `"DC": 3`) or the default for all DCs `replication_factor: 3`. See https://opensource.docs.scylladb.com/stable/cql/ddl.html#networktopologystrategy.

So in case a single slice of hosts is provided, I use `replication_factor` as a key to set a default. The logic can be pushed down, but it'll still have to discard the dummy datacenter name and replace it with the default.

test/e2e/set/scyllacluster/scyllacluster_external_seeds.go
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/utils/helpers.go
```diff
@@ -506,62 +506,96 @@ func GetMemberServiceSelector(scyllaClusterName string) labels.Selector {
...
+			continue
 		}
 
-		if !allSeeAllAsUN {
-			framework.Infof("ScyllaDB nodes have not reached status consistency yet. Statuses:\n%s", strings.Join(infoMessages, ","))
+		scyllaClient, hosts, err := GetScyllaClient(ctx, client, sc)
+		if err != nil {
+			return nil, fmt.Errorf("can't get scylla client: %w", err)
 		}
+		defer scyllaClient.Close()
```
tnozicka, line 546: defer in for loop :(

please, wrap it in a function

test/e2e/utils/datainserter.go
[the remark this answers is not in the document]
tnozicka: ok, now I see where the bad API is coming from - it's mixing two schemas into single field, making one key to be special and likely breaking if you actually name on of the DCs `replication_factor`.

given the 2 schemas are merged into one, the variables didn't make sense, so say  `for dc, hosts := range dcHosts {` would need to be `for dcOrRFMagicValue, hosts := range dcHosts {`, so I'll leave it be. thanks

test/e2e/utils/helpers.go
rzetelskik: done

test/e2e/utils/helpers.go
```diff
@@ -506,62 +506,96 @@ func GetMemberServiceSelector(scyllaClusterName string) labels.Selector {
...
-		allSeeAllAsUN := true
-		infoMessages := make([]string, 0, len(hosts))
-		var errs []error
-		for _, h := range sortedHosts {
-			s, err := scyllaClient.Status(ctx, h)
-			if err != nil {
-				return true, fmt.Errorf("can't get scylla status on node %q: %w", h, err)
-			}
+func GetScyllaHostsByDCAndWaitForFullQuorum(ctx context.Context, dcClientMap map[string]corev1client.CoreV1Interface, scs []*scyllav1.ScyllaCluster) (map[string][]string, error) {
+	allHosts := make(map[string][]string, 0)
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/sidecar/identity/member.go
```diff
@@ -88,8 +92,12 @@ func (m *Member) GetSeed(ctx context.Context, coreClient v1.CoreV1Interface) (st
 
 	svc, err := coreClient.Services(m.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
-	return svc.Spec.ClusterIP, nil
+	res := make([]string, 0, len(externalSeeds)+1)
+	copy(res, externalSeeds)
```
tnozicka, line 99: return values need to be processed or explicitly ignored.
(to me, using append looks more idiomatic and consistent but I don't feel strong)

tnozicka:
/approve
(and lgtm)

/assign @zimnx

zimnx (approved):
/lgtm

PR #1327, opened 2023-08-17
Fix variable defaulting and undefined variables in docs Makefile

tnozicka:

Using unconditional assignment was overwriting env vars, which for poetry resulted in taking  the wrong binary that didn't exist.
`POETRY := $(HOME)/.poetry/bin/poetry`
```
 /tmp/tmp.tuESVb54FZ/.poetry/bin/poetry run sphinx-build -b dirhtml -d _build/doctrees  --keep-going source -W _build/dirhtml
make: /tmp/tmp.tuESVb54FZ/.poetry/bin/poetry: No such file or directory
```
Also there were some undefined variables that needed fixing
```
 + make --warn-undefined-variables dirhtml SPHINXOPTS=--keep-going
Makefile:10: warning: undefined variable 'OS'
Makefile:17: warning: undefined variable 'PAPEROPT_'
Makefile:19: warning: undefined variable 'PAPEROPT_' 
```
~And the spaces after assignments matter in make (taken as literal strings including the leading space).~

~Hopefully, defaulting the paper to a4 works as the default.~

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,22 +1,32 @@
 # You can set these variables from the command line.
-POETRY        := $(HOME)/.poetry/bin/poetry
-SPHINXBUILD   := $(POETRY) run sphinx-build
-SPHINXOPTS    :=
-PAPER         :=
-BUILDDIR      := _build
-SOURCEDIR     := source
-
-# For Windows users
 ifeq ($(OS),Windows_NT)
-    POETRY = $(APPDATA)\Python\Scripts\poetry
+# For Windows users
+POETRY ?= $(APPDATA)\Python\Scripts\poetry
+else
+POETRY ?= $(HOME)/.poetry/bin/poetry
 endif
 
+SPHINXBUILD ?= $(POETRY) run sphinx-build
+SPHINXOPTS ?=
+PAPER ?=
+BUILDDIR ?= _build
+SOURCEDIR ?= source
+OS ?=
+
+
 # Internal variables.
-PAPEROPT_a4     := -D latex_paper_size=a4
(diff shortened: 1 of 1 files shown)


docs/Makefile
```diff
@@ -1,22 +1,23 @@
...
 ifeq ($(OS),Windows_NT)
-    POETRY = $(APPDATA)\Python\Scripts\poetry
+    POETRY =$(APPDATA)\Python\Scripts\poetry
 endif
 
 # Internal variables.
-PAPEROPT_a4     := -D latex_paper_size=a4
-PAPEROPT_letter := -D latex_paper_size=letter
-ALLSPHINXOPTS   := -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) $(SOURCEDIR) -W
+PAPEROPT_a4 ?=-D latex_paper_size=a4
```
rzetelskik, line 16: This section is described as `Internal variables`. I don't think overwriting them should be allowed?

tnozicka: you can still overwrite then even with `:=` if you do `make PAPEROPT_a4=foo`. internals should be lowercase and there are some options to enforce it but I don't feel it's worth the effort. I can put these back though.

tnozicka:
> make manual says Leading whitespace characters are discarded from your input before substitution of variable references and function calls.

I though there was a case where it was doing double spaces but I can't recall the case and all other I've tried work fine, so I take that back, thx

rzetelskik:
/lgtm
/approve

thanks!

rzetelskik (approved):

tnozicka:
docs only, everything except e2e is green and these are not relevant

PR #1331, opened 2023-08-17
Remove GH presubmits in favour of Prow

tnozicka:
GH presumbits are ignored by Prow (it can't retrigger them on base branch changing so they'd be stale).


tnozicka:
/hold
(I'd like to check the context manually the first time when it's tagged.)

tnozicka:
green and ready to start a new era :rocket: 

/hold cancel

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,129 @@
...
+kubectl -n scylla-manager rollout status deployment/scylla-manager-controller
+
+ingress_address="$( kubectl -n haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' )"
+
+kubectl create namespace e2e --dry-run=client -o yaml | kubectl_create -f -
+kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o yaml | kubectl_create -f -
+kubectl create -n e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o yaml | kubectl_create -f -
+
+kubectl -n e2e run --restart=Never --image="${SO_IMAGE}" --labels='app=e2e' --command=true e2e -- bash -euExo pipefail -O inherit_errexit -c "mkdir /tmp/artifacts && scylla-operator-tests run scylla-operator/conformance/parallel --parallelism=60 --loglevel=2 --color=false --artifacts-dir=/tmp/artifacts --feature-gates=\"${SCYLLA_OPERATOR_FEATURE_GATES}\" --override-ingress-address=\"${ingress_address}\" && touch /tmp/done && until [[ -f '/tmp/exit' ]]; do sleep 1; done"
```
zimnx, line 109: Compared to existing CI, this doesn't run serial suite at all, and there're no runs without feature flags.

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,129 @@
...
+    nodeSelector:
+      scylla.scylladb.com/node-type: scylla
+    tolerations:
+    - effect: NoSchedule
+      key: role
+      operator: Equal
+      value: scylla-clusters
+EOF
+
+kubectl_create -n local-csi-driver -f=https://raw.githubusercontent.com/scylladb/k8s-local-volume-provisioner/dcfceb52d9122192e50236463060cb5151d90dc4/deploy/kubernetes/local-csi-driver/{00_namespace,10_csidriver,10_driver_serviceaccount,10_provisioner_clusterrole,20_provisioner_clusterrolebinding,50_daemonset}.yaml
```
rzetelskik, line 85: Can't we just have a copy of manifests in `.ci` directory, as we do now with `actions`?

rzetelskik (commented):
What's up with the commit name "Move e2e business logic into its repo"?

hack/.ci/run-e2e-gke.sh
tnozicka: I'll add the serial suite, cost wise PRs will move to all features enabled and we can add some of the other combinations as periodics

tnozicka:
> Move e2e business logic into its repo

It used to live in the presubmit definitions, now it calls the script from the target repo. I'll reword it as it relates to the interim that didn't make it here

hack/.ci/run-e2e-gke.sh
tnozicka: ok, I'll add it, this was a snapshot from our own repo for the interim while it should eventually make it to the NodeConfig

tnozicka: done

hack/.ci/run-e2e-gke.sh
tnozicka: done

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,134 @@
...
+EOF
+
+kubectl apply --server-side -f ./pkg/api/scylla/v1alpha1/scylla.scylladb.com_nodeconfigs.yaml
+kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com
+
+if [[ "${SO_DISABLE_NODECONFIG}" == "true" ]]; then
+  echo "Skipping NodeConfig creation"
+else
+  kubectl_create -f - << EOF
+apiVersion: scylla.scylladb.com/v1alpha1
```
rzetelskik, line 70: nit: NodeConfig could also be moved to a manifest file

tnozicka: done

hack/.ci/run-e2e-gke.sh
rzetelskik: Until we have periodics or postsubmits you've mentioned, isn't it prone to introducing regressions to only run suites with features enabled, compared to how it's set up now?

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,110 @@
...
+    - patch
+EOF
+
+kubectl apply --server-side -f ./pkg/api/scylla/v1alpha1/scylla.scylladb.com_nodeconfigs.yaml
+kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com
+
+if [[ "${SO_DISABLE_NODECONFIG}" == "true" ]]; then
+  echo "Skipping NodeConfig creation"
+else
+  kubectl_create -f - ./hack/.ci/manifests/cluster/nodeconfig.yaml
```
rzetelskik, line 69:
```suggestion
  kubectl_create -f ./hack/.ci/manifests/cluster/nodeconfig.yaml
```

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,110 @@
...
+kubectl -n scylla-manager patch --field-manager="${field_manager}" deployment/scylla-manager-controller --type=json -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--loglevel=4"}]'
+kubectl -n scylla-manager rollout status deployment/scylla-manager-controller
+
+ingress_address="$( kubectl -n haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' )"
+
+kubectl create namespace e2e --dry-run=client -o yaml | kubectl_create -f -
+kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o yaml | kubectl_create -f -
+kubectl create -n e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o yaml | kubectl_create -f -
+
+kubectl -n e2e run --restart=Never --image="${SO_IMAGE}" --labels='app=e2e' --command=true e2e -- bash -euExo pipefail -O inherit_errexit -c "mkdir /tmp/artifacts && scylla-operator-tests run '${SO_SUITE}' --parallelism=60 --loglevel=2 --color=false --artifacts-dir=/tmp/artifacts --feature-gates='${SCYLLA_OPERATOR_FEATURE_GATES}' --override-ingress-address='${ingress_address}' && touch /tmp/done && until [[ -f '/tmp/exit' ]]; do sleep 1; done"
```
zimnx, line 90: `parallelism` shouldn't be hardcoded here, because it overrides default suite settings. For `serial` suite, we always want it to be `1`.

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,110 @@
...
+kubectl -n e2e wait --for=condition=Ready pod/e2e
+
+# Setup artifacts transfer when finished and unblock the e2e pod when done.
+(
+  kubectl -n e2e exec pod/e2e -- bash -euEo pipefail -O inherit_errexit -c "until [[ -f /tmp/done ]]; do sleep 1; done; ls -l /tmp/artifacts"
```
zimnx, line 95: `/tmp/artifacts` could be extracted to variable, as it's used in multiple places

hack/.ci/run-e2e-gke.sh
```diff
@@ -0,0 +1,110 @@
...
+fi
+
+if [ -z ${SO_IMAGE+x} ]; then
+  echo "SO_IMAGE can't be empty" > /dev/stderr
+  exit 2
+fi
+
+SO_DISABLE_NODECONFIG=${SO_DISABLE_NODECONFIG:-false}
+
+field_manager=hack
```
zimnx, line 21:
```suggestion
field_manager=run-e2e-script
```

hack/.ci/run-e2e-gke.sh
tnozicka: it is. the set is fairly limited.

hack/.ci/run-e2e-gke.sh
tnozicka: yeah, it shouldn't but it still run in serial just wasted 59 dummy workers

hack/.ci/run-e2e-gke.sh
tnozicka: I don't think that's worth it, especially with the nested shells and reasoning about where it gets evaluated.

tnozicka: but the 60 is a leftover, thx

tnozicka: done

hack/.ci/run-e2e-gke.sh
tnozicka: done

zimnx (approved):

PR #1332, opened 2023-08-17
Remove haproxy tunning that breaks GKE

tnozicka:
Most of the haproxy option were just initial attempts. The ones removed here were breaking deployment on GKE, so for now we'll go back to the defaults.


tnozicka:
parallel tests flaked on https://github.com/scylladb/scylla-operator/issues/1336 (flakes on master as well), not related

zimnx (approved):
/lgtm

PR #1333, opened 2023-08-17
Ignore not found errors when gathering artifacts

tnozicka:
Some API objects may naturally be gone between the listing and collection time. Like events with TTL or other GCed objects. There is nothing to do with those anyways and we shouldn't fail the run.


rzetelskik (commented):
/lgtm
/approve

zimnx (approved):
/lgtm

Issue #821, opened 2021-10-22
Support sha references to scylla images


tnozicka:
**What should the feature do:**
Currently scylla images can be specified only with a tag that has to be a semver. This dosn't work for named references like `latest`, rolling update references like `4.4` (sort of) and for sha references. As images can be overwritten sha reference is the only reliable way to always reference the same image.

As we need to know the information in the operator we can't just download and run the image to find out but when buliding the image we could set an image label that the operator would retrieve. The label is stored in the image manifest that can be downloaded separately from a container registry at low cost and the result can be cached.

Needs cooperation from scylla.

(Look at libs like `github.com/containers/image/v5 `)

tnozicka:
@rzetelskik can you link the PoC you did?

I don't thing we've written a proposal but (for the record) the next step would be to get a label into scylla container images.

rzetelskik:
@tnozicka if you mean the PoC for fetching the manifests and extracting labels then this is the the code using [containers/image](https://github.com/containers/image) in a CLI tool mimicking skopeo: [rzetelskik/labels-poc](https://github.com/rzetelskik/labels-poc).

Should I go ahead and start adding label extraction to the operator?

> the next step would be to get a label into scylla container images

This isn't on us though, is it?

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2455.

PR #1346, opened 2023-08-22
Add full support for named and tagged image references

### LINKED ISSUE #1310: ScyllaCluster doesn't support image references with registry port
**Describe the bug**
Operator fails to transition cluster into Available state when registry contains host port.

Consider a scenario where you have a `scyllaCluster`, and instead of pulling images from the official image hub, you pull them through `someproxy:1234/proxy`:

```
kind: ScyllaCluster
(...)
  spec:
  (...)
  repository: someproxy:1234/proxy/scylladb/scylla
  tag: 5.1.8
```

Even though operator starts all scylladb nodes (pods are in running state) it does not transition to correct state due to following error
`E0720 23:49:19.448247       1 scyllacluster/status.go:86] "Can't get scylla version" err="parse scylla container version: Invalid image name:  someproxy:1234/proxy/scylladb/scylla:5.1.8" ScyllaCluster="scylladb/default"`

scyllaCluster status
```
(...)
- lastTransitionTime: "2023-07-21T11:09:37Z"
    message: 'StatefulSetControllerAvailable: Racks "db-0, db-1, db-2" are not in
      the desired version'
    observedGeneration: 1
    reason: StatefulSetControllerAvailable_RacksNotAtDesiredVersion
    status: "False"
    type: Available
(...)
```

**Additional context**
Upon investigating the operator's code, it appears that the function at https://github.com/scylladb/scylla-operator/blob/master/pkg/naming/names.go#L103 needs to be modified to fully support the Docker image naming convention.

As per the official Docker documentation (https://docs.docker.com/engine/reference/commandline/tag/), it is allowed to have a port in the image name:
`[HOST[:PORT_NUMBER]/]PATH`

I propose the following code modification:
```
func ImageToVersion(image string) (string, error) {
	parts := strings.Split(image, ":")
	if len(parts) < 2 {
		return "", errors.New(fmt.Sprintf("Invalid image name: %s", image))
	}
	version := parts[len(parts)-1]
	if len(version) == 0 {
		return "", errors.New(fmt.Sprintf("Image version is empty: %s", image))
	}

	return parts[len(parts)-1], nil
}
```
I trust you understand the concept.

rzetelskik:
**Description of your changes:** As described in https://github.com/scylladb/scylla-operator/issues/1310, the way we now parse images to extract Scylla version from tags does not comply with image references specification. This PR adds full support for named and tagged image references as a workaround until we move to labels (https://github.com/scylladb/scylla-operator/issues/821).

Resolves #1310


zimnx (commented):
two nits

pkg/naming/names_test.go
```diff
@@ -0,0 +1,55 @@
...
+		{
+			name:          "digested reference",
+			image:         "docker.io/scylladb/scylla@sha256:12a95529d94498d8f56fc4596f8ac570e961618fab7ab5cf3e3df781c2b8fd33",
+			expectedError: fmt.Errorf("invalid, non-tagged image reference: docker.io/scylladb/scylla@sha256:12a95529d94498d8f56fc4596f8ac570e961618fab7ab5cf3e3df781c2b8fd33"),
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			version, err := ImageToVersion(tc.image)
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/naming/names_test.go
```diff
@@ -0,0 +1,55 @@
...
+package naming
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+)
+
+func Test_ImageToVersion(t *testing.T) {
+	tcs := []struct {
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/naming/names.go
```diff
@@ -101,11 +102,16 @@ func IndexFromName(n string) (int32, error) {
...
-	parts := strings.Split(image, ":")
-	if len(parts) != 2 || len(parts[1]) == 0 {
-		return "", errors.New(fmt.Sprintf("Invalid image name: %s", image))
+	named, err := reference.ParseNormalizedNamed(image)
+	if err != nil {
+		return "", fmt.Errorf("can't parse image: %w", err)
+	}
+	tagged, ok := named.(reference.NamedTagged)
+	if !ok {
+		return "", fmt.Errorf("invalid, non-tagged image reference: %s", image)
```
tnozicka, line 111: nit: when a type doesn't implement an interface, it might be useful to enrich the message with the type that failed as well

pkg/naming/names_test.go
```diff
@@ -0,0 +1,61 @@
...
+			image:           "someproxy:1234/proxy/scylladb/scylla:5.1.15",
+			expectedVersion: "5.1.15",
+			expectedError:   nil,
+		},
+		{
+			name:          "digested reference",
+			image:         "docker.io/scylladb/scylla@sha256:12a95529d94498d8f56fc4596f8ac570e961618fab7ab5cf3e3df781c2b8fd33",
+			expectedError: fmt.Errorf("invalid, non-tagged image reference: docker.io/scylladb/scylla@sha256:12a95529d94498d8f56fc4596f8ac570e961618fab7ab5cf3e3df781c2b8fd33"),
+		},
+	}
```
tnozicka, line 43: can you add a tag@sha256 as well? does it succeed parsing?

rzetelskik: Good point, yes. Added.

pkg/naming/names.go

pkg/naming/names.go
```diff
@@ -101,11 +103,16 @@ func IndexFromName(n string) (int32, error) {
...
-	parts := strings.Split(image, ":")
-	if len(parts) != 2 || len(parts[1]) == 0 {
-		return "", errors.New(fmt.Sprintf("Invalid image name: %s", image))
+	named, err := reference.ParseNormalizedNamed(image)
+	if err != nil {
+		return "", fmt.Errorf("can't parse image: %w", err)
+	}
+	tagged, ok := named.(reference.NamedTagged)
+	if !ok {
+		return "", fmt.Errorf("invalid, non-tagged image reference of type %s: %s", reflect.TypeOf(named).String(), image)
```
tnozicka, line 112: %T

tnozicka (approved):
/approve
/lgtm

thanks!

PR #1352, opened 2023-08-25
Limit parallelism in e2e test suites in GH Action postsubmits

zimnx:
Default was increased to 42 which is way too high for slow GH Actions workers.


rzetelskik:
/lgtm
/approve

thanks!

PR #1359, opened 2023-08-29
Add new ways of exposing ScyllaCluster outside 

### LINKED ISSUE #1270: Expose ScyllaDB to external clients and other datacenters
Currently, scylla operator exposes every ScyllaDB nodes with a service of type ClusterIP and sets the ClusterIP to be published by ScyllaDB nodes. ClusterIPs are virtual IPs that can't be reached out from outside of the cluster. Given they are virtual it's even hard to hack to route the external traffic to the cluster manually. It doesn't even matter if you setup LBs to expose the nodes manually because ScyllaDB will publish the internal IP to the external clients that will pick those up and fail.

To be able to reach the nodes from another DC or an external client, we need scylla to publish different IPs. Depending on the use case, the IP adresses can come from the Pod IP (when https://github.com/scylladb/scylladb/issues/6403 gets implemented) or different types of services, like LoadBalancer. With all the other service options we don't support, some of them control whether say the LB provides private or public IP. Maybe we should add a serviceTemplate field or a metadata template (to avoid conflicts) so it's easily controllable.

To sum up the expose options:
- Pod IPs allow exposing  pods in internal IPs to other DCs
- Service.type==LoadBalancer can allocate both external and internal IPs (internal ones are still useful to avoid VPC peering)
- Possibly this could be ingress based URL

Ideally we should abstract the current logic not to use IPs, or at least dynamically fetch them when necessary (like with `replace_address_first_boot`), so it we can then implement a switch specifying which ones to use. Picking the source of them will likely be less work then the abstraction work.

If we decide to make a switch specifying which type to use, it would allow us to implement this and maintain backwards compatibility. Given we'll eventually need both PodIP and LB, the switch seems reasonable to start with.

This needs to start with a design doc.
```[tasklist]
### Tasks
- [ ] https://github.com/scylladb/scylla-operator/pull/1297
- [ ] https://github.com/scylladb/scylla-operator/issues/1277
- [ ] https://github.com/scylladb/scylla-operator/issues/1315
- [ ] https://github.com/scylladb/scylladb/issues/6403
- [ ] https://github.com/gocql/gocql/issues/831#issuecomment-272060425
- [ ] https://github.com/scylladb/scylla-operator/issues/1319
- [ ] https://github.com/scylladb/scylla-operator/issues/1380
```

zimnx:
ExposeOptions of ScyllaCluster allows for choosing which Service type
is created for each cluster node, and which address is broadcasted to
clients and nodes. This allows for exposing ScyllaCluster to clients
from external networks, and provides an important building block for
connecting two distinct ScyllaClusters.

Fixes #1270


zimnx:
unit test failure is not related to this PR

/override ci/prow/unit

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -5,25 +5,248 @@ package scyllacluster
...
+					},
+				},
+			},
+			validateService: func(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster) {
+				serviceList, err := f.KubeClient().CoreV1().Services(sc.Namespace).List(ctx, metav1.ListOptions{
+					LabelSelector: utils.GetMemberServiceSelector(sc.Name).String(),
+				})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				svc := serviceList.Items[0]
```
rzetelskik, line 186: you're listing the services in every instance of `validateService` - can't you just get the service in Describe and pass it to the function as a parameter?

rzetelskik (commented):
first pass

zimnx:
sync

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -99,6 +101,25 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
 
 	if spec.ExposeOptions != nil {
+		supportsExposingViaService, err := scyllafeatures.VersionSupports(spec.Version, scyllafeatures.ExposingScyllaClusterViaServiceOtherThanClusterIP)
+		if err != nil {
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("version"), fmt.Errorf("can't determine if version %q supports exposing via Service: %w", spec.Version, err)))
+		}
+
+		if !supportsExposingViaService {
+			if spec.ExposeOptions.NodeService != nil && spec.ExposeOptions.NodeService.Type != scyllav1.ClusterIPServiceType {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("exposeOptions", "nodeService", "type"), spec.ExposeOptions.NodeService.Type, fmt.Sprintf("can't expose ScyllaCluster via Service other than ClusterIP with given ScyllaDB version, upgrade to latest one first")))
```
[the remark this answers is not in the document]
zimnx: it's `>=5.2.0` for OS, and `>=2023.1.0` for Enterprise, we would need to distinguish. Using `latest` is common and encourages users to update to latest version that may contain fixes we aren't aware today.

pkg/sidecar/identity/member.go
```diff
@@ -20,32 +21,77 @@ type Member struct {
...
+func NewMember(service *corev1.Service, pod *corev1.Pod, nodesAddressType, clientAddressType scyllav1.BroadcastAddressType) (*Member, error) {
+	var externalIP string
+	if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
+		if len(service.Status.LoadBalancer.Ingress) == 0 {
+			return nil, fmt.Errorf("LoadBalancer Service %q is missing endpoints in status", naming.ObjRef(service))
+		}
+
+		firstIngress := service.Status.LoadBalancer.Ingress[0]
+		if len(firstIngress.Hostname) != 0 {
+			externalIP = firstIngress.Hostname
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -108,13 +108,118 @@ type ScyllaClusterSpec struct {
...
+}
+
+type NodeServiceType string
+
+const (
+	// HeadlessServiceType means nodes will be exposed via Headless Service.
+	HeadlessServiceType NodeServiceType = "Headless"
+
+	// ClusterIPServiceType means nodes will be exposed via ClusterIP Service.
+	ClusterIPServiceType NodeServiceType = "ClusterIP"
```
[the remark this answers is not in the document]
zimnx: done, also for BroadcastAddressTypes

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -5,25 +5,248 @@ package scyllacluster
...
 	f := framework.NewFramework("scyllacluster")
 
+	type entry struct {
+		exposeOptions            *scyllav1.ExposeOptions
+		memberServiceCreatedHook func(ctx context.Context, client corev1client.CoreV1Interface, svc *corev1.Service)
+		validateService          func(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster)
+		validateScyllaConfig     func(ctx context.Context, configClient *scyllaclient.ConfigClient, svc *corev1.Service, pod *corev1.Pod)
+	}
+
+	g.DescribeTable("should expose cluster", func(e *entry) {
```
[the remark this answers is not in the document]
zimnx: done

pkg/scyllafeatures/scyllafeatures.go
```diff
@@ -16,7 +16,8 @@ var (
 type ScyllaFeature string
 
 const (
-	ReplacingNodeUsingHostID ScyllaFeature = "ReplacingNodeUsingHostID"
+	ReplacingNodeUsingHostID                          ScyllaFeature = "ReplacingNodeUsingHostID"
+	ExposingScyllaClusterViaServiceOtherThanClusterIP ScyllaFeature = "ExposingScyllaClusterViaServiceOtherThanClusterIP"
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -112,6 +133,96 @@ func ValidateExposeOptions(options *scyllav1.ExposeOptions, fldPath *field.Path)
...
+
+	nodeServiceType := scyllav1.ClusterIPServiceType
+	if nodeService != nil {
+		nodeServiceType = nodeService.Type
+	}
+
+	// Skipping an error when chosen option type is unsupported as it won't help anyhow users reading it.
+	allowedNodeServiceTypes, ok := allowedNodeServiceTypesByBroadcastAddressType[options.Type]
+	if ok && !helpers.ContainsItem(allowedNodeServiceTypes, nodeServiceType) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("type"), options.Type, fmt.Sprintf("cannot broadcast address not available in chosen node service type, allowed types for given broadcast address are: %v", allowedNodeServiceTypes)))
```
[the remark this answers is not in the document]
zimnx: rephrased

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -99,6 +101,25 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("version"), fmt.Errorf("can't determine if version %q supports exposing via Service: %w", spec.Version, err)))
+		}
+
+		if !supportsExposingViaService {
+			if spec.ExposeOptions.NodeService != nil && spec.ExposeOptions.NodeService.Type != scyllav1.ClusterIPServiceType {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("exposeOptions", "nodeService", "type"), spec.ExposeOptions.NodeService.Type, fmt.Sprintf("can't expose ScyllaCluster via Service other than ClusterIP with given ScyllaDB version, upgrade to latest one first")))
+			}
+			if spec.ExposeOptions.BroadcastOptions != nil {
+				if spec.ExposeOptions.BroadcastOptions.Clients != nil && spec.ExposeOptions.BroadcastOptions.Clients.Type != scyllav1.ServiceClusterIPBroadcastAddressType {
+					allErrs = append(allErrs, field.Invalid(fldPath.Child("exposeOptions", "broadcastOptions", "clients", "type"), spec.ExposeOptions.BroadcastOptions.Clients.Type, fmt.Sprintf("can't broadcast address other than ClusterIP to clients with given ScyllaDB version, upgrade to latest one first")))
```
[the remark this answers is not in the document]
zimnx: done

pkg/cmd/operator/sidecar.go
```diff
@@ -116,6 +129,20 @@ func (o *SidecarOptions) Validate() error {
 		}
 	}
 
+	supportedBroadcastAddressTypes := []scyllav1.BroadcastAddressType{
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

pkg/api/scylla/validation/cluster_validation_test.go
```diff
@@ -253,6 +545,160 @@ func TestValidateScyllaClusterUpdate(t *testing.T) {
...
+				sc.Spec.ExposeOptions = nil
+				return sc
+			}(),
+			expectedErrorList: field.ErrorList{
+				&field.Error{Type: field.ErrorTypeInvalid, Field: "spec.exposeOptions.nodeService.type", BadValue: (*v1.NodeServiceType)(nil), Detail: `field is immutable`},
+			},
+			expectedErrorString: `spec.exposeOptions.nodeService.type: Invalid value: "null": field is immutable`,
+		},
+		{
+			name: "node service type is cannot be changed",
```
rzetelskik, line 570:
```suggestion
			name: "node service type cannot be changed",
```

pkg/api/scylla/validation/cluster_validation_test.go
```diff
@@ -253,6 +545,160 @@ func TestValidateScyllaClusterUpdate(t *testing.T) {
 			},
 			expectedErrorString: `[spec.datacenter.racks[0]: Forbidden: rack "rack-0" can't be removed because it still has members that have to be scaled down to zero first, spec.datacenter.racks[1]: Forbidden: rack "rack-1" can't be removed because it still has members that have to be scaled down to zero first, spec.datacenter.racks[2]: Forbidden: rack "rack-2" can't be removed because it still has members that have to be scaled down to zero first]`,
 		},
+		{
+			name: "node service type is cannot be unset",
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/validation/cluster_validation_test.go
zimnx: done

pkg/cmd/operator/sidecar.go
```diff
@@ -243,7 +280,7 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 			return o.kubeClient.CoreV1().Pods(o.Namespace).Watch(ctx, options)
 		},
 	}
-	klog.V(2).InfoS("Waiting for Pod To have scylla ContainerID set", "Pod", naming.ManualRef(o.Namespace, o.ServiceName))
+	klog.V(2).InfoS("Waiting for Pod IP Address and to have scylla ContainerID set", "Pod", naming.ManualRef(o.Namespace, o.ServiceName))
```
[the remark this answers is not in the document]
zimnx: done

pkg/sidecar/config/config.go
```diff
@@ -230,15 +235,41 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
+		"seeds":              pointer.Ptr(strings.Join(seeds, ",")),
+		"developer-mode":     &devMode,
+		"overprovisioned":    &overprovisioned,
+		"smp":                pointer.Ptr(strconv.Itoa(s.cpuCount)),
+		"prometheus-address": &prometheusAddress,
 	}
+
+	var broadcastAddress, broadcastRPCAddress string
+
+	switch s.nodesBroadcastAddressType {
```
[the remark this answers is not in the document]
zimnx: yep, leftover after refactoring. removed it. thanks

pkg/sidecar/config/config.go
```diff
@@ -230,15 +235,41 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
+		broadcastAddress = m.PodIP
+	case scyllav1.BroadcastAddressTypeServiceClusterIP:
+		broadcastAddress = m.ClusterIP
+	case scyllav1.BroadcastAddressTypeServiceLoadBalancerIngressIP:
+		broadcastAddress = m.ExternalIP
+	default:
+		return nil, fmt.Errorf("unsupported nodes broadcast address type: %q", s.nodesBroadcastAddressType)
+	}
+
+	switch s.clientsBroadcastAddressType {
```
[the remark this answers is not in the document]
zimnx: removed

pkg/sidecar/config/config.go
```diff
@@ -65,9 +66,11 @@ type ScyllaConfig struct {
 	scyllaRackDCPropertiesConfigMapPath string
 	cpuCount                            int
 	externalSeeds                       []string
+	nodesBroadcastAddressType           scyllav1.BroadcastAddressType
+	clientsBroadcastAddressType         scyllav1.BroadcastAddressType
 }
 
-func NewScyllaConfig(m *identity.Member, kubeClient kubernetes.Interface, scyllaClient scyllaversionedclient.Interface, cpuCount int, externalSeeds []string) *ScyllaConfig {
+func NewScyllaConfig(m *identity.Member, kubeClient kubernetes.Interface, scyllaClient scyllaversionedclient.Interface, cpuCount int, externalSeeds []string, nodesBroadcastAddressType, clientsBroadcastAddressType scyllav1.BroadcastAddressType) *ScyllaConfig {
```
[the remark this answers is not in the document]
zimnx: yep, removed

zimnx:
dump doesn't contain scylla-operator logs required to validate why webhook test failed.

pkg/controller/scyllacluster/resource.go
```diff
@@ -882,6 +925,20 @@ func stringOrDefault(str, def string) string {
...
 }
 
+func ptrValueOrDefault[T any](v *T, def T) T {
+	if v != nil {
+		return *v
+	}
+	return def
+}
+
+func copyPtr[T any](v *T) *T {
```
rzetelskik, line 935: Not a fan of the name of this func, the name implies the function returns a copy of the address?

pkg/controller/scyllacluster/resource.go
```diff
@@ -882,6 +925,20 @@ func stringOrDefault(str, def string) string {
 	return def
 }
 
+func ptrValueOrDefault[T any](v *T, def T) T {
```
rzetelskik, line 928: please use a verb for the func name, e.g. getValueOrDefault?

pkg/controller/scyllacluster/resource.go
```diff
@@ -110,6 +110,33 @@ func MemberService(sc *scyllav1.ScyllaCluster, rackName, name string, oldService
...
+		ns := sc.Spec.ExposeOptions.NodeService
+
+		switch ns.Type {
+		case scyllav1.NodeServiceTypeClusterIP:
+			serviceType = corev1.ServiceTypeClusterIP
+		case scyllav1.NodeServiceTypeLoadBalancer:
+			serviceType = corev1.ServiceTypeLoadBalancer
+		case scyllav1.NodeServiceTypeHeadless:
+			serviceType = corev1.ServiceTypeClusterIP
+			clusterIP = corev1.ClusterIPNone
```
rzetelskik, line 130: nit: move to a separate block for readability
```go
		switch ns.Type {
		case scyllav1.NodeServiceTypeClusterIP:
			serviceType = corev1.ServiceTypeClusterIP
		case scyllav1.NodeServiceTypeLoadBalancer:
			serviceType = corev1.ServiceTypeLoadBalancer
		case scyllav1.NodeServiceTypeHeadless:
			serviceType = corev1.ServiceTypeClusterIP
                }

                if ns.Type == scyllav1.NodeServiceTypeHeadless {
			clusterIP = corev1.ClusterIPNone
                }
```

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -236,14 +293,23 @@ func (scc *Controller) syncCerts(
...
-	var servingDNSNames []string
 	for _, domain := range sc.Spec.DNSDomains {
 		servingDNSNames = append(servingDNSNames, naming.GetCQLProtocolSubDomain(domain))
 
 		for _, hostID := range hostIDs {
 			servingDNSNames = append(servingDNSNames, naming.GetCQLHostIDSubDomain(hostID, domain))
 		}
 	}
+
+	// Sign for every node DNS name and discovery endpoint.
```
rzetelskik, line 304: I think I don't understand why it's necessary to introduce it in this PR? Could you please explain?

pkg/sidecar/identity/member.go
```diff
@@ -97,7 +143,11 @@ func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, ex
 
 	res := make([]string, 0, len(externalSeeds)+1)
 	res = append(res, externalSeeds...)
-	res = append(res, svc.Spec.ClusterIP)
+	if svc.Spec.ClusterIP == corev1.ClusterIPNone {
```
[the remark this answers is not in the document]
zimnx: It's not the same across the DC during rolling updates. We are reasoning here about other node, not ourselves so we don't have information about which broadcastAddressType it is using.

pkg/controller/scyllacluster/sync_certs.go
zimnx: When node service type is Headless, we end up with chicken egg problem for serving certificate. Certificate cannot have empty SANs, and we need to sign it for PodIP. Pod IP is allocated after Pod initialization, for which we need all non-optional volumes to be available. Serving cert is non-optional, and it should stay like that. 

To solve it, we sign certificate with in-cluster DNS names which is what is expected to be in the cert, but was missing.

pkg/cmd/operator/sidecar.go
```diff
@@ -118,6 +127,14 @@ func (o *SidecarOptions) Validate() error {
 		}
 	}
 
+	if !helpers.ContainsItem(validation.SupportedBroadcastAddressTypes, scyllav1.BroadcastAddressType(o.NodesBroadcastAddressTypeString)) {
```
tnozicka, line 130: nit: I guess this could live alongside the time, say `NewBroadcastAddressType(t string) (BroadcastAddressType, error)`

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+
+const (
+	// BroadcastAddressTypePodIP selects the IP address from Pod.
+	BroadcastAddressTypePodIP BroadcastAddressType = "PodIP"
+
+	// BroadcastAddressTypeServiceClusterIP selects the IP address from Service.spec.ClusterIP
+	BroadcastAddressTypeServiceClusterIP BroadcastAddressType = "ServiceClusterIP"
+
+	// BroadcastAddressTypeServiceLoadBalancerIngressIP selects the IP address or Hostname from Service.status.ingress[0]
+	BroadcastAddressTypeServiceLoadBalancerIngressIP BroadcastAddressType = "ServiceLoadBalancerIngressIP"
```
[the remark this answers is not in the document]
zimnx: good idea, changed.

pkg/sidecar/identity/member.go
```diff
@@ -20,32 +21,77 @@ type Member struct {
 	Name string
 	// Namespace of the Pod
 	Namespace string
-	// IP of the Pod
-	IP string
+	// PodIP IP of the Pod
+	PodIP string
+	// ExternalIP is an ExternalIP of LoadBalancer Service
```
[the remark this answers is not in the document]
zimnx: done

pkg/sidecar/identity/member.go
```diff
@@ -20,32 +21,77 @@ type Member struct {
...
+	ExternalIP string
```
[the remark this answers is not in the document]
zimnx: done

pkg/sidecar/identity/member_test.go
```diff
@@ -51,72 +52,161 @@ func TestMember_GetSeeds(t *testing.T) {
...
+			expectSeeds:                []string{"1.2.3.4"},
 		},
 		{
-			name:          "bootstrap with external seeds and first created Pod when all Pods from DC are down and external seeds are provided",
-			memberName:    firstPod.Name,
-			memberIP:      firstService.Spec.ClusterIP,
-			objects:       []runtime.Object{firstPod, firstService, secondPod, secondService, thirdPod, thirdService},
-			externalSeeds: []string{"10.0.1.1", "10.0.1.2"},
-			expectSeeds:   []string{"10.0.1.1", "10.0.1.2", secondService.Spec.ClusterIP},
+			name:      "bootstraps with itself using node-to-node identifier of LoadBalancer External IP type when cluster is empty",
```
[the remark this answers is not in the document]
zimnx: added

pkg/controller/scyllacluster/resource.go
zimnx: changed the name

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go
zimnx: good idea, done

pkg/cmd/operator/sidecar.go
```diff
@@ -218,6 +238,23 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
...
+					o.nodesBroadcastAddressType == scyllav1.BroadcastAddressTypeServiceLoadBalancerIngressAddress {
+					if len(service.Status.LoadBalancer.Ingress) == 0 {
+						klog.V(4).InfoS("LoadBalancer Service is awaiting for public endpoint")
+						return false, nil
+					}
+				}
+
+				if o.clientsBroadcastAddressType == scyllav1.BroadcastAddressTypeServiceClusterIP ||
+					o.nodesBroadcastAddressType == scyllav1.BroadcastAddressTypeServiceClusterIP {
+				}
```
tnozicka, line 256: empty if

pkg/cmd/operator/sidecar.go
```diff
@@ -256,6 +293,11 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 			case watch.Added, watch.Modified:
 				pod = e.Object.(*corev1.Pod)
 
+				if len(pod.Status.PodIP) == 0 {
```
tnozicka, line 296: I guess it may not be needed in all cases but probably doesn't hurt to always test the same path

pkg/controller/scyllacluster/resource.go
```diff
@@ -110,6 +110,36 @@ func MemberService(sc *scyllav1.ScyllaCluster, rackName, name string, oldService
...
+		ns := sc.Spec.ExposeOptions.NodeService
+
+		switch ns.Type {
+		case scyllav1.NodeServiceTypeClusterIP:
+			serviceType = corev1.ServiceTypeClusterIP
+		case scyllav1.NodeServiceTypeLoadBalancer:
+			serviceType = corev1.ServiceTypeLoadBalancer
+		case scyllav1.NodeServiceTypeHeadless:
+			serviceType = corev1.ServiceTypeClusterIP
+		}
```
tnozicka, line 130: explicit default handling for safety

pkg/controller/scyllacluster/resource.go
```diff
@@ -110,6 +110,36 @@ func MemberService(sc *scyllav1.ScyllaCluster, rackName, name string, oldService
 		}
 	}
 
+	var internalTrafficPolicy *corev1.ServiceInternalTrafficPolicy
```
tnozicka, line 113: would it be more readable to start by defining a Service and use the fields directly then extra vars?

pkg/controller/scyllacluster/resource.go
tnozicka: interesting :) I'd have expected to set all options specific to this service type at the `case` with the service type

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -191,23 +191,80 @@ func (scc *Controller) syncCerts(
...
-		ips := svc.Spec.ClusterIPs
-		if len(ips) == 0 {
-			klog.V(2).InfoS("No IP available for member", "Service", klog.KObj(svc))
-			continue
-		}
-		parsedIPs, err := helpers.ParseClusterIPs(ips)
-		if err != nil {
-			return progressingConditions, err
+	if sc.Spec.ExposeOptions != nil && sc.Spec.ExposeOptions.BroadcastOptions != nil && sc.Spec.ExposeOptions.BroadcastOptions.Clients != nil {
+		switch sc.Spec.ExposeOptions.BroadcastOptions.Clients.Type {
```
tnozicka, line 197: should we sign them for all that are available? say a clusteIP should work as an initial contact point even if you publish PodIPs and vice versa

tnozicka (commented):
I keep getting diff outdated so I'll pause here for a bit but got through about a half of it

pkg/controller/scyllacluster/sync_certs.go
rzetelskik: I see, thanks!

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -99,6 +108,25 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
 	}
 
 	if spec.ExposeOptions != nil {
+		supportsExposingViaService, err := scyllafeatures.VersionSupports(spec.Version, scyllafeatures.ExposingScyllaClusterViaServiceOtherThanClusterIP)
+		if err != nil {
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("version"), fmt.Errorf("can't determine if version %q supports exposing via Service: %w", spec.Version, err)))
```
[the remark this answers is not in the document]
zimnx: I agree it can be an issue, but not a big one. This rule is enforced only when user requests exposing, not by default. Images published by Scylla, both concrete versions and rolling tags are parsed correctly - except `latest` but this could be added as a special case here, as both `latest` (oss and enterprise) are enough up-to-date to support what we require here. 
Suggested solution, to move it to runtime wouldn't change anything. We don't have a better way of telling which version of scylla we deploy, before we actually deploy, run and check from the Scylla API. Adding metadata check from image registry would be out of scope of this PR.

pkg/controller/scyllacluster/resource.go
rzetelskik: I think my preference comes from being used to pattern matching, where you'd just assign the result of the case like:
```
let serviceType = case ns.Type of
        scyllav1.NodeServiceTypeClusterIP -> corev1.ServiceTypeClusterIP
        ...
```
Personally I think adding an extra assignment in just one of the cases is easy to miss, but I do agree that it's somewhat redundant to add an extra check later. Idk, the closest thing to the above in Go would probably be a function converting NodeServiceType to ServiceType?

pkg/controller/scyllacluster/resource.go
rzetelskik: I think `dupPtr` has the same issue. Idk, maybe just `copyValue`, `copyReferencedValue`? The fact that it returns a pointer doesn't need to be reflected in the name imo, but what it does internally (i.e. copy a dereferenced value) should.

pkg/cmd/operator/sidecar.go
zimnx: With function that converts from string, implementing validation with correct error is harder and there's no point of making it harder.
Compare how webook validation would look like:
```
_, err := scyllav1.NewBroadcastAddressType(string(options.Type))
if err != nil {
    // is it NotSupported or sometheing else happened?
}
```
compared to explicit check:
```
if !helpers.ContainsItem(SupportedBroadcastAddressTypes, options.Type) {
    allErrs = append(allErrs, field.NotSupported(fldPath.Child("type"), options.Type, helpers.ConvertSlice(SupportedBroadcastAddressTypes, helpers.ToString[scyllav1.BroadcastAddressType])))
}
```
The latter one is cleaner, `SupportedBroadcastAddressTypes` is necessary to generate nice error message.

pkg/controller/scyllacluster/resource.go
zimnx: changed to `copyReferencedValue`

pkg/controller/scyllacluster/resource.go
zimnx: brought back initial code, if list of stuff needed to be done for particular case, having them all in one branch is more readable.

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
 	// cql specifies expose options for CQL SSL backend.
 	// EXPERIMENTAL. Do not rely on any particular behaviour controlled by this field.
 	// +optional
 	CQL *CQLExposeOptions `json:"cql,omitempty"`
+
+	// nodeService controls properties of Service dedicated for each ScyllaCluster node.
+	// +kubebuilder:default:={type:"ClusterIP"}
+	NodeService *NodeServiceTemplate `json:"nodeService,omitempty"`
+
+	// BroadcastOptions defines how ScyllaDB node publishes its IP address to other nodes and clients
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+	// Check Kubernetes corev1.Service documentation about semantic of this field.
+	// +optional
+	AllocateLoadBalancerNodePorts *bool `json:"allocateLoadBalancerNodePorts,omitempty"`
+
+	// loadBalancerClass controls value of corev1/Service.spec.loadBalancerClass of each node Service.
+	// Check Kubernetes corev1.Service documentation about semantic of this field.
+	// +optional
+	LoadBalancerClass *string `json:"loadBalancerClass,omitempty"`
+
+	// internalTrafficPolicy controls value of corev1/Service.spec.internalTrafficPolicy of each node Service.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+
+type NodeServiceTemplate struct {
+	// annotations is a custom key value map merged with every node Service annotations.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// type is the Kubernetes Service type.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Value is immutable"
+	Type NodeServiceType `json:"type"`
```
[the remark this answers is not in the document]
zimnx: Removed CEL as it's beta from 1.25 and we can't implement all the rules using it. Entire required validation is in webhook.

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+	// +optional
+	PodIP *PodIPAddressOptions `json:"podIP"`
+}
+
+// NodeBroadcastOptions hold options related to addresses broadcasted by ScyllaDB node.
+type NodeBroadcastOptions struct {
+	// nodes specifies options related to the address that is broadcasted for communication with other nodes.
+	// This field controls the `broadcast_address` value in ScyllaDB config.
+	// +kubebuilder:default:={type:"ServiceClusterIP"}
+	Nodes *BroadcastOptions `json:"nodes"`
```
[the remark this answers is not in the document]
zimnx: yeah it shouldn't be, converted to bare struct

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+
+	// clients specifies options related to the address that is broadcasted for communication with clients.
+	// This field controls the `broadcast_rpc_address` value in ScyllaDB config.
+	// +kubebuilder:default:={type:"ServiceClusterIP"}
+	Clients *BroadcastOptions `json:"clients"`
```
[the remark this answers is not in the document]
zimnx: converted to bare struct

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+	BroadcastAddressTypeServiceLoadBalancerIngressIP BroadcastAddressType = "ServiceLoadBalancerIngressIP"
+)
+
+// BroadcastOptions hold options related to address broadcasted by ScyllaDB node.
+type BroadcastOptions struct {
+	// type of broadcasted address.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Value is immutable"
+	Type BroadcastAddressType `json:"type"`
+
+	// podIP hold options related to Pod IP address.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+	// BroadcastAddressTypeServiceClusterIP selects the IP address from Service.spec.ClusterIP
+	BroadcastAddressTypeServiceClusterIP BroadcastAddressType = "ServiceClusterIP"
+
+	// BroadcastAddressTypeServiceLoadBalancerIngressIP selects the IP address or Hostname from Service.status.ingress[0]
+	BroadcastAddressTypeServiceLoadBalancerIngressIP BroadcastAddressType = "ServiceLoadBalancerIngressIP"
+)
+
+// BroadcastOptions hold options related to address broadcasted by ScyllaDB node.
+type BroadcastOptions struct {
+	// type of broadcasted address.
```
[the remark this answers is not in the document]
zimnx: rephrased

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Value is immutable"
+	Type BroadcastAddressType `json:"type"`
```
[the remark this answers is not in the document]
zimnx: Removed CEL as it's beta from 1.25 and we can't implement all the rules using it. Entire required validation is in webhook.

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -111,13 +111,121 @@ type ScyllaClusterSpec struct {
...
+	// interfaceName specifies interface name within a Pod from which address is taken from.
+	InterfaceName string `json:"interfaceName,omitempty"`
+}
+
+// PodIPAddressOptions hold options related to Pod IP address.
+type PodIPAddressOptions struct {
+	// sourceType specifies source of the Pod IP.
+	// +kubebuilder:default:="Status"
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Value is immutable"
+	Source PodIPSourceType `json:"source"`
```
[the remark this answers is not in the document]
zimnx: Removed CEL as it's beta from 1.25 and we can't implement all the rules using it. Entire required validation is in webhook.

pkg/cmd/operator/sidecar.go
zimnx: removed

pkg/cmd/operator/sidecar.go
zimnx: NewMember reads it so it's better to wait until it's available in all cases

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go
zimnx: yep, looks better. thanks

pkg/controller/scyllacluster/sync_certs.go
zimnx: done

rzetelskik:
@zimnx thanks for the updates
/approve

pkg/api/scylla/validation/cluster_validation.go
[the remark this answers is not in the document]
tnozicka: > as both latest (oss and enterprise) are enough up-to-date to support what we require here.

latest can be arbitrarily old if you have it already pulled

> Suggested solution, to move it to runtime wouldn't change anything.

It allows us to do time consuming / async operations like fetching the image manifest and checking the LABELS. Validation is special as it can't afford to reach to external system or retry because it block the client calls.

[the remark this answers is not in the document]
zimnx: I changed it to runtime check, even though I consider it a degradation of UX.

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -4,26 +4,258 @@ package scyllacluster
...
+
+			hostsIPs, err := helpers.ParseClusterIPs(serviceAndPodIPs)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			var servingDNSNames []string
+			services, err := f.KubeClient().CoreV1().Services(sc.Namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: labels.SelectorFromSet(naming.ClusterLabels(sc)).String(),
+			})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			for _, svc := range services.Items {
```
tnozicka, line 102: pls assert the length so this silently doesn't pass with 0 or less svcs then we expect

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -4,26 +4,258 @@ package scyllacluster
...
+						Type: scyllav1.BroadcastAddressTypeServiceClusterIP,
+					},
+					Clients: scyllav1.BroadcastOptions{
+						Type: scyllav1.BroadcastAddressTypeServiceClusterIP,
+					},
+				},
+			},
+			validateService: func(svc *corev1.Service) {
+				o.Expect(svc.Spec.Type).To(o.Equal(corev1.ServiceTypeClusterIP))
+				parsedClusterIP := net.ParseIP(svc.Spec.ClusterIP)
```
tnozicka, line 134: tiny nit: can you assert it's not empty before parsing it for nicer failure?

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -4,26 +4,258 @@ package scyllacluster
...
+				o.Expect(parsedClusterIP).ToNot(o.BeNil())
+			},
+			validateScyllaConfig: func(ctx context.Context, configClient *scyllaclient.ConfigClient, svc *corev1.Service, pod *corev1.Pod) {
+				broadcastAddress, err := configClient.BroadcastAddress(ctx)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
```
tnozicka, line 140: nit: - `\n` (the next paragraph uses that pattern)

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -4,26 +4,258 @@ package scyllacluster
...
```
tnozicka, line 173: and here

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -4,26 +4,258 @@ package scyllacluster
...
+				},
+			},
+			validateService: func(svc *corev1.Service) {
+				o.Expect(svc.Spec.Type).To(o.Equal(corev1.ServiceTypeClusterIP))
+				o.Expect(svc.Spec.ClusterIP).To(o.Equal(corev1.ClusterIPNone))
+			},
+			validateScyllaConfig: func(ctx context.Context, configClient *scyllaclient.ConfigClient, svc *corev1.Service, pod *corev1.Pod) {
+				broadcastAddress, err := configClient.BroadcastAddress(ctx)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
```
tnozicka, line 208: and here

test/e2e/set/scyllacluster/scyllacluster_expose.go
```diff
@@ -4,26 +4,258 @@ package scyllacluster
...
+				o.Expect(err).NotTo(o.HaveOccurred())
+				o.Expect(broadcastRPCAddress).To(o.Equal(pod.Status.PodIP))
+			},
+		}),
+		g.Entry("using LoadBalancer Service, and broadcast external IP to client and PodIP to nodes", &entry{
+			exposeOptions: &scyllav1.ExposeOptions{
+				NodeService: &scyllav1.NodeServiceTemplate{
+					Type: scyllav1.NodeServiceTypeLoadBalancer,
+					// Change to non-default LB class to avoid conflicts on status when running in cloud
+					LoadBalancerClass: pointer.Ptr("my-lb-class"),
```
tnozicka, line 221: nit: If it truly doesn't matter, can it be a random string so it runs inside customer clusters where an ingress class named `my-lb-class` is not impossible?

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -292,7 +310,6 @@ var _ = g.Describe("ScyllaCluster", func() {
 				o.Expect(initialSecret.Data).To(o.HaveKey("tls.key"))
 				o.Expect(initialSecret.Data["tls.crt"]).NotTo(o.BeEmpty())
 				o.Expect(initialSecret.Data["tls.key"]).NotTo(o.BeEmpty())
-				o.Expect(initialSecret.Annotations).To(o.HaveKeyWithValue("certificates.internal.scylla-operator.scylladb.com/refresh-reason", "needs new cert"))
```
tnozicka, line 295: as the cert was generated in phases, was this no longer deterministic?

test/e2e/utils/helpers.go
```diff
@@ -445,21 +449,91 @@ func GetHostsAndUUIDs(ctx context.Context, client corev1client.CoreV1Interface,
...
+}
+
+func GetNodesPodIPs(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster) ([]string, error) {
+	clusterPods, err := client.Pods(sc.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			naming.ClusterNameLabel: sc.Name,
+		}).String(),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("can't get member services: %w", err)
```
tnozicka, line 525: pods

test/e2e/utils/helpers.go
```diff
@@ -445,21 +449,91 @@ func GetHostsAndUUIDs(ctx context.Context, client corev1client.CoreV1Interface,
...
+	}
+
+	ipAddresses := make([]string, 0, len(clusterPods.Items))
+
+	for _, pod := range clusterPods.Items {
+		ipAddresses = append(ipAddresses, pod.Status.PodIP)
+	}
+
+	return ipAddresses, nil
```
tnozicka, line 534: should this fail if the len(pods) != racks*replicas?

tnozicka (commented):
thanks for the updates, got through it all this time, rest lgtm

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -191,23 +192,55 @@ func (scc *Controller) syncCerts(
...
+				if parsedIP == nil {
+					return progressingConditions, fmt.Errorf("can't parse ingress IP %q", ingressStatus.IP)
+				}
+
+				ipAddresses = append(ipAddresses, parsedIP)
+			}
+
+			if len(ingressStatus.Hostname) != 0 {
+				servingDNSNames = append(servingDNSNames, ingressStatus.Hostname)
+			}
```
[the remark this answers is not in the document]
zimnx: added conditions when required type is not yet available

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -191,23 +192,55 @@ func (scc *Controller) syncCerts(
...
 		}
-		parsedIPs, err := helpers.ParseClusterIPs(ips)
+
+		pod, err := scc.podLister.Pods(sc.Namespace).Get(svc.Name)
 		if err != nil {
-			return progressingConditions, err
+			if apierrors.IsNotFound(err) {
+				continue
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -191,23 +192,55 @@ func (scc *Controller) syncCerts(
...
 
-		ipAddresses = append(ipAddresses, parsedIPs...)
+		if len(pod.Status.PodIP) != 0 {
+			parsedIP := net.ParseIP(pod.Status.PodIP)
+			if parsedIP == nil {
+				return progressingConditions, fmt.Errorf("can't parse Pod %q IP %q", naming.ObjRef(pod), pod.Status.PodIP)
+			}
+
+			ipAddresses = append(ipAddresses, parsedIP)
+		}
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/sync_services.go
```diff
@@ -184,9 +188,27 @@ func (scc *Controller) syncServices(
...
+		if err != nil {
+			return nil, fmt.Errorf("can't determine if ScyllaDB version %q supports exposing via Service other than ClusterIP: %w", sc.Spec.Version, err)
+		}
+
+		if !supportsExposing {
+			scc.eventRecorder.Eventf(
+				sc,
+				corev1.EventTypeWarning,
+				"InvalidScyllaDBVersion",
+				fmt.Sprintf("Requested ScyllaDB version %q does not support exposing via Service other than ClusterIP, use the latest one", sc.Spec.Version),
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/sync_statefulsets.go
```diff
@@ -463,6 +464,26 @@ func (scc *Controller) syncStatefulSets(
...
+					sc,
+					corev1.EventTypeWarning,
+					"InvalidScyllaDBVersion",
+					fmt.Sprintf("Requested ScyllaDB version %q does not support broadcasting other address than ClusterIP, use the latest one", sc.Spec.Version),
+				)
+				return nil, fmt.Errorf("requested ScyllaDB version %q does not support broadcasting other address than ClusterIP, use the latest one", sc.Spec.Version)
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
zimnx: done

pkg/sidecar/identity/member.go
```diff
@@ -20,32 +21,77 @@ type Member struct {
...
+
+		firstIngress := service.Status.LoadBalancer.Ingress[0]
+		if len(firstIngress.Hostname) != 0 {
+			externalIP = firstIngress.Hostname
+		}
+
+		if len(firstIngress.IP) != 0 {
+			externalIP = firstIngress.IP
+		}
+	}
```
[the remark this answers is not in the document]
zimnx: added error, before this function is called we have expilict wait so it must be available here

pkg/sidecar/identity/member.go
```diff
@@ -97,7 +143,11 @@ func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, ex
 
 	res := make([]string, 0, len(externalSeeds)+1)
 	res = append(res, externalSeeds...)
-	res = append(res, svc.Spec.ClusterIP)
+	if svc.Spec.ClusterIP == corev1.ClusterIPNone {
+		res = append(res, pod.Status.PodIP)
+	} else {
+		res = append(res, svc.Spec.ClusterIP)
+	}
```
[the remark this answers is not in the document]
zimnx: done

pkg/sidecar/identity/member_test.go
```diff
@@ -51,72 +52,181 @@ func TestMember_GetSeeds(t *testing.T) {
...
+			memberPod: func() *corev1.Pod {
+				pod := firstPod.DeepCopy()
+				pod.Status.PodIP = "1.2.3.4"
+				return pod
+			}(),
+			memberService:              firstService,
+			memberClientsBroadcastType: scyllav1.BroadcastAddressTypeServiceClusterIP,
+			memberNodesBroadcastType:   scyllav1.BroadcastAddressTypePodIP,
+			objects:                    []runtime.Object{firstPod, firstService},
+			expectSeeds:                []string{"1.2.3.4"},
```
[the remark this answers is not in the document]
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_expose.go
zimnx: done

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: Before this change PodIP wasn't added to certificate at all. PodIP is available after Pod initializes so before SC rolls out, cert created and then updated so reason no longer matches.

test/e2e/utils/helpers.go
zimnx: done

test/e2e/utils/helpers.go
zimnx: why not, but not here as this is just a helper. Added this check on caller side.

pkg/controller/scyllacluster/sync.go
```diff
@@ -199,6 +199,11 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
 		return scc.updateStatus(ctx, sc, status)
 	}
 
+	err = validateScyllaCluster(sc, scc.eventRecorder)
+	if err != nil {
+		return fmt.Errorf("ScyllaCluster %q did not pass validation: %w", naming.ObjRef(sc), err)
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/sync_certs.go
```diff
@@ -191,23 +192,80 @@ func (scc *Controller) syncCerts(
...
+
+		if pod != nil && len(pod.Status.PodIP) != 0 {
+			parsedIP := net.ParseIP(pod.Status.PodIP)
+			if parsedIP == nil {
+				return progressingConditions, fmt.Errorf("can't parse Pod %q IP %q", naming.ObjRef(pod), pod.Status.PodIP)
+			}
+
+			ipAddresses = append(ipAddresses, parsedIP)
+		} else if podIPRequired {
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/validation.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/scyllafeatures"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+func validateScyllaCluster(sc *scyllav1.ScyllaCluster, eventRecorder record.EventRecorder) error {
```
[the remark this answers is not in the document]
zimnx: done

tnozicka (approved):
thanks for the updates!

/approve
/lgtm

:rocket:

PR #1361, opened 2023-08-30
Align release schedule with team sprints

tnozicka:
When a GA makes it through the process sooner, we need to keep the schedule aligned with our sprints. This PR aligns the schedules. (Sprint 15 ends on Oct 15, so I postponed the release to the next working day.)


zimnx (approved):
/lgtm

PR #1369, opened 2023-09-05
Skip systemd dependent unit test when systemd is not available

zimnx:
when unit test is run inside a container where systemd is not
available the test is failing. It works fine when test binary is ran directly on the host, 
and with NodeConfig because NC DaemonSet chroots into host where systemd is present.

### DIFF
--- a/pkg/systemd/daemon_test.go
+++ b/pkg/systemd/daemon_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"testing"
 
+	godbus "github.com/godbus/dbus/v5"
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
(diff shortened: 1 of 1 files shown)


tnozicka (changes requested):
/hold

I don't know, this feels out of place.  Conditional unit tests not executed in the CI seem weird and probably the place where it breaks at some point.

If a unit test needs system wide installation of another component it may not be a good unit test.

AFAIK, systemd runs fine inside containers we just don't start it, neither I am sure whether we should but it's an option.

tnozicka:
hm, I don't see a quick way out of it and we need to get our unit suite working. I've filled https://github.com/scylladb/scylla-operator/issues/1379 but let's not take the conditional unit as a precedence.

Please reference the issue in the code change.

/hold cancel

zimnx:
> Please reference the issue in the code change.

Added reference.

tnozicka (approved):
/approve
/lgtm

thanks

PR #1371, opened 2023-09-05
Use generic pointer.Ptr in entire project

zimnx:
Thanks to generics we can use pointer.Ptr without specifying required type of return value. All usages of previous typed functions were changed to use the generic one.


zimnx:
Failure is unrelated to this PR.

/override ci/prow/unit

tnozicka (approved):
/approve
/lgtm

tnozicka:
/retest
autoscaler picked this node for scale down :(

Issue #1231, opened 2023-05-05
Flake: Scylla Manager integration [It] should discover cluster and sync tasks


rzetelskik:
https://github.com/scylladb/scylla-operator/actions/runs/4885577292/jobs/8729441531?pr=1230
[e2e-artifacts-parallel-alpha.tar.lz4.zip](https://github.com/scylladb/scylla-operator/files/11404414/e2e-artifacts-parallel-alpha.tar.lz4.zip)
[logs_13091.zip](https://github.com/scylladb/scylla-operator/files/11404415/logs_13091.zip)

```
[FAILED] [576.223 seconds]
I0505 05:42:08.396946       1 tests/tests_run.go:418] "Process finished" ID=6 Command="/usr/bin/scylla-operator-tests run scylla-operator/conformance/parallel --loglevel=2 --artifacts-dir=/home/runner/work/_temp/e2e-artifacts --flake-attempts=0 --timeout=60m --feature-gates=AllAlpha=true,AllBeta=true --override-ingress-address=10.96.88.229 --parallel-server-address=127.0.0.1:45553 --loglevel=0 --parallel-shard=6" ProcessState="exit status 1"
Scylla Manager integration [It] should discover cluster and sync tasks
I0505 05:42:08.398194       1 tests/tests_run.go:418] "Process finished" ID=13 Command="/usr/bin/scylla-operator-tests run scylla-operator/conformance/parallel --loglevel=2 --artifacts-dir=/home/runner/work/_temp/e2e-artifacts --flake-attempts=0 --timeout=60m --feature-gates=AllAlpha=true,AllBeta=true --override-ingress-address=10.96.88.229 --parallel-server-address=127.0.0.1:45553 --loglevel=0 --parallel-shard=13" ProcessState="exit status 0"
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:23
I0505 05:42:08.398992       1 tests/tests_run.go:418] "Process finished" ID=10 Command="/usr/bin/scylla-operator-tests run scylla-operator/conformance/parallel --loglevel=2 --artifacts-dir=/home/runner/work/_temp/e2e-artifacts --flake-attempts=0 --timeout=60m --feature-gates=AllAlpha=true,AllBeta=true --override-ingress-address=10.96.88.229 --parallel-server-address=127.0.0.1:45553 --loglevel=0 --parallel-shard=10" ProcessState="exit status 0"

  Timeline >>
[...]

Issue #749, opened 2021-08-06
Set up sysctls when tuning nodes


tnozicka:
**What should the feature do:**
We should tune sysctls related to scylla on the scylla nodes. Ideally we need to keep a list in sync with what's in regular scylla deployment but enforcing that may not be realistic. Also some values may need to be raised as we could run multiple scylla pods per kubernetes node.

**What is use case behind this feature:**
Some sysctls can be set directly on our CR but only those that are namespaced (https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#setting-sysctls-for-a-pod).
`fs.inotify.max_user_instances` is one of the cases that aren't namespaced and scylla needs them https://github.com/scylladb/scylla/issues/7700

**Additional Information:**
https://scylladb-users.slack.com/archives/CCC9J41FW/p1628195109003700
```[tasklist]
### Tasks
- [ ] https://github.com/scylladb/scylla-operator/issues/1372
- [ ] https://github.com/scylladb/scylla-operator/issues/1373
- [ ] https://github.com/scylladb/scylla-operator/issues/1385
```

mykaul:
Feasible for 1.9?

PR #1382, opened 2023-09-06
Increase fs.aio-max-nr on CI worker nodes

rzetelskik:
**Description of your changes:** As discovered during the debugging of https://github.com/scylladb/scylla-operator/issues/1231, our CI suffers from too few allowable concurrent requests with the current parallelism of the e2e suite. Until https://github.com/scylladb/scylla-operator/issues/749 is completed, this PR deploys a DaemonSet on worker nodes of our CI to increase the `fs.aio-max-nr` as a workaround.

Resolves https://github.com/scylladb/scylla-operator/issues/1231

### DIFF
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -60,6 +60,45 @@ kubectl_create -f - <<EOF
     - patch
 EOF
 
+# FIXME: remove the workaround once https://github.com/scylladb/scylla-operator/issues/749 is done
+kubectl_create -n default -f - <<EOF
+  apiVersion: apps/v1
+  kind: DaemonSet
+  metadata:
+    name: sysctl
+  spec:
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: sysctl
+    template:
+      metadata:
+        labels:
+          app.kubernetes.io/name: sysctl
+      spec:
+        containers:
+        - name: sysctl
+          securityContext:
+            privileged: true
+          image: "${SO_IMAGE}"
+          imagePullPolicy: IfNotPresent
+          command:
+          - /usr/bin/bash
+          - -euExo
+          - pipefail
+          - -O
+          - inherit_errexit
+          - -c
+          args:
+          - |
+            sysctl fs.aio-max-nr=0xffffffff
+
+            sleep infinity &
+            wait
+        nodeSelector:
+          scylla.scylladb.com/node-type: scylla
+EOF
+kubectl -n default rollout status daemonset/sysctl
+
 kubectl apply --server-side -f ./pkg/api/scylla/v1alpha1/scylla.scylladb.com_nodeconfigs.yaml
 kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com
 


hack/.ci/run-e2e-gke.sh
```diff
@@ -60,6 +60,45 @@ kubectl_create -f - <<EOF
...
+          command:
+          - /usr/bin/bash
+          - -euExo
+          - pipefail
+          - -O
+          - inherit_errexit
+          - -c
+          args:
+          - |
+            sysctl fs.aio-max-nr=0xffffffff
```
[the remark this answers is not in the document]
tnozicka: Thanks, our CI clusters are not in the guaranteed class nor receive any significant traffic during the run, so the idea was more about fitting as many as we can without being artificially limited on this.

Does it have a relation to CPU count or the size of the cluster in general? When we recommend tuning this properly with https://github.com/scylladb/scylla-operator/issues/749 we need some metric to determine the value for the node while we don't know how many or how big clusters will run there. OR there are more advance options like advertising it as hardware resource using device plugins https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/ but that may be an overkill, at least at this point.

This PR is just about unstucking the CI that runs in the best effort QoS class / overcommitted.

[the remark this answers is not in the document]
rzetelskik: Yup, that's just a workaround aimed at eliminating flakes until it's supported by NodeConfig - see https://github.com/scylladb/scylla-operator/issues/1231. Hence an over the top value of max unsigned long.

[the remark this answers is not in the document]
tnozicka: might worth to put the reasons for this into a comment

hack/.ci/run-e2e-gke.sh
```diff
@@ -60,6 +60,46 @@ kubectl_create -f - <<EOF
...
+
+            sleep infinity &
+            wait
```
tnozicka, line 96: how is this different from `sleep infinity` on the foreground

hack/.ci/run-e2e-gke.sh
```diff
@@ -60,6 +60,46 @@ kubectl_create -f - <<EOF
...
+        nodeSelector:
+          kubernetes.io/os: linux
+          pool: workers
```
tnozicka, line 99: I think `"scylla.scylladb.com/node-type": "scylla"` would be more fitting and stable

rzetelskik: I did that initially but changed to `pool: workers` to align with existing manifests: https://github.com/scylladb/scylla-operator/blob/25735c38e04ee9deb0fa5f471f2ce7951aff5b16/hack/.ci/manifests/namespaces/local-csi-driver/50_daemonset.yaml#L19. Does `pool: workers` have a different meaning, i.e. is worker pool potentially a superset of scylla node pool? Anyway `"scylla.scylladb.com/node-type": "scylla"` is neater so I'll change it.

hack/.ci/run-e2e-gke.sh
rzetelskik: sigint handling?

hack/.ci/run-e2e-gke.sh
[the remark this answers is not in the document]
rzetelskik: Added a PR description, a FIXME in the script and created a tracking issue. Let me know if that suffices.

hack/.ci/run-e2e-gke.sh
tnozicka: thanks. not really, but `scylla.scylladb.com/node-type` is more standard and we use it in our docs. `pool: workers` for probably for some convenience and could go away.

hack/.ci/run-e2e-gke.sh
tnozicka: interesting, sending `SIGINT` to `sleep` works for me
```
 tnozicka  ~  sleep infinity
^C
 tnozicka  ~  SIGINT  
```

rzetelskik: Doesn't ^C send a SIGINT to the entire process group?
`kill -s INT -$PGID` as opposed to `kill -s INT $PID`

rzetelskik: Shouldn't you kill the bash process in that case?

```bash
~ cat script.sh
#!/bin/bash
sleep infinity
~ ./script.sh &
~ ps o pid,ppid,pgid,comm
   6467    6289    6467 zsh
  13292    6467   13292 script.sh
  13294   13292   13292 sleep
  13500    6467   13500 ps
~ kill -s INT 13292
~ ps o pid,ppid,pgid,comm
   6467    6289    6467 zsh
  13292    6467   13292 script.sh
  13294   13292   13292 sleep
  13931    6467   13931 ps
~ kill -s INT -13292
[1]  + 13292 interrupt  ./script.sh
```

PR #1388, opened 2023-09-08
Fix a race condition in seed election algorithm

### LINKED ISSUE #1349: ScyllaCluster can bootstrap into multiple independent cluster on a race
**Describe the bug**

**To Reproduce**
1. node-0 starts and bootstraps
2. node-1 pods is created but before it starts, dc-b-node-0 gets evicted
3. node-1 has no external seeds (as for this proposal) and hits
https://github.com/scylladb/scylla-operator/blob/d8e7d3c7580bbdd613ca673b115137a70c0f9294/pkg/sidecar/identity/member.go#L74-L77
4. node-1 bootstraps from its own IP and form a separate cluster

**Expected behavior**
The self seed can only be used on node with index 0.

rzetelskik:
**Description of your changes:** Currently, the seed election algorithm checks for other existing Pods on startup and only chooses its own identity for seed provisioning if there are no other pods. As described in #1349, this can lead to a race condition where eviction of a Pod with lower ordinal can cause a split-brain cluster to form, as more than one Pod will consider itself the first in the cluster. This PR fixes the issue by verifying the ordinals of the node and the rack it belongs to. To accommodate for these changes, the unit tests also need some tweaks. 

To avoid an import cycle between helpers and naming packages, the array package is moved to a helpers subpackage. Since `array` didn't reflect its contents, it's also renamed to `slices`.

Resolves #1349

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -195,8 +195,11 @@ func servicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
 // StatefulSetForRack make a StatefulSet for the rack.
 // existingSts may be nil if it doesn't exist yet.
 func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existingSts *appsv1.StatefulSet, sidecarImage string) (*appsv1.StatefulSet, error) {
-	matchLabels := naming.RackLabels(r, c)
-	rackLabels := naming.RackLabels(r, c)
+	rackLabels, err := naming.RackLabels(r, c)
+	if err != nil {
+		return nil, fmt.Errorf("can't get rack labels: %w", err)
+	}
+	matchLabels := helpers.ShallowCopyMap(rackLabels)
 	rackLabels[naming.ScyllaVersionLabel] = c.Spec.Version
 
 	placement := r.Placement
--- a/pkg/helpers/collections.go
+++ b/pkg/helpers/collections.go
@@ -18,3 +18,11 @@ func GetMapValues[M ~map[K]V, K comparable, V any](m M) []V {
 	}
 	return res
 }
+
+func ShallowCopyMap[M ~map[K]V, K comparable, V any](m M) M {
+	res := make(M, len(m))
+	for k, v := range m {
+		res[k] = v
+	}
+	return res
+}
--- a/pkg/helpers/slices/slices.go
+++ b/pkg/helpers/slices/slices.go
@@ -65,14 +65,18 @@ func ContainsItem[T comparable](slice []T, item T) bool {
 	return Contains(slice, IdentityFunc(item))
 }
 
-func Find[T comparable](slice []T, filterFunc func(T) bool) (T, bool) {
+func Find[T any](slice []T, filterFunc func(T) bool) (T, int, bool) {
 	for i := range slice {
 		if filterFunc(slice[i]) {
-			return slice[i], true
+			return slice[i], i, true
 		}
 	}
 
-	return *new(T), false
+	return *new(T), 0, false
+}
+
+func FindItem[T comparable](slice []T, item T) (T, int, bool) {
+	return Find(slice, IdentityFunc(item))
 }
 
 func Flatten[T any](xs [][]T) []T {
--- a/pkg/helpers/slices/slices.go
+++ b/pkg/helpers/slices/slices.go
@@ -1,11 +1,13 @@
-package helpers
+// Copyright (c) 2023 ScyllaDB
 
-func ToArray[T any](objs ...T) []T {
+package slices
+
+func ToSlice[T any](objs ...T) []T {
 	res := make([]T, 0, len(objs))
 	return append(res, objs...)
 }
 
-func ConvertToArray[To, From any](convert func(From) To, objs ...From) []To {
+func ConvertToSlice[To, From any](convert func(From) To, objs ...From) []To {
 	res := make([]To, 0, len(objs))
 
 	for i := range objs {
@@ -16,7 +18,7 @@ func ConvertToArray[To, From any](convert func(From) To, objs ...From) []To {
 }
 
 func ConvertSlice[To, From any](slice []From, convert func(From) To) []To {
-	return ConvertToArray(convert, slice...)
+	return ConvertToSlice(convert, slice...)
 }
 
 func Filter[T any](array []T, filterFunc func(T) bool) []T {
--- a/pkg/naming/labels.go
+++ b/pkg/naming/labels.go
@@ -1,7 +1,11 @@
 package naming
 
 import (
+	"fmt"
+	"strconv"
+
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/labels"
 )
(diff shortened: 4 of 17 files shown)


pkg/controller/scyllacluster/resource.go
```diff
@@ -195,8 +195,11 @@ func servicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
 // StatefulSetForRack make a StatefulSet for the rack.
 // existingSts may be nil if it doesn't exist yet.
 func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existingSts *appsv1.StatefulSet, sidecarImage string) (*appsv1.StatefulSet, error) {
-	matchLabels := naming.RackLabels(r, c)
-	rackLabels := naming.RackLabels(r, c)
+	rackLabels, err := naming.RackLabels(r, c)
+	if err != nil {
+		return nil, fmt.Errorf("can't get rack labels: %w", err)
+	}
+	matchLabels := helpers.CopyMap(rackLabels)
```
zimnx, line 202: there's no need for copying, RackLabels returns different map instance for each call.

pkg/naming/names.go
```diff
@@ -100,6 +100,16 @@ func IndexFromName(n string) (int32, error) {
 	return int32(index), nil
 }
 
+func RackOrdinalFromName(name string, sc *scyllav1.ScyllaCluster) (int, error) {
+	for i := range sc.Spec.Datacenter.Racks {
+		if sc.Spec.Datacenter.Racks[i].Name == name {
+			return i, nil
+		}
+	}
```
zimnx, line 108: nit: a little bit shorter
```suggestion
	for i, r := range sc.Spec.Datacenter.Racks {
		if r.Name == name {
			return i, nil
		}
	}
```

pkg/sidecar/identity/member.go
```diff
@@ -33,19 +35,29 @@ type Member struct {
...
 
-func NewMemberFromObjects(service *corev1.Service, pod *corev1.Pod) *Member {
+func NewMemberFromObjects(service *corev1.Service, pod *corev1.Pod) (*Member, error) {
+	rackOrdinalString, ok := pod.Labels[naming.RackOrdinalLabel]
+	if !ok {
+		return nil, fmt.Errorf("pod %q is missing %q label", naming.ObjRef(pod), naming.RackOrdinalLabel)
+	}
+	rackOrdinal, err := strconv.Atoi(rackOrdinalString)
+	if err != nil {
+		return nil, fmt.Errorf("can't get rack ordinal from label: %w", err)
```
zimnx, line 45:
```suggestion
		return nil, fmt.Errorf("can't get rack ordinal from label %q: %w", rackOrdinalString, err)
```

pkg/sidecar/identity/member.go
```diff
@@ -73,6 +85,16 @@ func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, ex
 
 	if len(otherPods) == 0 {
 		// We are the only one, assuming first bootstrap.
+
```
zimnx, line 88: How about we early return at the beginning if rackOrdinal and podOrdinal == 0? I think it would make this logic a little bit cleaner, and it would save us one listing - cheap one but still.

rzetelskik: I've implemented it that way initially but it breaks replacement - a node can't be its own seed if its replacing a node, so in a single DC (no external seeds) it'd cause Scylla to fail on bootstrap.

zimnx: I see, it makes sense. Lets leave it as is then.

pkg/controller/scyllacluster/resource.go
zimnx: i haven't noticed this is different variable, it's fine then.

pkg/naming/names.go
rzetelskik: Shorter, but the copy here is a bit unnecessary?

pkg/sidecar/identity/member.go

pkg/naming/names.go
tnozicka: I'd lean to readability first, compiler's job is to optimize the copy if it's not needed (+ the N is totally negligible)

pkg/helpers/collections.go
```diff
@@ -18,3 +18,11 @@ func GetMapValues[M ~map[K]V, K comparable, V any](m M) []V {
 	}
 	return res
 }
+
+func CopyMap[M ~map[K]V, K comparable, V any](m M) M {
```
tnozicka, line 22: to prevent future mishapps, I'd like to see `shallow` in its name

pkg/naming/names.go
```diff
@@ -100,6 +100,16 @@ func IndexFromName(n string) (int32, error) {
...
+
+	return -1, fmt.Errorf("can't find rack %q in ScyllaCluster %q", name, ObjRef(sc))
```
tnozicka, line 110: usually, when erroring out this is the default value like `nil, 0, ""` or at least I don't see the special value used

pkg/naming/names.go
```diff
@@ -100,6 +100,16 @@ func IndexFromName(n string) (int32, error) {
 	return int32(index), nil
 }
 
+func RackOrdinalFromName(name string, sc *scyllav1.ScyllaCluster) (int, error) {
```
tnozicka, line 103: sounds like `FindIndex` that could share implementation with https://github.com/scylladb/scylla-operator/blob/18e1ecc18233a952560ccac46507115b93911a77/pkg/helpers/array.go#L72-L80

pkg/sidecar/identity/member.go
```diff
@@ -73,6 +85,16 @@ func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, ex
...
+		podOrdinal, err := naming.IndexFromName(m.Name)
+		if err != nil {
+			return nil, fmt.Errorf("can't get pod index from name: %w", err)
+		}
+
+		if m.RackOrdinal != 0 || podOrdinal != 0 {
```
tnozicka, line 94: sounds like rack ordinal could be shortcut it at the start before even getting the pod ordinal

pkg/helpers/collections.go
rzetelskik: done

pkg/naming/names.go
rzetelskik: removed func in favor of generic FindIndex

pkg/naming/names.go

pkg/naming/names.go
rzetelskik: fair enough, done

pkg/helpers/array/array.go
```diff
@@ -1,4 +1,6 @@
-package helpers
+// Copyright (c) 2023 ScyllaDB
+
+package array
```
zimnx, line 3: `array` in Go is a slice with fixed size. This package contains functions working on slices, not arrays.  I would suggest to rename package and functions to use correct type name. 
Name of the package should be plural to be consistent with similar packages `bytes`, `strings` from stdlib.

rzetelskik: I agree, changed

pkg/helpers/slices/slices.go
```diff
@@ -63,14 +65,31 @@ func ContainsItem[T comparable](slice []T, item T) bool {
...
+	}
+
+	return *new(T), false
+}
+
+func FindItem[T comparable](slice []T, item T) (T, bool) {
+	return Find(slice, IdentityFunc(item))
+}
+
+func FindIndex[T any](slice []T, filterFunc func(T) bool) (int, bool) {
```
tnozicka, line 81: might be nice to share the implementation
```golang
func FindIndex[T any](slice []T, filterFunc func(T) bool) (T, int, bool) {
// for ...
}

func Find[T any](slice []T, filterFunc func(T) bool) (T, bool) {
    v, _, err := FindIndex(slice, filterFunc)
    return v, err
}
```
Or have `Find to return the index` and update the call places

rzetelskik: I'm leaning towards having a single function, changed

rzetelskik:
No logs from the serial suite.
/test e2e-gke-serial

pkg/naming/labels.go
```diff
@@ -26,12 +30,23 @@ func DatacenterLabels(c *scyllav1.ScyllaCluster) map[string]string {
...
 	recLabels := ScyllaLabels()
 	rackLabels := DatacenterLabels(c)
 	rackLabels[RackNameLabel] = r.Name
+	_, rackOrdinal, ok := slices.Find(c.Spec.Datacenter.Racks, func(rack scyllav1.RackSpec) bool {
+		if rack.Name == r.Name {
+			return true
+		}
 
-	return mergeLabels(rackLabels, recLabels)
+		return false
```
tnozicka, line 42: nit: for a one shot like this maybe it would be enought to `return rack.Name == r.Name`?

tnozicka (approved):
/approve

pkg/naming/labels.go
rzetelskik: obviously, thanks for noticing that

zimnx (approved):
/lgtm

Issue #1399, opened 2023-09-18
Failing test:  gocql panics in E2E test suite with `panic: scylla: <ip>:9042 invalid number of shards`


rzetelskik:
We've consistently been hitting a panic from gocql in our E2E test suite on master.

https://github.com/scylladb/scylla-operator/actions/runs/6178751661/job/16772628746#step:3:3702
```
E0914 00:01:49.220811       1 tests/tests_run.go:447] "Process failed" err="process failed" ID=6 Command="/usr/bin/scylla-operator-tests run scylla-operator/conformance/parallel --loglevel=2 --artifacts-dir=/home/runner/work/_temp/e2e-artifacts --flake-attempts=5 --timeout=110m --feature-gates= --override-ingress-address=10.101.34.231 --parallelism=8 --parallel-server-address=127.0.0.1:34643 --loglevel=0 --random-seed=1694646811 --parallel-shard=6" ProcessState="exit status 2" Logs=<
...
	panic: scylla: 10.105.211.70:9042 invalid number of shards
	
	goroutine 997 [running]:
	github.com/gocql/gocql.(*scyllaConnPicker).Put(0xc0005b1440, 0xc000904000)
		github.com/gocql/gocql@v1.6.0/scylla.go:400 +0x448
	github.com/gocql/gocql.(*hostConnPool).connect(0xc00058b1f0)
		github.com/gocql/gocql@v1.6.0/connectionpool.go:554 +0x2b0
	github.com/gocql/gocql.(*hostConnPool).fill(0xc00058b1f0)
		github.com/gocql/gocql@v1.6.0/connectionpool.go:397 +0x159
	github.com/gocql/gocql.(*policyConnPool).addHost(0xc0003decd0, 0xc000b849c0)
		github.com/gocql/gocql@v1.6.0/connectionpool.go:255 +0x1cf
	github.com/gocql/gocql.(*Session).startPoolFill(0xc000685000, 0x0?)
		github.com/gocql/gocql@v1.6.0/events.go:214 +0x2d
	created by github.com/gocql/gocql.(*controlConn).setupConn
		github.com/gocql/gocql@v1.6.0/control.go:301 +0x26a
 >
Error: all processes have finished but the suite still isn't done
```

https://github.com/scylladb/scylla-operator/suites/16143937950/artifacts/921837495

See also:
https://github.com/scylladb/scylla-operator/actions/runs/6191772438/job/16810747030#step:3:3178
[...]

PR #1401, opened 2023-09-18
Annotate E2E namespaces with parallel process and full text of a ginkgo node

rzetelskik:
**Description of your changes:** As proved by https://github.com/scylladb/scylla-operator/issues/1399, the logs printed out by a failed process of out test runner make it difficult to backtrack an issue to a particular test. This PR annotates the Namespaces created in the framework with the number of the parallel process running the spec as well as the full text of the ginkgo node. It should help with backtracking the failure to an offending test.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strconv"
 	"time"
 
 	g "github.com/onsi/ginkgo/v2"
@@ -148,6 +149,7 @@ func (f *Framework) setupNamespace(ctx context.Context) {
 		return names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", f.name))
 	}
 	name := generateName()
+	sr := g.CurrentSpecReport()
 	err := wait.PollImmediate(2*time.Second, 30*time.Second, func() (bool, error) {
 		var err error
 		// We want to know the name ahead, even if the api call fails.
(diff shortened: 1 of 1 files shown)


test/e2e/framework/framework.go
```diff
@@ -141,6 +142,15 @@ func (f *Framework) CommonLabels() map[string]string {
 	}
 }
 
+func (f *Framework) CurrentSpecAnnotations() map[string]string {
+	sr := g.CurrentSpecReport()
+
+	return map[string]string{
+		"ginkgo-parallel-process": strconv.Itoa(sr.ParallelProcess),
+		"ginkgo-full-text":        fmt.Sprintf("%q", sr.FullText()),
```
tnozicka, line 150: why do we need to escape it?

test/e2e/framework/framework.go
```diff
@@ -141,6 +142,15 @@ func (f *Framework) CommonLabels() map[string]string {
 	}
 }
 
+func (f *Framework) CurrentSpecAnnotations() map[string]string {
```
tnozicka, line 145: seems more private but maybe just inlining it into setupNamespace would be best to avoid having to merge maps when we want to add an annotation

rzetelskik: I did it this way so that it could be repurposed, but I don't really see the need for it now. Changed.

test/e2e/framework/framework.go
rzetelskik: we don't, changed

PR #1363, opened 2023-09-01
Update dependencies

tnozicka:
**Description of your changes:**
This PR updates our dependencies to the latest stable levels.

PR #1408, opened 2023-09-20
Revert gocql bump to v1.11.1

rzetelskik:
Gocql version was bumped from v1.7.3 to v1.11.1 in https://github.com/scylladb/scylla-operator/pull/1363. Unfortunately it quite consistently breaks one of the e2e tests in postsubmits in our GitHub CI: see https://github.com/scylladb/scylla-operator/issues/1399. This PR reverts the version back to v1.7.3.


tnozicka (approved):
/approve
/lgtm

See if you could rephrase the title, this is a bit ambiguous and I think in the end the version you care about, if reading it in the release notes, is where it ended (aka to v1.7.3)

tnozicka:
> This makes little sense to me. @avelanarius - can you work with the operator team to figure this one out ? Otherwise, we are less likely to find it and be able to move forward.

@mykaul we have a consistent reproducer on our local environment - what's the point of keeping the periodic CI tests broken and firing alerts given we are tracking this in #1399 and @rzetelskik is about to file a driver issue?

mykaul:
> > This makes little sense to me. @avelanarius - can you work with the operator team to figure this one out ? Otherwise, we are less likely to find it and be able to move forward.
> 
> @mykaul we have a consistent reproducer on our local environment - what's the point of keeping the periodic CI tests broken and firing alerts given we are tracking this in #1399 and @rzetelskik is about to file a driver issue?

If it's consistent, I'd like to see it worked on and fixed promptly. We can revert to get the CI running, but I'm uncomfortable with releasing a new operator release with this revert.

tnozicka:
> but I'm uncomfortable with releasing a new operator release with this revert.

I am fine holding the release if you are prepared to have the features coming in 1.11 hold off, if the drivers aren't fixed in time. Code freeze is planned for October 2nd.

FWIW, [operator v1.10.0 was also using gocql 1.7.3](https://github.com/scylladb/scylla-operator/blame/v1.10.0/go.sum#L370-L371) so the reverted version was never included in a release of the operator + it seems to be a driver issue and the gocql is used **only** in e2e test, not the operator itself - so the risk is our e2e will have issues or this is somehow not just a driver issue, while everything points that way.

mykaul:
If it's just in the E2E, no worries then.

PR #1418, opened 2023-09-25
Replace recommonmark with MyST-Parser in docs

rzetelskik:
**Description of your changes:** This PR replaces sphinx's deprecated extension `recommonmark` with `myst-parser`, as per https://sphinx-theme.scylladb.com/stable/configuration/markdown.html#recommonmark-parser.

Referential PR: https://github.com/scylladb/sphinx-scylladb-theme/pull/803

**Which issue is resolved by this Pull Request**:
Resolves: #1420 

/cc @dgarcia360


docs/source/conf.py
```diff
@@ -53,13 +51,10 @@
...
-def setup(sphinx):
-    sphinx.add_config_value('recommonmark_config', {
-        'enable_eval_rst': True,
-        'enable_auto_toc_tree': False,
-    }, True)
-    sphinx.add_transform(AutoStructify)
+# -- Options for myst parser ----------------------------------------
+
+myst_enable_extensions = ["colon_fence"]
+suppress_warnings = ["myst.header", "ref.myst"]
```
tnozicka, line 57: not sure we should just suppress those, at least without a tracking issue

tnozicka (commented):
/approve

annastuchlik (commented):
It looks good to me, but I'd like @dgarcia360 to review it, too. We have the issue to move from Recommonmark to Myst on our plate anyway.

docs/source/conf.py
rzetelskik: https://github.com/scylladb/scylla-operator/issues/1421
https://github.com/scylladb/scylla-operator/issues/1422
done

tnozicka (approved):
/lgtm

/hold
(for @dgarcia360  to have a look as well)

/assign @dgarcia360

zimnx (approved):

tzach:
I was under the impression the dependencies comes from https://github.com/scylladb/sphinx-scylladb-theme, so we do not have to update each project independently (in addition to upgrading the theme)
@dgarcia360 is it the case?

rzetelskik:
@tzach it's ahead of time, yes, but we have a custom workflow, so we update independently anyway.

tnozicka:
> I was under the impression the dependencies comes from https://github.com/scylladb/sphinx-scylladb-theme, so we do not have to update each project independently

@tzach While this may be desired from a product side, it has a side effect and it's in contrast with building a stable product and having a reproducible CI. If you run the same job again, it should produce the same results and changes in other repos shouldn't break yours without an explicit update to your repo and running it's CI. This is why we lock the dependencies both for Golang and Python.

dgarcia360:
@rzetelskik @annastuchlik Locally works fine. However, there is an issue when I execute the `make multiversion` command - an error raises for every branch, except `master`:

image

I'm currently evaluating whether removing support for recommonmark in previous versions could be a contributing factor to the issue we are experiencing. It ispossible that maintaining support for it in previous versions might be necessary. I'll be assessing this in the context of https://github.com/scylladb/sphinx-scylladb-theme/issues/168

rzetelskik:
/hold
The existing docs should also be updated to use the new dependency once the issues are resolved.

rzetelskik:
Rebased on top of current master.

zimnx:
I wonder, don't we need old parser for multiversion documentation? Documentation framework is taken from master, but content from particular version is taken from branch. If old version documentation contain old syntax, we won't be able to parse it correctly when parser is removed.

tnozicka:
/hold
(To avoid merging this prematurely, we don't test multiversion in the CI and this might have the potential to break it. Aka needs manual verification.)

rzetelskik:
> don't we need old parser for multiversion documentation

Yes, but I was thinking we would backport the syntax changes?

rzetelskik:
Let's wait for @dgarcia360 to solve https://github.com/scylladb/sphinx-scylladb-theme/issues/168 then.

rzetelskik:
@annastuchlik while the conf.py is read from the checked out branch, the old branches will still use the recommonmark syntax. As David mentioned earlier, running multiversion will also result in errors for other branches:
```
poetry run sphinx-multiversion source _build/dirhtml
...
Failed load config for refs/heads/v1.11 from /tmp/tmpv2c_gc8w/99c012bb7aff89a94701703f411a70568a30fd81/docs/source
...
```
It's not as simple as just having the two extensions installed though, I suppose we'd have to conditionally register extensions for different versions somehow.

dgarcia360:
Let's keep this pull request on hold - it still fails locally when running the multiversion command.

Now that we can use the latest version of Myst Parser, I'm evaluating how to conditionally register the extensions per version as part of https://github.com/scylladb/sphinx-scylladb-theme/issues/168

zimnx (approved):
/assign tnozicka

tnozicka (approved):
/lgtm
thanks

Issue #1429, opened 2023-09-28
`Dump cluster state` fails in postsubmits due to an invalid podman command


rzetelskik:
https://github.com/scylladb/scylla-operator/pull/1368 introduced an invalid podman command into `Dump cluster state` job in `Run E2E` github action, which causes our postsubmits to fail:

```console
+ sudo timeout -v 10m podman run --user=1001:127 --rm --entrypoint=/usr/bin/scylla-operator -v=/home/runner/work/_temp/e2e-artifacts:/home/runner/work/_temp/e2e-artifacts:rw -v=/home/runner/.kube/config:/kubeconfig:ro -e=KUBECONFIG=/kubeconfig docker.io/scylladb/scylla-operator:ci run must-gather --all-resources --dest-dir=/home/runner/work/_temp/e2e-artifacts/must-gather --loglevel=2
time="2023-09-27T17:36:43Z" level=warning msg="Error validating CNI config file /etc/cni/net.d/100-crio-bridge.conflist: [plugin bridge does not support config version \"1.0.0\"]"
time="2023-09-27T17:36:43Z" level=warning msg="Error validating CNI config file /etc/cni/net.d/200-loopback.conflist: [plugin loopback does not support config version \"1.0.0\"]"
Error: unknown command "run" for ""

Did you mean this?
	run-webhook-server

Run ' --help' for usage.
Error: unknown command "run" for ""

Did you mean this?
	run-webhook-server

Error: Process completed with exit code 1.
```

https://github.com/scylladb/scylla-operator/actions/runs/6328830467/job/17188040626#step:3:3368

PR #1430, opened 2023-09-28
Fix must-gather podman command in postsubmit's 'Dump cluster state' job

rzetelskik:
Current must-gather podman command is failing on master. This PR fixes the command. It also creates the destination directory beforehand as otherwise the command fails with `Error: destination directory <dest-dir> doesn't exist`

Resolves https://github.com/scylladb/scylla-operator/issues/1429


zimnx (approved):
/lgtm

Issue #1240, opened 2023-05-10
`add node` operation fails if a node with it's name was already `replaced` and `decommissioned`


vponomaryov:
**Describe the bug**
The following `nemesis` scenario was run in the setup:
- Terminate EC2 instance that hosts the last Scylla pod
- Wait for the recreation of the K8S node by the AWS platform
- Decrease the replica number by 1 in the `scyllacluster` object configuration
- Wait for the pod deletion
- Increase the replica number  by 1 in the `scyllacluster` object configuration
- Wait for the pod readiness

So, we see following in the logs:

```
< t:2023-04-30 02:41:39,550 f:__init__.py     l:2471 c:sdcm.cluster_k8s     p:DEBUG > Replace `/spec/datacenter/racks/0/members' with `2' in sct-cluster-2's spec
...
INFO  2023-04-30 02:47:45,637 [shard 3] api - decommission
INFO  2023-04-30 02:47:45,638 [shard 0] storage_service - decommission[2557cfff-74fa-4bdb-8c2c-989da968cd69]: Started decommission operation: node=ce0091e0-f732-48f2-9038-18aaac5c8cee/172.20.114.237
INFO  2023-04-30 02:47:45,638 [shard 0] storage_service - decommission[2557cfff-74fa-4bdb-8c2c-989da968cd69]: sync_nodes=decommission, ignore_nodes=ce0091e0-f732-48f2-9038-18aaac5c8cee
INFO  2023-04-30 02:47:45,649 [shard 0] storage_service - DECOMMISSIONING: starts
...
INFO  2023-04-30 02:48:29,045 [shard 0] gossip - 60000 ms elapsed, 172.20.75.146 gossip quarantine over
INFO  2023-04-30 02:48:29,131 [shard 0] storage_service - Set host_id=5bee7f55-2980-459f-80c2-f2d93ac3a6ea to be owned by node=172.20.75.146
...
INFO  2023-04-30 02:49:38,673 [shard 0] gossip - Node 172.20.114.237 will be removed from gossip at [2023-05-03 02:49:38]
...
INFO  2023-04-30 02:50:09,259 [shard 0] storage_service - DECOMMISSIONING: done
...
< t:2023-04-30 02:50:54,895 f:__init__.py     l:2471 c:sdcm.cluster_k8s     p:DEBUG > Replace `/spec/datacenter/racks/0/members' with `3' in sct-cluster-2's spec
...
[...]

PR #1431, opened 2023-09-29
Fix logs in orphaned PV controller

rzetelskik:
**Description of your changes:** This PR fixes the logging in orphaned PV controller by:
1. Logging a manual ref for a non-existing ScyllaCluster instead of currently used `KObj`, which prints an empty string given a nil argument.
2. Changing manual refs to `KObj` where possible.
3. Logging `KObj` instead of the entire struct, which currently makes the logs very unreadable and difficult to work with.

The unreadability of the logs can be observed in any of the runs posted in https://github.com/scylladb/scylla-operator/issues/1240.


zimnx (approved):
Good catch.
/lgtm

PR #1442, opened 2023-10-03
Remove explicit copying of replace label between Service instances

rzetelskik:
**Description of your changes:** Currently the replace label is copied explicitly between Service instances. This can result in the replace label being carried over past the end of the replacement procedure in case the replace label is not removed from the original Service on time, which can in turn break the node, as observed in https://github.com/scylladb/scylla-operator/issues/1240. This PR removes the superfluous explicit copy of the replacement label. It's necessary in neither the old nor the new replacement procedure. In the old procedure, the label will be picked up from ScyllaCluster status, while the new procedure doesn't terminate services. This change should fix the issue for the new replacement procedure, as the replacement label will never be carried over to a new Service instance. Unfortunately, this PR doesn't fix the deprecated procedure.

Resolves https://github.com/scylladb/scylla-operator/issues/1240

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -73,14 +73,10 @@ func MemberService(sc *scyllav1.ScyllaCluster, rackName, name string, oldService
 	svcLabels[naming.RackNameLabel] = rackName
 	svcLabels[naming.ScyllaServiceTypeLabel] = string(naming.ScyllaServiceTypeMember)
 
-	// Copy the old replace label, if present.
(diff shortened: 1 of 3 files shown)


tnozicka (commented):
I am happy to get rid of this but given @zimnx has been around this area the most lately, I'd like him to sign the change.

/approve
/assign @zimnx

rzetelskik:
@zimnx the change required some adjustment to unit tests, please check whether the changes to the current behaviour is acceptable

zimnx (approved):
/lgtm

PR #1443, opened 2023-10-04
Enable orphaned PV e2e test on envs with multiple Nodes

### LINKED ISSUE #1303: ScyllaCluster Orphaned PV controller test needs to support more than one node
Historically, `ScyllaCluster Orphaned PV controller` is skipped when there is more then one node.
https://github.com/scylladb/scylla-operator/blame/afc059575c7d607d388b97155d6a8f4f857d327c/test/e2e/set/scyllacluster/scyllacluster_pv.go#L82-L87

This is unfortunately true even if there is only one node that it can be scheduled to but an extra infra node with taints.

We should look into removing the node length limitation or at least checking scheduling constrains as a backup.

(Otherwise this test will be skipped by our new CI.)

zimnx:
Orphaned PV E2E test was skipped when number of Nodes in the Kubernetes cluster was higher than 1. This is because test logic required spawning a Pod that must land on the same Node where particular Scylla Pods are landing.

To ensure this behavior on multiple Node environment, we set required PodAffinity on ScyllaCluster Pods, which will attract Pods to Nodes serving PVC consumer Pod which are driving the test logic.

We have a special provisioner with 'hacky' logic in the test, due to NodeAffinity on PVC being immutable. Without it, we would need to actually delete Nodes from the test to trigger logic we test, making test disruptive, serial and enforcing at least 3 Nodes in Kube cluster. Because node count requirement is resource expensive, hard to manage on developer environments I decided to stick with the provisioner.

Fixes #1303

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_pv.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_pv.go
@@ -76,22 +76,26 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", func() {
 		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
 		defer cancel()
 
-		nodes, err := f.KubeAdminClient().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
-		o.Expect(err).NotTo(o.HaveOccurred())
-
-		if len(nodes.Items) != 1 {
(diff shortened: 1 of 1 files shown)


tnozicka (commented):
/lgtm

PR #1459, opened 2023-10-10
Reenable E2E test for shard-aware ports

### LINKED ISSUE #1452: Enable testing shard aware CQL port
```
We have such test but it's skipped due to NAT on Service which prevents from determining to which shard client is trying to connect. Since we have PodIPs available we should be able to enable it preventing Service overlay NATting the traffic.
```
https://github.com/scylladb/scylla-operator/pull/1450#issuecomment-1750438223

zimnx:
It was disabled due to ClusterIP Service NATting the traffic, which prevented from shard-aware ports to assign correct shard. Today we are able configure broadcasting of ScyllaCluster to use PodIPs which doesn't NAT.

Resolves #1452

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
@@ -13,6 +13,7 @@ import (
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	"github.com/scylladb/gocqlx/v2"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllaclusterfixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/scylladb/scylla-operator/test/e2e/utils"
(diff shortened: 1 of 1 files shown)


test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -28,23 +29,29 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				},
+				Nodes: scyllav1.BroadcastOptions{
+					Type: scyllav1.BroadcastAddressTypePodIP,
+				},
+			},
+		}
 
-		// Ensure 2 shards.
+		// Ensure number of shards.
 		sc.Spec.Datacenter.Racks[0].Resources.Limits[corev1.ResourceCPU] = resource.MustParse(fmt.Sprintf("%d", nrShards))
```
rzetelskik, line 55: Shouldn't requests also be set to the same value to have 4 cores guaranteed?

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
```diff
@@ -59,8 +66,6 @@ var _ = g.Describe("ScyllaCluster", func() {
 
 		hosts := getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
 		o.Expect(hosts).To(o.HaveLen(1))
-		di := insertAndVerifyCQLData(ctx, hosts)
```
rzetelskik, line 62: why?

zimnx: it creates additional connection and writes data to cluster which makes debugging harder. CQL session required by test is created below.

test/e2e/set/scyllacluster/scyllacluster_shardawareness.go
zimnx: Setting `requests` won't guarantee it either. Number of shards is controlled by the limit of the container:
https://github.com/scylladb/scylla-operator/blob/aa89840c2f7ba428d2a3e15952c6313c38302296/pkg/controller/scyllacluster/resource.go#L436-L445

https://github.com/scylladb/scylla-operator/blob/eb144ac50175c040e486f47eb9468908181ce678/pkg/sidecar/config/config.go#L237

tnozicka (approved):
/approve
/lgtm

PR #1461, opened 2023-10-10
Add E2E test verifying TLS cert reload doesn't interrupt ongoing traffic

zimnx:
Test makes sure we or ScyllaDB doesn't interrupt ongoing user traffic when TLS certificate is reloaded.

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_tls.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_tls.go
@@ -9,8 +9,10 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/gocql/gocql"
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
@@ -18,6 +20,7 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/crypto"
 	"github.com/scylladb/scylla-operator/pkg/features"
 	"github.com/scylladb/scylla-operator/pkg/helpers"
+	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"github.com/scylladb/scylla-operator/pkg/kubecrypto"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/pointer"
@@ -31,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
@@ -110,6 +114,7 @@ var _ = g.Describe("ScyllaCluster", func() {
 				scData, err := runtime.Encode(scheme.Codecs.LegacyCodec(scyllav1.GroupVersion), sc)
 				o.Expect(err).NotTo(o.HaveOccurred())
 				// TODO: Use generated Apply method when our clients have it.
+				//       Ref: https://github.com/scylladb/scylla-operator/issues/1474
 				sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Patch(ctx, sc.Name, types.ApplyPatchType, scData, metav1.PatchOptions{
 					FieldManager: f.FieldManager(),
 					Force:        pointer.Ptr(true),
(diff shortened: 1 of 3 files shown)


test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +250,154 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+						ServerName:         host,
+						InsecureSkipVerify: false,
+						Certificates:       []tls.Certificate{adminTLSCert},
+						RootCAs:            servingCAPool,
+					},
+				}
+				session, err := gocqlx.WrapSession(clusterConfig.CreateSession())
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				di := insertAndVerifyCQLData(ctx, hosts, utils.WithSession(&session))
```
rzetelskik, line 329: could we have `WithConfig` instead of `WithSession`, and skip creating it here?

rzetelskik, line 329: Also, maybe instead of reading in the loop here, we could have a method that reads continuously in the background and takes context/stop channel as parameter? Seems like something useful in the longer run

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +250,154 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				defer di.Close()
+
+				for reads := 0; ; reads++ {
```
rzetelskik, line 332: should this use a ticker instead?

zimnx:
Requires: https://github.com/scylladb/scylla-operator-release/pull/39

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: ticker for counting iterations?

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: done

zimnx: done

test/e2e/set/scyllacluster/scyllacluster_tls.go
rzetelskik: discussed internally - there is no rate limiting here so loop is fine, resolved

test/e2e/utils/datainserter.go
```diff
@@ -157,8 +159,6 @@ func (di *DataInserter) AwaitSchemaAgreement(ctx context.Context) error {
 }
 
 func (di *DataInserter) Read() ([]*TestData, error) {
-	framework.Infof("Reading data from table %s", di.table.Name())
```
rzetelskik, line 161: Why delete it if Read() is not used in continuous reading?

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +248,124 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				},
+			}
+
+			di := insertAndVerifyCQLData(ctx, hosts, utils.WithClusterConfig(clusterConfig))
+			defer di.Close()
+
+			trafficCtx, trafficCtxCancel := context.WithCancel(ctx)
+			defer trafficCtxCancel()
+
+			stopTraffic, err := di.StartContinuousReads(trafficCtx)
```
rzetelskik, line 315: Shouldn't the read data also be verified? It could also be moved to a single helper func then, like `insertAndVerifyCQLDataContinuously`?

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +248,124 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+
+		framework.By("Adding new DNS domain to rotate certificates")
+		testDNSDomain := fmt.Sprintf("%s.scylla-operator.scylladb.com", rand.String(6))
+
+		sc.ManagedFields = nil
+		sc.ResourceVersion = ""
+		sc.Spec.DNSDomains = append(sc.Spec.DNSDomains, testDNSDomain)
+		scData, err := runtime.Encode(scheme.Codecs.LegacyCodec(scyllav1.GroupVersion), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		// TODO: Use generated Apply method when our clients have it.
```
rzetelskik, line 332: Shall we have an issue for it? I feel like these TODOs get lost in the void otherwise

test/e2e/utils/datainserter.go
```diff
@@ -173,6 +173,58 @@ func (di *DataInserter) Read() ([]*TestData, error) {
...
+			}
+
+			notifyTrafficStarted()
+		}
+	}()
+
+	select {
+	case <-time.After(10 * time.Second):
+		cancel()
+		return nil, fmt.Errorf("traffic not started before timeout")
```
rzetelskik, line 213: shouldn't this also call q.Release() and wait for the goroutine?

zimnx:
manager flake
/retest

test/e2e/utils/datainserter.go
zimnx: yep, done

test/e2e/utils/datainserter.go
zimnx: leftover from refactoring, brough it back, thanks

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: done

rzetelskik: Let's link the issue in this comment as well

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +249,128 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				Config: &tls.Config{
+					ServerName:         host,
+					InsecureSkipVerify: false,
+					Certificates:       []tls.Certificate{adminTLSCert},
+					RootCAs:            servingCAPool,
+				},
+			}
+
+			di := insertAndVerifyCQLData(ctx, hosts, utils.WithClusterConfig(clusterConfig))
+			defer di.Close()
```
tnozicka, line 313: defer in a for loop (globally)

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +249,128 @@ var _ = g.Describe("ScyllaCluster", func() {
 		}
 	})
 
+	g.It("should rotate TLS certificate without disrupting ongoing CQL traffic", func() {
```
tnozicka, line 252: can this be tested inside `should rotate TLS certificates before they expire`?

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -244,6 +249,128 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				},
+			}
+
+			di := insertAndVerifyCQLData(ctx, hosts, utils.WithClusterConfig(clusterConfig))
+			defer di.Close()
+
+			trafficCtx, trafficCtxCancel := context.WithCancel(ctx)
+			defer trafficCtxCancel()
+
+			stopTraffic, err := di.StartContinuousReads(trafficCtx)
```
tnozicka, line 318: is there any reason not to use it in any other non disruptive test as well?

test/e2e/utils/datainserter.go
```diff
@@ -174,16 +174,76 @@ func (di *DataInserter) Read() ([]*TestData, error) {
 	return res, nil
 }
 
+func (di *DataInserter) StartContinuousReads(ctx context.Context) (func() error, error) {
```
tnozicka, line 177: can we support multiple readers withing a datainserter or some higher level struct, so we don't have to pass lambdas with workgroups around?

test/e2e/set/scyllacluster/scyllacluster_tls.go
tnozicka: (if we want to move it elsewhere I think it would be helpful to use an abstraction layer that reduces the injection point to a minimum)

test/e2e/utils/datainserter.go
zimnx: what's the usecase for it and what's the benefit compared to mulitple datainserters? Test should decide when to stop, so function stopping have to be returned anyway.

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: Test isolation, should test verifying upgrade fail when tls cert reload causes traffic disruption? It's easier to triage issues when test focuses on particular behavior, not on everything around.

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: why these two should be coupled?

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: What's wrong with these? They line up on stack as expected.

test/e2e/utils/datainserter.go
```diff
@@ -74,11 +75,9 @@ func NewMultiDCDataInserter(dcHosts map[string][]string, options ...DataInserter
 		option(di)
 	}
 
-	if di.session == nil {
-		err := di.SetClientEndpoints(slices.Flatten(helpers.GetMapValues(dcHosts)))
-		if err != nil {
-			return nil, fmt.Errorf("can't set client endpoints: %w", err)
-		}
+	err := di.SetClientEndpoints(slices.Flatten(helpers.GetMapValues(dcHosts)))
```
rzetelskik, line 80: nit: can we move map helpers to a separate package (similar to pkg/helpers/slices) while we're at it? So `maps.Values` instead of `helpers.GetMapValues`, or use https://pkg.go.dev/golang.org/x/exp/maps#Values

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+	g.It("should rotate TLS certificate without disrupting ongoing CQL traffic", func() {
+		if !utilfeature.DefaultMutableFeatureGate.Enabled(features.AutomaticTLSCertificates) {
+			g.Skip(fmt.Sprintf("Skipping because %q feature is disabled", features.AutomaticTLSCertificates))
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
+
+		sc := f.GetDefaultScyllaCluster()
+		o.Expect(sc.Spec.Datacenter.Racks).To(o.HaveLen(1))
```
rzetelskik, line 277: we know what the manifest is and we don't assert this in other tests, I'd skip it

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+		sc.Spec.Datacenter.Racks[0].Members = 1
+
+		framework.By("Creating a ScyllaCluster")
+		sc, err := f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to rollout (RV=%s)", sc.ResourceVersion)
```
rzetelskik, line 284:
```suggestion
		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
```

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		verifyScyllaCluster(ctx, f.KubeClient(), sc)
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		hosts, _, err := utils.GetBroadcastRPCAddressesAndUUIDs(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(hosts).To(o.HaveLen(1))
+		host := hosts[0]
+
+		servingCABundleConfigMap, err := f.KubeClient().CoreV1().ConfigMaps(f.Namespace()).Get(ctx, fmt.Sprintf("%s-local-serving-ca", sc.Name), metav1.GetOptions{})
```
rzetelskik, line 297: nit: use naming.GetScyllaClusterLocalServingCAName

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		servingCACerts, _ := verification.VerifyAndParseCABundle(servingCABundleConfigMap)
+		o.Expect(servingCACerts).To(o.HaveLen(1))
+
+		adminClientSecret, err := f.KubeClient().CoreV1().Secrets(f.Namespace()).Get(ctx, fmt.Sprintf("%s-local-user-admin", sc.Name), metav1.GetOptions{})
```
rzetelskik, line 302: nit: use naming.GetScyllaClusterLocalUserAdminCertName

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				err := stopTraffic()
+				o.Expect(err).NotTo(o.HaveOccurred())
+			}()
+		}
+
+		framework.By("Adding new DNS domain to rotate certificates")
+		testDNSDomain := fmt.Sprintf("%s.scylla-operator.scylladb.com", rand.String(6))
+
+		sc.ManagedFields = nil
+		sc.ResourceVersion = ""
```
rzetelskik, line 349: why is this being set?

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				InsecureSkipVerify: false,
+				Certificates:       []tls.Certificate{adminTLSCert},
+				RootCAs:            servingCAPool,
+			})
+			eo.Expect(err).NotTo(o.HaveOccurred())
+			eo.Expect(serverCerts).NotTo(o.BeEmpty())
+
+			eo.Expect(serverCerts[0].DNSNames).To(o.Satisfy(func(dnsNames []string) bool {
+				return slices.Contains(dnsNames, func(dnsName string) bool {
+					return strings.HasSuffix(dnsName, testDNSDomain)
```
rzetelskik, line 375: why test for HasSuffix instead of equality?

test/e2e/set/scyllacluster/scyllacluster_tls.go
```diff
@@ -260,6 +265,130 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				})
+			}))
+		}).WithTimeout(5 * 60 * time.Second).WithPolling(1 * time.Second).Should(o.Succeed())
+		framework.Infof("Node %q reloaded certificates", host)
+
+		const trafficAfterCertUpdateDuration = 5 * time.Second
+		framework.By("Waiting %s before killing loaders", trafficAfterCertUpdateDuration)
```
rzetelskik, line 382: nit: add an explicit comment about why this is necessary

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: This test only relies on fact of having single node cluster, hence it's asserted.

test/e2e/utils/datainserter.go
zimnx: maybe not as part of this PR. It uses what existing codebase has.

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: reusing functions generating names in tests is error prone.

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: ditto

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: to generate nicer patch

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: because cert is also signed for individual nodes. We need to detect cert change, so when test domain is added, it means it changed.

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: There're 29 other messages like this one. Leaving as is for consistency.

test/e2e/set/scyllacluster/scyllacluster_tls.go
zimnx: added

zimnx:
```
   STEP: Waiting for ScyllaCluster to sync repair task deletion with Scylla Manager @ 04/16/24 19:10:20.409
  [FAILED] in [It] - github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:211 @ 04/16/24 19:13:20.413 
  [...]
  "timed out waiting for the condition",
  ```
Known Manager flake #1694 
/retest

test/e2e/utils/datainserter.go
rzetelskik: ok, doesn't have to be here

test/e2e/set/scyllacluster/scyllacluster_tls.go
rzetelskik: what's the value in making mistakes consistent? We have to start somewhere

zimnx: Feel free to open PR fixing it. This one is not about it.

test/e2e/set/scyllacluster/scyllacluster_tls.go
tnozicka: The defer scope should be around the variable it uses, this way it piles up on the stack in the whole test function and also makes Goland angry.

test/e2e/set/scyllacluster/scyllacluster_tls.go
tnozicka: each scyllacluster takes a lot of time to rollout and the certs are already available in the other test so this can fluently continue where the setup test has finished and save resources + execution time for the suite

test/e2e/set/scyllacluster/scyllacluster_tls.go
tnozicka: We have to make sure all of our operations are HA by default. That also means we'd have to double the number of tests. I think this should run by default in every test and could be checked at the end of the test so it can be clearly distinguished. But no non-disruptive test should be allowed to break HA invariant.

test/e2e/set/scyllacluster/scyllacluster_tls.go
tnozicka: I don't think previous misspellings are a good reason to grow the set, unless that's part of some related parsing.

test/e2e/set/scyllacluster/scyllacluster_tls.go
tnozicka: there is also an option to use global invariants checkers and generate a synthetic ginko failure

PR #1470, opened 2023-10-12
Bump Go version to 1.21

zimnx:
Scylla Operator is built with Go 1.21.
Minimal required version for building is Go 1.21.


tnozicka (approved):
/approve
/lgtm

PR #1472, opened 2023-10-13
Use the same tag version of gen-release-notes when preparing them

zimnx:
Discrepancy between dependencies versions may cause compilation failures


tnozicka (commented):
/approve
/lgtm

PR #1493, opened 2023-10-19
Set ScyllaCluster placement requirements on cleanup Jobs

### LINKED ISSUE #1490: Cleanup jobs cannot be scheduled on tainted nodes
**Describe the bug**

When all nodes in cluster are tainted, cleanup jobs cannot run anywhere due to missing toleration.

**To Reproduce**
1. Taint all nodes in cluster, specify toleration for some taint in ScyllaCluster object
2. Create ScyllaCluster object
3. Cleanup Jobs are created but stuck in Pending state forever.

**Expected behavior**
Cleanup Jobs run on nodes tolerated by ScyllaCluster.

zimnx:
Missing placement might cause cleanup Jobs to be stuck in Pending state forever, preventing ScyllaCluster from fully boostrapping. Placement requirements are copied into cleanup Jobs from ScyllaCluster spec.

Resolves #1490

### DIFF
--- a/pkg/controller/scyllacluster/resource_test.go
+++ b/pkg/controller/scyllacluster/resource_test.go
@@ -1876,6 +1876,382 @@ func TestMakeJobs(t *testing.T) {
 			},
 			expectedConditions: nil,
 		},
+		{
+			name: "cleanup job has the same placement requirements as ScyllaCluster",
+			cluster: func() *scyllav1.ScyllaCluster {
+				cluster := basicScyllaCluster.DeepCopy()
+				cluster.Spec.Datacenter.Racks[0].Placement = &scyllav1.PlacementSpec{
+					NodeAffinity: &corev1.NodeAffinity{
+						RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+							NodeSelectorTerms: []corev1.NodeSelectorTerm{
+								{
+									MatchExpressions: []corev1.NodeSelectorRequirement{
+										{
+											Key:      "topology.kubernetes.io/zone",
+											Operator: corev1.NodeSelectorOpIn,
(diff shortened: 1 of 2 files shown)


pkg/controller/scyllacluster/resource_test.go
```diff
@@ -1876,6 +1876,194 @@ func TestMakeJobs(t *testing.T) {
 			},
 			expectedConditions: nil,
 		},
+		{
+			name: "cleanup job has the same placement requirements as ScyllaCluster",
+			cluster: func() *scyllav1.ScyllaCluster {
+				cluster := basicScyllaCluster.DeepCopy()
+				cluster.Spec.Datacenter.Racks[0].Placement = &scyllav1.PlacementSpec{
```
tnozicka, line 1883: should we have 2 racks to make sure we assign the correct, corresponding placement options?

zimnx: Good idea, added.

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -1876,6 +1876,380 @@ func TestMakeJobs(t *testing.T) {
...
+								},
+							},
+							Tolerations: []corev1.Toleration{
+								{
+									Key:      "role",
+									Operator: corev1.TolerationOpEqual,
+									Value:    "scylla-clusters",
+									Effect:   corev1.TaintEffectNoSchedule,
+								},
+							},
```
tnozicka, line 1988: please make the different from rack[0] as well

zimnx: done

tnozicka (approved):
/approve
/lgtm

thanks!

PR #1507, opened 2023-10-24
Fix preview target in docs readme

tnozicka:
`multiversionpreview` doesn't actually build the docs on its own so the instructions don't work on a clean repository + it doesn't make sense to checkout the other branches. We should use the `preview` target that builds the docs and live reloads them.


rzetelskik (approved):
/lgtm
thanks!

Issue #1415, opened 2023-09-25
Update scylladb/gocql dependency


tnozicka:
We have had to rollback scylladb/gocql because of https://github.com/scylladb/gocql/issues/145 in https://github.com/scylladb/scylla-operator/pull/1408 . When https://github.com/scylladb/gocql/issues/145 is fixed we should update the dependency.

PR #1508, opened 2023-10-24
Make E2E suite expose ScyllaClusters on PodIPs by default

### LINKED ISSUE #1416: Switch e2e suite to use PodIPs
E2E suite should by default run cluster using PodIPs, although the option should be configurable with a flag so we can run it in different configuration in a periodic.

rzetelskik:
**Description of your changes:** This PR makes expose options configurable in Scylla Operator's test runner, adds support for exposing ScyllaClusters (both broadcast addresses and broadcast rpc addresses) over PodIP in E2E test suite and makes it a default setup.

Resolves #1416 

/kind feature
/priority important-soon

/hold
for https://github.com/scylladb/scylla-operator/issues/1415, gocql bump is required to pass CI

### DIFF
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -16,6 +16,21 @@ if [ -z ${SO_IMAGE+x} ]; then
   exit 2
 fi
 
+if [ -z ${SO_SCYLLACLUSTER_NODE_SERVICE_TYPE+x} ]; then
+  echo "SO_SCYLLACLUSTER_NODE_SERVICE_TYPE can't be empty" > /dev/stderr
+  exit 2
+fi
+
+if [ -z ${SO_SCYLLACLUSTER_NODES_BROADCAST_ADDRESS_TYPE+x} ]; then
+  echo "SO_SCYLLACLUSTER_NODES_BROADCAST_ADDRESS_TYPE can't be empty" > /dev/stderr
+  exit 2
+fi
+
+if [ -z ${SO_SCYLLACLUSTER_CLIENTS_BROADCAST_ADDRESS_TYPE+x} ]; then
+  echo "SO_SCYLLACLUSTER_CLIENTS_BROADCAST_ADDRESS_TYPE can't be empty" > /dev/stderr
+  exit 2
+fi
+
 if [ -z ${ARTIFACTS+x} ]; then
   echo "ARTIFACTS can't be empty" > /dev/stderr
   exit 2
@@ -156,7 +171,7 @@ ingress_controller_address="$( kubectl -n haproxy-ingress get svc haproxy-ingres
 
 kubectl create -n e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o yaml | kubectl_create -f -
 
-kubectl -n e2e run --restart=Never --image="${SO_IMAGE}" --labels='app=e2e' --command=true e2e -- bash -euExo pipefail -O inherit_errexit -c "function wait-for-artifacts { touch /tmp/done && until [[ -f '/tmp/exit' ]]; do sleep 1; done } && trap wait-for-artifacts EXIT && mkdir /tmp/artifacts && scylla-operator-tests run '${SO_SUITE}' --loglevel=2 --color=false --artifacts-dir=/tmp/artifacts --feature-gates='${SCYLLA_OPERATOR_FEATURE_GATES}' --ingress-controller-address='${ingress_controller_address}' --ingress-controller-ingress-class-name='${ingress_class_name}' --ingress-controller-custom-annotations='${ingress_custom_annotations}'"
+kubectl -n e2e run --restart=Never --image="${SO_IMAGE}" --labels='app=e2e' --command=true e2e -- bash -euExo pipefail -O inherit_errexit -c "function wait-for-artifacts { touch /tmp/done && until [[ -f '/tmp/exit' ]]; do sleep 1; done } && trap wait-for-artifacts EXIT && mkdir /tmp/artifacts && scylla-operator-tests run '${SO_SUITE}' --loglevel=2 --color=false --artifacts-dir=/tmp/artifacts --feature-gates='${SCYLLA_OPERATOR_FEATURE_GATES}' --ingress-controller-address='${ingress_controller_address}' --ingress-controller-ingress-class-name='${ingress_class_name}' --ingress-controller-custom-annotations='${ingress_custom_annotations}' --scyllacluster-node-service-type='${SO_SCYLLACLUSTER_NODE_SERVICE_TYPE}' --scyllacluster-nodes-broadcast-address-type='${SO_SCYLLACLUSTER_NODES_BROADCAST_ADDRESS_TYPE}' --scyllacluster-clients-broadcast-address-type='${SO_SCYLLACLUSTER_CLIENTS_BROADCAST_ADDRESS_TYPE}'"
 kubectl -n e2e wait --for=condition=Ready pod/e2e
 
 # Setup artifacts transfer when finished and unblock the e2e pod when done.
--- a/pkg/cmd/tests/options.go
+++ b/pkg/cmd/tests/options.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/spf13/cobra"
 	apierrors "k8s.io/apimachinery/pkg/util/errors"
@@ -15,11 +17,29 @@ type IngressControllerOptions struct {
 	CustomAnnotations map[string]string
 }
 
+type ScyllaClusterOptions struct {
+	NodeServiceType             string
+	NodesBroadcastAddressType   string
+	ClientsBroadcastAddressType string
+}
+
+var supportedNodeServiceTypes = []scyllav1.NodeServiceType{
+	scyllav1.NodeServiceTypeHeadless,
+	scyllav1.NodeServiceTypeClusterIP,
+}
+
+var supportedBroadcastAddressTypes = []scyllav1.BroadcastAddressType{
+	scyllav1.BroadcastAddressTypePodIP,
+	scyllav1.BroadcastAddressTypeServiceClusterIP,
+}
+
 type TestFrameworkOptions struct {
 	ArtifactsDir                 string
 	DeleteTestingNSPolicyUntyped string
 	DeleteTestingNSPolicy        framework.DeleteTestingNSPolicyType
 	IngressController            *IngressControllerOptions
+	ScyllaClusterOptionsUntyped  *ScyllaClusterOptions
+	scyllaClusterOptions         *framework.ScyllaClusterOptions
 }
 
 func NewTestFrameworkOptions() TestFrameworkOptions {
@@ -27,6 +47,11 @@ func NewTestFrameworkOptions() TestFrameworkOptions {
 		ArtifactsDir:                 "",
 		DeleteTestingNSPolicyUntyped: string(framework.DeleteTestingNSPolicyAlways),
 		IngressController:            &IngressControllerOptions{},
+		ScyllaClusterOptionsUntyped: &ScyllaClusterOptions{
+			NodeServiceType:             string(scyllav1.NodeServiceTypeHeadless),
+			NodesBroadcastAddressType:   string(scyllav1.BroadcastAddressTypePodIP),
+			ClientsBroadcastAddressType: string(scyllav1.BroadcastAddressTypePodIP),
+		},
 	}
 }
 
@@ -43,6 +68,18 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(&o.IngressController.Address, "ingress-controller-address", "", o.IngressController.Address, "Overrides destination address when sending testing data to applications behind ingresses.")
 	cmd.PersistentFlags().StringVarP(&o.IngressController.IngressClassName, "ingress-controller-ingress-class-name", "", o.IngressController.IngressClassName, "Ingress class name under which ingress controller is registered")
 	cmd.PersistentFlags().StringToStringVarP(&o.IngressController.CustomAnnotations, "ingress-controller-custom-annotations", "", o.IngressController.CustomAnnotations, "Custom annotations required by the ingress controller")
+	cmd.PersistentFlags().StringVarP(&o.ScyllaClusterOptionsUntyped.NodeServiceType, "scyllacluster-node-service-type", "", o.ScyllaClusterOptionsUntyped.NodeServiceType, fmt.Sprintf("Kubernetes service type that the ScyllaCluster nodes are exposed with. Allowed values are [%s].", strings.Join(
+		slices.ConvertSlice(supportedNodeServiceTypes, slices.ToString[scyllav1.NodeServiceType]),
+		", ",
+	)))
+	cmd.PersistentFlags().StringVarP(&o.ScyllaClusterOptionsUntyped.NodesBroadcastAddressType, "scyllacluster-nodes-broadcast-address-type", "", o.ScyllaClusterOptionsUntyped.NodesBroadcastAddressType, fmt.Sprintf("Type of address that the ScyllaCluster nodes broadcast for communication with other nodes. Allowed values are [%s].", strings.Join(
+		slices.ConvertSlice(supportedBroadcastAddressTypes, slices.ToString[scyllav1.BroadcastAddressType]),
+		", ",
+	)))
+	cmd.PersistentFlags().StringVarP(&o.ScyllaClusterOptionsUntyped.ClientsBroadcastAddressType, "scyllacluster-clients-broadcast-address-type", "", o.ScyllaClusterOptionsUntyped.ClientsBroadcastAddressType, fmt.Sprintf("Type of address that the ScyllaCluster nodes broadcast for communication with clients. Allowed values are [%s].", strings.Join(
+		slices.ConvertSlice(supportedBroadcastAddressTypes, slices.ToString[scyllav1.BroadcastAddressType]),
+		", ",
+	)))
 }
 
 func (o *TestFrameworkOptions) Validate() error {
@@ -56,6 +93,18 @@ func (o *TestFrameworkOptions) Validate() error {
 		errors = append(errors, fmt.Errorf("invalid DeleteTestingNSPolicy: %q", p))
 	}
 
+	if !slices.ContainsItem(supportedNodeServiceTypes, scyllav1.NodeServiceType(o.ScyllaClusterOptionsUntyped.NodeServiceType)) {
+		errors = append(errors, fmt.Errorf("invalid scylla-cluster-node-service-type: %q", o.ScyllaClusterOptionsUntyped.NodeServiceType))
+	}
+
+	if !slices.ContainsItem(supportedBroadcastAddressTypes, scyllav1.BroadcastAddressType(o.ScyllaClusterOptionsUntyped.NodesBroadcastAddressType)) {
+		errors = append(errors, fmt.Errorf("invalid scylla-cluster-nodes-broadcast-address-type: %q", o.ScyllaClusterOptionsUntyped.NodesBroadcastAddressType))
+	}
+
+	if !slices.ContainsItem(supportedBroadcastAddressTypes, scyllav1.BroadcastAddressType(o.ScyllaClusterOptionsUntyped.ClientsBroadcastAddressType)) {
+		errors = append(errors, fmt.Errorf("invalid scylla-cluster-clients-broadcast-address-type: %q", o.ScyllaClusterOptionsUntyped.ClientsBroadcastAddressType))
+	}
+
 	return apierrors.NewAggregate(errors)
 }
 
@@ -64,6 +113,14 @@ func (o *TestFrameworkOptions) Complete() error {
 
 	// Trim spaces so we can reason later if the dir is set or not
 	o.ArtifactsDir = strings.TrimSpace(o.ArtifactsDir)
+
+	o.scyllaClusterOptions = &framework.ScyllaClusterOptions{
+		ExposeOptions: framework.ExposeOptions{
+			NodeServiceType:             scyllav1.NodeServiceType(o.ScyllaClusterOptionsUntyped.NodeServiceType),
+			NodesBroadcastAddressType:   scyllav1.BroadcastAddressType(o.ScyllaClusterOptionsUntyped.NodesBroadcastAddressType),
+			ClientsBroadcastAddressType: scyllav1.BroadcastAddressType(o.ScyllaClusterOptionsUntyped.ClientsBroadcastAddressType),
+		},
+	}
 
 	return nil
 }
--- a/pkg/cmd/tests/tests_run.go
+++ b/pkg/cmd/tests/tests_run.go
@@ -261,6 +261,7 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
 		RestConfig:            o.RestConfig,
 		ArtifactsDir:          o.ArtifactsDir,
 		DeleteTestingNSPolicy: o.DeleteTestingNSPolicy,
+		ScyllaClusterOptions:  o.scyllaClusterOptions,
 	}
 	if o.IngressController != nil {
 		framework.TestContext.IngressController = &framework.IngressController{
--- a/test/e2e/fixture/scylla/scyllacluster.yaml.tmpl
+++ b/test/e2e/fixture/scylla/scyllacluster.yaml.tmpl
@@ -6,6 +6,14 @@ spec:
   agentVersion: 3.2.5
   version: 5.4.0
   developerMode: true
+  exposeOptions:
+    nodeService:
+      type: {{ .nodeServiceType }}
+    broadcastOptions:
+      nodes:
+        type: {{ .nodesBroadcastAddressType }}
+      clients:
+        type: {{ .clientsBroadcastAddressType }}
   datacenter:
     name: us-east-1
     racks:
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -14,7 +14,9 @@ import (
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllaclientset "github.com/scylladb/scylla-operator/pkg/client/scylla/clientset/versioned"
+	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -141,6 +143,19 @@ func (f *Framework) CommonLabels() map[string]string {
 		"e2e":       "scylla-operator",
 		"framework": f.name,
 	}
+}
+
+func (f *Framework) GetDefaultScyllaCluster() *scyllav1.ScyllaCluster {
+	renderArgs := map[string]any{
+		"nodeServiceType":             TestContext.ScyllaClusterOptions.ExposeOptions.NodeServiceType,
+		"nodesBroadcastAddressType":   TestContext.ScyllaClusterOptions.ExposeOptions.NodesBroadcastAddressType,
+		"clientsBroadcastAddressType": TestContext.ScyllaClusterOptions.ExposeOptions.ClientsBroadcastAddressType,
+	}
+
+	sc, _, err := scyllafixture.ScyllaClusterTemplate.RenderObject(renderArgs)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	return sc
 }
 
 func (f *Framework) setupNamespace(ctx context.Context) {
--- a/test/e2e/framework/meta.go
+++ b/test/e2e/framework/meta.go
@@ -7,7 +7,8 @@ import (
 )
 
 const (
-	SerialLabelName = "Serial"
+	SerialLabelName            = "Serial"
+	RequiresClusterIPLabelName = "RequiresClusterIP"
 )
 
 var (
@@ -15,4 +16,5 @@ var (
 		g.Serial,
 		g.Label(SerialLabelName),
 	}
+	RequiresClusterIP = g.Label(RequiresClusterIPLabelName)
 )
--- a/test/e2e/framework/testcontext.go
+++ b/test/e2e/framework/testcontext.go
@@ -3,6 +3,7 @@
 package framework
 
 import (
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	restclient "k8s.io/client-go/rest"
 )
 
@@ -22,9 +23,20 @@ type IngressController struct {
 	CustomAnnotations map[string]string
 }
 
+type ScyllaClusterOptions struct {
+	ExposeOptions ExposeOptions
+}
+
+type ExposeOptions struct {
+	NodeServiceType             scyllav1.NodeServiceType
+	NodesBroadcastAddressType   scyllav1.BroadcastAddressType
+	ClientsBroadcastAddressType scyllav1.BroadcastAddressType
+}
+
 type TestContextType struct {
 	RestConfig            *restclient.Config
 	ArtifactsDir          string
 	DeleteTestingNSPolicy DeleteTestingNSPolicyType
 	IngressController     *IngressController
+	ScyllaClusterOptions  *ScyllaClusterOptions
 }
--- a/test/e2e/set/scyllacluster/scyllacluster_auth.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_auth.go
@@ -12,7 +12,6 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/scyllaclient"
-	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/scylladb/scylla-operator/test/e2e/utils"
 	"gopkg.in/yaml.v2"
@@ -30,7 +29,7 @@ var _ = g.Describe("ScyllaCluster authentication", func() {
 		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
 		defer cancel()
 
-		sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc := f.GetDefaultScyllaCluster()
 		sc.Spec.Datacenter.Racks[0].Members = 1
 
 		framework.By("Creating a ScyllaCluster")
@@ -44,7 +43,10 @@ var _ = g.Describe("ScyllaCluster authentication", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
-		hosts := getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		hosts, hostIDs, err := utils.GetBroadcastRPCAddressesAndUUIDs(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(hosts).To(o.HaveLen(1))
 		di := insertAndVerifyCQLData(ctx, hosts)
 		defer di.Close()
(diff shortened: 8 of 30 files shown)


test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -117,7 +119,18 @@ var _ = g.Describe("ScyllaCluster authentication", func() {
...
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
-		o.Expect(hosts).To(o.ConsistOf(getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)))
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		oldHosts := hosts
+		oldHostIDs := hostIDs
+		hosts, hostIDs, err = utils.GetBroadcastRPCAddressesAndUUIDs(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(hosts).To(o.HaveLen(len(oldHosts)))
+		o.Expect(hostIDs).To(o.ConsistOf(oldHostIDs))
```
tnozicka, line 129: won't this be different for PodIPs and generally depending on the exposure type?
I think we don't need the check here, just the test that explicitly replace nodes

test/e2e/set/scyllacluster/scyllacluster_evictions.go
```diff
@@ -37,7 +36,10 @@ var _ = g.Describe("ScyllaCluster evictions", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
-		hosts := getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		hosts, err := utils.GetBroadcastRPCAddresses(ctx, f.KubeClient().CoreV1(), sc)
```
tnozicka, line 41: I wonder (in general) whether we shouldn't keep a client open and test it works / updates the hostid during the rolling restart). real workloads will refresh hosts from scratch only on client restarts

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -72,13 +73,32 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+					Nodes: scyllav1.BroadcastOptions{
+						Type: scyllav1.BroadcastAddressTypeServiceClusterIP,
+					},
+					Clients: scyllav1.BroadcastOptions{
+						Type: scyllav1.BroadcastAddressTypeServiceClusterIP,
+					},
+				},
+			}
+		}
+
```
tnozicka, line 99: if we ask the test suite to be run with a scylla version and some exposure type, this should better fail or skip, but probably fail

tnozicka (commented):
let's wait for the updated drivers, I don't think some of the steps should be necessary

pkg/cmd/tests/tests_run.go
```diff
@@ -270,6 +272,45 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+			},
+		}
+	default:
+	}
+
+	sc, _, err := scyllafixture.ScyllaClusterTemplate.RenderObject(renderArgs)
+	if err != nil {
+		return fmt.Errorf("can't render ScyllaClusterTemplate: %w", err)
+	}
+	framework.TestContext.DefaultScyllaCluster = sc
```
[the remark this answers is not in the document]
rzetelskik: I was trying to do so but couldn't really figure out a nice way to initialize it once and avoid rendering the template for every framework instance

test/e2e/framework/framework.go
```diff
@@ -143,6 +144,10 @@ func (f *Framework) CommonLabels() map[string]string {
 	}
 }
 
+func (f *Framework) GetDefaultScyllaCluster() *scyllav1.ScyllaCluster {
+	return TestContext.DefaultScyllaCluster.DeepCopy()
```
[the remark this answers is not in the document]
rzetelskik: Why would you do it this way? I wanted to avoid rendering it more than once for the entire suite (https://github.com/scylladb/scylla-operator/pull/1508/files/8031e49a51c0e3ef31c83e88fa38730d9e2ffb5f#r1372055888).

test/e2e/set/scyllacluster/scyllacluster_auth.go
rzetelskik: > won't this be different for PodIPs and generally depending on the exposure type?

It shouldn't influence the UUIDs of the nodes. This was done in an effort to avoid depending on the exposure type.

> I think we don't need the check here, just the test that explicitly replace nodes

I tried not to reduce coverage and this (and almost every other) test checks that the hosts didn't change after a rollout.

test/e2e/set/scyllacluster/scyllacluster_evictions.go
rzetelskik: I agree, and I removed the code reestablishing the session (`di.SetClientEndpoints`) wherever possible, so basically any test running more than one node. This tests only inserts and verifies data once, but many tests have extra checks `verifyCQLData(ctx, di)` without resetting the connection through their duration.

test/e2e/set/scyllacluster/scyllacluster_replace.go
rzetelskik: I get the skip, but why would you want to fail it? Do you mean we should *expect* the test to fail for particular setups?

pkg/cmd/tests/tests_run.go
```diff
@@ -270,6 +272,45 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+			},
+			BroadcastOptions: &scyllav1.NodeBroadcastOptions{
+				Nodes: scyllav1.BroadcastOptions{
+					Type: scyllav1.BroadcastAddressTypePodIP,
+				},
+				Clients: scyllav1.BroadcastOptions{
+					Type: scyllav1.BroadcastAddressTypePodIP,
+				},
+			},
+		}
```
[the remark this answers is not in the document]
rzetelskik: I suppose you mean nodeService.Type, broadcastOptions.Nodes.Type and broadcastOptions.Clients.Type all being independently configurable by flags?

pkg/cmd/tests/tests_run.go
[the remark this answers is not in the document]
tnozicka: > but couldn't really figure out a nice way to initialize it once and avoid rendering the template for every framework instance

generally a singleton pattern or https://pkg.go.dev/sync#example-Once
that said of the few renders is totally negligible to any code complexity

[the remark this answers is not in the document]
rzetelskik: I thought a singleton would be a bit of an unnecessary overhead compared to just doing it there, but maybe that's better for separation of concerns. I'll give it a try

test/e2e/framework/framework.go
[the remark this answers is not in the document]
rzetelskik: copy of https://github.com/scylladb/scylla-operator/pull/1508#discussion_r1371907162, let's resolve there

pkg/cmd/tests/options.go
```diff
@@ -15,18 +15,28 @@ type IngressControllerOptions struct {
...
-	ArtifactsDir                 string
-	DeleteTestingNSPolicyUntyped string
-	DeleteTestingNSPolicy        framework.DeleteTestingNSPolicyType
-	IngressController            *IngressControllerOptions
+	ArtifactsDir                   string
+	DeleteTestingNSPolicyUntyped   string
+	DeleteTestingNSPolicy          framework.DeleteTestingNSPolicyType
+	IngressController              *IngressControllerOptions
+	ScyllaDBClusterExposureUntyped string
+	ScyllaDBClusterExposure        ScyllaDBClusterExposureType
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/cmd/tests/options.go
```diff
@@ -43,6 +53,13 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(&o.IngressController.Address, "ingress-controller-address", "", o.IngressController.Address, "Overrides destination address when sending testing data to applications behind ingresses.")
 	cmd.PersistentFlags().StringVarP(&o.IngressController.IngressClassName, "ingress-controller-ingress-class-name", "", o.IngressController.IngressClassName, "Ingress class name under which ingress controller is registered")
 	cmd.PersistentFlags().StringToStringVarP(&o.IngressController.CustomAnnotations, "ingress-controller-custom-annotations", "", o.IngressController.CustomAnnotations, "Custom annotations required by the ingress controller")
+	cmd.PersistentFlags().StringVarP(&o.ScyllaDBClusterExposureUntyped, "scylladb-cluster-exposure", "", o.ScyllaDBClusterExposureUntyped, fmt.Sprintf("ScyllaCluster exposure type. Allowed values are [%s].", strings.Join(
+		[]string{
+			string(ScyllaDBClusterExposureTypePodIP),
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/cmd/tests/tests_run.go
```diff
@@ -270,6 +272,45 @@ func (o *RunOptions) run(ctx context.Context, streams genericclioptions.IOStream
...
+			BroadcastOptions: &scyllav1.NodeBroadcastOptions{
+				Nodes: scyllav1.BroadcastOptions{
+					Type: scyllav1.BroadcastAddressTypeServiceClusterIP,
+				},
+				Clients: scyllav1.BroadcastOptions{
+					Type: scyllav1.BroadcastAddressTypeServiceClusterIP,
+				},
+			},
+		}
+	default:
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/cmd/tests/tests_run.go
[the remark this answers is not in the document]
rzetelskik: Moved to suggested place and changed validation a bit

pkg/helpers/networking.go
```diff
@@ -5,7 +5,7 @@ import (
 	"net"
 )
 
-func ParseClusterIP(s string) (net.IP, error) {
+func ParseIP(s string) (net.IP, error) {
```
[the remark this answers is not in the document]
rzetelskik: https://github.com/scylladb/scylla-operator/pull/1524

test/e2e/utils/helpers.go
```diff
@@ -436,41 +437,134 @@ func GetScyllaConfigClient(ctx context.Context, client corev1client.CoreV1Interf
...
+	for _, svc := range serviceList.Items {
+		broadcastRPCAddress, err := GetBroadcastRPCAddress(ctx, client, sc, &svc)
 		if err != nil {
-			return nil, nil, fmt.Errorf("can't create scylla config client with host %q: %w", host, err)
+			return nil, nil, fmt.Errorf("can't get broadcast rpc address of Service %q: %w", naming.ObjRef(&svc), err)
 		}
-		clientBroadcastedAddress, err := configClient.BroadcastRPCAddress(ctx)
+
+		broadcastRPCAddresses = append(broadcastRPCAddresses, broadcastRPCAddress)
+		uuids = append(uuids, svc.Annotations[naming.HostIDAnnotation])
```
zimnx, line 540: ditto

pkg/cmd/tests/tests_run.go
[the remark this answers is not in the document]
zimnx: +1, current options are very limited compared to how flexible API is

test/e2e/utils/helpers.go
```diff
@@ -436,41 +437,134 @@ func GetScyllaConfigClient(ctx context.Context, client corev1client.CoreV1Interf
...
+	serviceList, err := client.Services(sc.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: GetMemberServiceSelector(sc.Name).String(),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return slices.ConvertSlice(serviceList.Items, func(svc corev1.Service) string {
+		return svc.Annotations[naming.HostIDAnnotation]
+	}), nil
```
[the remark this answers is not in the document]
rzetelskik: There's a dedicated e2e testing this behaviour - I think we can trust it in the other tests.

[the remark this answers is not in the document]
zimnx: When feature of storing HostID in annotation breaks, only e2e testing this feature should fail. When tests doesn't rely on other tests it makes triaging much easier.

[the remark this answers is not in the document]
rzetelskik: Generally I disagree with the notion that the other tests should not use that feature because some other test is using it - it's supposed to isolate it, not be the only one using it. I do agree with easier bug tracking though and it's simple to implement it with scylla client anyway, so I changed it.

pkg/cmd/tests/tests_run.go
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/utils/helpers.go
rzetelskik: done

rzetelskik:
https://github.com/scylladb/scylla-operator/pull/1528 merged
/hold cancel

zimnx (approved):
lgtm, thanks
/assign tnozicka

rzetelskik:
@tnozicka ping

test/e2e/set/scyllacluster/scyllacluster_pv.go
```diff
@@ -348,13 +350,8 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", func() {
...
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
 
-		oldHosts := hosts
-		hosts = getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
-		o.Expect(hosts).To(o.HaveLen(len(oldHosts)))
-		o.Expect(hosts).To(o.ConsistOf(oldHosts))
-		err = di.SetClientEndpoints(hosts)
-		o.Expect(err).NotTo(o.HaveOccurred())
```
tnozicka, line 357: should we check the UUIDs?

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -72,13 +73,32 @@ var _ = g.Describe("ScyllaCluster", func() {
...
-		sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc := f.GetDefaultScyllaCluster()
 		sc.Spec.Repository = e.scyllaImageRepository
 		sc.Spec.Version = e.scyllaVersion
 		sc.Spec.Datacenter.Racks[0].Members = 3
 
+		supportsExposing, err := scyllafeatures.Supports(sc, scyllafeatures.ExposingScyllaClusterViaServiceOtherThanClusterIP)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		if !supportsExposing {
```
tnozicka, line 84: why here and not when generating the default cluster? ExposingScyllaClusterViaServiceOtherThanClusterIP seems to be a generic issue with configuration

tnozicka: or should this be `ReplacingNodeUsingHostID`?

test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -42,8 +41,12 @@ var _ = g.Describe("ScyllaCluster", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
-		hosts := getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		hosts, hostIDs, err := utils.GetBroadcastRPCAddressesAndUUIDs(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(hosts).To(o.HaveLen(1))
+		o.Expect(hostIDs).To(o.HaveLen(1))
```
tnozicka, line 49: tiny nit: I'd technically rather assert `hostIDs` and `hosts` length for equality (globally)

test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -173,12 +186,17 @@ var _ = g.Describe("ScyllaCluster", func() {
...
-		hosts = getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+		oldHostIDs = hostIDs
+		hosts, hostIDs, err = utils.GetBroadcastRPCAddressesAndUUIDs(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(oldHosts).To(o.HaveLen(2))
+		o.Expect(oldHostIDs).To(o.HaveLen(2))
 		o.Expect(hosts).To(o.HaveLen(1))
-		o.Expect(oldHosts).To(o.ContainElements(hosts))
+		o.Expect(hostIDs).To(o.HaveLen(1))
+		o.Expect(oldHostIDs).To(o.ContainElements(hostIDs))
```
tnozicka, line 199: shouldn't this be `ConsistOf` so it implies the sets should be equal in random order? I think `ContainElements` means that the `oldHostIDs` could have extra items.
(globally)

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
```diff
@@ -37,9 +37,28 @@ var _ = g.Describe("ScyllaCluster upgrades", func() {
 			ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
 			defer cancel()
 
-			sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+			sc := f.GetDefaultScyllaCluster()
 			sc.Spec.Version = e.initialVersion
 
+			supportsExposing, err := scyllafeatures.Supports(sc, scyllafeatures.ExposingScyllaClusterViaServiceOtherThanClusterIP)
```
tnozicka, line 43: why does this relates to upgrades in particular?

test/e2e/utils/helpers.go
```diff
@@ -436,41 +436,139 @@ func GetScyllaConfigClient(ctx context.Context, client corev1client.CoreV1Interf
...
+
+func GetBroadcastAddress(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster, svc *corev1.Service) (string, error) {
+	host := svc.Spec.ClusterIP
+
+	if host == corev1.ClusterIPNone {
+		pod, err := client.Pods(sc.Namespace).Get(ctx, svc.Name, metav1.GetOptions{})
+		if err != nil {
+			return "", fmt.Errorf("can't get pod %q: %w", naming.ManualRef(sc.Namespace, svc.Name), err)
 		}
+		host = pod.Status.PodIP
```
tnozicka, line 468: should this check that it's not empty? + that pod isn't pending, terminated or restarting

test/e2e/utils/helpers.go
```diff
@@ -436,41 +436,139 @@ func GetScyllaConfigClient(ctx context.Context, client corev1client.CoreV1Interf
...
+	}
+
+	return broadcastAddress, nil
+}
+
+func GetUUIDs(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster) ([]string, error) {
+	_, uuids, err := GetBroadcastRPCAddressesAndUUIDs(ctx, client, sc)
+
+	return uuids, err
+}
```
tnozicka, line 487: I'd personally drop this as you can always ignore the return var on the caller side and it takes the same args but I don't feels strong

test/e2e/utils/helpers.go
```diff
@@ -436,41 +436,139 @@ func GetScyllaConfigClient(ctx context.Context, client corev1client.CoreV1Interf
...
+
+func GetBroadcastRPCAddresses(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster) ([]string, error) {
+	serviceList, err := client.Services(sc.Namespace).List(ctx, metav1.ListOptions{
```
tnozicka, line 490: maybe this is where it should branch based on which exposing type is chosen. you don't needs services to read pods and technically it's controlled by expose option config, not whether service has clusterip

test/e2e/set/scyllacluster/scyllacluster_scaling.go
rzetelskik: They won't be equal - the cluster was just scaled down (2->1).

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
rzetelskik: We only have one minor version of Scylla OS which is capable of exposing nodes on PodIPs, so we can't run the upgrade test with PodIPs, and have to modify the default cluster accordingly (which I assume is better than skipping the test).

pkg/cmd/tests/options.go
```diff
@@ -15,18 +17,41 @@ type IngressControllerOptions struct {
...
+	scyllav1.BroadcastAddressTypePodIP,
+	scyllav1.BroadcastAddressTypeServiceClusterIP,
+}
+
 type TestFrameworkOptions struct {
 	ArtifactsDir                 string
 	DeleteTestingNSPolicyUntyped string
 	DeleteTestingNSPolicy        framework.DeleteTestingNSPolicyType
 	IngressController            *IngressControllerOptions
+	ScyllaClusterUntyped         *ScyllaClusterOptions
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/cmd/tests/options.go
```diff
@@ -43,6 +68,18 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(&o.IngressController.Address, "ingress-controller-address", "", o.IngressController.Address, "Overrides destination address when sending testing data to applications behind ingresses.")
 	cmd.PersistentFlags().StringVarP(&o.IngressController.IngressClassName, "ingress-controller-ingress-class-name", "", o.IngressController.IngressClassName, "Ingress class name under which ingress controller is registered")
 	cmd.PersistentFlags().StringToStringVarP(&o.IngressController.CustomAnnotations, "ingress-controller-custom-annotations", "", o.IngressController.CustomAnnotations, "Custom annotations required by the ingress controller")
+	cmd.PersistentFlags().StringVarP(&o.ScyllaClusterUntyped.NodeServiceTypeUntyped, "scylla-cluster-node-service-type", "", o.ScyllaClusterUntyped.NodeServiceTypeUntyped, fmt.Sprintf("Kubernetes service type that the ScyllaDB cluster nodes are exposed with. Allowed values are [%s].", strings.Join(
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/framework/framework.go
```diff
@@ -143,6 +145,18 @@ func (f *Framework) CommonLabels() map[string]string {
 	}
 }
 
+func (f *Framework) GetDefaultScyllaCluster() *scyllav1.ScyllaCluster {
+	renderArgs := map[string]any{}
+	renderArgs["nodeServiceType"] = TestContext.ScyllaCluster.ExposeOptions.NodeServiceType
+	renderArgs["nodesBroadcastAddressType"] = TestContext.ScyllaCluster.ExposeOptions.NodesBroadcastAddressType
+	renderArgs["clientsBroadcastAddressType"] = TestContext.ScyllaCluster.ExposeOptions.ClientsBroadcastAddressType
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/framework/testcontext.go
```diff
@@ -22,9 +23,20 @@ type IngressController struct {
 	CustomAnnotations map[string]string
 }
 
+type ScyllaCluster struct {
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/set/scyllacluster/scyllacluster_auth.go
```diff
@@ -117,7 +119,18 @@ var _ = g.Describe("ScyllaCluster authentication", func() {
...
 
 		verifyScyllaCluster(ctx, f.KubeClient(), sc)
-		o.Expect(hosts).To(o.ConsistOf(getScyllaHostsAndWaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)))
+		waitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		oldHosts := hosts
+		oldHostIDs := hostIDs
+		hosts, hostIDs, err = utils.GetBroadcastRPCAddressesAndUUIDs(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(hosts).To(o.HaveLen(len(oldHosts)))
```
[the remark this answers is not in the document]
rzetelskik: changed to explicit size check, removed the `oldHosts` variable

test/e2e/set/scyllacluster/scyllacluster_pv.go
rzetelskik: done

test/e2e/set/scyllacluster/scyllacluster_scaling.go
rzetelskik: equal length follows from the `GetBroadcastRPCAddressesAndUUIDs` function, i.e. otherwise you'd receive an error. I suppose I can remove the explicit check for both and only check it for one, since the other follows implicitly

test/e2e/utils/helpers.go
rzetelskik: done

test/e2e/set/scyllacluster/scyllacluster_replace.go
rzetelskik: I think there should only be one default, uninfluenced by the needs of specific tests. In other words if a test has some specific changes it should apply them directly like it does. 
Exposing requires `ReplacingNodeUsingHostID` so it's the same really, and I think since this influences the way we configure exposing in the test, the name is more fitting.

test/e2e/utils/helpers.go
rzetelskik: added check for emptiness. Do we really care about status though? We could just let the client deal with those cases

test/e2e/fixture/scylla/scyllacluster.yaml.tmpl
```diff
@@ -6,6 +6,20 @@ spec:
   version: 5.2.7
   agentVersion: 3.1.0
   developerMode: true
+  exposeOptions:
+    nodeService:
+      {{- with .nodeServiceType }}
+      type: {{ . }}
+      {{- end }}
```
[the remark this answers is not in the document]
rzetelskik: I can't recall the reason, might be a leftover from when I had an additional nested struct there. Changed.

rzetelskik:
/hold
for https://github.com/scylladb/scylla-operator/issues/1638

rzetelskik:
https://github.com/scylladb/scylla-operator/issues/1638 landed
/hold cancel

rzetelskik:
@tnozicka ping

pkg/cmd/tests/options.go
```diff
@@ -15,18 +17,41 @@ type IngressControllerOptions struct {
...
+	scyllaClusterOptions         *framework.ScyllaClusterOptions
 }
 
 func NewTestFrameworkOptions() TestFrameworkOptions {
 	return TestFrameworkOptions{
 		ArtifactsDir:                 "",
 		DeleteTestingNSPolicyUntyped: string(framework.DeleteTestingNSPolicyAlways),
 		IngressController:            &IngressControllerOptions{},
+		ScyllaClusterOptionsUntyped: &ScyllaClusterOptions{
+			NodeServiceType:             string(scyllav1.NodeServiceTypeHeadless),
```
tnozicka, line 51: Why headless? Normally, I'd tell people to use at least cluster IP to have discovery endpoint I assume we could use it in some tests at some point as well.

test/e2e/set/scyllacluster/scyllacluster_replace.go
tnozicka: if I run a suite with `--scyllacluster-nodes-broadcast-address-type=PodIP` and a accidentally a wrong scylla version I don't want it to silently pass - it needs to fail or we can have a green CI job and wrongly think it's working

test/e2e/utils/helpers.go
```diff
@@ -436,41 +436,136 @@ func GetScyllaConfigClient(ctx context.Context, client corev1client.CoreV1Interf
...
-				return nil, nil, fmt.Errorf("can't get pod %q: %w", naming.ManualRef(sc.Namespace, s.Name), err)
-			}
-			host = pod.Status.PodIP
+		broadcastAddresses = append(broadcastAddresses, broadcastAddress)
+	}
+
+	return broadcastAddresses, nil
+}
+
+func GetBroadcastAddress(ctx context.Context, client corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster, svc *corev1.Service) (string, error) {
```
tnozicka, line 460: there is the new `controllerhelpers.GetScyllaBroadcastAddress`

pkg/cmd/tests/options.go
tnozicka: I suppose this test the newest path, although it wouldn't be a production "default", so the answer would likely be we need the testing matrix eventually

rzetelskik: Now that you mentioned it I'm not sure it's that good of a default: it's the most "disruptive" in regard to some of our procedures, e.g. the ClusterIP-based replace procedure. On one hand it's a good thing, as it doesn't let any behaviour based on ClusterIPs slip (say e.g. the Scylla host for upgrades issue), on the other hand these test *will* fail with Headless services unless we run them conditionally.

tnozicka: >  these test will fail with Headless services

which tests do you expect to fail on setups that don't use clusterIP? the CI looked green

rzetelskik: https://github.com/scylladb/scylla-operator/pull/1508#discussion_r1434219342 - the clusterIP-based replace tests fail without explicit override of exposeOptions (or other ways of circumventing it like skipping the tests or having sub-suites)
```
LAST SEEN   TYPE      REASON                   OBJECT                      MESSAGE
11s         Warning   InvalidScyllaDBVersion   scyllacluster/basic-25b5d   Can't proceed with the rollout. Specified ScyllaDB version "2022.2.12" does not support exposing via Service other than ClusterIP. Please use different ScyllaDB version.
```

test/e2e/utils/helpers.go
rzetelskik: done

test/e2e/set/scyllacluster/scyllacluster_replace.go
rzetelskik: > if I run a suite with --scyllacluster-nodes-broadcast-address-type=PodIP and a accidentally a wrong scylla version I don't want it to silently pass - it needs to fail or we can have a green CI job and wrongly think it's working

Agreed. I added a conditional label filter to the test runner instead, ptal.

pkg/cmd/tests/options.go
tnozicka: i thought we can just use newer scylladb there

rzetelskik: To my understanding the purpose of these tests is to specifically test the support for old procedure. After all we declare support for Scylla OS >= 5.0. The HostID-based procedure is only possible with >= 5.2.0.

tnozicka: Where do our test use scylladb < 5.2.0? Manifest is fixed to `5.4.0` and upgrades start from `5.2.11`.

When we wire CI to run multiple scenarios (scylla versions) I expect we give it the right combination of scylla version and expose type to test.

tnozicka: ok, I see the replace test has one more version spec - eventually the scylla version should be an argument but right now we mix it and that's probably why our reasoning here differs

rzetelskik: https://github.com/scylladb/scylla-operator/blob/master/test/e2e/set/scyllacluster/scyllacluster_replace.go#L186
https://github.com/scylladb/scylla-operator/blob/master/test/e2e/set/scyllacluster/scyllacluster_replace.go#L198

rzetelskik: And that's aligned with the test plan: https://github.com/scylladb/scylla-operator/tree/master/enhancements/proposals/1311-replace-nodes-using-hostid#test-plan.

tnozicka: ok, let's skip the test and run a the other variant in a different job

test/e2e/utils/helpers.go
tnozicka: it could give you a better message on why it is empty but you can find out from the dump too

test/e2e/framework/meta.go
```diff
@@ -7,12 +7,14 @@ import (
 )
 
 const (
-	SerialLabelName = "Serial"
+	SerialLabelName            = "Serial"
+	RequiresClusterIPLabelName = "RequiresClusterIP"
```
tnozicka, line 11: `ClusterIPOnly` would read nicely in the logs

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -72,11 +72,15 @@ var _ = g.Describe("ScyllaCluster", func() {
...
 
-		sc := scyllafixture.BasicScyllaCluster.ReadOrFail()
+		sc := f.GetDefaultScyllaCluster()
 		sc.Spec.Repository = e.scyllaImageRepository
 		sc.Spec.Version = e.scyllaVersion
 		sc.Spec.Datacenter.Racks[0].Members = 3
 
+		if e.procedure == clusterIPProcedure && (sc.Spec.ExposeOptions == nil || sc.Spec.ExposeOptions.NodeService == nil || sc.Spec.ExposeOptions.NodeService.Type == scyllav1.NodeServiceTypeHeadless) {
+			g.Skip("Skipping because ClusterIP-based replace procedure requires member services to have ClusterIPs")
+		}
```
tnozicka, line 82: nit
```suggestion
		isHeadlessNodeService := sc.Spec.ExposeOptions == nil || sc.Spec.ExposeOptions.NodeService == nil || sc.Spec.ExposeOptions.NodeService.Type == scyllav1.NodeServiceTypeHeadless
		if e.procedure == clusterIPProcedure && isHeadlessNodeService {
			g.Skip("Skipping because ClusterIP-based replace procedure can only be tested on ScyllaClusters exposed on ClusterIPs")
		}
```

tnozicka (approved):
thanks Kacper
/approve
/lgtm

/hold
for presubmit setup

test/e2e/framework/meta.go
rzetelskik: Wouldn't it be misleading in regard to the service type though? E.g. service of type LoadBalancer (or any other extension of ClusterIP) would also handle these tests

test/e2e/set/scyllacluster/scyllacluster_replace.go
rzetelskik: done

tnozicka: the main point was to make the condition smaller, but I don't see the variable extraction done

test/e2e/framework/meta.go
tnozicka: it's a label for a particular test that can only run with scyllaclusters using cluster ip exposure type. (`Only` is often used as suffix for test that are skipped when the config/platform doesn't support it)

tnozicka: or were you pointing to the need for type prefix like `SCClusterIPOnly`?

tnozicka: I guess in can be ambiguous in a different place. `RequiresClusterIP` is fine.

rzetelskik: > it's a label for a particular test that can only run with scyllaclusters using cluster ip exposure type.

for the NodeServiceType that would also be misleading, no? As the `NodeServiceTypeLoadBalancer` would also handle it

hack/.ci/run-e2e-gke.sh
```diff
@@ -23,6 +23,10 @@ fi
 
 SO_DISABLE_NODECONFIG=${SO_DISABLE_NODECONFIG:-false}
 
+SO_SCYLLACLUSTER_NODE_SERVICE_TYPE=${SO_SCYLLACLUSTER_NODE_SERVICE_TYPE:-Headless}
+SO_SCYLLACLUSTER_NODES_BROADCAST_ADDRESS_TYPE=${SO_SCYLLACLUSTER_NODES_BROADCAST_ADDRESS_TYPE:-PodIP}
+SO_SCYLLACLUSTER_CLIENTS_BROADCAST_ADDRESS_TYPE=${SO_SCYLLACLUSTER_CLIENTS_BROADCAST_ADDRESS_TYPE:-PodIP}
```
tnozicka, line 28: please make these required so we can be sure it takes the values from a presubmit

rzetelskik: done

test/e2e/set/scyllacluster/scyllacluster_replace.go
rzetelskik: ok, I only noticed the comment at first. done

tnozicka (commented):
/lgtm

tnozicka:
/hold
for https://github.com/scylladb/scylla-operator-release/pull/87

tnozicka:
/hold cancel

https://github.com/scylladb/scylla-operator-release/pull/87 has landed and prow config has been updated

Issue #1422, opened 2023-09-25
Make docs' header levels increase consecutively


rzetelskik:
Currently, we're suppressing MyST's `myst.header` to avoid warnings like the following:
```
/go/github.com/scylladb/scylla-operator/docs/source/helm.md:10:Non-consecutive header level increase; H1 to H3 [myst.header]
```
We should make the header levels consecutive to remove the suppression.

PR #1512, opened 2023-10-30
Fix hyperlinks in multi datacenter documentation

zimnx:
MultiDC documentation lives at depth 1, where links refer to root documents if it was at depth 2.


rzetelskik:
Did you actually try that? Because I remember it was a bit funky to get it working with recommonmark, and it required that to work

zimnx:
This link with single `../` works fine on deployed documentation:
https://github.com/scylladb/scylla-operator/blob/v1.11/docs/source/multidc/eks.md?plain=1#L65
Compared with https://github.com/scylladb/scylla-operator/blob/v1.11/docs/source/multidc/eks.md?plain=1#L44 which doesn't work because it has `../../`

rzetelskik:
> This link with single ../ works fine on deployed documentation:
https://github.com/scylladb/scylla-operator/blob/v1.11/docs/source/multidc/eks.md?plain=1#L65

It doesn't point to a specific heading

zimnx:
Yep, seems like links to headings are different than usual links. And they work differently when single and multi version documentation is built :( 

I updated PR with links which doesn't work on single version preview, but works on multiversion.

tnozicka:
@annastuchlik do you know any way around the relative links with sections issue ^?

tnozicka:
> It doesn't validate whether links points to something resolvable. Currently our deployed multidc docs have links pointing to 404 page.

@dgarcia360 do you know whether there is something in sphinx or the related machinery that could validate local links and avoid shipping 404s?

rzetelskik:
Since David prepared https://github.com/scylladb/scylla-operator/pull/1514 and MyST parser resolves the issue, should we drop this PR?

tnozicka:
> @dgarcia360 do you know whether there is something in sphinx or the related machinery that could validate local links and avoid shipping 404s?

https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#automatically-label-sections

the less convenient syntax
```
`{ref}`Custom title <guides/cross-referencing-with-sphinx:explicit targets>`
```
 for links validates the target exists, feels like we should use that

tnozicka:
> I don't know if there is a workaround for now

The only workaround I found is to use RTS style links inside Markdown as posted above - that validates links with anchors.

> I think we should wait for https://github.com/scylladb/scylla-operator/pull/1514 to be merged.

I've already tried with MyST and while it creates correct links, it doesn't validate the target path exists, if you use anchors (`#section`)

> they can be used in Mardown as {#tag}, but I'm not sure if we can successfully parse them at the moment

parsing is fine as far as creating the hyperlinks but the target validation is still not working

> In RST, we create an anchor, for example, .. _support-for-x:, and create a link using the anchor:

Yeah, I thought using RST may be a way out, but the headings syntax just sucks so much :( But I think embedding the RST syntax into Markdown get us the closest. Similarly we can use 
````
```{literalinclude} ../../../examples/monitoring/v1alpha1/scylladbmonitoring.yaml
   :language: yaml
   :linenos:
```
````
and actually test the same example with e2e.

tnozicka:
we need to get this unstuck to have the docs ready for the v1.11. @zimnx did something like https://github.com/scylladb/scylla-operator/pull/1512#issuecomment-1790388696 work?
https://recommonmark.readthedocs.io/en/latest/auto_structify.html#embed-restructuredtext

dgarcia360:
I would merge https://github.com/scylladb/scylla-operator/pull/1514 first before changing the links format.

> do you know whether there is something in sphinx or the related machinery that could validate local links and avoid shipping 404s?

Removing [myst.xref_missing](https://github.com/scylladb/scylla-operator/pull/1514/files#diff-ee84de99ec0ab708d3b5cc1cbda282aade6f76f468ceaae983434a30515db7dfR33) from  `suppress_warnings` configuration property should start to validate links.

> the less convenient syntax

I use the [ref directive](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#reference-roles) to link to anchors in other documents.  I'm not aware of a better way.

rzetelskik:
> Removing [myst.xref_missing](https://github.com/scylladb/scylla-operator/pull/1514/files#diff-ee84de99ec0ab708d3b5cc1cbda282aade6f76f468ceaae983434a30515db7dfR33) from suppress_warnings configuration property should start to validate links.

Just to follow up on this, we have an issue for getting rid of the suppression https://github.com/scylladb/scylla-operator/issues/1422.

tnozicka:
> @tnozicka Have you tried that without the suppression?

it seems to work with that, so I we can use the markdown style links

> I use the [ref directive](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#reference-roles) to link to anchors in other documents. I'm not aware of a better way.

I suppose the markdown native links would be even more fitting, now that the validation works. Or is there a difference?

PR #1515, opened 2023-10-31
Avoid service enpoints size limit in monitoring examples

tnozicka:
We should select member services when configuring ScyllaDBMonitoring. For small clusters the identity Endpoints set is identical, but there is a limit to the number of endpoints aggregated there. Prometheus doesn't care about publishing not ready endpoints and does its own health checks

Future issues with larger clusters.


tnozicka:
https://github.com/scylladb/scylla-operator/issues/1516
/retest

zimnx (approved):
/lgtm

PR #1521, opened 2023-11-02
Use correct key in error message when getting agent token

### LINKED ISSUE #1519: Logs: Invalid Manager Agent config warn is not clear
I have deployed Scylla Operator, Scylla Manager and a Scylla cluster. The three of them are up and running, however in the operator logs the following lines appear repeatedly:
```
E1102 10:55:13.552546       1 scyllacluster/controller.go:234] syncing key 'scylla/scylla' failed: can't sync agent token: can't get agent token: secret "scylla/scylla-agent-config" is missing "auth-token.yaml" data
I1102 10:55:13.552682       1 record/event.go:294] "Event occurred" object="scylla/scylla" fieldPath="" kind="ScyllaCluster" apiVersion="scylla.scylladb.com/v1" type="Warning" reason="InvalidManagerAgentConfig" message="Can't gent agent token: can't get agent token: secret \"scylla/scylla-agent-config\" is missing \"auth-token.yaml\" data
```
Note that I already created the `scylla-agent-config` secret in the `scylla` namespace where the cluster is.
The secret was something like:
```yaml
apiVersion: v1
data:
  auth-token.yaml: xxx #base64-encoded `auth-token: <the token>`
kind: Secret
# ...
type: Opaque
```

After taking a look at the Operator's code [here](https://github.com/scylladb/scylla-operator/blob/964418f71f87518440fe6767b07d3cc936ac46d7/pkg/helpers/agent.go#L27-L34), I found that the secret key should be  `scylla-manager-agent.yaml` rather than `auth-token.yaml`, and updating the secret like this indeed makes the Manager Agent config valid:
 
```yaml
apiVersion: v1
data:
  scylla-manager-agent.yaml: xxx #the same as before
kind: Secret
# ...
type: Opaque
```

The log line could be updated in order to make clear that the secret key should be `scylla-manager-agent.yaml`.

For reference, I am running all deployments on k8s 1.26 using v1.8.1 Helm charts, but I noticed that the relevant code for this has not changed up to and including master. 
Thank you for your outstanding work on the Operator!

tnozicka:
This PR fixes error message when getting agent token from config file, to refer to the same key it's asserting.

Resolves #1519

### DIFF
--- a/pkg/helpers/agent.go
+++ b/pkg/helpers/agent.go
@@ -27,7 +27,7 @@ func ParseTokenFromConfig(data []byte) (string, error) {
 func GetAgentAuthTokenFromAgentConfigSecret(secret *corev1.Secret) (string, error) {
 	configData, ok := secret.Data[naming.ScyllaAgentConfigFileName]
 	if !ok {
-		return "", fmt.Errorf("secret %q is missing %q data", naming.ObjRef(secret), naming.ScyllaAgentAuthTokenFileName)
+		return "", fmt.Errorf("secret %q is missing %q data", naming.ObjRef(secret), naming.ScyllaAgentConfigFileName)
 	}
 
 	return ParseTokenFromConfig(configData)


zimnx (approved):
/lgtm

PR #1522, opened 2023-11-02
Fix GH Action installing kubernetes

zimnx:
Newest GH Action Worker images already contain kubernetes repository key, gpg complains when output file exists.
kubectl in 1.28 version is also avaiailable, we have to first remove it to install our desired version.


tnozicka (approved):
/approve
/lgtm

thanks!

PR #1541, opened 2023-11-08
Generate API docs

### LINKED ISSUE #1526: Generate API reference in docs
At this point, our docs have [an outdated example of ScyllaCluster API](https://operator.docs.scylladb.com/stable/scylla-cluster-crd.html). This is historical, manually written and outdated. We also have more APIs / CRs then just ScyllaClusters.

Doing this manually is unmaintainable. But there is a way to generate the API docs automatically, as many sites do, like
- [OpenShift API reference](https://docs.openshift.com/container-platform/4.14/rest_api/index.html)
- [Kubernetes API reference](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/)

With some effort we can get swagger / openapi for our CRDs which can be used with an existing API docs generator. Given we use shpinx, https://github.com/sphinx-contrib/openapi would seem like a natural choice but it didn't work when I tried (see https://github.com/sphinx-contrib/openapi/issues/144) for details.

If that doesn't get fixed we could generate markdown, like kube does https://github.com/kubernetes/website/blob/5c744748410a7f1d4ebaa03800ace6c2c4f672e2/content/en/blog/_posts/2020-12-04-gsod-2020-improving-api-reference-experience.md#markdown-pages

tnozicka:
The PR add a tool (`gen-api-reference`) that can generate API docs from CRDs. While it has it's quirks, like the inlined objects from other or the same API group, it is marginally better then not having to the docs at all.  Compared to native Swagger/OpenAPI this is aware of API groups and versions.

The next step would be to generate the docs base on Golang types that don't inline referenced objects but this still gets us 95% of the way and with the internal links to objects it's quite usable.

Resolves #1526


tnozicka:
> Why did you choose generating rst instead of markdown?

We need the RST directives so embedding it in markdown just makes it harder or possibly bumping into "convertor" issues. Some like toc may not be even supported there. That said, it's limited to the templates.

> Couldn't it be handled similarly to how https://github.com/kubernetes-sigs/reference-docs/tree/master/gen-resourcesdocs handles it?

I think we'll eventually use a similar generator from GoType, I didn't realize CRDs lack some of the information initially and this was quick to get. I'd like it to be slightly less manually but will see when I do v2 on types one day.

(You can cut this PR some slack as we'll eventually replace it.)

rzetelskik:
> We need the RST directives so embedding it in markdown just makes it harder or possibly bumping into "convertor" issues. Some like toc may not be even supported there. That said, it's limited to the templates.

toc is supported - mystparser docs are dogfooding https://github.com/executablebooks/MyST-Parser/blob/master/docs/index.md. Having said that our indexes are still in rst so not an issue, this can be converted later, I was just curious if there was a known blocker. I'd personally rather see all our docs in markdown than the other way round but let's not get into this here.

tnozicka:
Good, I am fine if this eventually reconciles one way or the other, if we decide to unify the rest. If we sort it out for the rest, the templates are easy to switch.

tnozicka:
@rzetelskik gentle ping

pkg/cmd/generateapireference/run.go
```diff
@@ -0,0 +1,250 @@
...
+		err := os.Mkdir(groupDir, 0777)
+		if err == nil {
+			klog.V(2).InfoS("Created group directory", "Path", groupDir)
+		} else if os.IsExist(err) {
+			klog.V(2).InfoS("Using existing group directory", "Path", groupDir)
+		} else {
+			return fmt.Errorf("can't create group directory %q: %w", groupDir, err)
+		}
+
+		data, err := assets.RenderTemplate(templates[gvIndexTemplateFileName], map[string]string{"Group": group})
```
rzetelskik, line 222: Not in practice since `templateFiles` is hardcoded, but that's a potential nil dereference. Up to you if you want to assert that

pkg/cmd/generateapireference/run_test.go
```diff
@@ -0,0 +1,139 @@
...
+					Properties: map[string]apiextensionsv1.JSONSchemaProps{
+						"name": {
+							Type: "string",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "nested arrays containing object",
```
rzetelskik, line 124: leftover?

docs/source/helm.md
```diff
@@ -163,7 +163,7 @@ racks:
 Above cluster will use 4.3.0 Scylla, 2.2.1 Scylla Manager Agent sidecar and will have a single rack having 2 nodes. 
 Each node will have a single CPU and 1 GiB of memory.
 
-For other customizable fields, please refer to [ScyllaCluster CRD definition](scylla-cluster-crd.md).
+For other customizable fields, please refer to [ScyllaCluster CRD definition](api-reference/groups/scylla.scylladb.com/scyllaclusters.rst).
```
rzetelskik, line 166: nit: not an addition but `ScyllaCluster CRD definition` is redundant.

docs/source/manager.md
```diff
@@ -160,7 +160,7 @@ Add following task definition to Cluster spec:
       interval: "1d"
 ```
 
-For full task definition configuration consult [Scylla Cluster CRD](scylla-cluster-crd.md).
+For full task definition configuration consult [Scylla Cluster CRD](api-reference/groups/scylla.scylladb.com/scyllaclusters.rst).
```
rzetelskik, line 163: nit: not an addition but s/Scylla Cluster CRD/ScyllaCluster CRD
Also maybe leaving the link text empty wouldn't be a bad idea?
image

rzetelskik (commented):
lgtm overall, a few comments.
I wonder - would it be possible to indent nested fields, like Kube API reference does? Or does that require moving to parsing go files?

image

docs/source/manager.md
tnozicka: > Also maybe leaving the link text empty wouldn't be a bad idea?

possibly, but on the other hand some changes in the other file might make this look worse over time without us noticing

tnozicka:
> would it be possible to indent nested fields

I think they selectively inline some definitions and the others. This is tricky and you may end up defining some multiple times. I'd rather not go into it here.

docs/source/manager.md
tnozicka: space fixed

docs/source/helm.md
tnozicka: done

pkg/cmd/generateapireference/run_test.go
tnozicka: done

pkg/cmd/generateapireference/run.go
tnozicka: restructured

docs/source/api-reference/index.rst
```diff
@@ -0,0 +1,13 @@
+API Reference
+=============
+
+All APIs provided by Scylla Operator are defined using CRDs and adhere to Kubernetes API standards. You can find out how Kubernetes style REST APIs work on `<https://kubernetes.io/docs/reference/using-api/>`_ and `<https://kubernetes.io/docs/reference/access-authn-authz/>`_
```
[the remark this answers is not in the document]
tnozicka: done

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1546#issuecomment-1860319092
/retest

pkg/cmd/generateapireference/run.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	kindTemplate, err := o.parseTemplate(kindTemplateFileName)
+	if err != nil {
+		return err
+	}
+
+	templates := map[string]*template.Template{}
+	for _, tmplFileName := range templateFiles {
+		tmplFile := filepath.Join(o.TemplatesDir, tmplFileName)
```
rzetelskik, line 220: let's use `parseTemplate` func here

rzetelskik (commented):
one more comment, other than that lgtm
/assign zimnx

pkg/cmd/generateapireference/run.go
tnozicka: actually, I forgot to remove this whole block, thx

tnozicka: removed

rzetelskik: right, I didn't notice it wasn't even used anymore

rzetelskik:
/approve
and lgtm

tnozicka:
> one more comment, other than that lgtm
/assign zimnx

@rzetelskik based on your previous comment (and resolving the referenced nit) I suppose this is now only awaiting your official tag?

PR #1545, opened 2023-11-09
Update support matrix

zimnx:
There's no change in between v1.10 and v1.11 in terms of API requirements. 
Monitoring version is set to one managed by the Operator.
We might consider dropping support for older releases in next release as lot of them are already way past their end of life.


tnozicka (approved):
/approve
/lgtm

PR #1532, opened 2023-11-06
Move aggregated deploy manifests to deploy folder

tnozicka:
**Description of your changes:**
Aggregating all deploy manifests allows users to do `kubectl apply --server-side -f https://github.com/scylladb/scylla-operator/raw/master/deploy/operator.yaml` that complements the local `kubectl apply --server-side -f deploy/operator/`

Moving the aggregated manifests to deploy folder ensures they can be found and referenced in out install scenarios.

This also add an aggregated version of dev setup for manager. 

Resolves #1533

Issue #1525, opened 2023-11-03
Flake: ScyllaCluster [It] nodes are cleaned up after horizontal scaling


rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1524/pull-scylla-operator-e2e-gke-parallel/1720397617084502016#1:test-build-log.txt%3A1594
https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1524/pull-scylla-operator-e2e-gke-parallel/1720397617084502016/

```console
• [FAILED] [771.623 seconds]
ScyllaCluster [It] nodes are cleaned up after horizontal scaling
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_cleanup.go:29
...
[FAILED] Expected
      <[]utils.ObserverEvent[*k8s.io/api/batch/v1.Job] | len:3, cap:15>: [
...
      ]
  to have length 2
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_cleanup.go:109 @ 11/03/23 11:31:40.032
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.glob..func2.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_cleanup.go:109 +0xd45
```

/kind flake
/priority important-longterm

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1552/pull-scylla-operator-e2e-gke-parallel/1722988328061505536#1:test-build-log.txt%3A1565

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/841/pull-scylla-operator-e2e-gke-parallel/1729123170528530432

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/841/pull-scylla-operator-e2e-gke-parallel/1729135508577062912

rzetelskik:
[...]

PR #1552, opened 2023-11-10
Generate aggregated third-party manifests

tnozicka:
Aggregating the manifest allows using kubectl -f with remote file.

Resolves #

/hold
needs https://github.com/scylladb/scylla-operator/pull/1532


tnozicka:
sound like infra hickup around cloning source code or something
/retest

tnozicka:
/hold cancel
#1532 landed

rzetelskik (commented):
lgtm overall, but shouldn't you generate it for local-csi-driver (./examples/common/local-csi-driver/*.yaml to ./examples/common/local-csi-driver.yaml) as well?

tnozicka:
actually, I think that should move to deploy folder (eventually to be gone and managed by nodeconfig) so we can reply on the storage class in your manifest, so it works out of the box

rzetelskik:
> I think that should move to deploy folder (eventually to be gone and managed by nodeconfig) 

The CSI driver is not a hard requirement of ours, only what we recommend for an optimal performance, but there's nothing stopping you from using a different storage provisioner, is there? I can see moving it to `deploy/`, especially as it's not really *third-party*, but I don't really see it as a part of NodeConfig; I guess that's a topic for a separate discussion though. 
My point was that the manifests for the local-csi-driver are also used in our docs so if that's why you want to concatenate the examples then you're going to hit this as well.

tnozicka:
I think it's outside of scope of this PR. It's not a hard requirement but our default setup doesn't work with NVMes without editing the manifest. It should be the other way around.

tnozicka:
thx, I've filled https://github.com/scylladb/scylla-operator/issues/1554 for local-csi-driver

rzetelskik:
https://github.com/scylladb/scylla-operator/issues/1525#issuecomment-1805957503
/test e2e-gke-parallel

PR #1555, opened 2023-11-10
[WIP] Integrate provisioner to deploy folder and make sure our default deployments and examples use it

### LINKED ISSUE #1554: Our examples don't use our NVMe storage by default or get stuck
Our primary story is to run ScyllaDB on NVMes, yet our deploy manifests and examples either get the default storage class from the cloud (NAS, not our local NVE storage) or, if there is none, they get stuck even if they deploy the local-csi-driver. This is surprising and not easy to debug for someone who has spin this up for the first time and may not be familiar with Kubernetes too much.

We can't easily instruct people to switch default storage classes, nor it's correct as it may break the other apps.

We should make sure that when the user applies our manifests or examples they by default use our storage class (`scylladb-local-xfs`). If someone wants to make changes they are free to do so but the main story should work by default without need to intervene.

This should also remove the default storage class hack from our deploy scripts.

Requires:
- [x] https://github.com/scylladb/scylla-operator/issues/1556

tnozicka:
This make sure our primary deployment on NVMes work while it still retains the option for people to select their own storage class.

This fixes setups on systems without a default storage class, where it didn't provision PVCs because our storage class isn't mark as default to avoid conflicts with existing provisioners.

This also makes sure that even on GKE / EKS people don't deploy local-csi-driver and then later findout their scyllacluster are runing on NAS instead of NVMs. The explicit storage class avoid it.

Resolves #1554


tnozicka:
I assume this will be chllenging without the integration with the nodeconfig that provides xfs formatting

tnozicka:
needs to wait for https://github.com/scylladb/scylla-operator/issues/1556

PR #1562, opened 2023-11-13
Collect CRDs in must-gather

tnozicka:
This PR add CRDs collection to must-gather so we can asses whether our CRDs are up to data as users frequently forget to update them because of helm deficiencies. It also helps to see what other extensions are present in the cluster in case of an interference.


rzetelskik (approved):
good idea, lgtm
/assign zimnx

zimnx (approved):
/lgtm

PR #1563, opened 2023-11-13
Automatically create explicit dest-dir for must-gather

tnozicka:
When using `must-gather` with explicit `--dest-dir=`, this PR allows automatically creating that dir, if it doesn't exist. This helps with the UX so someone iterating on that can only use `rm` but no longer needs the extra `mkdir` command.

### DIFF
--- a/pkg/cmd/operator/gatherbase.go
+++ b/pkg/cmd/operator/gatherbase.go
@@ -77,12 +77,8 @@ func (o *GatherBaseOptions) Validate() error {
 			if len(files) > 0 {
 				errs = append(errs, fmt.Errorf("destination directory %q is not empty", o.DestDir))
 			}
-		} else {
-			if os.IsNotExist(err) {
-				errs = append(errs, fmt.Errorf("destination directory %q doesn't exist", o.DestDir))
-			} else {
-				errs = append(errs, fmt.Errorf("can't stat destination directory %q: %w", o.DestDir, err))
-			}
+		} else if !os.IsNotExist(err) {
+			errs = append(errs, fmt.Errorf("can't stat destination directory %q: %w", o.DestDir, err))
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/operator/gatherbase.go
```diff
@@ -107,13 +103,24 @@ func (o *GatherBaseOptions) Complete() error {
...
 
+	ignoreDestDirExists := false
 	if len(o.DestDir) == 0 {
 		o.DestDir = fmt.Sprintf("%s-%s", o.GathererName, utilrand.String(12))
-		err := os.Mkdir(o.DestDir, 0770)
-		if err != nil {
-			return fmt.Errorf("can't create destination directory %q: %w", o.DestDir, err)
-		}
+	} else {
+		// We have already made sure that the dir doesn't exist or is empty in validation.
```
rzetelskik, line 110: Wouldn't the logic be simpler if you moved DestDir creation before verification?

rzetelskik (commented):
/lgtm

pkg/cmd/operator/gatherbase.go
rzetelskik: /hold

tnozicka: I don't think it should create dest dir when there are invalid arguments. The point of verification as a first step is to make sure the inputs are correct before taking any action.

rzetelskik: Fair enough. In that case I'd personally move the destDir creation out of the Validate func though.

tnozicka: it's inside "Complete()" function...

rzetelskik: ah, nvm, didn't notice the lines skipped by github

rzetelskik (approved):
/lgtm

PR #1564, opened 2023-11-13
Add ScyllaCluster and scylla config examples

tnozicka:
Adds a simple (single file per resource) example for ScyllaCluster and an example of scylla config. This will be later used in a demo.


rzetelskik (approved):
If it's not meant as the unified manifest for docs but just for the demo purposes then ok.
/lgtm

Issue #1476, opened 2023-10-13
Flake: ScyllaCluster [It] should connect to cluster via Ingresses


tnozicka:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1456/pull-scylla-operator-e2e-gke-parallel/1712825884106297344#1:test-build-log.txt%3A1394
```
 • [FAILED] [493.231 seconds]
ScyllaCluster [It] should connect to cluster via Ingresses
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_expose.go:47
  Timeline >>
  STEP: Creating a new namespace @ 10/13/23 13:57:27.987
  Oct 13 13:57:28.126: INFO: Created namespace "e2e-test-scyllacluster-25s7q-rxxkg".
  STEP: Waiting for service account token Secret "e2e-user-token" in namespace "e2e-test-scyllacluster-25s7q-rxxkg". @ 10/13/23 13:57:28.312
  STEP: Waiting for default ServiceAccount in namespace "e2e-test-scyllacluster-25s7q-rxxkg". @ 10/13/23 13:57:28.962
  STEP: Creating a ScyllaCluster @ 10/13/23 13:57:29.122
  STEP: Waiting for ScyllaCluster to deploy @ 10/13/23 13:57:29.227
  Oct 13 14:00:25.783: INFO: ScyllaCluster e2e-test-scyllacluster-25s7q-rxxkg/basic-mh9kd (RV=10407) is rolled out
  STEP: Verifying the ScyllaCluster @ 10/13/23 14:00:25.783
  Oct 13 14:00:25.833: INFO: Found 1 pvc(s) in namespace "e2e-test-scyllacluster-25s7q-rxxkg"
  Oct 13 14:00:25.833: INFO: Found 1 pvc(s) for ScyllaCluster "e2e-test-scyllacluster-25s7q-rxxkg/basic-mh9kd"
  STEP: Nodes reloaded the certificate @ 10/13/23 14:00:26.024
  STEP: Waiting for the ScyllaCluster(s) to reach consistency ALL @ 10/13/23 14:00:26.024
  Oct 13 14:00:26.150: INFO: ScyllaDB nodes have reached status consistency.
  STEP: Connecting via e2e-test-scyllacluster-25s7q-rxxkg.private.nodes.scylladb.com domain @ 10/13/23 14:00:26.22
[...]

PR #1565, opened 2023-11-13
Generelize GKE NodeConfig so it can be faked on libvirt

tnozicka:
This PR allows to fake NVMe drives on libvirt and still use the same config. I don't think we need to mess with the model by default.


tnozicka:
https://github.com/scylladb/scylla-operator/issues/1476#issuecomment-1808515998
/retest

PR #1569, opened 2023-11-14
Fix error aggregation in scyllacluster controller job sync

rzetelskik:
**Description of your changes:** We don't aggregate and check for errors in ScyllaCluster controller's job sync. This PR fixes it.

/kind bug
/priority important-soon


zimnx (approved):
good catch, thanks!
/approve
/lgtm

PR #1598, opened 2023-11-27
Gracefully evict Pods on ScyllaDB node replacements

### LINKED ISSUE #732: Revert forced deletion compensating kube eviction bug
We had to fallback to forced pod deletion because of kubernetes eviction bug https://github.com/kubernetes/kubernetes/issues/103970 When it gets fixed we should revert the commit (TODO: ref) and use safe eviction.

tnozicka:
With https://github.com/kubernetes/kubernetes/issues/103970 fixed in our supported version we can finally gracefully evict Pods and respect PDBs.

This adds stability and unblocks the path to mass node replacement.

Resolves #732


zimnx (approved):
/lgtm

Issue #1028, opened 2022-09-15
Flake: ScyllaCluster should allow to build connection pool using shard aware ports


tnozicka:
https://github.com/scylladb/scylla-operator/actions/runs/3056629503/jobs/4931011388#step:12:664

```
↺ [FLAKEY TEST - TOOK 2 ATTEMPTS TO PASS] [SLOW TEST] [716.582 seconds]
ScyllaCluster
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_shardawareness.go:25
  should allow to build connection pool using shard aware ports
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_shardawareness.go:30

  Begin Captured StdOut/StdErr Output >>
    I0914 23:21:03.448313      47 framework/framework.go:249] "Namespace removed." Namespace="e2e-test-scyllacluster-jmx8m-nqlnz"
    I0914 23:26:03.711932      47 framework/framework.go:249] "Namespace removed." Namespace="e2e-test-scyllacluster-jmx8m-psjsw"
  << End Captured StdOut/StdErr Output

  Begin Captured GinkgoWriter Output >>
    [BeforeEach] ScyllaCluster
      github.com/scylladb/scylla-operator/test/e2e/framework/framework.go:57
    STEP: Creating a new namespace 09/14/22 23:14:07.13
    Sep 14 23:14:07.185: INFO: Created namespace "e2e-test-scyllacluster-jmx8m-nqlnz".
    STEP: Waiting for ServiceAccount "e2e-user" in namespace "e2e-test-scyllacluster-jmx8m-nqlnz". 09/14/22 23:14:07.215
    STEP: Waiting for default ServiceAccount in namespace "e2e-test-scyllacluster-jmx8m-nqlnz". 09/14/22 23:14:08.878
    [It] should allow to build connection pool using shard aware ports
      github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_shardawareness.go:30
    STEP: Creating a ScyllaCluster 09/14/22 23:14:08.883
    STEP: Waiting for the ScyllaCluster to deploy 09/14/22 23:14:08.9
    Sep 14 23:17:38.950: INFO: ScyllaCluster e2e-test-scyllacluster-jmx8m-nqlnz/basic-x5qk6 (RV=4358) is rolled out
[...]

PR #1612, opened 2023-11-30
Automatically detect number of cpus from cgroup in Scylla Operator binaries

zimnx:

Golang runtime doesn't understand cgroups hence it assumes all cpus are
available for running goroutines.
Since applications on Kubernetes are running under restricted resources
we should detect cgroups settings and align Go runtime to discovered settings.
This should lower resource consumption, and allow us to run on even more limited resources.

### DIFF
--- a/cmd/scylla-operator-tests/scylla-operator-tests.go
+++ b/cmd/scylla-operator-tests/scylla-operator-tests.go
@@ -7,11 +7,11 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
-	"runtime"
 	"time"
 
 	cmd "github.com/scylladb/scylla-operator/pkg/cmd/tests"
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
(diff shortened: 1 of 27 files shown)


zimnx:
> Did we actually hit any throttling-related issues in the past?

No, we haven't, as we don't have any observability, nor any container running on low resources having a liveness/readiness probe so any slowness is not visible. 

As part of other PR I will add such low resource container with probes. I hit multiple times issue with such container being started but not even printing starting log message. After I configured Go runtime properly I haven't noticed it again. Can't tell whether I was lucky or it indeed fixed what I saw. 

This is a known issue, fixing it wouldn't hurt.

tnozicka (approved):
if it helps you I am fine going ahead with it and reverting this when https://github.com/golang/go/issues/33803 gets fixed

/approve
/lgtm

zimnx:
Flake: https://github.com/scylladb/scylla-operator/issues/1028#issuecomment-1838868867
/test e2e-gke-parallel

PR #1615, opened 2023-12-01
Limit scylla manager API calls duration

### LINKED ISSUE #1433: Flake: Scylla Manager integration [It] should discover cluster and sync tasks
```console
• [FAILED] [3599.606 seconds]
Scylla Manager integration [It] should discover cluster and sync tasks
...
  [FAILED] Unexpected error:
      <wait.errInterrupted>: 
      timed out waiting for the condition
      {
          cause: <*errors.errorString | 0xc00033d280>{
              s: "timed out waiting for the condition",
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:88 @ 09/30/23 00:42:37.578

  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.glob..func18.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:88 +0x8e4
```

https://github.com/scylladb/scylla-operator/actions/runs/6358101136/job/17270108645#step:3:3867
https://github.com/scylladb/scylla-operator/suites/16722273178/artifacts/955646400

tnozicka:
Unfortunately, scylla manager retires call internally and some calls can take ages. This was timing out a context in scylla manager controller.
```
E1201 08:53:03.866606       1 manager/sync.go:137] "Failed to execute action" err="[POST /cluster/{cluster_id}/tasks][400] PostClusterClusterIDTasks default  &{Message:create backup target: location is not accessible: 10.105.71.19: giving up after 2 attempts: after 30s: context deadline exceeded TraceID:H9iX4_oNS4SeVNkLug5wKQ}" action="add task &{ClusterID: Enabled:true ID: Name:daily backup Properties:map[location:[s3:bucket] retention:3] Schedule:0xc0008ba5d0 Tags:[] Type:backup}"
```
(so two independent 30s timeouts met)
The fix worked for me even with the same 30s timeout but I felt we should better raise it as well given the timings and manager API behaviour.

Resolves #1433

### DIFF
--- a/pkg/cmd/operator/manager_controller.go
+++ b/pkg/cmd/operator/manager_controller.go
@@ -123,7 +123,12 @@ func (o *ManagerControllerOptions) Complete() error {
 
 	// TODO: Use https and wire certs.
 	url := fmt.Sprintf("http://%s/api/v1", naming.ScyllaManagerServiceName)
-	managerClient, err := mermaidclient.NewClient(url, &http.Transport{})
+	managerClient, err := mermaidclient.NewClient(url, &http.Client{
+		// Limit manager calls by default to a higher bound.
+		// Individual calls can still be further limited using context.
+		// Manager is prone to extremely long calls because it (unfortunately) retries errors internally.
+		Timeout: 15 * time.Second,
+	})
 	if err != nil {
 		return fmt.Errorf("can't build manager client: %w", err)
 	}
--- a/pkg/controller/manager/controller.go
+++ b/pkg/controller/manager/controller.go
@@ -34,7 +34,11 @@ const (
 	ControllerName = "ScyllaManagerController"
 	// maxSyncDuration enforces preemption. Do not raise the value! Controllers shouldn't actively wait,
 	// but rather use the queue.
-	maxSyncDuration = 30 * time.Second
+	// Unfortunately, Scylla Manager calls are synchronous, internally retried and can take ages.
+	// Contrary to what it should be, this needs to be quite high.
+	// Given we reconcile tasks individually with an API call,
+	// this may not be enough for N tasks but should eventually make it.
+	maxSyncDuration = 2 * time.Minute
 )
 
 var (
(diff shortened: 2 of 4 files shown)


zimnx:
Thanks
/approve
/lgtm

pkg/controller/manager/controller.go
```diff
@@ -34,7 +34,11 @@ const (
 	ControllerName = "ScyllaManagerController"
 	// maxSyncDuration enforces preemption. Do not raise the value! Controllers shouldn't actively wait,
 	// but rather use the queue.
-	maxSyncDuration = 30 * time.Second
+	// Unfortunately, Scylla Manager calls are synchronous, internally retried and can take ages.
+	// Contrary to what it should be, this needs to be quite high.
+	// Given we reconcile tasks individually with an API call,
+	// this may not be enough for N task but should eventually make it.
+	maxSyncDuration = 2 * time.Minute
```
rzetelskik, line 41: does that imply that individual tasks can take longer than 30 seconds? Otherwise requeuing on our side should reach the same result?

pkg/cmd/operator/manager_controller.go
```diff
@@ -123,7 +123,12 @@ func (o *ManagerControllerOptions) Complete() error {
 
 	// TODO: Use https and wire certs.
 	url := fmt.Sprintf("http://%s/api/v1", naming.ScyllaManagerServiceName)
-	managerClient, err := mermaidclient.NewClient(url, &http.Transport{})
+	managerClient, err := mermaidclient.NewClient(url, &http.Client{
+		// Limit manager calls by default to a higher bound.
```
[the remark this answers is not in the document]
tnozicka: only by the 30s in the context from the sync loop, but that made everything fail

pkg/controller/manager/controller.go
tnozicka: we do more then 1 manager API call per sync loop

tnozicka: it has been failing permanently on my local setup so this definitely helps. We can file a new one if there is still something left and it flakes again.

PR #1624, opened 2023-12-08
Fix StatefulSet rolled out condition

zimnx:
Condition used to determine whether given StatefulSet is fully rolled out was missing two conditions, causing false positivies. Operator considered StatefulSet being rolled out and set ScyllaCluster rollout success prematurely.

### DIFF
--- a/pkg/controllerhelpers/apps.go
+++ b/pkg/controllerhelpers/apps.go
@@ -21,7 +21,15 @@ func IsStatefulSetRolledOut(sts *appsv1.StatefulSet) (bool, error) {
 		return false, nil
 	}
 
+	if sts.Status.Replicas != *sts.Spec.Replicas {
+		return false, nil
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/controllerhelpers/apps.go
```diff
@@ -25,6 +25,14 @@ func IsStatefulSetRolledOut(sts *appsv1.StatefulSet) (bool, error) {
 		return false, nil
 	}
 
+	if sts.Status.AvailableReplicas < *sts.Spec.Replicas {
+		return false, nil
+	}
+
+	if sts.Status.Replicas < *sts.Spec.Replicas {
```
tnozicka, line 32: how can there be less replicas then availableReplicas checked in the previous steps?

zimnx: it can't, rearranged and changed to require equality.

tnozicka (approved):
/approve
/lgtm
thanks

PR #1630, opened 2023-12-12
Disable Secret redaction in E2E dumps

zimnx:
E2E dump contains temporary data which is not sensitive.
Redaction prevents from inspecting Operator output which
prevents from debugging CI flakes.

Prerequisites:

- [ ] https://github.com/scylladb/scylla-operator/pull/1629

/hold for #1629

### DIFF
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -36,7 +36,7 @@ function kubectl_create {
 }
 
 function gather-artifacts {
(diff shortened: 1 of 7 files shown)


tnozicka:
I feel like we should be careful about this, say you endup configuring S3 backups or something else there.

Can you elaborate on the case where you needed the private data and see whether there is a different way to approach it?

zimnx:
> I feel like we should be careful about this, say you endup configuring S3 backups or something else there.

We should use bindings to ServiceAccounts not store plain passwords in secrets.

> Can you elaborate on the case where you needed the private data and see whether there is a different way to approach it?

As part of debugging of "connecting via Ingress" E2E I couldn't confirm that the connection config generated by Operator is correct because it was redacted. It will be an issue with every Secret produced by Operator.

tnozicka:
> We should use bindings to ServiceAccounts not store plain passwords in secrets.

using workload identity is available only on some platforms and for some use cases. Storing sensitive information in secrets is what they are meant for.

> As part of debugging of "connecting via Ingress" E2E flake I couldn't confirm that the connection config generated by Operator is correct because it was redacted. It will be an issue with every Secret produced by Operator.

I think there should be other ways to find out if the connection config is working or specific failures for those. Collecting secrets is a tricky job that I'd rather avoid as the consequences are severe. We should figure out other ways to get more context info or live with trying some stuff locally.

PR #1634, opened 2023-12-12
Fix default ServiceAccount awaiting in E2E framework

zimnx:
Namespace is cluster-wide resource, so it doesn't have 'Namespace' field set. Code awaiting for 'default' ServiceAccount in e2e Namespace should use Name of Namespace.


rzetelskik (approved):
/lgtm
thanks!

PR #1603, opened 2023-11-28
Update Scylla Manager to 3.2.5

tnozicka:
**Description of your changes:**
Updates Scylla Manager to `3.2.5`.

**Which issue is resolved by this Pull Request:**
Resolves #1588

PR #1637, opened 2023-12-12
Bump Scylla-manager version to be 3.2.5

vponomaryov:

Bump the scylla-manager for the following reasons:
- The `3.2.5` is the latest available and recommended scylla-manager version
- It is multiarch and supports `ARM` instances as well as `X86`-based.

So, any new user deploying the default manager and it's agent on `ARM` layout won't get failures caused by the instance types incompatibility issues.


zimnx:
Closing as duplicate of https://github.com/scylladb/scylla-operator/pull/1603

PR #1639, opened 2023-12-13
Rollback StatefulSet selector

zimnx:
Immutable StatefulSet selector was changed as part of #1388. 
This caused issues on upgrades because Operator tried to update immutable selector field.

This rolls back StatefulSet selector to one used in <v1.11.0 versions, but keeps added labels as is.

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
@@ -214,13 +215,23 @@ func servicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
 
 // StatefulSetForRack make a StatefulSet for the rack.
 // existingSts may be nil if it doesn't exist yet.
-func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existingSts *appsv1.StatefulSet, sidecarImage string) (*appsv1.StatefulSet, error) {
-	rackLabels, err := naming.RackLabels(r, c)
+func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existingSts *appsv1.StatefulSet, sidecarImage string, rackOrdinal int) (*appsv1.StatefulSet, error) {
+	rackLabels, err := naming.RackSelectorLabels(r, c)
 	if err != nil {
 		return nil, fmt.Errorf("can't get rack labels: %w", err)
 	}
-	matchLabels := helpers.ShallowCopyMap(rackLabels)
+	rackLabels[naming.RackOrdinalLabel] = strconv.Itoa(rackOrdinal)
 	rackLabels[naming.ScyllaVersionLabel] = c.Spec.Version
+
+	selectorLabels, err := naming.RackSelectorLabels(r, c)
+	if err != nil {
+		return nil, fmt.Errorf("can't get selector labels: %w", err)
+	}
+
+	pvcLabels, err := naming.RackSelectorLabels(r, c)
+	if err != nil {
+		return nil, fmt.Errorf("can't get PVC Template labels: %w", err)
+	}
 
 	placement := r.Placement
 	if placement == nil {
@@ -247,7 +258,7 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 			// Use a common Headless Service for all StatefulSets
 			ServiceName: naming.HeadlessServiceNameForCluster(c),
 			Selector: &metav1.LabelSelector{
-				MatchLabels: matchLabels,
+				MatchLabels: selectorLabels,
 			},
 			PodManagementPolicy: appsv1.OrderedReadyPodManagement,
 			UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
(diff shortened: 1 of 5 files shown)


zimnx:
I will validate manually whether upgrade succeeds

/hold

zimnx:
Upgrade succeeds

/hold cancel

pkg/controller/scyllacluster/resource.go
```diff
@@ -219,9 +221,25 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 	if err != nil {
 		return nil, fmt.Errorf("can't get rack labels: %w", err)
 	}
-	matchLabels := helpers.ShallowCopyMap(rackLabels)
```
rzetelskik, line 222: nit: Should we also remove the ShallowCopyMap function with its only usage gone?

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -579,29 +579,24 @@ func TestStatefulSetForRack(t *testing.T) {
...
 		return map[string]string{
 			"app":                          "scylla",
 			"app.kubernetes.io/managed-by": "scylla-operator",
 			"app.kubernetes.io/name":       "scylla",
 			"scylla/cluster":               "basic",
 			"scylla/datacenter":            "dc",
 			"scylla/rack":                  "rack",
-			"scylla/rack-ordinal":          "0",
+			"scylla/scylla-version":        "",
+			"scylla/rack-ordinal":          fmt.Sprintf("%d", ordinal),
```
rzetelskik, line 591: nit: strconv.Itoa

rzetelskik (approved):
two nits, lgtm
/assign tnozicka

pkg/controller/scyllacluster/resource.go
```diff
@@ -219,9 +221,25 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
...
+	_, rackOrdinal, ok := slices.Find(c.Spec.Datacenter.Racks, func(rack scyllav1.RackSpec) bool {
```
tnozicka, line 224: nit: I suppose at this place you could just get the index as argument (the caller has it) but I don't feel strong

pkg/controller/scyllacluster/resource.go
```diff
@@ -219,9 +221,25 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
...
+		return rack.Name == r.Name
+	})
+	if !ok {
+		return nil, fmt.Errorf("can't find ordinal of rack %q in ScyllaCluster %q", r.Name, naming.ObjRef(c))
+	}
+	rackLabels[naming.RackOrdinalLabel] = strconv.Itoa(rackOrdinal)
 	rackLabels[naming.ScyllaVersionLabel] = c.Spec.Version
 
+	selectorLabels, err := naming.RackLabels(r, c)
```
tnozicka, line 233: I think `RackLabels` might need to be renamed to `RackSelectorLabels` or something like that, so there is a clear distinction + a comment not to change these.

pkg/controller/scyllacluster/resource.go
```diff
@@ -219,9 +221,25 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
...
+	if err != nil {
+		return nil, fmt.Errorf("can't get selector labels: %w", err)
+	}
+
+	pvcLabels, err := naming.RackLabels(r, c)
```
tnozicka, line 238: I suppose this could be more then just the selector labels, I guess the ordinal could be useful label as well. e.g. to list first node pvc in all clusters

pkg/controller/scyllacluster/resource_test.go
```diff
@@ -617,11 +612,18 @@ func TestStatefulSetForRack(t *testing.T) {
 			Spec: appsv1.StatefulSetSpec{
 				Replicas: pointer.Ptr(int32(0)),
 				Selector: &metav1.LabelSelector{
-					MatchLabels: newBasicStatefulSetLabels(),
+					MatchLabels: map[string]string{
+						"app":                          "scylla",
+						"app.kubernetes.io/managed-by": "scylla-operator",
+						"app.kubernetes.io/name":       "scylla",
```
tnozicka, line 618: we might want to keep the selector labels minimal (none of them is technically needed/used; note we have ownerRefs)

pkg/controller/scyllacluster/resource_test.go
zimnx: why?

pkg/controller/scyllacluster/resource.go
zimnx: PVCTemplate is immutable

pkg/controller/scyllacluster/resource_test.go
zimnx: we might, but it's immutable so we cannot change it now.

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go

pkg/controller/scyllacluster/resource.go
zimnx: done

pkg/controller/scyllacluster/resource_test.go
tnozicka: yep, let's keep it in mind if we ever go for cascade delete

pkg/controller/scyllacluster/resource_test.go
tnozicka: avoids parsing the format string (but we can spare the extra cycle in test :) )

tnozicka (approved):
/approve
/lgtm

thanks!

PR #1641, opened 2023-12-13
Determine host for Scylla client using ScyllaCluster expose options

### LINKED ISSUE #1638: Upgrade procedure over PodIPs gets clusters stuck in a degraded state
### What happened?

If you try to upgrade a ScyllaCluster exposed over PodIPs, it will get stuck in a degraded state.
See e.g. https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1508/pull-scylla-operator-e2e-gke-parallel/1734923770008702976#1:test-build-log.txt%3A2515.

Example conditions from a local run:
```
  - lastTransitionTime: "2023-12-12T17:08:32Z"
    message: 'StatefulSetControllerDegraded: service e2e-test-scyllacluster-jfc9h-l7xbw/basic-qs2p2-us-east-1-rack-0-0
      doesn''t have a ClusterIP'
    observedGeneration: 2
    reason: StatefulSetControllerDegraded_Error
    status: "True"
    type: Degraded
```

The upgrade procedure should support ScyllaClusters exposed over PodIPs.

- Create a ScyllaCluster exposed over PodIPs and wait for it to roll out.
- Upgrade Scylla version.

### Scylla Operator version

master

<details>
```console
Client Version: v1.28.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.3
```

Kubernetes platform info:
n/a
</details>

https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1508/pull-scylla-operator-e2e-gke-parallel/1734923770008702976/

_No response_

rzetelskik:
**Description of your changes:** Currently, the functions used to get a host for a Scylla client connection lack support for clusters exposed over PodIPs. Since the functions are used in the StatefulSet controller, it breaks the ScyllaCluster ugprade procedure for clusters exposed over PodIPs. This PR makes the controller use expose options to determine the address to use as a forced host for Scylla client.

Resolves #1638

/kind bug
/priority critical-urgent

### DIFF
--- a/pkg/controller/scyllacluster/sync_statefulsets.go
+++ b/pkg/controller/scyllacluster/sync_statefulsets.go
@@ -145,7 +145,7 @@ func (scc *Controller) beforeUpgrade(ctx context.Context, sc *scyllav1.ScyllaClu
 	klog.V(2).InfoS("Running pre-upgrade hook", "ScyllaCluster", klog.KObj(sc))
 	defer klog.V(2).InfoS("Finished running pre-upgrade hook", "ScyllaCluster", klog.KObj(sc))
 
-	hosts, err := controllerhelpers.GetRequiredScyllaHosts(sc, services)
+	hosts, err := controllerhelpers.GetRequiredScyllaHosts(sc, services, scc.podLister)
 	if err != nil {
 		return true, err
 	}
@@ -184,7 +184,7 @@ func (scc *Controller) afterUpgrade(ctx context.Context, sc *scyllav1.ScyllaClus
 	klog.V(2).InfoS("Running post-upgrade hook", "ScyllaCluster", klog.KObj(sc))
 	defer klog.V(2).InfoS("Finished running post-upgrade hook", "ScyllaCluster", klog.KObj(sc))
 
-	hosts, err := controllerhelpers.GetRequiredScyllaHosts(sc, services)
+	hosts, err := controllerhelpers.GetRequiredScyllaHosts(sc, services, scc.podLister)
 	if err != nil {
 		return err
 	}
@@ -230,8 +230,13 @@ func (scc *Controller) beforeNodeUpgrade(ctx context.Context, sc *scyllav1.Scyll
 	}
 
 	// Drain the node.
+	podName := naming.PodNameFromService(svc)
+	pod, err := scc.podLister.Pods(sc.Namespace).Get(podName)
+	if err != nil {
+		return false, fmt.Errorf("can't get pod %q: %w", naming.ManualRef(sc.Namespace, podName), err)
+	}
 
-	host, err := controllerhelpers.GetScyllaIPFromService(svc)
+	host, err := controllerhelpers.GetScyllaHost(sc, svc, pod)
 	if err != nil {
 		return true, err
 	}
@@ -294,7 +299,6 @@ func (scc *Controller) beforeNodeUpgrade(ctx context.Context, sc *scyllav1.Scyll
 	// https://github.com/kubernetes/kubernetes/issues/67250
 	// Kubernetes can't evict pods when DesiredHealthy == 0 and it's already down, so we need to use DELETE
 	// to succeed even when having just one replica.
-	podName := svcName
 	klog.V(2).InfoS("Deleting Pod", "ScyllaCluster", klog.KObj(sc), "Pod", naming.ManualRef(sc.Namespace, podName))
 	err = scc.kubeClient.CoreV1().Pods(sc.Namespace).Delete(ctx, podName, metav1.DeleteOptions{})
 	if err != nil {
@@ -311,7 +315,19 @@ func (scc *Controller) beforeNodeUpgrade(ctx context.Context, sc *scyllav1.Scyll
 }
 
 func (scc *Controller) afterNodeUpgrade(ctx context.Context, sc *scyllav1.ScyllaCluster, sts *appsv1.StatefulSet, ordinal int32, services map[string]*corev1.Service) error {
-	host, err := controllerhelpers.GetScyllaHost(sts.Name, ordinal, services)
+	svcName := fmt.Sprintf("%s-%d", sts.Name, ordinal)
+	svc, ok := services[svcName]
+	if !ok {
+		return fmt.Errorf("missing service %q", naming.ManualRef(sc.Namespace, svcName))
+	}
+
+	podName := naming.PodNameFromService(svc)
+	pod, err := scc.podLister.Pods(sc.Namespace).Get(podName)
+	if err != nil {
+		return fmt.Errorf("can't get pod %q: %w", naming.ManualRef(sc.Namespace, podName), err)
+	}
+
+	host, err := controllerhelpers.GetScyllaHost(sc, svc, pod)
 	if err != nil {
 		return err
 	}
--- a/pkg/controllerhelpers/scylla.go
+++ b/pkg/controllerhelpers/scylla.go
@@ -15,46 +15,79 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/errors"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	corev1schedulinghelpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 )
 
-func GetScyllaIPFromService(svc *corev1.Service) (string, error) {
-	if svc.Spec.Type != corev1.ServiceTypeClusterIP {
-		return "", fmt.Errorf("service %s is of type %q instead of %q", naming.ObjRef(svc), svc.Spec.Type, corev1.ServiceTypeClusterIP)
+func GetScyllaHost(sc *scyllav1.ScyllaCluster, svc *corev1.Service, pod *corev1.Pod) (string, error) {
+	// Assume API's default.
+	nodeBroadcastAddressType := scyllav1.BroadcastAddressTypeServiceClusterIP
+	if sc.Spec.ExposeOptions != nil && sc.Spec.ExposeOptions.BroadcastOptions != nil {
+		nodeBroadcastAddressType = sc.Spec.ExposeOptions.BroadcastOptions.Nodes.Type
 	}
 
-	if svc.Spec.ClusterIP == corev1.ClusterIPNone {
-		return "", fmt.Errorf("service %s doesn't have a ClusterIP", naming.ObjRef(svc))
-	}
-
-	return svc.Spec.ClusterIP, nil
+	return GetScyllaBroadcastAddress(nodeBroadcastAddressType, svc, pod)
 }
 
-func GetScyllaHost(statefulsetName string, ordinal int32, services map[string]*corev1.Service) (string, error) {
-	svcName := fmt.Sprintf("%s-%d", statefulsetName, ordinal)
-	svc, found := services[svcName]
-	if !found {
-		return "", fmt.Errorf("missing service %q", svcName)
-	}
+func GetScyllaBroadcastAddress(broadcastAddressType scyllav1.BroadcastAddressType, svc *corev1.Service, pod *corev1.Pod) (string, error) {
+	switch broadcastAddressType {
+	case scyllav1.BroadcastAddressTypeServiceLoadBalancerIngress:
+		if len(svc.Status.LoadBalancer.Ingress) < 1 {
+			return "", fmt.Errorf("service %q does not have an ingress status", naming.ObjRef(svc))
+		}
 
-	ip, err := GetScyllaIPFromService(svc)
-	if err != nil {
-		return "", err
-	}
+		if len(svc.Status.LoadBalancer.Ingress[0].IP) != 0 {
+			return svc.Status.LoadBalancer.Ingress[0].IP, nil
+		}
 
-	return ip, nil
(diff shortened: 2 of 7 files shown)


pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,114 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/errors"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1schedulinghelpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 )
 
-func GetScyllaIPFromService(svc *corev1.Service) (string, error) {
+func GetScyllaHost(ctx context.Context, sc *scyllav1.ScyllaCluster, coreV1Client corev1client.CoreV1Interface, svc *corev1.Service) (string, error) {
```
zimnx, line 24: lets cover it with unit tests

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,114 @@ import (
...
+		}
+
+		return host, nil
+	default:
+		return "", fmt.Errorf("unsupported broadcast address type: %q", sc.Spec.ExposeOptions.BroadcastOptions.Clients.Type)
+	}
+}
+
+func getServiceClusterIP(svc *corev1.Service) (string, error) {
 	if svc.Spec.Type != corev1.ServiceTypeClusterIP {
```
zimnx, line 77: This is quite unexpected error for function called this way. For example LB Service has ClusterIP, but user of this function wouldn't be able to get it.

pkg/controllerhelpers/scylla.go
rzetelskik: done

pkg/controllerhelpers/scylla.go
rzetelskik: thanks, remove this check

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,110 @@ import (
...
+			if err != nil {
+				return "", fmt.Errorf("can't get pod for service %q: %w", naming.ObjRef(svc), err)
+			}
+
+			host, err := getPodStatusPodIP(pod)
+			if err != nil {
+				return "", fmt.Errorf("can't get PodIP from pod %q status: %w", naming.ObjRef(pod), err)
+			}
+
+			return host, nil
```
tnozicka, line 60: this is similar to what we do to create member config and it should likely live in a shared function

zimnx (approved):
lgtm
/assign tnozicka

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,110 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/errors"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1schedulinghelpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 )
 
-func GetScyllaIPFromService(svc *corev1.Service) (string, error) {
-	if svc.Spec.Type != corev1.ServiceTypeClusterIP {
-		return "", fmt.Errorf("service %s is of type %q instead of %q", naming.ObjRef(svc), svc.Spec.Type, corev1.ServiceTypeClusterIP)
+func GetScyllaHost(ctx context.Context, sc *scyllav1.ScyllaCluster, coreV1Client corev1client.CoreV1Interface, svc *corev1.Service) (string, error) {
```
tnozicka, line 24: Maybe we should make a function that gets a tuple client+node IPs and use the same in one in member configuration

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,110 @@ import (
...
+		return "", fmt.Errorf("pod does not have a PodIP")
 	}
 
-	ip, err := GetScyllaIPFromService(svc)
-	if err != nil {
-		return "", err
+	return pod.Status.PodIP, nil
+}
+
+func getServiceExternalAddress(svc *corev1.Service) (string, error) {
```
tnozicka, line 92: is this different from https://github.com/scylladb/scylla-operator/blob/eb144ac50175c040e486f47eb9468908181ce678/pkg/sidecar/identity/member.go#L55-L73

pkg/controllerhelpers/scylla.go
rzetelskik: >  a function that gets a tuple client+node IPs

how would that work? I can see it taking service and pod objects but don't see how this could work with the addresses directly

pkg/controllerhelpers/scylla.go
rzetelskik: not really, the two can be made into a shared function

pkg/controllerhelpers/scylla.go
tnozicka: i wrote the "gets" from the callers point of view, so the function would return the tuple.

rzetelskik:
refactored to share code between member configuration and scylla helpers

zimnx (approved):

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,78 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/errors"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1schedulinghelpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 )
 
-func GetScyllaIPFromService(svc *corev1.Service) (string, error) {
-	if svc.Spec.Type != corev1.ServiceTypeClusterIP {
-		return "", fmt.Errorf("service %s is of type %q instead of %q", naming.ObjRef(svc), svc.Spec.Type, corev1.ServiceTypeClusterIP)
+func GetScyllaHost(ctx context.Context, sc *scyllav1.ScyllaCluster, coreV1Client corev1client.CoreV1Interface, svc *corev1.Service) (string, error) {
```
tnozicka, line 24: > coreV1Client corev1client.CoreV1Interface

why is this using a live client and not a lister?
(applies globally)

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,78 @@ import (
...
+	// Assume API's default.
+	if sc.Spec.ExposeOptions == nil || sc.Spec.ExposeOptions.BroadcastOptions == nil {
+		if svc.Spec.ClusterIP == corev1.ClusterIPNone {
+			return "", fmt.Errorf("service %q does not have a ClusterIP", naming.ObjRef(svc))
+		}
+
+		return svc.Spec.ClusterIP, nil
```
tnozicka, line 31: this branch is also in `GetScyllaBroadcastAddress` - could this just default the broadcast option object to scyllav1.BroadcastAddressTypeServiceClusterIP?

pkg/controllerhelpers/scylla.go
```diff
@@ -15,46 +16,78 @@ import (
...
+		if len(svc.Status.LoadBalancer.Ingress[0].IP) != 0 {
+			return svc.Status.LoadBalancer.Ingress[0].IP, nil
+		}
+
+		if len(svc.Status.LoadBalancer.Ingress[0].Hostname) != 0 {
+			return svc.Status.LoadBalancer.Ingress[0].Hostname, nil
+		}
+
+		return "", fmt.Errorf("service %q does not have an external address", naming.ObjRef(svc))
+	case scyllav1.BroadcastAddressTypeServiceClusterIP:
```
tnozicka, line 59: nit: `\n` would help split the cases

pkg/controller/scyllacluster/sync_statefulsets.go
```diff
@@ -311,7 +310,12 @@ func (scc *Controller) beforeNodeUpgrade(ctx context.Context, sc *scyllav1.Scyll
 }
 
 func (scc *Controller) afterNodeUpgrade(ctx context.Context, sc *scyllav1.ScyllaCluster, sts *appsv1.StatefulSet, ordinal int32, services map[string]*corev1.Service) error {
-	host, err := controllerhelpers.GetScyllaHost(sts.Name, ordinal, services)
+	svcName := fmt.Sprintf("%s-%d", sts.Name, ordinal)
```
tnozicka, line 313: naming.MemberServiceName?

pkg/sidecar/identity/member.go
```diff
@@ -173,32 +132,10 @@ func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, ex
...
-		}
-		res = append(res, svc.Spec.ClusterIP)
-	case scyllav1.BroadcastAddressTypePodIP:
-		if len(pod.Status.PodIP) == 0 {
-			return nil, fmt.Errorf("pod %q does not have IP address, despite being set as broadcasted", naming.ObjRef(pod))
-		}
-		res = append(res, pod.Status.PodIP)
-	default:
-		return nil, fmt.Errorf("unsupported node broadcast address type %q", m.NodesBroadcastAddressType)
+	localSeed, err := controllerhelpers.GetScyllaBroadcastAddress(m.NodesBroadcastAddressType, svc, pod)
```
tnozicka, line 135: this seems to be called in the "constructor", maybe you can save the result on the object rather then the type?

tnozicka (changes requested):

pkg/sidecar/identity/member.go
rzetelskik: > maybe you can save the result on the object rather then the type?

could you please rephrase that? I've read it like 10 times and don't understand what you mean by that

pkg/controller/scyllacluster/sync_statefulsets.go
rzetelskik: `naming.MemberServiceName` takes a rack spec, so I'd have to get the rack name from the statefulset, get the rack spec from SC to then get the statefulset name from rack inside the func, it seems completely redundant to me.

tnozicka: no worries, seems like some of our helpers take bigger scope then they need

pkg/sidecar/identity/member.go
tnozicka: This function is already executed in `NewMember`, only the type is stored and the function is executed again here with `m.NodesBroadcastAddressType`. I am suggesting to cache the result instead of the argument (`m.NodesBroadcastAddressType`) and avoid running it a second time.
https://github.com/scylladb/scylla-operator/blob/53078ab596ba79e5f0b01475b31840f9aff6c5ce/pkg/sidecar/identity/member.go#L59-L65

rzetelskik: oh, it's called for a different node here (`pod := otherPods[0]`)

tnozicka: thx, my bad. from the diff, it seemed like the same one

rzetelskik: I see that I accidentally moved the comment explaining the use of member's node broadcast address type for it (`// Assume nodes share broadcast address type and it's immutable.`) so I'll fix that

pkg/controllerhelpers/scylla.go
rzetelskik: I figured I'll just make it take a pod as argument - easier to test

pkg/controllerhelpers/scylla.go
rzetelskik: done

tnozicka (commented):
/approve
(lgtm, thanks!)

/assign @zimnx

tnozicka (approved):

pkg/sidecar/identity/member.go
```diff
@@ -173,32 +132,11 @@ func (m *Member) GetSeeds(ctx context.Context, coreClient v1.CoreV1Interface, ex
...
-	default:
-		return nil, fmt.Errorf("unsupported node broadcast address type %q", m.NodesBroadcastAddressType)
+	// Assume nodes share broadcast address type and it's immutable.
+	localSeed, err := controllerhelpers.GetScyllaBroadcastAddress(m.NodesBroadcastAddressType, svc, pod)
+	if err != nil {
+		return nil, fmt.Errorf("can't get node broadcast address for service %q: %w", naming.ObjRef(svc), err)
 	}
 
-	return res, nil
+	return append(res, localSeed), nil
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/controller/scyllacluster/sync_statefulsets.go
```diff
@@ -230,8 +230,12 @@ func (scc *Controller) beforeNodeUpgrade(ctx context.Context, sc *scyllav1.Scyll
 	}
 
 	// Drain the node.
+	pod, err := scc.podLister.Pods(sc.Namespace).Get(naming.PodNameFromService(svc))
+	if err != nil {
+		return false, fmt.Errorf("can't get pod for service %q: %w", naming.ObjRef(svc), err)
```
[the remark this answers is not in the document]
rzetelskik: done

zimnx (approved):
/lgtm

Please backport it to 1.11. Thanks!

PR #1644, opened 2023-12-15
[WIP] Removed force restarts in haproxy example

tnozicka:
The haproxy example shouldn't restart every 66s.


zimnx:
> The haproxy example shouldn't restart every 66s.

This parameter controls how long to wait for inflight connections to close when restart is required, not restart interval.

tnozicka:
hm, I saw it constantly restarting on its own. I'll have another look, thx.

PR #1649, opened 2023-12-20
Loosen validation for LoadBalancerClass in NodeServiceTemplate

zimnx:
NameIsDNSSubdomain was too strict to validate this field, preventing from setting load balancer classes available in cloud environments.


tnozicka (approved):
/approve
/lgtm

thanks

zimnx:
Flake: https://github.com/scylladb/scylla-operator/issues/1525#issuecomment-1864678537

/test e2e-gke-parallel

PR #1653, opened 2023-12-21
add webhook server resource defaults

nevlkv:
If helm chart deployed in k8s with checking mandatory resources section  we needs  setting LimitRange in the namespace


zimnx:
Please don't change what's the default. I'm fine with making webhook resources configurable, but default limits shouldn't be hardcoded.

PR #1655, opened 2023-12-22
Support update of StatefulSet selector 

zimnx:
When Pods managed by existing StatefulSet matches new selector, StatefulSet
is orphan deleted and recreated.

/cc @tnozicka

### DIFF
--- a/pkg/resourceapply/apps.go
+++ b/pkg/resourceapply/apps.go
@@ -2,9 +2,13 @@ package resourceapply
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/scylladb/scylla-operator/pkg/pointer"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
 	appsv1listers "k8s.io/client-go/listers/apps/v1"
 	"k8s.io/client-go/tools/record"
@@ -17,7 +21,31 @@ func ApplyStatefulSetWithControl(
 	required *appsv1.StatefulSet,
 	options ApplyOptions,
 ) (*appsv1.StatefulSet, bool, error) {
-	return ApplyGeneric[*appsv1.StatefulSet](ctx, control, recorder, required, options)
+	return ApplyGenericWithHandlers[*appsv1.StatefulSet](
+		ctx,
+		control,
+		recorder,
+		required,
+		options,
+		nil,
+		func(required *appsv1.StatefulSet, existing *appsv1.StatefulSet) (string, *metav1.DeletionPropagation, error) {
+			if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
+				existingPodLabels := existing.Spec.Template.Labels
+				requiredSelector, err := metav1.LabelSelectorAsSelector(required.Spec.Selector)
+				if err != nil {
+					return "", nil, fmt.Errorf("can't parse required StatefulSet selector: %w", err)
+				}
+
+				if !requiredSelector.Matches(labels.Set(existingPodLabels)) {
+					return "", nil, fmt.Errorf("required StatefulSet selector %q doesn't match existing Pod Labels set %v", requiredSelector, existingPodLabels)
+				}
+
+				return "spec.selector is immutable", pointer.Ptr(metav1.DeletePropagationOrphan), nil
+			}
+
+			return "", nil, nil
+		},
+	)
 }
 
 func ApplyStatefulSet(
@@ -56,11 +84,11 @@ func ApplyDaemonSetWithControl(
 		required,
 		options,
 		nil,
-		func(required *appsv1.DaemonSet, existing *appsv1.DaemonSet) string {
+		func(required *appsv1.DaemonSet, existing *appsv1.DaemonSet) (string, *metav1.DeletionPropagation, error) {
 			if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
-				return "spec.selector is immutable"
+				return "spec.selector is immutable", nil, nil
 			}
-			return ""
+			return "", nil, nil
 		},
 	)
 }
--- a/pkg/resourceapply/apps_test.go
+++ b/pkg/resourceapply/apps_test.go
@@ -46,8 +46,17 @@ func TestApplyStatefulSet(t *testing.T) {
 			},
 			Spec: appsv1.StatefulSetSpec{
 				Replicas: pointer.Ptr(int32(3)),
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"foo": "bar",
+					},
+				},
 				Template: corev1.PodTemplateSpec{
-					ObjectMeta: metav1.ObjectMeta{},
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"foo": "bar",
+						},
+					},
 					Spec: corev1.PodSpec{
 						Containers: []corev1.Container{
 							{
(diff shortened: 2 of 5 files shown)


tnozicka (commented):
I think we should make a safety check that the new selector matches the old  podLabels or this can go quite wrong at some point

pkg/resourceapply/apps.go
```diff
@@ -17,7 +17,39 @@ func ApplyStatefulSetWithControl(
...
+			}
+			if !equality.Semantic.DeepEqual(existing.Spec.UpdateStrategy, required.Spec.UpdateStrategy) {
+				return "spec.updateStrategy is immutable"
+			}
+			if !equality.Semantic.DeepEqual(existing.Spec.RevisionHistoryLimit, required.Spec.RevisionHistoryLimit) {
+				return "spec.revisionHistoryLimit is immutable"
+			}
+			if !equality.Semantic.DeepEqual(existing.Spec.Ordinals, required.Spec.Ordinals) {
+				return "spec.ordinals is immutable"
+			}
```
tnozicka, line 48: I think I'd limit this to explicit immutable changes that we require because of a reason (like when we broke the selector upgrade). In general these are immutable because changing them requires action or has consequences that often needs to be compensated for. Limiting this avoids testing the unused path, debating the procedures and possibly accidentally breaking something in the future. This is more important for sts as it serves the scylla workload.

so in this case: if a selector is still selecting the old pods we can orphan delete the sts and recreate it

pkg/resourceapply/apps.go
```diff
@@ -17,7 +17,39 @@ func ApplyStatefulSetWithControl(
...
 ) (*appsv1.StatefulSet, bool, error) {
-	return ApplyGeneric[*appsv1.StatefulSet](ctx, control, recorder, required, options)
+	return ApplyGenericWithHandlers[*appsv1.StatefulSet](
+		ctx,
+		control,
+		recorder,
+		required,
+		options,
+		nil,
+		func(required *appsv1.StatefulSet, existing *appsv1.StatefulSet) string {
```
tnozicka, line 27: the selector recreate is only graceful, if done with orphan delete, so i think we need it enforce at this or lower level, not the caller that can forget to set it.

tnozicka: could the propagation policy be a return value here?

zimnx: done

pkg/resourceapply/apps.go
zimnx: I agree, done

zimnx:
> I think we should make a safety check that the new selector matches the old podLabels or this can go quite wrong at some point

Good idea, done.

pkg/resourceapply/apps.go
```diff
@@ -17,7 +20,31 @@ func ApplyStatefulSetWithControl(
...
+				requiredSelector, err := metav1.LabelSelectorAsSelector(required.Spec.Selector)
+				if err != nil {
+					return "", metav1.DeletePropagationBackground, fmt.Errorf("can't parse required StatefulSet selector: %w", err)
+				}
+
+				if !requiredSelector.Matches(labels.Set(existingPodLabels)) {
+					return "", metav1.DeletePropagationBackground, fmt.Errorf("required StatefulSet selector %s doesn't match existing Pod Labels set %v", required.Spec.Selector, existingPodLabels)
+				}
+
+				return "spec.selector is immutable", metav1.DeletePropagationOrphan, nil
```
tnozicka, line 42: nit: returning empty string would give user the native error from sts or are you trying to compensate for it not being specific enough?

pkg/resourceapply/helpers.go
```diff
@@ -244,7 +244,7 @@ func ApplyGenericWithHandlers[T kubeinterfaces.ObjectInterface](
 	required T,
 	options ApplyOptions,
 	projectFunc func(required *T, existing T),
-	getRecreateReasonFunc func(required T, existing T) string,
+	getRecreateReasonFunc func(required T, existing T) (string, metav1.DeletionPropagation, error),
```
tnozicka, line 247: i think this would make the changes and UX nicer
```suggestion
	getRecreateReasonFunc func(required T, existing T) (string, *metav1.DeletionPropagation, error),
```

pkg/resourceapply/apps.go
zimnx: Non empty recreate reason controls whether deletion happens and is being part of the log message associated with it.
https://github.com/scylladb/scylla-operator/blob/master/pkg/resourceapply/helpers.go#L321-L327

pkg/resourceapply/helpers.go
zimnx: done

pkg/resourceapply/apps.go
tnozicka: I've mixed the lines, disregard this

pkg/resourceapply/apps_test.go
```diff
@@ -493,6 +497,60 @@ func TestApplyStatefulSet(t *testing.T) {
 			expectedErr:     nil,
 			expectedEvents:  []string{"Normal StatefulSetUpdated StatefulSet default/test updated"},
 		},
+		{
+			name: "deletes and creates the StatefulSet when selector differs",
+			existing: []runtime.Object{
+				newSts(),
```
tnozicka, line 503: i think this should have a selector set to test the matching later on

pkg/resourceapply/apps_test.go
```diff
@@ -493,6 +497,60 @@ func TestApplyStatefulSet(t *testing.T) {
 			expectedErr:     nil,
 			expectedEvents:  []string{"Normal StatefulSetUpdated StatefulSet default/test updated"},
 		},
+		{
+			name: "deletes and creates the StatefulSet when selector differs",
```
tnozicka, line 501: when selector is changed and still matches the old pods

pkg/resourceapply/apps_test.go
```diff
@@ -493,6 +497,60 @@ func TestApplyStatefulSet(t *testing.T) {
...
+				"Normal StatefulSetDeleted StatefulSet default/test deleted",
+				"Normal StatefulSetCreated StatefulSet default/test created",
+			},
+		},
+		{
+			name: "apply fails when StatefulSet selector differs and existing Pod labels doesn't match new selector",
+			existing: []runtime.Object{
+				func() *appsv1.StatefulSet {
+					sts := newSts()
+					sts.Spec.Template.Labels = nil
```
tnozicka, line 536: closer to a real scenario there would be at least one label

pkg/resourceapply/helpers.go
```diff
@@ -326,7 +335,6 @@ func ApplyGenericWithHandlers[T kubeinterfaces.ObjectInterface](
 			"Ref", naming.ObjRefWithUID(existing),
 		)
 
-		propagationPolicy := metav1.DeletePropagationBackground
```
tnozicka, line 329: if propagationPolicy == nil {
			propagationPolicy = pointer.Ptr(metav1.DeletePropagationBackground)
		}

pkg/resourceapply/helpers.go
```diff
@@ -315,8 +315,17 @@ func ApplyGenericWithHandlers[T kubeinterfaces.ObjectInterface](
...
-		recreateReason = getRecreateReasonFunc(requiredCopy, existing)
+		var overridePropagationPolicy *metav1.DeletionPropagation
+		recreateReason, overridePropagationPolicy, err = getRecreateReasonFunc(requiredCopy, existing)
+		if err != nil {
+			return *new(T), false, fmt.Errorf("can't get recreate reason: %w", err)
+		}
+
+		if overridePropagationPolicy != nil {
+			propagationPolicy = *overridePropagationPolicy
+		}
```
tnozicka, line 328:
```suggestion
		recreateReason, propagationPolicy, err = getRecreateReasonFunc(requiredCopy, existing)
		if err != nil {
			return *new(T), false, fmt.Errorf("can't get recreate reason: %w", err)
		}
```

pkg/resourceapply/helpers.go
```diff
@@ -315,8 +315,17 @@ func ApplyGenericWithHandlers[T kubeinterfaces.ObjectInterface](
 	}
 
 	var recreateReason string
+	propagationPolicy := metav1.DeletePropagationBackground
```
tnozicka, line 318:
```suggestion
	var propagationPolicy *metav1.DeletionPropagation
```

tnozicka (approved):
few nits, rest lgtm
/approve

pkg/resourceapply/apps_test.go
tnozicka: I suppose this could be in the default manifest so it better resembles an sts that would pass validation

pkg/resourceapply/helpers.go
tnozicka: (gh didn't allow me to suggest a change over non-changed+deleted line so this is in 3 changes but the essentially jsut eliminate the need for the override variable)

pkg/resourceapply/apps_test.go
zimnx: done

pkg/resourceapply/apps_test.go
zimnx: done

pkg/resourceapply/apps_test.go
zimnx: done

pkg/resourceapply/helpers.go
zimnx: done

pkg/resourceapply/helpers.go
zimnx: done

pkg/resourceapply/helpers.go
zimnx: done

pkg/resourceapply/apps_test.go
```diff
@@ -47,7 +47,11 @@ func TestApplyStatefulSet(t *testing.T) {
 			Spec: appsv1.StatefulSetSpec{
```
tnozicka, line 47: empty selector is invalid sts

https://github.com/kubernetes/kubernetes/blob/c3384191ea6410121339a48d1c68ba6f4b3ab31a/pkg/apis/apps/validation/validation.go#L139-L143

zimnx: done

tnozicka (approved):
/lgtm

thanks!

PR #1664, opened 2023-12-28
Inherit and customize ScyllaCluster labels and annotations

### LINKED ISSUE #1592: Extend API with customizable labels+annotations
Epic https://github.com/scylladb/scylla-operator/issues/1590

tnozicka:
This PR allows users to set metadata for several managed objects as described in [SOEP-1665](https://github.com/scylladb/scylla-operator/blob/9fbd70d/enhancements/proposals/1665-metadata/README.md)

Resolves #1592 #1668


pkg/api/scylla/v1/types_cluster.go
```diff
@@ -21,8 +21,22 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type ObjectTemplateMetadata struct {
+	// labels is a custom key value map that gets merged with managed object annotations.
```
zimnx, line 25: s/annotations/labels

pkg/controller/scyllacluster/resource.go
```diff
@@ -896,14 +971,19 @@ func MakeAgentAuthTokenSecret(c *scyllav1.ScyllaCluster, authToken string) (*cor
...
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      naming.AgentAuthTokenSecretName(c.Name),
 			Namespace: c.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				*metav1.NewControllerRef(c, scyllaClusterControllerGVK),
 			},
-			Labels: naming.ClusterLabels(c),
+			Labels:      labels,
+			Annotations: maps.Clone(c.Annotations),
```
zimnx, line 986: nit: lets make it consistent that labels and annotations are defined in variables and these variables are used in spec

zimnx:
Also, please add support for these fields in Helm chart template

pkg/controller/scyllacluster/resource.go
tnozicka: tbh, I am not seeing the value in that as it add extra jump / dereference for the reader and doesn't use the value + it' not extensible as the `Clone` and empty map pattern differ anyways.

pkg/api/scylla/v1/types_cluster.go
tnozicka: done, thx

pkg/controller/scyllacluster/resource.go
zimnx: I meant to be consistent with style within single commit. In other places (MakeServiceAccount, MakeRoleBinding) there's an extra variable defined - which I like more because adding anything new values requires shorter diff - and in other there's explicit clone.

zimnx:
Thanks for updates.
/lgtm

Issue #1590, opened 2023-11-23
Allow configuring annotation and labels for managed resources


tnozicka:
We should allow defining global labels and annotation for all objects that we create and merge those with our metadata and also allow finer grained setting just for PodTemplate (e.g. for Datadog scrape labels). Also possibly identity and member services separately.

### Related Issues
- [x] #918
- [x] #1024

```[tasklist]
- [ ] https://github.com/scylladb/scylla-operator/issues/1592
- [ ] https://github.com/scylladb/scylla-operator/issues/1666
- [ ] https://github.com/scylladb/scylla-operator/issues/1668
```

tnozicka:
done

PR #1665, opened 2023-12-28
Add proposal for specifying ScyllaCluster metadata

### LINKED ISSUE #1666: Add proposal for metadata configuration on managed objects
Add proposal for epic #1590

tnozicka:
Proposal for managing metadata (labels+annotation) on ScyllaClusters.

Epic https://github.com/scylladb/scylla-operator/issues/1590

Resolves #1666

### DIFF
--- a/enhancements/proposals/1665-metadata/README.md
+++ b/enhancements/proposals/1665-metadata/README.md
@@ -1,0 +1,145 @@
+# Support for ScyllaDB Alternator
+
+## Summary
+
+This proposal aims to introduce basic support for handling metadata in managed objects by ScyllaClusters.
+It aims to make minimal changes to satisfy the use cases presented so far.
+
+## Motivation
+
+ScyllaCluster users need to specify labels and annotations for various reasons including external integrations or object discovery / tracking.
+
+### Goals
+
+- Inherit ScyllaCluster labels and annotation into managed objects (aka default labels)
+- Allow specifying explicit labels and annotations for node services
+- Allow specifying explicit labels and annotations for pods
+- Allow specifying explicit labels and annotations for initial persistent volume claims
+- Allow specifying explicit labels and annotations for ingresses
+
+### Non-Goals
+
+- Allow unique metadata for every single object we create 
+
+## Proposal
+
+I'd like to use the existing ScyllaCluster.metadata (labels and annotations) and inherit them in every object we create,
+unless there is an override.
+In the case when using an override, the ScyllaCluster.metadata won't be used.
+
+I propose to extend the ScyllaCluster API to allow overriding metadata for pods, node services and ingresses.
+
+In case there is a conflict with managed metadata, the managed metadata will win.
+
+These changes are purely intended for metadata and shall not change the selectors used.
+
+### User Stories
+
+#### DataDog metrics
+As a user I want to configure [DataDog metrics](https://docs.datadoghq.com/containers/kubernetes/prometheus/?tab=kubernetesadv2#setup) for certain pods.
+This is done by adding `ad.datadoghq.com/<CONTAINER_IDENTIFIER>.checks` to Pod's annotations.
+
+#### Third Party Service and Ingress options
+As a user I want to configure additional options for my service or ingress.
+These third party options are usually controlled by adding annotations or labels, e.g.:
+- [`networking.gke.io/load-balancer-type: "Internal"` annotation for GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing)
+- [`service.beta.kubernetes.io/aws-load-balancer-eip-allocations: eipalloc-xxxxxxxxxxxxxxxxx,eipalloc-yyyyyyyyyyyyyyyyy` annotation for assigning elastic IPs](https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html)
+- [Ingress Controller sharding using labels and selectors in OCP](https://docs.openshift.com/container-platform/4.14/networking/ingress-sharding.html)
+
+#### Cost tracking
+As a user I want to add a label to all of my resources and use it to track costs.
+
+#### PagerDuty
+As a user I want to add extra label to my workload to route dynamic PagerDuty alerts.
+
+#### PVC properties
+As a user I want to add extra annotations to configure additional properties for PVCs, like "ebs.csi.aws.com/iops" or "ebs.csi.aws.com/throughput".
+
+### Risks and Mitigations
+
+Not known.
+
+## Design Details
+
+### API changes for ScyllaClusters.scylla.scylladb.com/v1
+
+The following API changes show how we'd approximately extend `scyllaclusters.scylla.scylladb.com/v1.spec` specification. Only new fields are shown for existing structs.
+
+```golang
+package v1
+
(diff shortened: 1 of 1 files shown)


enhancements/proposals/1665-metadata/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+These changes are purely intended for metadata and shall not change the selectors used.
+
+### User Stories
+
+#### DataDog metrics
+As a user I want to configure [DataDog metrics](https://docs.datadoghq.com/containers/kubernetes/prometheus/?tab=kubernetesadv2#setup) for certain pods.
+This is done by adding `ad.datadoghq.com/<CONTAINER_IDENTIFIER>.checks` to Pod's annotations.
+
+#### Third Party Service and Ingress options
```
rzetelskik, line 41: Isn't this already covered with `.spec.exposeOptions.cql.ingress.annotations` and `.spec.exposeOptions.nodeService.annotations`?

rzetelskik, line 41: Isn't this already covered with `.spec.exposeOptions.cql.ingress.annotations` and `.spec.exposeOptions.nodeService.annotations`? Which ones take precedence in case of a conflict? Are the ones coming from `.spec.exposeOptions` considered managed?

enhancements/proposals/1665-metadata/README.md
```diff
@@ -0,0 +1,122 @@
...
+It aims to make minimal changes to satisfy the use cases presented so far.
+
+## Motivation
+
+ScyllaCluster users need to specify labels and annotations for various reasons including external integrations or object discovery / tracking.
+
+### Goals
+
+- Inherit ScyllaCluster labels and annotation into managed objects (aka default labels)
+- Allow specifying distinct labels and annotations for node services
```
rzetelskik, line 15: Wouldn't `distinct` suggest you'd be able to specify it per pod/service/ingress?

enhancements/proposals/1665-metadata/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Non-Goals
+
+- Allow unique metadata for every single object we create 
+
+## Proposal
+
+I'd like to use the existing ScyllaCluster.metadata (labels and annotations) and inherit them in every object we create,
+unless there is an override.
+In the case when using an override, the ScyllaCluster.metadata won't be used.
```
rzetelskik, line 27: Wouldn't it be more intuitive to merge the two? With the priority being managed>object-defined>cluster-defined. I can imagine wanting to have some labels/annotations applied to all objects and then a subset applied to say services only.

enhancements/proposals/1665-metadata/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+type IngressOptions struct {
+	ObjectTemplateMetadata `json:",inline"`
+	
+	// ...
+}
+
+
+type NodeServiceTemplate struct {
+    // ObjectTemplateMetadata "replaces" / embeds Labels
```
rzetelskik, line 93: Looks like a leftover?

enhancements/proposals/1665-metadata/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Implementation History
+
+- 2023-12-28: Initial enhancement proposal
+
+## Drawbacks
+
+Not known.
+
+## Alternatives
```
rzetelskik, line 120: A per-object (i.e. per specific member e.g.) metadata could also be listed as a non-viable alternative.

enhancements/proposals/1665-metadata/README.md
tnozicka: it's only for the purposes of this proposal to show that this removes a field named `annotations` (I'll fix the typo)
see https://github.com/scylladb/scylla-operator/pull/1664/files#diff-73814bffab13ea418b67e2c20f449963e2240a1a07ce288ca3c5919fccf10272L185

enhancements/proposals/1665-metadata/README.md
rzetelskik: Ok, I understand it now with https://github.com/scylladb/scylla-operator/pull/1665/commits/599c666ba179996ef56d0d66aea5a3444dc4de11#r1438170089. Resolved.

enhancements/proposals/1665-metadata/README.md
rzetelskik: I see, thanks.

enhancements/proposals/1665-metadata/README.md
tnozicka: you wouldn't be able to remove an annotation used on the top level object

enhancements/proposals/1665-metadata/README.md
tnozicka: > Isn't this already covered with .spec.exposeOptions.cql.ingress.annotations and .spec.exposeOptions.nodeService.annotations?

yes, both are reused in the embedded struct

> Which ones take precedence in case of a conflict?

there is no merge between those two, `In the case when using an override, the ScyllaCluster.metadata won't be used.`

> Or are the ones coming from .spec.exposeOptions considered managed?

no, managed are the ones defined in operator binary

enhancements/proposals/1665-metadata/README.md
tnozicka: added

rzetelskik:
lgtm
/assign zimnx

tnozicka:
/override e2e-gke-parallel 
/override e2e-gke-parallel-clusterip

WIP to include PVC annotations

enhancements/proposals/1665-metadata/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+	// ...
+}
+
+type RackSpec struct {
+	// volumeClaimMetadata controls shared metadata for the volume claim for this rack.
+	// At this point, this field is applied only for the initial claim and is not reconciled during its lifetime.
+	// Note that this may get fixed in the future and this behaviour shouldn't be relied on.
+	// +optional
+	VolumeClaimMetadata *ObjectTemplateMetadata `json:"volumeClaimMetadata,omitempty"`
```
rzetelskik, line 94: Did you consider placing this in `rackSpec.storage`?  E.g. `rackSpec.storage.capacity` specifies the requested size of that same PVC. It doesn't seem perfect either, buy maybe a bit more concise? Wdyt?

tnozicka: I have but given how it's structured today it would kind of loose the metadata vs. spec separation. I kind of wish this API was modeled similarly to Kubernetes resources but here we are. There is a lot of space to make this better in v2.

Picking the less bad option - given the API has only `storage` field, we might rename `StorageSpec` struct to just `Storage` and put metadata struct into it. So it would be more like `storage.metadata.annotations`. Is that what you'd prefer?

rzetelskik: I don't have a strong preference, although 'storage.metadata.annotations' seems more intuitive to me. I guess pick whatever you think will be easier to transpile to v2.

rzetelskik:
/lgtm
Thanks!

Issue #1671, opened 2024-01-03
Extend Scylla Manager E2E with backup and manual restore procedure


rzetelskik:
### What should the feature do?

Until recently, we weren't testing the automated backup procedure due to lack of access to object storage. We should extend the existing E2E test for Scylla Manager with an actual backup to object storage and a manual restore procedure.

### What is the use case behind this feature?

Verifying correctness of automated backup and manual restore procedure using Scylla Manager.

Issue #1478, opened 2023-10-16
Flake: ScyllaDBMonitoring [It] should setup monitoring stack


rzetelskik:
```console
• [FAILED] [2800.181 seconds]
ScyllaDBMonitoring [It] should setup monitoring stack
github.com/scylladb/scylla-operator/test/e2e/set/scylladbmonitoring/scylladbmonitoring.go:33
...
  [FAILED] Timed out after 300.000s.
  The function passed to Eventually failed at github.com/scylladb/scylla-operator/test/e2e/set/scylladbmonitoring/scylladbmonitoring.go:150 with:
  Unexpected error:
      <*url.Error | 0xc000ba7440>: 
      Get "https://10.96.192.8:9142/api/v1/targets": EOF
      {
          Op: "Get",
          URL: "https://10.96.192.8:9142/api/v1/targets",
          Err: <*errors.errorString | 0x34a8000>{s: "EOF"},
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scylladbmonitoring/scylladbmonitoring.go:184 @ 10/13/23 16:13:20.552

  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scylladbmonitoring.glob..func1.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scylladbmonitoring/scylladbmonitoring.go:184 +0x1504
```

https://github.com/scylladb/scylla-operator/actions/runs/6509877307/job/17682537281#step:3:5678
https://github.com/scylladb/scylla-operator/suites/17202489789/artifacts/983189835

tnozicka:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1532/pull-scylla-operator-e2e-gke-parallel/1722576840066863104#1:test-build-log.txt%3A696

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1614/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1748305604050423808

rzetelskik:
[...]

PR #1702, opened 2024-01-24
Switch from in-tree mermaid client to out-of-tree manager client

### LINKED ISSUE #1701: Manager client needs to be updated

The manager client needs to be updated, as the current one is missing restore-related methods.

Test the manual restore procedure in our e2e suite (https://github.com/scylladb/scylla-operator/issues/1671).

_No response_

rzetelskik:
The manager client needs to be updated, as the current one is missing restore-related methods. The restore-related methods are required for https://github.com/scylladb/scylla-operator/issues/1671.
Instead of updating the in-tree mermaid client, this PR removes the in-tree dependency and switches to out-of-tree manager client vendored from Scylla Manager repository.

Resolves #1701 

/priority important-soon
/kind feature
/area dependency
/cc


rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1702/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1750114355044683777#1:test-build-log.txt%3A1351
```shell
  [FAILED] Expected
      <[]*models.TaskListItem | len:0, cap:0>: nil
  to have length 1
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:99 @ 01/24/24 11:36:39.531
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.glob..func18.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:99 +0xcf6
```
Looking into what caused this

rzetelskik:
https://github.com/scylladb/scylla-operator/issues/1478#issuecomment-1910075162
/retest

rzetelskik:
The issue was caused by me initially missing the [`nowSafety`](https://github.com/scylladb/scylla-operator/pull/1702/commits/3073f9d6de62f3b1c4101c59cb59e7415f927cc2#diff-0180809a1db5cac9409f3932796fc7b311f20a4011e9c4c6efd167c26afdb273R23) parameter. This is outdated and somewhat broken, but I'll address it separately. This PR is keeping it as it's only supposed to migrate the client out of tree and update it.

pkg/controller/manager/sync_action.go
```diff
@@ -488,3 +489,15 @@ func (a *updateTaskAction) Execute(ctx context.Context, client *mermaidclient.Cl
...
+func messageOf(err error) string {
+	err = errors.Cause(err)
+	switch v := err.(type) {
+	case interface {
+		GetPayload() *models.ErrorResponse
+	}:
+		return v.GetPayload().Message
+	}
+	return err.Error()
+}
```
zimnx, line 503: why it's moved here? feels like better place is managerclient

pkg/controller/manager/types_old.go
```diff
@@ -177,3 +186,60 @@ func unescapeFilters(strs []string) []string {
 	}
 	return strs
 }
+
+// parseStartDate parses the supplied string as a strfmt.DateTime.
+func parseStartDate(value string) (strfmt.DateTime, error) {
```
zimnx, line 191: ditto, why manager logic interpreting start date is part of user package?

pkg/controller/manager/types_old.go
```diff
@@ -177,3 +186,60 @@ func unescapeFilters(strs []string) []string {
...
+
+var (
+	byteCountRe          = regexp.MustCompile(`([0-9]+(?:\.[0-9]+)?)(B|[KMGTPE]iB)`)
+	byteCountReValueIdx  = 1
+	byteCountReSuffixIdx = 2
+)
+
+// parseByteCount returns byte count parsed from input string.
+// This is opposite of StringByteCount function.
+func parseByteCount(s string) (int64, error) {
```
zimnx, line 222: ditto

pkg/controller/manager/sync_action.go
rzetelskik: > feels like better place is managerclient

what do you mean? managerclient is a vendored package

pkg/controller/manager/types_old.go
rzetelskik: it's not related to manager really - it used to be that way, but it was moved to sctool. https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/pkg/command/flag/type.go#L102. This is essentially parsing a value from ScyllaCluster API, which has the same syntax

pkg/controller/manager/types_old.go
rzetelskik: again, moved away from manager https://github.com/scylladb/scylla-manager/commit/edbbf605148b9f765cc79e72def5e69c69683377, still part of our API though

pkg/controller/manager/types_old.go
zimnx: I see, I though it's part of Scylla Manager API, but looks like these date and size extensions are part of CLI logic.

pkg/controller/manager/sync_action.go
zimnx: right, my bad. I forgot it's vendored now.

zimnx (approved):
/approve
/assign tnozicka

tnozicka (approved):
/approve
/lgtm
thanks

PR #1706, opened 2024-01-26
Replace nowSafety buffer with zero-value datetime in manager task scheduling

### LINKED ISSUE #1705: "now" syntax is broken in manager task translation
### What happened?

Currently, when translating manager tasks, the operator adds a "safety buffer" to a current time for special "now" syntax. This behaviour is generally broken, but this specific case is handled by sctool and manager server by sending a zero-value datetime, so that the server can handle it on its side.

We should replace the safety buffer with zero-value datetime, like sctool does: https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/pkg/command/flag/type.go#L107.

n/a

### Scylla Operator version

master

n/a

n/a

This behaviour is generally broken, the start-date deprecated (?) without a viable alternative.
See https://github.com/scylladb/scylla-manager/issues/3700, https://github.com/scylladb/scylla-manager/issues/3701, https://github.com/scylladb/scylla-manager/issues/3702, for which I'll open separate, corresponding issues.

rzetelskik:
Currently, when translating manager tasks, the operator adds a "safety buffer" to a current time for special "now" syntax. This PR doesn't fix the entire "now+duration" syntax issue, but is limited to this specific case.

Resolves #1705

/cc
/kind bug

### DIFF
--- a/pkg/controller/manager/types_old.go
+++ b/pkg/controller/manager/types_old.go
@@ -19,10 +19,6 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/util/duration"
 )
 
-const (
-	nowSafety = 30 * time.Second
-)
-
 type RepairTask v1.RepairTaskStatus
 
 func (r RepairTask) ToManager() (*managerclient.Task, error) {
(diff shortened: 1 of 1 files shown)


zimnx:
Code reference it's calculated on server side: https://github.com/scylladb/scylla-manager/blob/389b408234232cd831bf4d8f48d49ced8578bf38/pkg/service/scheduler/service.go#L284-L285

/lgtm

tnozicka (approved):
/approve

Issue #1684, opened 2024-01-12
Limit ScyllaDB restarts on updates


tnozicka:
At this point, every scylla-operator release causes a rolling restart on scylla to update the sidecar that is run alongside scylla, even when none of the sidecar logic has changed.

We should split the sidecar logic into it's own repo and image that would provide probes and starter logic.

Note that at this point, there are [parts of the operator logic within the sidecar that shouldn't be there[(#604) and we need to remove them first.

### Requires
- #604

### Tasks
```[tasklist]
- [ ] https://github.com/scylladb/scylla-operator/issues/1714
- [ ] https://github.com/scylladb/scylla-operator/issues/1720
```

Issue #604, opened 2021-05-21
Move scylla config creation into the operator


tnozicka:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
Scylla config should be built in the operator, then represented as a CM/Secret and mounted into the container. Same goes for the args. Currently it is build in a sidecar which has several disadvantages:
- the sidecar has to run exactly the same image as the operator which is impossible to enforce and breaks with rolling tags
- if any remote call fails it cases a restart
- we can't trigger rolling restarts on changes
- wastes live api call, the operator already has caches with data that needs to be substituted
- is build again for every replica, instead of the operator doing it once
- if config building fails we end up with 1 replica down, while the operator would find out sooner and would stop before trying the first replica
- it can't be revisioned
- needs extra RBAC
- it doesn't honor the Three laws of controllers (ownerRefs, adoption, release)
- it needs additional arguments to know e.g. the manager-agent secret location, while the operator already has all of the prerequisites and caches
- we can't prevent a rollout with invalid config

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2329.

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2466.

PR #1713, opened 2024-02-06
Move ScyllaDB main config file generation from sidecar to operator

### LINKED ISSUE #1714: Move scylladb config creation from sidecar to the operator
ScyllaDB config should be created in the operator where all the data and informers are available without taking extra cpu or requiring binary copies (sidecar).

tnozicka:
This PR moves config generation from sidecar to the operator. This will allow us to move more flags into the config as well in the future and limit the sidecar scope, hopefully to zero one day.

Note: The Alternator work builds on top of this PR.

Fixes #1714

First step to #1684 and #604


pkg/controller/scyllacluster/sync_config.go
```diff
@@ -0,0 +1,48 @@
...
+) ([]metav1.Condition, error) {
+	var errs []error
+	var progressingConditions []metav1.Condition
+
+	cm, err := MakeManagedScyllaDBConfig(sc)
+	if err != nil {
+		return progressingConditions, fmt.Errorf("can't make managed scylladb config: %w", err)
+	}
+
+	cm.SetOwnerReferences([]metav1.OwnerReference{
```
zimnx, line 28: why this is outside of `MakeManagedScyllaDBConfig`?

pkg/sidecar/config/config.go
```diff
@@ -34,6 +30,7 @@ const (
 	configDirScyllaD                    = "/etc/scylla.d"
 	scyllaYAMLPath                      = configDirScylla + "/" + naming.ScyllaConfigName
 	scyllaYAMLConfigMapPath             = naming.ScyllaConfigDirName + "/" + naming.ScyllaConfigName
+	scyllaYAMLManagedConfigMapPath      = naming.ScyllaDBManagedConfigDir + "/scylladb-managed-config.yaml"
```
zimnx, line 33: `scylladb-managed-config.yaml` could be const, and then synchronized between this variable and actual ConfigMap content.

pkg/controller/scyllacluster/resource.go
```diff
@@ -396,6 +399,17 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
...
 								},
 							},
+							{
+								Name: "scylladb-managed-config",
+								VolumeSource: corev1.VolumeSource{
+									ConfigMap: &corev1.ConfigMapVolumeSource{
+										LocalObjectReference: corev1.LocalObjectReference{
+											Name: naming.GetScyllaDBManagedConfigCMName(c.Name),
+										},
+										Optional: ptr.To(false),
```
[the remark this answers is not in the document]
tnozicka: it wasn't available before but I don't mind changing that

pkg/controller/scyllacluster/sync_config.go
tnozicka: modeled after https://github.com/scylladb/scylla-operator/blob/ef4406fd838ed8db3eec6dd87786218ee265b88a/pkg/controller/scylladbmonitoring/sync_grafana.go#L477-L486

the idea behind that is that whoever creates it sets it, so possibly other controllers can use it too. I admit it's not that likely, so I can see this going both ways a bit.

pkg/sidecar/config/config.go
tnozicka: done

pkg/controller/scyllacluster/resource.go
```diff
@@ -272,6 +274,7 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 	}
 	rackTemplateAnnotations[naming.PrometheusScrapeAnnotation] = naming.LabelValueTrue
 	rackTemplateAnnotations[naming.PrometheusPortAnnotation] = "9180"
+	rackTemplateAnnotations["inputs-hash"] = inputsHash
```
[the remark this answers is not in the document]
tnozicka: done

pkg/controller/scyllacluster/sync_config.go
zimnx: I would rather keep it consistent at least across CRD. Having object construction split into multiple places is error prone. 
We can extract it out, once we actually need it.

zimnx:
PR title suggest entire config generation is moved to Operator, but only subset is. Lets fix that.

tnozicka:
> PR title suggest entire config generation is moved to Operator, but only subset is. Lets fix that.

Which ones aren't? There are command line flags but those are not config. If you had these in mind should I reword "config generation" as "config file generation"?

zimnx:
>Which ones aren't? 
Snitch config `cassandra-rackdc.properties` is not moved.

pkg/controller/scyllacluster/sync_config.go
tnozicka: done

tnozicka:
Renamed + created #1714 to track it, thx.

zimnx (approved):
Thanks!
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1718#issuecomment-1935835282
/retest

PR #1715, opened 2024-02-06
Collect namespace where NodeConfig Pods are executed

zimnx:
In case of any errors related to NodeConfig not being able to optimize nodes, we need logs from `scylla-operator-node-config` where NodeConfig Pods are running.
This namespace was missing in must-gather collector.

### DIFF
--- a/pkg/cmd/operator/mustgather.go
+++ b/pkg/cmd/operator/mustgather.go
@@ -194,6 +194,14 @@ var mustGatherSpecs = []struct {
 	},
 	{
 		GroupResource: schema.GroupResource{
+			Resource: "namespaces",
+			Group:    "",
+		},
+		Namespace: corev1.NamespaceAll,
+		Name:      "scylla-operator-node-tuning",
+	},
+	{
+		GroupResource: schema.GroupResource{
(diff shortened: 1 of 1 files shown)


pkg/cmd/operator/mustgather.go
```diff
@@ -192,6 +192,14 @@ var mustGatherSpecs = []struct {
 		Namespace: corev1.NamespaceAll,
 		Name:      "scylla-manager",
 	},
+	{
+		GroupResource: schema.GroupResource{
+			Resource: "namespaces",
+			Group:    "",
+		},
+		Namespace: corev1.NamespaceAll,
+		Name:      "scylla-operator-node-tuning",
```
tnozicka, line 201: does this need to be conditional as the namespace is not there in some configurations?

zimnx: no, collector ignores when resource we search for is not there:
https://github.com/scylladb/scylla-operator/blob/6c7657d58b464a013e04cb6abe31dd6a8fdaf539/pkg/gather/collect/collect.go#L468-L470

tnozicka (approved):
/approve
/lgtm

Issue #1426, opened 2023-09-26
Flake: NodeConfig Optimizations [It] should correctly project state for each scylla pod [Serial]


rzetelskik:
```
• [FAILED] [1252.316 seconds]
NodeConfig Optimizations [It] should correctly project state for each scylla pod [Serial]
github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_optimizations.go:145
  [FAILED] Unexpected error:
      <wait.errInterrupted>: 
      timed out waiting for the condition
      {
          cause: <*errors.errorString | 0xc0003b7000>{
              s: "timed out waiting for the condition",
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_optimizations.go:262 @ 09/26/23 10:16:46.865
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig.glob..func2.4()
    	github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_optimizations.go:262 +0x140f
```

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1423/pull-scylla-operator-e2e-gke-serial/1706604188068220928#1:test-build-log.txt%3A615
https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1423/pull-scylla-operator-e2e-gke-serial/1706604188068220928/

tnozicka:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1494/pull-scylla-operator-e2e-gke-serial/1714952752070660096#1:test-build-log.txt%3A645

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1508/pull-scylla-operator-e2e-gke-serial/1738219481697619968#1:test-build-log.txt%3A586

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1717/pull-scylla-operator-master-e2e-gke-serial/1757049657621483520

[...]

PR #1717, opened 2024-02-07
Collect only ServiceMonitors managed by given ScyllaDBMonitoring

### LINKED ISSUE #1716: ScyllaDBMonitoring Prometheus collects all ServiceMonitors from the same namespace
### What happened?

When ScyllaDBMonitoring is deployed, Prometheus object originating from it has empty serviceMonitorSelector, which targets every ServiceMonitor in the same namespace. 

Only ServiceMonitors managed by ScyllaDBMonitoring are targeted.

Deploy ScyllaDBMonitoring and custom ServiceMonitor within the same namespace.

### Scylla Operator version

master

n/a

-

_No response_

zimnx:
Prometheus object had empty serviceMonitorSelector which matched every ServiceMonitor within the same namespace.
This sets serviceMonitorSelector to match only ServiceMonitor's that are managed by the same ScyllaDBMonitoring.

Fixes #1716 

/cc tnozicka

### DIFF
--- a/assets/monitoring/prometheus/v1/prometheus.yaml
+++ b/assets/monitoring/prometheus/v1/prometheus.yaml
@@ -29,7 +29,8 @@ spec:
     httpConfig:
       http2: true
   serviceMonitorSelector:
-    matchLabels: {}
+    matchLabels:
+      scylla-operator.scylladb.com/scylladbmonitoring-name: "{{ .scyllaDBMonitoringName }}"
   affinity:
     {{- .affinity | toYAML | nindent 4 }}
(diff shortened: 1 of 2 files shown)


zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1718
/retest

tnozicka (approved):
/approve
/lgtm
thanks!

zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1426#issuecomment-1941406961
/retest

Issue #1080, opened 2022-10-31
Add e2e for alternator over HTTPS


tnozicka:
We need to cover the alternator backend over HTTPS.

tnozicka:
done in #1719

PR #1719, opened 2024-02-08
Fully support Alternator API and deprecate the old port option

### LINKED ISSUE #1082: Add option to enable different protocols
* Feature Request

It should enable allowing and disabling these protocols:
- Alternator
- Alternator over TLS

Being able to control the protocols exposed to the user.

**Additional Information:**
The non secure variant need to be disabled by default, at least eventually. For those currently expose we need to design a migration path.

tnozicka:
This PR adds proper API support for setting up Alternator API (in addition to CQL) as described in [SOEP-1405](https://github.com/scylladb/scylla-operator/blob/51bf0f4/enhancements/proposals/1405-alternator/README.md) and adds e2e tests.

Resolves #1082 #1080

```[tasklist]
### Requires
- [ ] https://github.com/scylladb/scylla-operator/pull/1713
- [ ] https://github.com/scylladb/scylla-operator/issues/1728
```


nyh:
> This PR adds proper API support for setting up Alternator API (in addition to CQL) as described in [SOEP-1405](https://github.com/scylladb/scylla-operator/blob/b9585c303d44b66caf7f750c9e1c13e592967a0e/enhancements/proposals/1405-alternator/README.md) and adds e2e tests.

I like the way you do those formal proposals. I read through it and have a few comments/questions about the proposal document:

1. I see you chose port 8443 for Alternator. Why not 443, i.e., the standard HTTPS port that AWS also uses, and just typing "https://" without a port number will get you 443, not 8443.
2. I didn't understand the proposal for a load balancer. I also didn't understand if you do use a load balancer, why does it need or want to communicate with the backend servers with SSL? Can't we assume that the communication between the load balancer and the backend servers is secure?
3. I didn't understand how this proposal configures authorization (i.e., signing of requests). Is signature-checking turned on by default? Is there some UI to add authorized keys, or there is some user-facing documentation on how to do this via CQL?

tnozicka:
> I like the way you do those formal proposals. I read through it and have a few comments/questions about the proposal document:

thanks @nyh 

>    I see you chose port 8443 for Alternator. Why not 443, i.e., the standard HTTPS port that AWS also uses, and just typing "https://" without a port number will get you 443, not 8443.

This uses the service / IP shared with other ScyllaDB services (cql, cql-ssl, REST API, ...) so our default choice is to follow ports used by ScyllaDB for consistency and to avoid conflicts in the future.

>    I didn't understand the proposal for a load balancer. I also didn't understand if you do use a load balancer, why does it need or want to communicate with the backend servers with SSL? Can't we assume that the communication between the load balancer and the backend servers is secure?

Local network is newer secure and all apps should use end to end encryption. All other options are just shortcuts to avoid solving it properly. They are also shortcuts to getting hacked :)

https://en.wikipedia.org/wiki/Zero_trust_security_model

>    I didn't understand how this proposal configures authorization (i.e., signing of requests). Is signature-checking turned on by default? Is there some UI to add authorized keys, or there is some user-facing documentation on how to do this via CQL?

If by authorization you mean for purposes of mTLS, then I didn't see a way to configure client-ca for the alternator in ScyllaDB config (contrary to CQL over TLS that has it), unfortunately. Can you elaborate on what do you think we should configure?

nyh:
> > I didn't understand how this proposal configures authorization (i.e., signing of requests). Is signature-checking turned on by default? Is there some UI to add authorized keys, or there is some user-facing documentation on how to do this via CQL?
> 
> If by authorization you mean for purposes of mTLS, then I didn't see a way to configure client-ca for the alternator in ScyllaDB config (contrary to CQL over TLS that has it), unfortunately. Can you elaborate on what do you think we should configure?

No, the DynamoDB API (like the rest of AWS) doesn't authenticate users via TLS: It uses TLS only used to authenticate the server, and to hide the content of the communication and to prevent modification to responses. To authenticate the user **and** his requests, AWS uses a signature algorithm (https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) which signs a request proving the signer's identity and his intention to run this specific query.

Alternator implements this signature v4 algorithm to allow only authenticated users to make requests, but it needs to be configured (the documentation explains how). TLS itself is **not** enough - any attacker can connect via TLS and send requests...

tnozicka:
Yep, this isn't meant as authentication nor authorization, only secure transport. Seems like it all depends on Scylla config that can be adjusted by the user, similarly to how you'd have to do it with ScyllaDB VM. Same way we use the default authenticators for CQL that are in the ScyllaDB config; for better or worse.

So the user would update his ScyllaDB config with `alternator_enforce_authorization: true` and use CQL to adjust `system_auth.roles`.

CQL, by default inherits ScyllaDB defaults, so no authentication either (modulo mTLS), so this isn't really different, yet I wonder that it may be easier to enforce the auth now as a default because we can't do it later. It would need an extra API field though.

tnozicka:
for xref, I have updated the proposal to use authorization by default and this PR as well
https://github.com/scylladb/scylla-operator/pull/1731

helm/scylla/templates/scyllacluster.yaml
```diff
@@ -22,8 +22,7 @@ spec:
   {{- end }}
   {{- if .Values.alternator.enabled }}
   alternator:
-    port: {{ .Values.alternator.port }}
-    writeIsolation: {{ .Values.alternator.writeIsolation }}
+    {{- toYaml .Values.alternator | nindent 4 }}
```
zimnx, line 25: wouldn't it try to set `enabled` field, and fail due to additional property?

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
+
+func ValidateTLSCertificate(cert *scyllav1.TLSCertificate, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	switch cert.Type {
+	case scyllav1.TLSCertificateTypeOperatorManaged:
+		if cert.OperatorManagedOptions != nil {
+			allErrs = append(allErrs, ValidateOperatorManagedTLSCertificateOptions(
+				cert.OperatorManagedOptions,
+				fldPath.Child("OperatorManagedOptions"),
```
zimnx, line 63: s/OperatorManagedOptions/operatorManagedOptions

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
+			)...)
+		}
+
+	case scyllav1.TLSCertificateTypeUserManaged:
+		if cert.UserManagedOptions != nil {
+			allErrs = append(allErrs, ValidateUserManagedTLSCertificateOptions(
+				cert.UserManagedOptions,
+				fldPath.Child("OperatorManagedOptions"),
```
zimnx, line 71: s/OperatorManagedOptions/userManagedOptions

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
+			)...)
+		} else {
+			allErrs = append(allErrs, field.Required(fldPath.Child("UserManagedOptions"), ""))
```
zimnx, line 74: s/UserManagedOptions/userManagedOptions

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
+		}
+
+	case "":
+		allErrs = append(allErrs, field.Required(fldPath.Child("Type"), ""))
```
zimnx, line 78: s/Type/type

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
+
+	default:
+		allErrs = append(
+			allErrs,
+			field.NotSupported(
+				fldPath.Child("Type"),
```
zimnx, line 84:
```suggestion
				fldPath.Child("type"),
```

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
-			for _, wi := range AlternatorSupportedWriteIsolation {
-				if spec.Alternator.WriteIsolation == wi {
-					found = true
-				}
-			}
-			if !found {
-				allErrs = append(allErrs, field.NotSupported(fldPath.Child("alternator", "writeIsolation"), spec.Alternator.WriteIsolation, AlternatorSupportedWriteIsolation))
-			}
-		}
+		allErrs = append(allErrs, ValidateAlternatorSpec(spec.Alternator, fldPath.Child("Alternator"))...)
```
zimnx, line 130:
```suggestion
		allErrs = append(allErrs, ValidateAlternatorSpec(spec.Alternator, fldPath.Child("alternator"))...)
```

pkg/controller/scyllacluster/resource.go
```diff
@@ -785,24 +830,29 @@ func containerPorts(c *scyllav1.ScyllaCluster) []corev1.ContainerPort {
...
+		if c.Spec.Alternator.Port != 0 || (enableHTTP != nil && *enableHTTP) {
+			insecurePort := int32(alternatorInsecurePort)
+			if c.Spec.Alternator.Port != 0 {
+				insecurePort = c.Spec.Alternator.Port
+			}
+			ports = append(ports, corev1.ContainerPort{
+				Name:          alternatorInsecurePortName,
+				ContainerPort: insecurePort,
+			})
+		}
```
zimnx, line 855: This is copy pasted from servicePorts, please extract common method of getting alternator ports and append them.

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -0,0 +1,225 @@
...
+			HTTPClient: httClient,
+			Credentials: credentials.StaticCredentialsProvider{
+				Value: awsCredentials,
+			},
+			RetryMaxAttempts: 1,
+		}
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		for _, alternatorURL := range []string{
+			fmt.Sprintf("https://%s:8043", svcIP),
```
zimnx, line 127: loop over single element slice? Looks like something is missing here, probably the identity service domain

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -0,0 +1,225 @@
...
+			AccessKeyID:     "alternator",
+			SecretAccessKey: "salted-password-42",
+		}
+		wrongAWSCredentials := awsCredentials
+		wrongAWSCredentials.SecretAccessKey += "-wrong"
+
+		err = cqlSession.Query(
+			`INSERT INTO system_auth.roles (role, salted_hash) VALUES (?, ?)`,
+			awsCredentials.AccessKeyID, awsCredentials.SecretAccessKey,
+		).Exec()
```
zimnx, line 100: nit: release query after last use

test/e2e/set/scyllacluster/scyllacluster_alternator.go
tnozicka: added a version with the DNS name that can still be run from our laptops

pkg/api/scylla/validation/cluster_validation.go
tnozicka: fixed, thx

pkg/api/scylla/validation/cluster_validation.go
tnozicka: fixed, thx

pkg/api/scylla/validation/cluster_validation.go
tnozicka: fixed, thx

tnozicka: fixed, thx

helm/scylla/templates/scyllacluster.yaml
tnozicka: fixed, thx

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -524,16 +524,66 @@ type Storage struct {
...
+	// +optional
+	AdditionalIPAddresses []string `json:"additionalIPAddresses,omitempty"`
+}
+
+type TLSCertificate struct {
+	// type determines the source of this certificate.
+	// +kubebuilder:validation:Enum="OperatorManaged";"UserManaged"
+	Type TLSCertificateType `json:"type"`
+
+	// userManagedCertificateOptions specify option for certificates manged by users.
```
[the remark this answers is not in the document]
tnozicka: fixed, thx

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -0,0 +1,238 @@
...
+					ServerName: tc.serverName,
+					RootCAs:    alternatorServingCAPool,
+				}
+			})
+			config := aws.Config{
+				HTTPClient: httClient,
+				Credentials: credentials.StaticCredentialsProvider{
+					Value: awsCredentials,
+				},
+				RetryMaxAttempts: 1,
```
zimnx, line 139: Does it fail with 0 retries?

tnozicka: fixed

zimnx (approved):
thanks for updates, lgtm
/assign @rzetelskik

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -0,0 +1,238 @@
...
+		"year":  year,
+	}
+}
+
+var _ = g.Describe("ScyllaCluster", func() {
+	defer g.GinkgoRecover()
+
+	f := framework.NewFramework("scyllacluster")
+
+	g.It("should setup Alternator API when enabled", func() {
```
rzetelskik, line 57: nit: set up

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -0,0 +1,238 @@
...
+				url:        fmt.Sprintf("https://%s:8043", svcIP),
+				serverName: "",
+			},
+			{
+				url:        fmt.Sprintf("https://%s:8043", svcIP),
+				serverName: fmt.Sprintf("%s-client.%s.svc", sc.Name, sc.Namespace),
+			},
+		} {
+
+			httClient := awshttp.NewBuildableClient().WithTransportOptions(func(transport *http.Transport) {
```
rzetelskik, line 128: s/httClient/httpClient

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -0,0 +1,238 @@
...
+		o.Expect(alternatorServingCACerts).To(o.HaveLen(1))
+		alternatorServingCAPool := x509.NewCertPool()
+		alternatorServingCAPool.AddCert(alternatorServingCACerts[0])
+
+		for _, tc := range []struct {
+			url        string
+			serverName string
+		}{
+			{
+				url:        fmt.Sprintf("https://%s:8043", svcIP),
```
rzetelskik, line 119: Shouldn't you also test it with the unsecured port for maintaining backwards compatibility?

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -40,23 +40,94 @@ func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
...
+					alternator.WriteIsolation,
+					AlternatorSupportedWriteIsolation,
+				),
+			)
+		}
+	}
+
+	if alternator.InsecureEnableHTTP != nil && *alternator.InsecureEnableHTTP {
+		if alternator.Port != 0 {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("port"), "deprecated port is not allowed to by specified"))
```
rzetelskik, line 117:
```suggestion
			allErrs = append(allErrs, field.Forbidden(fldPath.Child("port"), "deprecated port is not allowed to be specified"))
```

rzetelskik: actually why is this **only** forbidden in conjunction with InsecureEnableHTTP? It feels a bit counterintuitive. Is this a step towards ignoring the port?

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -524,16 +524,66 @@ type Storage struct {
...
+
+type UserManagedTLSCertificateOptions struct {
+	// secretName references a kubernetes.io/tls type secret containing the TLS cert and key.
+	SecretName string `json:"secretName"`
+}
+
+type OperatorManagedTLSCertificateOptions struct {
+	// additionalDNSNames represents external DNS names that the certificates should be signed for.
+	// +optional
+	AdditionalDNSNames []string `json:"additionalDNSNames,omitempty"`
```
rzetelskik, line 542: Is it me not being able to find it, or is it not being used/tested anywhere?

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -524,16 +524,66 @@ type Storage struct {
...
+
+	// additionalIPAddresses represents external IP addresses that the certificates should be signed for.
+	// +optional
+	AdditionalIPAddresses []string `json:"additionalIPAddresses,omitempty"`
```
rzetelskik, line 546: ditto

rzetelskik (commented):
left a couple of comments

test/e2e/set/scyllacluster/scyllacluster_alternator.go
tnozicka: That would mean spinning up another scyllaclusters with every combination of our 3 api options, so I covered those in a unit test and tested the production setup here.

pkg/api/scylla/validation/cluster_validation.go
tnozicka: the port is deprecated but we have to keep backwards compatibility, so we forbid it only in conjunction with the new options so it doesn't break storage migration (essentially when we know it's a manual edit using the new fields)

pkg/api/scylla/v1/types_cluster.go
tnozicka: nope, that's missing, thx

tnozicka: done

pkg/api/scylla/v1/types_cluster.go
tnozicka: done

test/e2e/set/scyllacluster/scyllacluster_alternator.go
tnozicka: fixed, thx

test/e2e/set/scyllacluster/scyllacluster_alternator.go
tnozicka: done

rzetelskik:
/lgtm
thanks

tnozicka:
seems like a kube-proxy traffic issue
/retest

tnozicka:
removed IP len check in the e2e that depends on exposing type, fixes https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1719/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1760697844819300352#1:test-build-log.txt%3A1023

tnozicka:
slow service lb update https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1719/pull-scylla-operator-master-e2e-gke-parallel/1760720462477791232#1:test-build-log.txt%3A1042
/retest

tnozicka:
/hold
(we might want to fix minReadySeconds for podips + identity service first)

tnozicka:
needs #1746

tnozicka:
/hold cancel
#1745 landed

PR #1724, opened 2024-02-13
Get cpuset from kubelet PodResources API 

### LINKED ISSUE #1722: Improve cpuset detection

When Cpuset is not available through CRI, Operator falls back to inspecting hardcoded list of paths in cgroups fs to detect cpuset of Scylla container. 
As slice naming may depend on configuration and CRI it might be that Operator won't be able to detect cpuset and hence unblock Scylla from running. 

We should be more futureproof and either:
* validate whether newer CRI version standardize cpuset information
* scan entire cgroup fs to not rely on particular slice naming
* read naming convention from kubelet configuration

Improve future-proofing of node tuning

_No response_

zimnx:

New method of getting cpuset is added to node tuning controller.
Kubelet provides cpuset information via PodResources API exposed via socket
within kubelet host directory. It gives reliable information about allocated cpus
compared to existing methods.
Old unreliable methods were replaced with it.

Resolves #1722


tnozicka:
We already require Kubernetes >=1.21, so kubelet with 1.20 API seems good and I'd prefer an API to replace the magic paths that don't work across the board.

zimnx:
> We already require Kubernetes >=1.21, so kubelet with 1.20 API seems good and I'd prefer an API to replace the magic paths that don't work across the board.

I agree, I'll update PR with with removal of old methods.

zimnx:
> I agree, I'll update PR with with removal of old methods.

Done

zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1028#issuecomment-1942072967
/retest

pkg/kubelet/podresourcesclient.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2024 ScyllaDB.
```
tnozicka, line 1: this should probably go to third_party dir unless we want to differ

tnozicka (commented):
Thanks for getting rid of the magic paths!

few nits, otherwise this lgtm

With Kube high API standards, I am a bit saddened that it's lacking a client for this one :(
https://github.com/kubernetes/kubernetes/issues/119817

pkg/kubelet/podresourcesclient.go
zimnx: third party contains code copy pasted from other repo where we couldn't vendor it. This one was manually written by me.

pkg/controller/nodetune/tune.go
```diff
@@ -110,37 +70,30 @@ func scyllaDataDirMountHostPaths(ctx context.Context, criClient cri.Client, scyl
...
-	// GKE: /sys/fs/cgroup/cpuset/kubepods/podfc060df5-82a2-4a5e-86c0-40aec54b2a09/e3e0fc65ca5a88c9f47078aa0f097053f4c0620b2134a903c124a9b015386505/cpuset.cpus
-	// minikube (cri-o): /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-pod74298416_16b6_47bb_ae8f_01ee4f91c525.slice/crio-4a839ac8074c771bd0f6d3732e4d5acd3e27acb818ae6895d767f06198a19596.scope/cpuset.cpus
-
-	return []string{
-		path.Join(cgroupMountpoint, fmt.Sprintf("/cpuset/kubepods.slice/kubepods-pod%s.slice/docker-%s.scope/cpuset.cpus", strings.ReplaceAll(podID, "-", "_"), containerID)),
-		path.Join(cgroupMountpoint, fmt.Sprintf("/cpuset/kubepods.slice/kubepods-pod%s.slice/cri-containerd-%s.scope/cpuset.cpus", strings.ReplaceAll(podID, "-", "_"), containerID)),
-		path.Join(cgroupMountpoint, fmt.Sprintf("/cpuset/kubepods/pod%s/%s/cpuset.cpus", podID, containerID)),
-		path.Join(cgroupMountpoint, fmt.Sprintf("/cpuset/kubepods.slice/kubepods-pod%s.slice/crio-%s.scope/cpuset.cpus", strings.ReplaceAll(podID, "-", "_"), containerID)),
-	}
+	klog.V(2).InfoS("Not found cpuset in kubelet PodResources API", "Pod", klog.KObj(scyllaPod), "Container", containerName)
```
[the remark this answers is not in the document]
zimnx: rephrased.

pkg/controller/nodetune/tune.go
```diff
@@ -110,37 +70,30 @@ func scyllaDataDirMountHostPaths(ctx context.Context, criClient cri.Client, scyl
...
+		return cpuset.CPUSet{}, fmt.Errorf("can't list pod resources: %w", err)
 	}
 
-	containerCpuSet, err := cpuset.Parse(cs.Info.RuntimeSpec.Linux.Resources.CPU.Cpus)
-	if err != nil {
-		return cpuset.CPUSet{}, fmt.Errorf("can't parse container %q cpuset %q, %w", cid, cs.Info.RuntimeSpec.Linux.Resources.CPU.Cpus, err)
+	for _, pr := range prs {
+		if pr.Namespace == scyllaPod.Namespace && pr.Name == scyllaPod.Name {
+			for _, cr := range pr.Containers {
+				if cr.Name == containerName {
```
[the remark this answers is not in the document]
zimnx: FindContainerWithName takes `[]corev1.Container` not `[]v1.ContainerResources`. Helper would be good if there would be another function using it, but since we only have one, and I don't see use case for another one I left it inlined.
I refactored the code to reduce indentation required though.

pkg/kubelet/podresourcesclient.go
tnozicka: ok, it seemed similar to kube, but there probably aren't many ways to do it

pkg/controller/nodetune/tune.go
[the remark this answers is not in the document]
tnozicka: my bad about the existing method. both external method and restructuring the nesting works for me, thx

pkg/controller/nodetune/tune.go
```diff
@@ -110,37 +71,35 @@ func scyllaDataDirMountHostPaths(ctx context.Context, criClient cri.Client, scyl
...
 
-	containerCpuSet, err := cpuset.Parse(cs.Info.RuntimeSpec.Linux.Resources.CPU.Cpus)
-	if err != nil {
-		return cpuset.CPUSet{}, fmt.Errorf("can't parse container %q cpuset %q, %w", cid, cs.Info.RuntimeSpec.Linux.Resources.CPU.Cpus, err)
-	}
+		cr, _, ok := slices.Find(pr.Containers, func(cr *v1.ContainerResources) bool {
+			return cr.Name == containerName
+		})
+		if !ok {
+			continue
```
tnozicka, line 89: nit: i think this would be useful to log, so we can later distinguish whether is doesn't find the podresourcs or a container within them

pkg/controller/nodetune/tune.go
```diff
@@ -110,37 +71,35 @@ func scyllaDataDirMountHostPaths(ctx context.Context, criClient cri.Client, scyl
...
+		}
 
-	return containerCpuSet, nil
-}
+		asInts := slices.ConvertSlice(cr.CpuIds, func(cpuid int64) int {
```
tnozicka, line 92: nit: `asInts` is kind of missing "what as ints" part, `cpuIDs` or `cpuIDInts` might fit better

pkg/controller/nodetune/tune.go
zimnx: done

tnozicka (approved):
thanks for the updates

/approve
/lgtm

PR #1728, opened 2024-02-13
Add identity service to ScyllaDB serving certs

### LINKED ISSUE #1723: Add identity service to ScyllaDB serving certs to support it as an initial contact point

When using ephemeral IPs, the only fixed endpoint for the users is the identity service. We should include its IP and DNS name in the ScyllaDB service certificate.

Using identity service as an initial contact point with ephemeral IPs and TLS.

_No response_

tnozicka:
To use the identity service as an initial contact point we needs its IP and DNS name to be present in ScyllaDB serving certs.

Resolves #1723


zimnx (approved):
/lgtm

PR #1731, opened 2024-02-16
Extend alternator proposal with using Authorization by default

### LINKED ISSUE #1732: Extend alternator proposal with using Authorization by default
Based on @nyh's [suggestion](https://github.com/scylladb/scylla-operator/pull/1719#issuecomment-1946024279) we should enable authorization for Alternator by default as part of #1095.

This would be hard to have as a default later on and stay backwards compatible.

tnozicka:
This PR extends the alternator proposal ([SOEP-1405](https://github.com/scylladb/scylla-operator/blob/b9585c303d44b66caf7f750c9e1c13e592967a0e/enhancements/proposals/1405-alternator/README.md)) with enabling Authorization by default. Unfortunately, it isn't on by default in ScyllaDB itself.

This would be hard to have as a default later on and stay backwards compatible.

Based on @nyh's [suggestion](https://github.com/scylladb/scylla-operator/pull/1719#issuecomment-1946024279)

Resolves #1732

/cc @zimnx

fyi, @tzach @mykaul

### DIFF
--- a/enhancements/proposals/1405-alternator/README.md
+++ b/enhancements/proposals/1405-alternator/README.md
@@ -21,6 +21,7 @@ we can break existing deployments. Same goes for the Service port.
 - Expose Alternator service using a load balanced endpoint (Service)
 - Expose Alternator over secure and static port
 - Expose Alternator on insecure port only if explicitly requested
+- Expose Alternator with Authorization on by default
 - Ensure e2e coverage for Alternator
 - Expose Alternator together with CQL (not exclusively)
 - Deprecate `scyllaclusters.scylla.scylladb.com/v1.spec.alternator.port`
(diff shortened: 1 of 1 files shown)


zimnx (approved):
/lgtm

nyh:
Please note that according to https://github.com/scylladb/scylladb/issues/17089, we have a problem that enabling authorization in Alternator reduces performance for an unknown reason having to do with service labels (the user who posted that report suggested to disable service levels, see https://github.com/scylladb/scylladb/pull/17055). I'm not sure if this is a real issue or not, but it might be worth looking into in case some users don't want authorization and will be hurt the the supposed (?) performance reduction.

tnozicka:
Noted, thx. we don't enable alternator by default so it won't affect CQL only clusters.

PR #1734, opened 2024-02-16
Enable internode compression by default

### LINKED ISSUE #1733: Enable inter node compression by default

[ScyllaDB Production Readiness Guidelines](https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/tips/production-readiness.html) suggests to enable internode compression. We should follow this guideline and enable it for all ScyllaCluster's.

Save network costs, follow recommendations.

_No response_

zimnx:
Scylla documentation points out it should be done for every production deployment.
All ScyllaClusters will have it enabled by default. 
This value is safe to update even for existing clusters, as compression is negotiated between nodes on connection setup.

Fixes #1733

### DIFF
--- a/assets/scylladb/managedconfig.cm.yaml
+++ b/assets/scylladb/managedconfig.cm.yaml
@@ -10,6 +10,7 @@ data:
     cluster_name: "{{ .ClusterName }}"
     rpc_address: "0.0.0.0"
     endpoint_snitch: "GossipingPropertyFileSnitch"
+    internode_compression: "all"
     {{- if .EnableTLS }}
     native_transport_port_ssl: 9142
     native_shard_aware_transport_port_ssl: 19142
--- a/pkg/controller/scyllacluster/resource_test.go
+++ b/pkg/controller/scyllacluster/resource_test.go
@@ -2578,6 +2578,7 @@ func Test_MakeManagedScyllaDBConfig(t *testing.T) {
 cluster_name: "foo"
 rpc_address: "0.0.0.0"
 endpoint_snitch: "GossipingPropertyFileSnitch"
+internode_compression: "all"
 `, "\n"),
 				},
 			},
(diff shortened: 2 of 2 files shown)


zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1028#issuecomment-1948512665

/retest

tzach:
I do not understand why this is done on Operator and not ScyllaDB core.

> Scylla documentation points out it should be done for every production deployment.

If this is the case, we need to either update the doc or core configuration.

tnozicka:
> I do not understand why this is done on Operator and not ScyllaDB core.

I'd wonder as well, but it's where we are.

Btw. the same goes for why ScyllaDB starts by default without enabling CQL or Alternator authorization... so, even in other cases, it's definitely not production ready out of the box.

tnozicka (approved):
/approve
/lgtm

/hold
(for @tzach's comment)

tnozicka:
/hold cancel
@tzach mentioned that while it would be good to have this in ScyllaDB we shouldn't block this as it help right now

nyh:
> > I do not understand why this is done on Operator and not ScyllaDB core.
> 
> I'd wonder as well, but it's where we are.
> 
> Btw. the same goes for why ScyllaDB starts by default without enabling CQL or Alternator authorization... so, even in other cases, it's definitely not production ready out of the box.

I think many of the defaults in Scylla core are fairly arbitrary, something which made sense at some point, and while users today prefer a different default, the Scylla developers simply don't know.
I had long arguments about that when I was developing "safe mode", which is now known as "guardrails". I suggested that Scylla core shouldn't enable any of these protections by default - it should just provide configurations mechanisms which people who are in touch with real users - such as Scylla Cloud - would need to use. But it turned out that Scylla cloud never sat down and decided which protections they wanted to use, so all they got was the protections we (Scylla core) turned on by default, which was basically nothing. So when more "guardrails" were recently added by @ptrsmrn we were more diligent in turning some of them on by default in Scylla core.

I think at some point, someone (@tzach ?) would need to look at the whole configuration issue more holistically - and consider which configuration options should be set to what in Scylla Core source code, Scylla Core scylla.yaml, Scylla Enterprise, Scylla Cloud, etc. - or decide that each of these products should choose their own set of defaults without regard to what the other projects chose.

Issue #1191, opened 2023-03-21
Add server side printers for ScyllaCluster


tnozicka:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
`kubectl get scyllacluster.scylla.scylladb.com` should print some meaningful info beyond its name and creationTimestamp. Likely `numberOfRacksReady/NumberOfRacks`, `totalMembersReady/totalMembers`, possibly others

**What is use case behind this feature:**
UX, compare e.g. a statefulset
```
$ kubectl get scyllacluster.scylla.scylladb.com,statefulset
NAME                                        AGE
scyllacluster.scylla.scylladb.com/example   33m

NAME                                            READY   AGE
statefulset.apps/example-us-east-1-us-east-1a   3/3     33m
statefulset.apps/prometheus-example             1/1     31m
```

**Additional Information:**

mykaul:
Can we extend kubectl with some information?
https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/

tnozicka:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns

PR #1737, opened 2024-02-19
Extend ScyllaCluster status with aggregated data for all racks

tnozicka:
This PR adds additional fields in ScyllaCluster status that describe the aggregated state of all racks.

This helps consumers of the API to determine the overall state easier and is required for #1191


zimnx:
Thanks
/lgtm

PR #1746, opened 2024-02-23
Always use default minReadySeconds so identity service works

### LINKED ISSUE #1745: Connection refused after startup when using identity service with PodIPs
With 1 node and PodIPs the identity service "load balancer" wasn't updated while ScyllaCLuster claimed to be deployed successfully and a first CQL connect failed with `connection refused` .

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1719/pull-scylla-operator-master-e2e-gke-parallel/1760720462477791232#1:test-build-log.txt%3A1042

CQL connections should succeed when ScyllaCluster report `Available=True, Progressing=False, Degraded=False`

tnozicka:
When using PodIPs, the operator was defaulting `minReadySeconds` to `0` because the node services don't have ClusterIP that would need a delay to propagate to the embedded load balancer but there is still the identity service that's used for initial discovery that needs to be updated. That service is actually even more important with the ephemeral PodIPs for clients, as you might otherwise loose all discovery endpoints, if your client is down during a rolling update.

There is still an option to adjust this on the API level, this changes only the "default" when unspecified.

Resolves #1745 

/cc @zimnx

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -340,12 +340,7 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
 		minTerminationGracePeriodSeconds = loadBalancerSyncPeriodSeconds
 		minReadySeconds = loadBalancerSyncPeriodSeconds
 	}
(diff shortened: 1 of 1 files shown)


zimnx (approved):
/lgtm

PR #1747, opened 2024-02-23
Accept custom NodeConfig in script running E2E

zimnx:

To allow different instance setups E2E script accepts parameter specifying path to NodeConfig which should be deployed alongside.

### DIFF
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -36,6 +36,7 @@ if [ -z ${ARTIFACTS+x} ]; then
   exit 2
 fi
 
+SO_NODECONFIG_PATH=${SO_NODECONFIG_PATH=./hack/.ci/manifests/cluster/nodeconfig.yaml}
 SO_DISABLE_NODECONFIG=${SO_DISABLE_NODECONFIG:-false}
 
 field_manager=run-e2e-script
@@ -144,10 +145,10 @@ kubectl -n default rollout status daemonset/sysctl
 kubectl apply --server-side -f ./pkg/api/scylla/v1alpha1/scylla.scylladb.com_nodeconfigs.yaml
 kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com
 
-if [[ "${SO_DISABLE_NODECONFIG}" == "true" ]]; then
+if [[ "${SO_DISABLE_NODECONFIG}" == "true"  ]] || [[ -z "${SO_NODECONFIG_PATH}" ]]; then
   echo "Skipping NodeConfig creation"
 else
-  kubectl_create -f ./hack/.ci/manifests/cluster/nodeconfig.yaml
+  kubectl_create -f "${SO_NODECONFIG_PATH}"
   kubectl_create -n local-csi-driver -f ./hack/.ci/manifests/namespaces/local-csi-driver/
 fi
 


hack/.ci/manifests/cluster/nodeconfig-pd.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: scylla.scylladb.com/v1alpha1
+kind: NodeConfig
+metadata:
+  name: cluster
+spec:
+  localDiskSetup:
+    loopDevices:
+      - name: persistent-volumes
+        size: 50Gi
```
tnozicka, line 9: is this because its not actually used? otherwise I was hoping for less (the local SSD are cheaper in the end if you actually use the space they give you)

tnozicka: i think we use like 40GiB disk in CI or something like that

hack/.ci/run-e2e-gke.sh
```diff
@@ -36,6 +36,7 @@ if [ -z ${ARTIFACTS+x} ]; then
   exit 2
 fi
 
+SO_NODECONFIG_DISK_TYPE=${SO_NODECONFIG_DISK_TYPE:-"nvme"}
```
rzetelskik, line 39: nit: you could check for supported disk types to fail early instead of on kubectl command

rzetelskik (commented):
one nit

hack/.ci/run-e2e-gke.sh
zimnx: I changed it to accepting path to NodeConfig instead suffix. This allows to specify custom NodeConfig suitable for given setup, which may be defined alongside kubernetes cluster spec.

hack/.ci/manifests/cluster/nodeconfig-pd.yaml
zimnx: Changed it to accept path to nodeconfig, this will allow to specify nodeconfig spec alongside kubernetes cluster spec.

hack/.ci/run-e2e-gke.sh
```diff
@@ -147,7 +148,7 @@ kubectl wait --for condition=established crd/nodeconfigs.scylla.scylladb.com
 if [[ "${SO_DISABLE_NODECONFIG}" == "true" ]]; then
```
tnozicka, line 148: It might be good to add `OR` here on `SO_NODECONFIG_PATH` being empty, so we can eventually drop the `SO_DISABLE_NODECONFIG` when we switch CI

zimnx: added `or` and changed how default value is assigned when explicit empty string is passed by.

hack/.ci/run-e2e-gke.sh
```diff
@@ -36,6 +36,7 @@ if [ -z ${ARTIFACTS+x} ]; then
   exit 2
 fi
 
+SO_NODECONFIG_PATH=${SO_NODECONFIG_PATH=./hack/.ci/manifests/cluster/nodeconfig.yaml}
```
tnozicka, line 39: this seems to work but elsewhere we use `:-` - what's the difference?

zimnx: `:-` sets variable to default value when variable is explicitly set to empty string, where `=` keep it as is.

tnozicka: thx

tnozicka (approved):
/approve
/lgtm

PR #1758, opened 2024-02-28
[WIP] Disable JMX service

### LINKED ISSUE #1762: Disable JMX service
JMX service is deprecated in ScyllaDB and not used in any of the flows with the Operator either. 

It exposes 2 insecure ports: 7199 on 127.0.0.1 and random port on `0.0.0.0`. Removing it helps us harden ScyllaDB security.

We should also not run multiple service in one container so this helps with container split as well.

tnozicka:
ScyllaDB JMX is deprecated and opens ports that are not secured. We should avoid exposing insecure ports by default.

This get rid of two insecure ports , the JMX service on local host and then a random port on all interfaces :facepalm:

As this is coming from ScyllaDB image/container we need to explicitly disable it in supervisord for now. When we get rid of using supervisord there and run services explicitly, this will become cleaner.

Resolves #1762

### DIFF
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


tnozicka:
/retest
https://github.com/scylladb/scylla-operator/issues/1767

tnozicka:
looked into https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1758/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1762835621090430976#1:test-build-log.txt%3A2280

but it need #1773 first to see why it failed, gonna try once more if its a flake (we have also switched networking config today)
/retest

tnozicka:
reproduced this locally - its the nodetool :(
```
$ kubectl logs -c e2e-drain-scylla -f pod/basic-8gn5q-us-east-1-us-east-1a-1
nodetool: Failed to connect to '127.0.0.1:7199' - ConnectException: 'Connection refused (Connection refused)'.
```

tnozicka:
fingers crossed for 6.0 dropping Java
/close

PR #1776, opened 2024-02-28
Return empty list of loop devices when `losetup` doesn't print anything

### LINKED ISSUE #1775: NodeConfig can't create raid device on EKS
### What happened?

NodeConfig trying to create raid device on EKS is stuck due to unsupported output of losetup.

```
E0228 13:44:09.870172       1 nodesetup/controller.go:192] syncing key 'cluster' failed: failed to sync nodeconfig "cluster": can't sync loop devices: can't get node config loop devices: can't list loop devices: failed to unmarshal losetup output "": unexpected end of JSON input
```

NodeConfig setting up local disks the way wanted.

Deploy latest scylla-operator, create NodeConfig setting up a raid device.

### Scylla Operator version

latest

EKS

-

_No response_

zimnx:
Version of `losetup` installed on EKS has different behavior when there are no loop devices on the node. On GKE and local env, it returns an empty list of loop devices as json output, where on EKS it doesn't print anything, even though we ask for json explicitly.
Losetup library now supports both behaviors.

Fixes #1775

### DIFF
--- a/pkg/util/losetup/losetup_test.go
+++ b/pkg/util/losetup/losetup_test.go
@@ -1,0 +1,109 @@
+// Copyright (c) 2024 ScyllaDB.
+
+package losetup
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/scylladb/scylla-operator/pkg/util/exectest"
+	testingexec "k8s.io/utils/exec/testing"
+)
+
+func TestListLoopDevices(t *testing.T) {
+	t.Parallel()
+
+	tt := []struct {
+		name                string
+		commands            []exectest.Command
+		expectedLoopDevices []*LoopDevice
+		expectedError       error
+	}{
+		{
+			name: "losetup fails with random error",
+			commands: []exectest.Command{
+				{
+					Cmd:    "losetup",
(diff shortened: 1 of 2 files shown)


pkg/util/losetup/losetup_test.go
```diff
@@ -0,0 +1,109 @@
...
+		expectedLoopDevices []*LoopDevice
+		expectedError       error
+	}{
+		{
+			name: "losetup fails with random error",
+			commands: []exectest.Command{
+				{
+					Cmd:    "losetup",
+					Args:   []string{"--all", "--list", "--json"},
+					Stdout: []byte("stdout error"),
```
tnozicka, line 30: tiny nit: i suppose since the error is within stderr, "stdout output" or something like that might be more fitting here

tnozicka (approved):
/approve
lgtm

pkg/util/losetup/losetup_test.go
zimnx: done

zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1426#issuecomment-1970951498
/retest

tnozicka (commented):
/lgtm
thanks

PR #1779, opened 2024-02-29
Disable unused sshd service in ScyllaDB container

### LINKED ISSUE #1763: Disable unused sshd service
ScyllaDB manager no longer needs ssh so we can finally drop the ssh service run within the ScyllaDB container.

This will help security as well as future container split.

tnozicka:
This PR disables the unused sshd service that comes from supervisord config in ScyllaDB image [that isn't used by ScyllaDB manager since 2.0](https://github.com/scylladb/scylla-operator/issues/1763#issuecomment-1969388936)

This helps with ScyllaDB latency because cgroup is shared by all the processes in that container and future process split.

Resolves #1763

### DIFF
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


mykaul:
I think it's the wrong fix - we do not need sshd in our container, and I'm sure we've removed it - @yaronkaikov  ?

tnozicka:
replaced with scylladb bump https://github.com/scylladb/scylla-operator/pull/1783

PR #1782, opened 2024-02-29
[v1.12] Return empty list of loop devices when `losetup` doesn't print anything

zimnx:
Backport of #1776


tnozicka (approved):
/approve
/lgtm

tnozicka:
/hold cancel 
#1776 landed

PR #1784, opened 2024-03-01
Extend Scylla Manager E2E test with progress check

rzetelskik:
At this point the E2E for Scylla Manager integration only tests that the task was scheduled. This PR extends the test with awaiting for task's completeness.

Part of the effort to improve scylla manager integration.

/kind feature
/priority important-longterm
/cc

### DIFF
--- a/test/e2e/set/scyllacluster/scyllamanager.go
+++ b/test/e2e/set/scyllacluster/scyllamanager.go
@@ -5,14 +5,19 @@ package scyllacluster
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"time"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	"github.com/scylladb/scylla-manager/v3/pkg/managerclient"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/scylladb/scylla-operator/test/e2e/utils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var _ = g.Describe("Scylla Manager integration", func() {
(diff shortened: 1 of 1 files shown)


test/e2e/set/scyllacluster/scyllamanager.go
```diff
@@ -85,20 +115,32 @@ var _ = g.Describe("Scylla Manager integration", func() {
...
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(tasks.TaskListItemSlice).To(o.HaveLen(1))
 		repairTask := tasks.TaskListItemSlice[0]
-		o.Expect(repairTask.Name).To(o.Equal(sc.Spec.Repairs[0].Name))
+		o.Expect(repairTask.Name).To(o.Equal(sc.Status.Repairs[0].Name))
+		o.Expect(repairTask.ID).To(o.Equal(sc.Status.Repairs[0].ID))
 		o.Expect(repairTask.Properties.(map[string]interface{})["parallel"].(json.Number).Int64()).To(o.Equal(int64(123)))
+
+		framework.By("Waiting for repair to finish")
+		err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 10*time.Minute, true, func(context.Context) (done bool, err error) {
```
tnozicka, line 136: should this be propagated to scyllacluster status? if yes and we miss the information, a todo with an issue link seems to be in order.

tnozicka (approved):
/approve
/asssing @zimnx

test/e2e/set/scyllacluster/scyllamanager.go
rzetelskik: to my knowledge we don't sync the tasks post-scheduling, so it's not that we miss the status propagation, we don't have the logic to handle the progress. I'm not sure if that could work with recurring tasks though. @zimnx mind sharing your thoughts on if that's viable?

tnozicka: yeah, with recurring tasks we'd have to parse it and schedule queue add at that point - probably not what we want to get into - I take this back :)

test/e2e/set/scyllacluster/scyllamanager.go
```diff
@@ -85,20 +115,32 @@ var _ = g.Describe("Scylla Manager integration", func() {
...
+			repairProgress, err := managerClient.RepairProgress(ctx, *sc.Status.ManagerID, repairTask.ID, "latest")
```
zimnx, line 137: add o.Expect validating whether sc.Status.ManagerID is not nil before this poll

rzetelskik: done

zimnx (approved):
/lgtm

rzetelskik:
flake https://github.com/scylladb/scylla-operator/issues/1525#issuecomment-1979722391
/retest

Issue #1694, opened 2024-01-18
[Flake] Scylla Manager integration [It] should register cluster and sync repair tasks


zimnx:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1691/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1747926847733108736

### Snippet of what failed.

```
 Scylla Manager integration [It] should discover cluster and sync tasks
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:23
  Timeline >>
  STEP: Creating a new namespace @ 01/18/24 10:36:49.96
  Jan 18 10:36:50.156: INFO: Created namespace "e2e-test-scyllacluster-pgwg6-tvn8t".
  STEP: Waiting for service account token Secret "e2e-user-token" in namespace "e2e-test-scyllacluster-pgwg6-tvn8t". @ 01/18/24 10:36:50.448
  STEP: Waiting for default ServiceAccount in namespace "e2e-test-scyllacluster-pgwg6-tvn8t". @ 01/18/24 10:36:51.262
  STEP: Waiting for kube-root-ca.crt in namespace "e2e-test-scyllacluster-pgwg6-tvn8t". @ 01/18/24 10:36:52.008
  STEP: Creating a ScyllaCluster @ 01/18/24 10:36:52.154
  STEP: Waiting for the ScyllaCluster to rollout (RV=6490) @ 01/18/24 10:36:52.205
  Jan 18 10:41:27.531: INFO: ScyllaCluster e2e-test-scyllacluster-pgwg6-tvn8t/basic-9lr8s (RV=12525) is rolled out
  STEP: Verifying the ScyllaCluster @ 01/18/24 10:41:27.532
  Jan 18 10:41:27.756: INFO: Found 1 pvc(s) in namespace "e2e-test-scyllacluster-pgwg6-tvn8t"
  Jan 18 10:41:27.756: INFO: Found 1 pvc(s) for ScyllaCluster "e2e-test-scyllacluster-pgwg6-tvn8t/basic-9lr8s"
  STEP: Waiting for the ScyllaCluster(s) to reach consistency ALL @ 01/18/24 10:41:27.902
  Jan 18 10:41:28.039: INFO: ScyllaDB nodes have reached status consistency.
  STEP: Inserting data @ 01/18/24 10:41:28.076
[...]

PR #1795, opened 2024-03-07
Update manager to 3.2.6 and require it for v1.12

tnozicka:
This PR updates manager version i our manifests and CI to from 3.2.5 to 3.2.6 which fixes TLS and other issues. Given the restore procedure was broken with TLS and manager 3.2.5, we set 3.2.6 as the minimal version as well.

Resolves #


rzetelskik (approved):
lgtm
/assign zimnx

zimnx (approved):
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-1983400091
/retest

tnozicka:
that might be a problem, I don't think requiring it retrospectively for v1.11 would be good. if it need the new API we might just live with not fixing this in v1.11.

rzetelskik:
> that might be a problem, I don't think requiring it retrospectively for v1.11 would be good. if it need the new API we might just live with not fixing this in v1.11.

It definitely requires the 3.2.6 client, but with older version of the server the new fields will just be ignored. Let's not bump the required version for 1.11 then.

tnozicka:
ok, if it's backward compatible we can land it as best effort :+1:

tnozicka:
#1814 landed
/retest

PR #1800, opened 2024-03-07
Pass auth token on Scylla Manager cluster update

### LINKED ISSUE #1799: Scylla Manager integration does not pass auth token on cluster update despite auth token mismatch
### What happened?

There is a bug in our Scylla Manager integration where the controller checks for auth token mismatch and issues an update request to Scylla Manager, but it doesn't set said auth token parameter. If the controller ever hits that, it will just keep sending the update requests.

https://github.com/scylladb/scylla-operator/blob/128676d78fcba4857613b44cd2f8bbbce1c25d7e/pkg/controller/manager/sync_action.go#L42

/priority important-longterm
/kind bug
/assign

Controller should pass auth token in the request.

n/a

### Scylla Operator version

master

n/a

n/a

n/a

rzetelskik:
**Description of your changes:** There is a bug in our Scylla Manager integration where the controller checks for auth token mismatch and issues an update request to Scylla Manager, but it doesn't set said auth token parameter. This PR sets the auth token parameter.

Resolves #1799

/kind bug
/priority important-longterm
/cc zimnx

### DIFF
--- a/pkg/controller/manager/sync_action.go
+++ b/pkg/controller/manager/sync_action.go
@@ -42,9 +42,10 @@ func runSync(ctx context.Context, cluster *scyllav1.ScyllaCluster, authToken str
 				if c.AuthToken != authToken {
 					actions = append(actions, &updateClusterAction{
 						cluster: &managerclient.Cluster{
-							ID:   c.ID,
-							Name: naming.ManagerClusterName(cluster),
-							Host: naming.CrossNamespaceServiceNameForCluster(cluster),
+							ID:        c.ID,
+							Name:      naming.ManagerClusterName(cluster),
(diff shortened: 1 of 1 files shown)


zimnx (approved):
Good catch!
/lgtm

tnozicka:
@rzetelskik what the fallout from this? do we need it in v1.12? how did it work without the token?

rzetelskik:
> what the fallout from this?

In theory that could cause the controller to keep sending update requests to the manager server

> do we need it in v1.12? how did it work without the token?

It didn't, but I suppose you'd only hit it if you lost the secret or manually deleted it. In that case you could manually fix it by deleting the cluster from manager and letting the controller sync again. It's better to backport it imo.

PR #1814, opened 2024-03-11
Use gocql client's default timeouts in tests

rzetelskik:
**Description of your changes:** Our E2E suite has gotten noticeably more flaky since we introduced a suite using PDs instead of the NVMes we've been using so far. All of the observed flakes are coming from the gocql client we're using for data consistency verification, see e.g. https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1614/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1767126234778570752#1:test-build-log.txt%3A761. This PR addresses that by removing the explicitly set timeouts and falling back to the client's default, which is set to accommodate the longest default server-side timeout.

Resolves #

/kind machinery
/priority important-soon

### DIFF
--- a/test/e2e/utils/datainserter.go
+++ b/test/e2e/utils/datainserter.go
@@ -180,8 +180,6 @@ func (di *DataInserter) GetExpected() []*TestData {
 
 func (di *DataInserter) createSession(hosts []string) error {
 	clusterConfig := gocql.NewCluster(hosts...)
-	clusterConfig.Timeout = 3 * time.Second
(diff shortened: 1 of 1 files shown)


zimnx (approved):
lgtm
/assign tnozicka

tnozicka (approved):
/approve
/lgtm
to go back to the defaults but we should look why it took more then 3s on PDs as that's not expected

PR #1816, opened 2024-03-12
Update tooManyFiles alerts to filter for only the scylla data mountpoint

Rumbles:
After implementing scylla operator, we started getting alerts for too many open files on our new clusters. I noticed that all the alerts were for `/etc/hosts` and other files that should not be generating alerts IMO. 
 - Added a filter similar to other alerts in the file to only alert when the mountpoint used for scylla data has too many open files
 - Corrected a typo in Instance

Looking on scylla-monitoring, it looks like similar changes must have been made already:

https://github.com/scylladb/scylla-monitoring/blob/master/prometheus/prom_rules/prometheus.rules.yml#L265

### DIFF
--- a/assets/monitoring/prometheus/v1/alerts.prometheusrule.yaml
+++ b/assets/monitoring/prometheus/v1/alerts.prometheusrule.yaml
@@ -270,24 +270,24 @@ spec:
         description: 'OOM Kill on {{ $labels.instance }}'
         summary: A process was terminated on Instance {{ $labels.instance }}
     - alert: tooManyFiles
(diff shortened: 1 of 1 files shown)


zimnx:
I think ScyllaDB Monitoring changes shouldn't cherry pick individual lines from ScyllaDB Monitoring.
If this is fixed in newer version, and it is as you pointed it out, we shall update dashboards, prometheus rules and everything which is part of ScyllaDB Monitoring.

tnozicka:
agreed, we should bump the full monitoring definitions or this will become a hard to maintain and introduces untested monitoring definitions
/hold

Rumbles:
That's a fair point, IMHO you want an automated process to bring in changes from the other repo periodically/automatically, as this will be a constant battle

tnozicka:
hasn't seen activity in 2 months
/close

PR #1818, opened 2024-03-12
Use real salted_hash in alternator e2e

### LINKED ISSUE #1808: Replace manual salted_hash in alternator e2e by enabling AuthN+AuthZ
When CQL Authentication and Authorization are enabled, `salted_hash` used as alternator token is present by default and there is no need to do CQL inserts. This is also how a regular user would get the token.

tnozicka:
`salted_hash` is supposed to come from ScyllaDB, unfortunatelly it wasn't populating it for the default user without explicitly enabling AuthN+AuthZ which lead to a confusion on who should populate the value.

This PR sets up CQL AuthN+AuthZ for alternator e2e which is how our users would do it. With that, it only needs to read the token (`salted_hash`).

Resolves #1808

### DIFF
--- a/test/e2e/fixture/scylla/registry.go
+++ b/test/e2e/fixture/scylla/registry.go
@@ -8,6 +8,7 @@ import (
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/assets"
 	"github.com/scylladb/scylla-operator/test/e2e/scheme"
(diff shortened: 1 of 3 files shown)


tnozicka:
@zimnx @rzetelskik ptal, this should go to v1.12 as well

zimnx (approved):
lgtm
/assign @rzetelskik

rzetelskik (approved):
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-1997529239
/retest

PR #1821, opened 2024-03-12
Fix update action loop break in manager controller

### LINKED ISSUE #1820: Faulty logic in manager controller status update logic on task update actions
### What happened?

Current implementation of Scylla Manager controller has a bug in the logic updating status of updated tasks. It will only update status of the task if it matches the first item in the slice.
https://github.com/scylladb/scylla-operator/blob/8e6efb77b60bfbe54cb00b23135e4a326a012313/pkg/controller/manager/sync_action.go#L497

/priority important-soon
/kind bug
/assign

The loop shouldn't be broken of after the first iteration.

Schedule >1 tasks of the same type and update them.

### Scylla Operator version

n/a

n/a

n/a

_No response_

rzetelskik:
**Description of your changes:** Current implementation of Scylla Manager controller has a bug in the logic updating status of updated tasks. It will only update status of the task if it matches the first item in the slice.

Resolves #1820 

/priority important-soon
/kind bug

### DIFF
--- a/pkg/controller/manager/sync_action.go
+++ b/pkg/controller/manager/sync_action.go
@@ -493,8 +493,8 @@ func (a *updateTaskAction) Execute(ctx context.Context, client *managerclient.Cl
 				if err != nil {
 					status.Repairs[i].Error = messageOf(err)
 				}
+				break
 			}
-			break
 		}
 	}
 	if a.task.Type == "backup" {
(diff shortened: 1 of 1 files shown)


zimnx (approved):
great catch!
/lgtm
lets backport it to 1.12

PR #1824, opened 2024-03-12
Parametrize e2e command

zimnx:
This allows for reusing e2e server and runner in other commands.

### DIFF
--- a/pkg/cmd/tests/options.go
+++ b/pkg/cmd/tests/options.go
@@ -3,9 +3,12 @@ package tests
 import (
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
+	"github.com/onsi/ginkgo/v2"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
 	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/spf13/cobra"
@@ -35,6 +38,8 @@ var supportedBroadcastAddressTypes = []scyllav1.BroadcastAddressType{
 }
 
 type TestFrameworkOptions struct {
+	genericclioptions.ClientConfig
+
 	ArtifactsDir                 string
 	DeleteTestingNSPolicyUntyped string
 	DeleteTestingNSPolicy        framework.DeleteTestingNSPolicyType
@@ -47,8 +52,9 @@ type TestFrameworkOptions struct {
 	gcsServiceAccountKey         []byte
 }
 
-func NewTestFrameworkOptions() TestFrameworkOptions {
-	return TestFrameworkOptions{
+func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent string) *TestFrameworkOptions {
+	return &TestFrameworkOptions{
+		ClientConfig:                 genericclioptions.NewClientConfig(userAgent),
 		ArtifactsDir:                 "",
 		DeleteTestingNSPolicyUntyped: string(framework.DeleteTestingNSPolicyAlways),
 		IngressController:            &IngressControllerOptions{},
@@ -65,6 +71,8 @@ func NewTestFrameworkOptions() TestFrameworkOptions {
 }
 
 func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
+	o.ClientConfig.AddFlags(cmd)
+
 	cmd.PersistentFlags().StringVarP(&o.ArtifactsDir, "artifacts-dir", "", o.ArtifactsDir, "A directory for storing test artifacts. No data is collected until set.")
 	cmd.PersistentFlags().StringVarP(&o.DeleteTestingNSPolicyUntyped, "delete-namespace-policy", "", o.DeleteTestingNSPolicyUntyped, fmt.Sprintf("Namespace deletion policy. Allowed values are [%s].", strings.Join(
 		[]string{
@@ -93,8 +101,13 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(&o.GCSServiceAccountKeyPath, "gcs-service-account-key-path", "", o.GCSServiceAccountKeyPath, "Path to a file containing a GCS service account key.")
 }
 
-func (o *TestFrameworkOptions) Validate() error {
+func (o *TestFrameworkOptions) Validate(args []string) error {
 	var errors []error
+
+	err := o.ClientConfig.Validate()
+	if err != nil {
+		errors = append(errors, err)
+	}
 
 	switch p := framework.DeleteTestingNSPolicyType(o.DeleteTestingNSPolicyUntyped); p {
 	case framework.DeleteTestingNSPolicyAlways,
@@ -127,7 +140,12 @@ func (o *TestFrameworkOptions) Validate() error {
 	return apierrors.NewAggregate(errors)
 }
 
-func (o *TestFrameworkOptions) Complete() error {
+func (o *TestFrameworkOptions) Complete(args []string) error {
+	err := o.ClientConfig.Complete()
+	if err != nil {
+		return err
+	}
+
 	o.DeleteTestingNSPolicy = framework.DeleteTestingNSPolicyType(o.DeleteTestingNSPolicyUntyped)
 
 	// Trim spaces so we can reason later if the dir is set or not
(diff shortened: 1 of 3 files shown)


pkg/test/e2e/cmd.go
```diff
@@ -1,4 +1,6 @@
-package tests
+// Copyright (c) 2024 ScyllaDB.
```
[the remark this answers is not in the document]
zimnx: Idea behind moving it to separate package was to split it from scyllaoperator specifics and discourage from adding non-generic stuff to it. 
This way, we have operator specific options, suites in `/pkg/cmd/tests` and generic Run command in `/pkg/test/e2e` which just "runs" anything feeded into.

I don't feel strong about it though.

rzetelskik (approved):
lgtm
/assign tnozicka

pkg/cmd/tests/tests.go
```diff
@@ -1,17 +1,44 @@
...
+
+	// Include suites
+	_ "github.com/scylladb/scylla-operator/test/e2e"
 )
 
 const (
 	EnvVarPrefix = "SCYLLA_OPERATOR_TESTS_"
 )
 
+var ScyllaOperatorSuites = ginkgotest.TestSuites{
```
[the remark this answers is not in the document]
zimnx: done

zimnx:
Cluster provisioning failed + manager flake https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2020503432
/retest

zimnx:
Cluster provisioning failed
/retest

zimnx:
manager flake https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2021190928
/retest

pkg/test/e2e/cmd.go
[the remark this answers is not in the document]
tnozicka: I don't feel enthusiastic about having commands outside of `pkg/cmd`, please keep the existing file structure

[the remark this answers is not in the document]
zimnx: moved it back.

zimnx:
Manager flake https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2032051928
/override ci/prow/e2e-gke-release-script-latest
/retest

zimnx:
Flake  -https://github.com/scylladb/scylla-operator/issues/1525#issuecomment-2034093381
/retest

pkg/cmd/tests/tests_run.go
```diff
@@ -113,18 +89,37 @@ func NewRunOptions(streams genericclioptions.IOStreams) *RunOptions {
...
+		testSuites = append(testSuites, cts)
+	}
+
+	testSuites = append(testSuites, &ginkgotest.TestSuite{
+		Name: "all",
+		Description: templates.LongDesc(`
+		Runs all tests.`,
+		),
+		DefaultParallelism: 42,
+	})
```
tnozicka, line 109: this should be part of the default suite as it used to

pkg/cmd/tests/tests_run.go
```diff
@@ -113,18 +89,37 @@ func NewRunOptions(streams genericclioptions.IOStreams) *RunOptions {
...
 	}
 }
 
-func NewRunCommand(streams genericclioptions.IOStreams) *cobra.Command {
-	o := NewRunOptions(streams)
+func NewRunCommand(streams genericclioptions.IOStreams, customTestSuites ginkgotest.TestSuites, customOptions ...CustomFrameworkOptions) *cobra.Command {
+	testSuites := make(ginkgotest.TestSuites, 0, len(customTestSuites)+1)
+	for _, cts := range customTestSuites {
+		testSuites = append(testSuites, cts)
+	}
```
tnozicka, line 101: I'd prefer we just give it and array of suites to run. The caller can use the public variable if he want to reuse some, it is more transparent and you can remove suites.

pkg/cmd/tests/tests_run.go
```diff
@@ -43,36 +40,9 @@ const (
...
-		LabelFilter:        fmt.Sprintf("%s", framework.SerialLabelName),
-		DefaultParallelism: 1,
-	},
-}
-
 type RunOptions struct {
 	genericclioptions.IOStreams
-	genericclioptions.ClientConfig
-	TestFrameworkOptions
+	CustomOptions []CustomFrameworkOptions
```
tnozicka, line 45: i think this should be the other way around. whoever is extending the options can embed them into their own struct (aka inheritence)

zimnx: done

pkg/cmd/tests/tests_run.go
zimnx: done

pkg/cmd/tests/tests_run.go
zimnx: done

zimnx:
manager flake 
/retest

pkg/cmd/tests/options.go
```diff
@@ -1,21 +1,390 @@
...
 )
 
+const (
+	parallelShardFlagKey            = "parallel-shard"
+	parallelServerAddressFlagKey    = "parallel-server-address"
+	randomSeedFlagKey               = "random-seed"
+	ginkgoOutputInterceptorModeNone = "none"
+)
+
+type RunOptions struct {
```
tnozicka, line 44: > Move RunOptions to options.go where it belongs 

Putting aside the non-descriptiveness of such commit message, I think they should live where they are, near to the struct methods are in the same file as the struct definition. This is even more true when subcommands are used - these aren't generic options for *-test command but only for run subcommand. Maybe renaming the `options.go` to `testoptions.go` would avoid the confusion.

Consistency wise, all the other option live in the command files:
https://github.com/scylladb/scylla-operator/blob/7d9d80edb6a2a1ae76b2efe83a63712e9067fb24/pkg/cmd/operator/operator.go#L41
https://github.com/scylladb/scylla-operator/blob/5b752cca9f222188614191c4da1418955e915199/pkg/cmd/operator/gather.go#L42
https://github.com/scylladb/scylla-operator/blob/eefa9572f41540e29b48392882af21ca0e5633b9/pkg/cmd/operator/sidecar.go#L46

Anyway, I see no reason to couple such move to the scope of this PR. It will be also smaller and easier to review, with less friction points.

zimnx: Consistency wise, we don't have any as there're commands having options structure in the same file, as well as in separate ones (genapireference, releasenotes, tests), so it's a matter of personal preference.
Removed this improvement (IMO), to avoid this pointless discussion.

pkg/cmd/tests/options.go
```diff
@@ -35,6 +39,9 @@ var supportedBroadcastAddressTypes = []scyllav1.BroadcastAddressType{
 }
 
 type TestFrameworkOptions struct {
+	genericclioptions.ClientConfig
+	RunOptions
```
tnozicka, line 43: Why is this reversing the existing inheritance?
https://github.com/scylladb/scylla-operator/pull/1824/files#diff-679d6ea01ce94872b44bde356e4eb82af41ca51988762be80625b957f8834a66L75

RunOptions are the top level entry point (named command) and contain an instance of TestFrameworkOptions (specific internal logic, not a named command).

tnozicka (changes requested):

pkg/cmd/tests/options.go
tnozicka: >  to avoid this pointless discussion

This is not appropriate, nor on point. The bar (and the point) goes the other way. You chose to couple this opinionated change to the **existing code** to this PR (which didn't need it and could have had a diff order of magnitude smaller) so the *argument* really applies vice versa.

pkg/cmd/tests/options.go
zimnx: yep, should be other way around. fixed

pkg/cmd/tests/tests_run.go
```diff
@@ -43,36 +40,9 @@ const (
...
-		Tests that ensure an Scylla Operator is working properly.
-		`),
-		LabelFilter:        fmt.Sprintf("%s", framework.SerialLabelName),
-		DefaultParallelism: 1,
-	},
-}
-
 type RunOptions struct {
 	genericclioptions.IOStreams
-	genericclioptions.ClientConfig
```
tnozicka, line 74: is this a leftover from https://github.com/scylladb/scylla-operator/pull/1824#discussion_r1590825458 and should be brought back as well?

pkg/cmd/tests/tests_run.go
```diff
@@ -43,36 +40,9 @@ const (
...
-	TestFrameworkOptions
+	*TestFrameworkOptions
```
tnozicka, line 45: why changing the embedding to a pointer?

pkg/cmd/tests/tests_run.go
zimnx: It's part of TestFrameworkOptions as framework requires it.

pkg/cmd/tests/tests_run.go
tnozicka: It's effectively inheritance and the object is "required" - the pointer makes it effectively optional and forces callers to check for nil on something that suppose to always be there.

zimnx: changed

pkg/cmd/tests/tests_run.go
```diff
@@ -90,14 +60,13 @@ type RunOptions struct {
 	ParallelServerAddress string
 	ParallelLogLevel      int32
 
+	testSuites    ginkgotest.TestSuites
```
tnozicka, line 63: I'd expect this to be public, so it's easily extensible as the other fields

pkg/cmd/tests/tests_run.go
```diff
@@ -113,18 +82,32 @@ func NewRunOptions(streams genericclioptions.IOStreams) *RunOptions {
...
 		ParallelServerAddress: "",
 		ParallelLogLevel:      0,
+
+		testSuites: testSuites,
 	}
 }
 
-func NewRunCommand(streams genericclioptions.IOStreams) *cobra.Command {
-	o := NewRunOptions(streams)
+var AllTestSuite = &ginkgotest.TestSuite{
```
tnozicka, line 90: this feels a bit off to be defined in a different file then the Suites

pkg/cmd/tests/tests_run.go
zimnx: done

pkg/cmd/tests/tests_run.go
zimnx: removed

tnozicka:
thanks for the updates
/approve
/lgtm

zimnx:
apiserver died + manager flake
/retest

zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1028#issuecomment-2112791896
/retest

PR #1837, opened 2024-03-15
Add CI script to deploy published releases

### LINKED ISSUE #1936: Add CI script to deploy published releases
We need a script that will be used in periodics to test our published images. It essentially codifies what needs to be done to deploy a particular scylla-operator version.

tnozicka:
This script will be used in periodics to test our published images. It essentially codifies what needs to be done to deploy a particular scylla-operator version. Some action could be simplified by restructuring manifest but it's out of the scope of this PR. Eventually, we should fix this and possibly have some smarter templating / publish manifests into releases with the replaced image. So take this PR as a start with what we have now, not where it should end.

Resolves #1936 

### Requires
- [x] #1901 
- [x] #1922
- [x] https://github.com/scylladb/scylla-operator-release/pull/193
- [x] https://github.com/scylladb/scylla-operator-release/pull/194

### DIFF
--- a/hack/ci-deploy-release.sh
+++ b/hack/ci-deploy-release.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+#
+# Copyright (C) 2024 ScyllaDB
+#
+# This script deploys scylla-operator stack.
+# Usage: ${0} <operator_image_ref>
+# (Avoid using rolling tags.)
+
+set -euxEo pipefail
+shopt -s inherit_errexit
+
+source "$( dirname "${BASH_SOURCE[0]}" )/lib/kube.sh"
+
+if [[ -n "${1+x}" ]]; then
+    operator_image_ref="${1}"
+else
+    echo "Missing operator image ref.\nUsage: ${0} <operator_image_ref>" >&2 >/dev/null
+    exit 1
+fi
+
+source_raw="$( skopeo inspect --format='{{ index .Labels "org.opencontainers.image.source" }}' "docker://${operator_image_ref}" )"
+if [[ -z "${source_raw}" ]]; then
+    echo "Image '${operator_image_ref}' is missing source label" >&2 >/dev/null
+    exit 1
+fi
+source_url="${source_raw/"://github.com/"/"://raw.githubusercontent.com/"}"
+
+revision="$( skopeo inspect --format='{{ index .Labels "org.opencontainers.image.revision" }}' "docker://${operator_image_ref}" )"
+if [[ -z "${revision}" ]]; then
+    echo "Image '${operator_image_ref}' is missing revision label" >&2 >/dev/null
+    exit 1
+fi
+
+ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}"
+
+kubectl_create -n=prometheus-operator -f="${source_url}/${revision}/examples/third-party/prometheus-operator.yaml"
+kubectl_create -n=haproxy-ingress -f="${source_url}/${revision}/examples/third-party/haproxy-ingress.yaml"
+
+kubectl_create -f="${source_url}/${revision}/examples/common/cert-manager.yaml"
+# Wait for cert-manager crd and webhooks
+kubectl wait --for condition=established --timeout=60s crd/certificates.cert-manager.io crd/issuers.cert-manager.io
+for d in cert-manager{,-cainjector,-webhook}; do
+    kubectl -n=cert-manager rollout status --timeout=5m "deployment.apps/${d}"
+done
+wait-for-object-creation cert-manager secret/cert-manager-webhook-ca
+
+mkdir "${ARTIFACTS}/operator"
+cat > "${ARTIFACTS}/operator/kustomization.yaml" << EOF
+resources:
+- ${source_url}/${revision}/deploy/operator.yaml
+patches:
+- patch: |-
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      name: scylla-operator
+      namespace: scylla-operator
+    spec:
+      template:
+        spec:
+          containers:
+          - name: scylla-operator
+            image: "${operator_image_ref}"
+            env:
+            - name: SCYLLA_OPERATOR_IMAGE
+              value: "${operator_image_ref}"
+EOF
+kubectl kustomize "${ARTIFACTS}/operator" | kubectl_create -n=scylla-operator -f=-
+
+# Manager needs scylla CRD registered and the webhook running
+kubectl wait --for condition=established crd/{scyllaclusters,nodeconfigs}.scylla.scylladb.com
+kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/scylla-operator
+kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/webhook-server
+
+if [[ -z "${SO_NODECONFIG_PATH:-}" ]]; then
+ echo "Skipping NodeConfig creation"
+else
+ kubectl_create -f="${SO_NODECONFIG_PATH}"
+fi
+
+kubectl_create -n=local-csi-driver -f="${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/"{00_namespace.yaml,00_scylladb-local-xfs.storageclass.yaml,10_csidriver.yaml,10_driver.serviceaccount.yaml,10_provisioner_clusterrole.yaml,20_provisioner_clusterrolebinding.yaml,50_daemonset.yaml}
+kubectl -n=local-csi-driver rollout status --timeout=5m daemonset.apps/local-csi-driver
+
+mkdir "${ARTIFACTS}/manager"
+cat > "${ARTIFACTS}/manager/kustomization.yaml" << EOF
+resources:
+- ${source_url}/${revision}/deploy/manager-prod.yaml
+patches:
+- target:
+    group: scylla.scylladb.com
+    version: v1
+    kind: ScyllaCluster
+    name: scylla-manager-cluster
+  patch: |
+    - op: replace
+      path: /spec/datacenter/racks/0/storage/storageClassName
+      value: scylladb-local-xfs
+EOF
+kubectl kustomize "${ARTIFACTS}/manager" | kubectl_create -n=scylla-manager -f=-
+
+wait-for-object-creation scylla-manager statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
+kubectl -n=scylla-manager rollout status --timeout=5m statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
+kubectl -n=scylla-manager rollout status --timeout=5m deployment.apps/scylla-manager
+kubectl -n=scylla-manager rollout status --timeout=5m deployment.apps/scylla-manager-controller
+
+kubectl -n=haproxy-ingress rollout status --timeout=5m deployment.apps/haproxy-ingress
+
+kubectl wait --for condition=established crd/{scyllaoperatorconfigs,scylladbmonitorings}.scylla.scylladb.com
+kubectl wait --for condition=established crd/{prometheuses,prometheusrules,servicemonitors}.monitoring.coreos.com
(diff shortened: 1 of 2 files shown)


tnozicka:
this can't be used by presubmits by definition as their merged git shas are not published - this is only for published releases

rzetelskik:
> this can't be used by presubmits by definition as their merged git shas are not published - this is only for published releases

I know, but the question was about how is this going to be integrated with CI, pointing at whether this PR is potentially missing any adjustments.

tnozicka:
> I know, but the question was about how is this going to be integrated with CI, pointing at whether this PR is potentially missing any adjustments.

This will be used only by new periodics based on already published images (we don't have these periodics yet). 
(Eventually, we can add an on demand presubmit using this script against a latest release to smoke test it.)

tnozicka:
added the extra presubmit in https://github.com/scylladb/scylla-operator-release/pull/157

zimnx (approved):
/lgtm

please fix CI script
/hold

rzetelskik:
https://github.com/scylladb/scylla-operator-release/pull/163

rzetelskik:
`time="2024-03-28T13:38:18Z" level=fatal msg="Error parsing image name \"docker://docker.io/scylladb/scylla-operator:latest@sha256:0c8896a666fdc0d273b92dbd1bd936140d1da34e6c25b9f7ac1260a7acc18f16\": Docker references with both a tag and digest are currently not supported"`

needs a fix

tnozicka:
fixed in https://github.com/scylladb/scylla-operator-release/pull/168

tnozicka:
https://github.com/scylladb/scylla-operator-release/pull/168 has landed
/retest

tnozicka:
cluster didn't spin up in time
/retest

tnozicka:
oh, I've tested this but seems like it depends on `extglob` which I had enabled and the image doesn't. need to fix that in the CI config

tnozicka:
https://github.com/scylladb/scylla-operator-release/pull/180 landed
/retest

tnozicka:
(this need more work on artifact collection and specifying nodeconfig on CI side)

hack/ci-deploy.sh
```diff
@@ -15,10 +15,10 @@ if [[ -z ${1+x} ]]; then
     exit 1
 fi
 
-ARTIFACTS_DIR=${ARTIFACTS_DIR:-$( mktemp -d )}
```
rzetelskik, line 18: I assume this is coming from https://github.com/scylladb/scylla-operator/pull/1922, please rebase

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,107 @@
...
+      namespace: scylla-operator
+    spec:
+      template:
+        spec:
+          containers:
+          - name: scylla-operator
+            image: "${operator_image_ref}"
+            env:
+            - name: SCYLLA_OPERATOR_IMAGE
+              value: "${operator_image_ref}"
```
rzetelskik, line 66: is this missing the tweaks from ci-deploy (https://github.com/scylladb/scylla-operator/blob/master/hack/ci-deploy.sh#L34-L35) on purpose? If so, is it superfluous there or accidentally missing here?

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,107 @@
...
+ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}"
+
+kubectl_create -n=prometheus-operator -f="${source_url}/${revision}/examples/third-party/prometheus-operator.yaml"
+kubectl_create -n=haproxy-ingress -f="${source_url}/${revision}/examples/third-party/haproxy-ingress.yaml"
+
+kubectl_create -f="${source_url}/${revision}/examples/common/cert-manager.yaml"
+# Wait for cert-manager crd and webhooks
+kubectl wait --for condition=established --timeout=60s crd/certificates.cert-manager.io crd/issuers.cert-manager.io
+for d in cert-manager{,-cainjector,-webhook}; do
+    kubectl -n=cert-manager rollout status --timeout=5m "deployment.apps/${d}"
```
rzetelskik, line 43: missing wait-for-object-creation from ci-deploy: https://github.com/scylladb/scylla-operator/blob/b7ca9a4fce4aa060cec0bb8139382fdf8d909928/hack/ci-deploy.sh#L48. Is it superfluous there or accidentally missing here?

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,107 @@
...
+kubectl kustomize "${ARTIFACTS}/operator" | kubectl_create -n=scylla-operator -f=-
+
+# Manager needs scylla CRD registered and the webhook running
+kubectl wait --for condition=established crd/{scyllaclusters,nodeconfigs}.scylla.scylladb.com
+kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/scylla-operator
+kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/webhook-server
+
+if [[ -n "${SO_NODECONFIG_PATH:-}" ]]; then
+    kubectl_create -f="${SO_NODECONFIG_PATH}"
+fi
```
rzetelskik, line 77: diverges from ci-deploy: https://github.com/scylladb/scylla-operator/blob/b7ca9a4fce4aa060cec0bb8139382fdf8d909928/hack/ci-deploy.sh#L60-L64

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,107 @@
...
+
+wait-for-object-creation scylla-manager statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
+kubectl -n=scylla-manager rollout status --timeout=5m statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
+kubectl -n=scylla-manager rollout status --timeout=5m deployment.apps/scylla-manager
+kubectl -n=scylla-manager rollout status --timeout=5m deployment.apps/scylla-manager-controller
+
+kubectl -n=haproxy-ingress rollout status --timeout=5m deployment.apps/haproxy-ingress
+
+kubectl wait --for condition=established crd/{scyllaoperatorconfigs,scylladbmonitorings}.scylla.scylladb.com
+kubectl wait --for condition=established crd/{prometheuses,prometheusrules,servicemonitors}.monitoring.coreos.com
```
rzetelskik, line 107: ditto https://github.com/scylladb/scylla-operator/blob/b7ca9a4fce4aa060cec0bb8139382fdf8d909928/hack/ci-deploy.sh#L81-L84

rzetelskik (changes requested):

hack/ci-deploy.sh
tnozicka: weird, fixed with a clean rebase

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,107 @@
...
+    group: scylla.scylladb.com
+    version: v1
+    kind: ScyllaCluster
+    name: scylla-manager-cluster
+  patch: |
+    - op: replace
+      path: /spec/datacenter/racks/0/storage/storageClassName
+      value: scylladb-local-xfs
+EOF
+kubectl kustomize "${ARTIFACTS}/manager" | kubectl_create -n=scylla-manager -f=-
```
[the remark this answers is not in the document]
tnozicka: I don't think those have to match as these are different use cases. This should be eventually dropped and all our manifest should pick the correct storage class without relying on the defaults or needing messing with the global classes. But our provisioner doesn't support the full disks yet, so we are not there yet.

hack/ci-deploy-release.sh
tnozicka: done

hack/ci-deploy-release.sh
tnozicka: nodeconfigs are waited for a bit earlier
`find` doesn't work with `https://`

tnozicka:
@rzetelskik please have a look at the comment replies, I'd like to land this and unblock your arm64 PR

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Manager needs scylla CRD registered and the webhook running
+kubectl wait --for condition=established crd/{scyllaclusters,nodeconfigs}.scylla.scylladb.com
+kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/scylla-operator
+kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/webhook-server
+
+if [[ -n "${SO_NODECONFIG_PATH:-}" ]]; then
+    kubectl_create -f="${SO_NODECONFIG_PATH}"
+fi
+
```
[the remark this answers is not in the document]
rzetelskik: it shouldn't be defined here then? https://github.com/scylladb/scylla-operator/pull/1837/files#diff-5da3c104525c31203c7505086de76d48b279733088c4ec836270ee1aabcde637R23

hack/ci-deploy-release.sh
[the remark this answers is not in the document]
rzetelskik: Imo having them diverge jut makes it harder to find the root cause if anything fails, and doing the same thing for the regular ci-deploy would be preferred?

hack/ci-deploy-release.sh
rzetelskik: should it be removed from the other script then? (not in this pr)

rzetelskik:
> @rzetelskik please have a look at the comment replies, I'd like to land this and unblock your arm64 PR

done

hack/ci-deploy-release.sh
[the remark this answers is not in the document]
tnozicka: good catch, that one shouldn't be there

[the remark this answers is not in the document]
tnozicka: fixed

hack/ci-deploy-release.sh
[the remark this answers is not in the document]
tnozicka: only temporarily, I've added an explicit mention to clean this up in https://github.com/scylladb/scylla-operator/issues/1554

hack/ci-deploy-release.sh
tnozicka: I think so

rzetelskik: should we have a tracker for this? Even if it's in the backlog

rzetelskik (approved):
lgtm
/assign zimnx
in case you want to have a second look at it, it changed since your review. Otherwise feel free to remove the hold.

hack/ci-deploy-release.sh
tnozicka: https://github.com/scylladb/scylla-operator/pull/1945

hack/ci-deploy-release.sh
```diff
@@ -0,0 +1,109 @@
...
+    name: scylla-manager-cluster
+  patch: |
+    - op: replace
+      path: /spec/datacenter/racks/0/storage/storageClassName
+      value: scylladb-local-xfs
+EOF
+kubectl kustomize "${ARTIFACTS}/manager" | kubectl_create -n=scylla-manager -f=-
+
+wait-for-object-creation scylla-manager statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
+kubectl -n=scylla-manager rollout status --timeout=5m statefulset.apps/scylla-manager-cluster-manager-dc-manager-rack
```
zimnx, line 102: nit: these could be replaced with waiting for our three aggregated conditions on ScyllaCluster without hardcoding names of Operator managed resources.

zimnx (approved):
One nit. Feel free to cancel hold even without applying, i'm fine with having both options. 

/lgtm
/hold

hack/ci-deploy-release.sh
tnozicka: It's a good point but it's in ci-deploy.sh as well, I'll send a followup.
/hold cancel

tnozicka: https://github.com/scylladb/scylla-operator/pull/1947

PR #1848, opened 2024-03-19
Add integrity information to must-gather

### LINKED ISSUE #1847: Compute integrity for must-gather dumps
We should add integrity information to must-gather dump to help us identify when it was tempered with. Removing some sensitive files can be justified but in my experience it is greatly misused but most importantly never reported as such, so we end up wondering whether the dumps are not present because they were removed, or because the collection didn't work properly.

tnozicka:
This PR add integrity information to must-gather dump. This will help to identify whether files have been removed or the collection failed.

Resolves #1847

### DIFF
--- a/pkg/cmd/operator/mustgather.go
+++ b/pkg/cmd/operator/mustgather.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"time"
 
@@ -319,6 +320,12 @@ func (o *MustGatherOptions) run(ctx context.Context) error {
 			}
 		}
 	}
+
+	integritySHA, err := collector.WriteIntegrityFile()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("can't write integrity file: %w", err))
+	}
+	klog.InfoS("Witten integrity file", "SHA", hex.EncodeToString(integritySHA))
 
 	return utilerrors.NewAggregate(errs)
 }
--- a/pkg/cmd/operator/mustgather_test.go
+++ b/pkg/cmd/operator/mustgather_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
+	"github.com/scylladb/scylla-operator/pkg/gather/collect"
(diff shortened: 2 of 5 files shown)


rzetelskik (commented):
I only had a brief look - wouldn't a merkle tree be a standard approach for this use case?

tnozicka:
> I only had a brief look - wouldn't a merkle tree be a standard approach for this use case?

It's valid but more for cryptography.

The approach here allows you to easily identify the places with changes with respect to the particular resource and couples well with linux CLI tools. Say, to verify a folder manually, you can run `find scylla-operator-must-gather/namespaces/scylla-operator/serviceaccounts -mindepth 1 -exec cat {} \; | sha512sum -` and you have the explicit count so you at least know how many files are missing.

rzetelskik:
> > I only had a brief look - wouldn't a merkle tree be a standard approach for this use case?
> 
> It's valid but more for cryptography.
> 
> The approach here allows you to easily identify the places with changes with respect to the particular resource and couples well with linux CLI tools. Say, to verify a folder manually, you can run find scylla-operator-must-gather/namespaces/scylla-operator/serviceaccounts -mindepth 1 -exec cat {} \; | sha512sum -

would this approach also allow you to check if a given file was originally a part of the archive?

> and you have the explicit count so you at least know how many files are missing.

is that actually a valuable information if you don't know which ones are missing?

Since you already put the work into this I don't feel strong about using the merkle tree (altough there probable are many existing implementations), I'm struggling to understand what is the purpose of this exactly if we don't have the capability of signing it (correct me if I'm wrong here)? The PR description says `This will help to identify whether files have been removed or the collection failed`, so it's working on an assumption that someone purposefully tampered with the archive, so they are already "malicious" - what's to stop them from tampering with the integrity information as well? Would you be able to tell the modified archive apart from an archive in which some manifests just weren't collected?

rzetelskik:
> > is that actually a valuable information if you don't know which ones are missing?
> 
> The most pressing concern is temper detection, after that point I'd usually just not proceed further.

Are you saying that we wouldn't investigate a customer issue if they removed **any** file, even if we weren't sure it was necessarily related to Scylla?

> > what's to stop them from tampering with the integrity information as well?
> 
> nothing, with any approach we take. they are tempering with the archive with the intent to limit the data to what they think is not related (usually wrongly), not to be intentfully malicious.

I mean, I get the point, but at this point you'd have to assume they deleted a related file and assume a somewhat hostile approach. I'm not sure how that'd turn out in practice.

Before we go into reviewing the implementation, @zimnx what are your thoughts on this?

PR #1849, opened 2024-03-19
Add Scheme Builder

zimnx:
It's useful when other project wants to add all types defined by Scylla Operator to its schema.

### DIFF
--- a/pkg/scheme/scheme.go
+++ b/pkg/scheme/scheme.go
@@ -25,15 +25,17 @@ var (
 			Strict: true,
 		},
 	)
+	localSchemeBuilder = runtime.SchemeBuilder{
+		kscheme.AddToScheme,
+		scyllav1.Install,
+		scyllav1alpha1.Install,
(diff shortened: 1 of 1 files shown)


pkg/scheme/scheme.go
```diff
@@ -25,15 +25,18 @@ var (
 			Strict: true,
 		},
 	)
+	localSchemeBuilder = runtime.SchemeBuilder{
+		scyllav1.Install,
+		scyllav1alpha1.Install,
+		cqlclientv1alpha1.Install,
+		monitoringv1.Install,
```
tnozicka, line 32: I think it should be all of them, including kube. I assume the intended to build a superset so is effectively a deep copy + it's hard to draw a line between monitoring being an external API and kube not being included.

zimnx: makes sense, added.

zimnx:
slow schema change
/retest

tnozicka (approved):
/approve
/lgtm

Issue #1752, opened 2024-02-26
Scylla manager controller deletes and creates tasks even without hitting conflicts


rzetelskik:
### What happened?

There's an issue with manager task synchronisation in scylla manager controller. On each reconciliation iteration, the controller goes through the tasks gathered from current manager state, and deletes the ones which are missing from ScyllaCluster status.
https://github.com/scylladb/scylla-operator/blob/c8901da8c23f21c078d1720ea31979afa35b1e8e/pkg/controller/manager/sync_action.go#L273

What that means in practice is that the tasks can be deleted right after they've been scheduled and saved in the object status. And it doesn't even require the status update to fail, it's enough that the next key in queue is the previous generation of the object without the updated status. This can result in many iterations of task creations and deletions.

Example logs:
```console
2024-02-26T14:50:26.589360595Z I0226 14:50:26.589344       1 manager/sync.go:93] "Started syncing ScyllaCluster" ScyllaCluster="e2e-test-scyllacluster-g6xgc-k
v4qg/basic-sh88x" startTime="2024-02-26 14:50:26.589328928 +0000 UTC m=+236.630081103"
2024-02-26T14:50:26.614357881Z I0226 14:50:26.614311       1 manager/sync.go:134] "Executing action" action="add task &{ClusterID: Enabled:true ID: Name:weekl
y backup Properties:map[location:[gcs:so-f2w8df7xf4c556cwbv64v5x8z96dxbf848cd62f5xfx5fxc25427c484d567] retention:1] Schedule:0xc0004d98f0 Tags:[] Type:backup}
"
2024-02-26T14:50:31.461267373Z I0226 14:50:31.461196       1 manager/sync.go:144] "Updating cluster status" new={"observedGeneration":2,"racks":{"us-east-1a":
{"version":"5.4.0","members":1,"readyMembers":1,"updatedMembers":1,"stale":false}},"managerId":"ed7d66c3-beee-4910-bc83-c06a41feebb7","backups":[{"name":"week
ly backup","startDate":"now","interval":"7d","numRetries":0,"location":["gcs:so-f2w8df7xf4c556cwbv64v5x8z96dxbf848cd62f5xfx5fxc25427c484d567"],"retention":1,"
id":"1d81d2e6-5330-434b-afbe-497c7f85b0dc","error":""}],"conditions":[{"type":"ServiceAccountControllerDegraded","status":"False","observedGeneration":2,"last
[...]

PR #1851, opened 2024-03-19
Support cron and timezone in manager task integration

### LINKED ISSUE #1739: Support `cron` in ScyllaCluster API for Scylla Manager tasks

With https://github.com/scylladb/scylla-manager/pull/3720 and existing, partial deprecation of `start_date`, we should add support for [`cron`](https://github.com/scylladb/scylla-manager/blob/d86640b0c3f64599be6a56b4590f294a26bd07f2/swagger/scylla-manager.json#L270) in our API for Scylla Manager tasks. 

ditto

_No response_

### LINKED ISSUE #1796: Deprecate "interval" in ScyllaCluster's manager task API

The `interval` parameter has been deprecated in the most recent Scylla Manager release - 3.2.6. We should deprecate it in our API as well.

ditto

_No response_

rzetelskik:
**Description of your changes:** This PR adds missing cron and related fields, currently missing from out API, to manager task specs and extends the unit tests to cover them.

Resolves #1739 

/kind feature
/priority important-longterm
/cc


rzetelskik:
/priority critical-urgent
since the bug fixes are a prerequisite for https://github.com/scylladb/scylla-operator/issues/1752

rzetelskik:
/cc @Michal-Leszczynski 
would you be able to verify if there's anything missing from manager task specs in our API?

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -323,6 +323,7 @@ type SchedulerTaskSpec struct {
 	StartDate string `json:"startDate,omitempty"`
 
 	// interval represents a task schedule interval e.g. 3d2h10m, valid units are d, h, m, s.
+	// Deprecated: please use cron instead.
```
tnozicka, line 326: this is not in scope in the description of this PR, please split this so its visible in the release notes

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -331,6 +332,23 @@ type SchedulerTaskSpec struct {
...
+	// +optional
+	Cron string `json:"cron,omitempty"`
+
+	// retryWait specifies an initial exponential backoff duration X[h|m|s].
+	// +optional
+	RetryWait string `json:"retryWait,omitempty"`
+
+	// timezone specifies the timezone of window and cron fields.
+	// +optional
+	Timezone string `json:"timezone,omitempty"`
```
tnozicka, line 345: `Therefore, we ask that pointers always be used with optional fields that do not have a built-in nil value.`
https://github.com/kubernetes/community/blob/e9f78a/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

pkg/api/scylla/v1/types_cluster.go
rzetelskik: ok, https://github.com/scylladb/scylla-operator/pull/1853

tnozicka (changes requested):
please fix the API to use pointers where needed

pkg/api/scylla/v1/types_cluster.go
rzetelskik: > that do not have a built-in nil value

""?

tnozicka: that's not `nil` value, but a `zero` value

rzetelskik: for instance: https://github.com/scylladb/scylla-operator/blob/1520cf77e2bd089ad3497264dbd86196a4586f29/pkg/api/scylla/v1/types_cluster.go#L106

rzetelskik: Ok, I see there are counterexamples in our api as well, so that's not really an argument. I'll amend it

rzetelskik: done

Michal-Leszczynski:
Feel free to ignore my comments if they are not related to this PR. I just wanted to highlight some fields that perhaps require additional attention and give context for some of them.

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -375,6 +392,14 @@ type RepairTaskSpec struct {
...
 	Host *string `json:"host,omitempty" mapstructure:"host,omitempty"`
+
+	// ignoreDownHosts specifies that the manager should not attempt to repair the nodes that are down, i.e. in status DN.
+	// +optional
+	IgnoreDownHosts bool `json:"IgnoreDownHosts,omitempty" mapstructure:"ignore_down_hosts,omitempty"`
+
+	// tags specifies an extra set of tags describing the task.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
 }
```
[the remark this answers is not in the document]
rzetelskik: > This is not repair-specific field

The api on our side doesn't really match SM's, so there isn't really a better place for it - the shared inline struct is `SchedulerTaskSpec`, which doesn't fit the case here. The fields in RepairTaskSpec/BackupTaskSpec seem to be a mix of mapped properties and some shared fields, some even taken out of the `Schedule`on SM side. tl;dr; it's a mess that we have to live with now for backwards compatibility

> Do you have a use case for this field?

Nope, I just didn't want to leave it out while synchronising all the other fields.

> first validate its existence on SM side.

Well, it's in the model https://github.com/scylladb/scylla-manager/commit/5a73edfd5b2c5b344829ddd58adec641d965d245

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -331,6 +331,23 @@ type SchedulerTaskSpec struct {
 	// +kubebuilder:default:=3
 	// +optional
 	NumRetries *int64 `json:"numRetries,omitempty"`
```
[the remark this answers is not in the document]
rzetelskik: It's an api convention, @tnozicka even pointed it out higher up in the comments https://github.com/scylladb/scylla-operator/pull/1851#discussion_r1531841953.
This particular field has a default, so it doesn't exactly make sense. Out of scope for this PR anyway.

rzetelskik:
> The now syntax can cause a lot of pain since it is parsed on the operator or sctool side. Maybe it's worth to discourage users from using via appropriate field description?

Agreed. Is that mentioned in sctool/sm? 

> The interval field is (or should be) deprecated by SM. Is it also deprecated here?

There's a separate PR in queue with interval deprecation. Best we can do is a field description though.

> Why is it string? SM supports float64 for backward compatibility, but all floats are converted to ints anyway, because they don't make sense anymore from Scylla POV.

To be honest - no idea. I don't think we can change this in this API version though.

pkg/api/scylla/v1/types_cluster.go
tnozicka: I agree mistakes were made :)

thanks for the update

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -408,6 +433,11 @@ type BackupTaskSpec struct {
 	// +optional
 	Retention int64 `json:"retention,omitempty" mapstructure:"retention,omitempty"`
 
+	// retention is the number of days for which a backup is stored. The next scheduled backup initiates a purge of expired backups.
+	// Can be used in tandem with `retention`.
+	// +optional
+	RetentionDays *int `json:"retentionDays,omitempty" mapstructure:"retention_days,omitempty"`
```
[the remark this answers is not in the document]
Michal-Leszczynski: But then there is a difference between operator and SM.
Running `sctool backup --retention-days 3` and corresponding (in a direct way) operator definition results in a different retention policy. Of course it's up to operator team to decide that, but I just wanted to highlight that.

pkg/api/scylla/v1/types_cluster.go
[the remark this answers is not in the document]
tnozicka: I would scope this down to the fields we see a meaning for. We can always add a field but never to remove one. So let's avoid adding them just because they are present in manager API. I don't see a big reason for `IgnoreDownHosts` either at this point.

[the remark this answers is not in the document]
rzetelskik: Removed the tags. As for `IgnoreDownHosts`, I think a reasonable rule of thumb is to allow setting whatever sctool allows for.

rzetelskik:
split mapstructure fix into a separate PR: https://github.com/scylladb/scylla-operator/pull/1854
/hold for https://github.com/scylladb/scylla-operator/pull/1854

pkg/api/scylla/v1/types_cluster.go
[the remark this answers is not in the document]
rzetelskik: @tnozicka after internal conversation I removed the other fields, only leaving `cron` and `timezone` related to it

rzetelskik:
ping @tnozicka

pkg/api/scylla/v1/types_cluster.go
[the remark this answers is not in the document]
tnozicka: > imo the only reasonable thing to do would be to drop the defaults on our side altogether

you can stop defaulting the field but you can't make it required

We should not use defaults, if we want the manager to accept optional values. But that ship has half sailed. We default at admission and after that point the values are stored and always used as explicitly specified.

And defaulting or the lack of shall not render object invalid.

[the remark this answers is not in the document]
rzetelskik: I'll verify which defaults could be dropped safely, but that's for a separate PR anyway, since this ended up just adding cron and timezone. So let's get back to it once all the fixes in queue land, I don't want to introduce extra variables.

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -421,6 +451,14 @@ type BackupTaskSpec struct {
...
 	// +optional
 	UploadParallel []string `json:"uploadParallel,omitempty" mapstructure:"upload_parallel,omitempty"`
+
+	// purgeOnly specifies that only a backup cleanup should be run.
+	// +optional
+	PurgeOnly bool `json:"purgeOnly,omitempty" mapstructure:"purge_only,omitempty"`
+
+	// tags specifies an extra set of tags describing the task.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
```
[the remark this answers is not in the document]
rzetelskik: removed

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -331,6 +331,14 @@ type SchedulerTaskSpec struct {
...
 	// +optional
 	NumRetries *int64 `json:"numRetries,omitempty"`
+
+	// cron specifies the task schedule as a cron expression. It supports an extended syntax including @monthly, @weekly, @daily, @midnight, @hourly, @every X[h|m|s].
+	// +optional
+	Cron *string `json:"cron,omitempty"`
+
+	// timezone specifies the timezone of cron field.
+	// +optional
+	Timezone *string `json:"timezone,omitempty"`
```
tnozicka, line 341: if this is truly only for `cron`, maybe it should be named `cronTimezone` to avoid implications on `startDate` field?

rzetelskik: It's also for 'window', which was excluded from this PR. I think naming it that way would kill potential extensibility. And I would refrain from interfering with the upstream api to much in general, it makes the ux worse imo.

tnozicka: > It's also for 'window',

ok, with that is makes sense.

tnozicka, line 341: please add validation making sure it's set only if the fields that it applies to are present

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -331,6 +331,14 @@ type SchedulerTaskSpec struct {
 	// +kubebuilder:default:=3
 	// +optional
 	NumRetries *int64 `json:"numRetries,omitempty"`
+
+	// cron specifies the task schedule as a cron expression. It supports an extended syntax including @monthly, @weekly, @daily, @midnight, @hourly, @every X[h|m|s].
+	// +optional
+	Cron *string `json:"cron,omitempty"`
```
tnozicka, line 337: is this mutually exclusive with start date? should you validate both are not set?

pkg/api/scylla/v1/types_cluster.go
rzetelskik: wouldn't it be better to make it transparent, let the client return an error and propagate it to status (globally)?

pkg/controller/manager/types_old.go
```diff
@@ -37,16 +38,18 @@ func (r RepairTask) ToManager() (*managerclient.Task, error) {
 		return nil, errors.Wrap(err, "parse start date")
 	}
 	t.Schedule.StartDate = &startDate
-
-	if _, err := duration.ParseDuration(r.Interval); err != nil {
-		return nil, errors.Wrap(err, "parse interval")
-	}
```
zimnx, line 43: is it part of webhook validation? If not, please move it there.

pkg/api/scylla/v1/types_cluster.go
rzetelskik: I don't see sctool doing this. Looking at https://github.com/scylladb/scylla-manager/pull/3720#issuecomment-1956621668 and https://github.com/scylladb/scylla-manager/issues/3750 I'm not even sure I understand how they interact with each other. @Michal-Leszczynski can you advise?

pkg/controller/manager/types_old.go
rzetelskik: I'll revert this change here as it's unrelated and I'll open a separate PR for it. Preferably after https://github.com/scylladb/scylla-operator/pull/1853 and https://github.com/scylladb/scylla-operator/pull/1850 land.

rzetelskik: reverted

pkg/api/scylla/v1/types_cluster.go
Michal-Leszczynski: @karol-kokoszka please correct me if I'm wrong, but `cron` works with `startDate` in a way that the first `cron` activation will scheduled after the `startDate` (not necessarily on the `startDate` itself). So the `startDate` can be perceived as a date when the `cron` task is added to SM.

karol-kokoszka: > but cron works with startDate in a way that the first cron activation will scheduled after the startDate

Indeed.

rzetelskik: so to my understanding cron is mutually exclusive with interval, but both can work with start date?

rzetelskik: Added validation for mutual exclusivity of cron and interval. A bit awkward since interval is not a pointer and defaulted.

tnozicka: yeah, I think even with the startDate it's weird if your `startDate` is say on Monday and cron says it should run on Fridays. Lets keep the validation tight as we can never go back, but loosening is possible if a use case is found that is more valuable then the conflict I pointed out here.

pkg/api/scylla/v1/types_cluster.go
tnozicka: >  the timezone will just be ignored

well, that suck even a bit more, since the user has a timezone set thinking all is well, but it silently runs in a different time then he though

pkg/api/scylla/v1/types_cluster.go
rzetelskik: I'll add the validation here, also sent a separate PR moving all existing validation to the validating webhook: https://github.com/scylladb/scylla-operator/pull/1868.
I also created a tracker for the other fields which won't be cover in these two PRs: https://github.com/scylladb/scylla-operator/issues/1869.

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -216,6 +216,41 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
...
+
+	return allErrs
+}
+
+func ValidateSchedulerTaskSpec(schedulerTaskSpec *scyllav1.SchedulerTaskSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if schedulerTaskSpec.Interval != nil {
+		if intervalDuration, err := duration.ParseDuration(*schedulerTaskSpec.Interval); err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("interval"), schedulerTaskSpec.Interval, "invalid interval format, valid units are d, h, m, s"))
```
zimnx, line 245: duplicated "invalid"

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -216,6 +216,41 @@ func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.
 	return allErrs
 }
 
+func ValidateRepairTaskSpec(repairTaskSpec *scyllav1.RepairTaskSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	_, err := strconv.ParseFloat(repairTaskSpec.Intensity, 64)
+	if err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("intensity"), repairTaskSpec.Intensity, "invalid intensity, it must be a float value"))
```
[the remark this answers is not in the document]
rzetelskik: that'll be a part of https://github.com/scylladb/scylla-operator/pull/1868, already removed there

pkg/api/scylla/validation/cluster_validation.go
rzetelskik: that'll be a part of https://github.com/scylladb/scylla-operator/pull/1868, already removed there

pkg/api/scylla/v1/types_cluster.go
rzetelskik: done

pkg/api/scylla/v1/types_cluster.go
rzetelskik: done

rzetelskik:
@zimnx @tnozicka I believe I have addressed all your comments - please have another go.

pkg/api/scylla/v1/types_cluster.go
rzetelskik: @tnozicka I wonder, should we also validate the extended cron format on our side? We don't control it, so if it changes upstream, we'll get out of sync

tnozicka: I'd start with a strict validation of the format and loosen if we bump into something. Kube seems to use a cron library as well, so this might be a good start https://github.com/kubernetes/kubernetes/blob/e800d15/go.mod#L58C2-L58C27

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -221,6 +224,35 @@ func ValidateRepairTaskSpec(repairTaskSpec *scyllav1.RepairTaskSpec, fldPath *fi
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("interval"), schedulerTaskSpec.Interval, "valid units are d, h, m, s"))
+		} else if intervalDuration != 0 {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("interval"), "can't be non-zero when cron is specified"))
+		}
+	}
+
+	if schedulerTaskSpec.Timezone != nil && schedulerTaskSpec.Cron == nil {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("timezone"), "can't be set when cron is not specified"))
+	}
+
```
tnozicka, line 255: please add timezone and cron format validation

rzetelskik: @tnozicka Scylla Manager accepts the `timezone` parameter, which, at least according to sctool docs, represents `Timezone of –cron and –window flag values`.

At the same time, it accepts `TZ` and `CRON_TZ` prefixes of `cron` parameter (see https://github.com/robfig/cron/blob/e843a09e5b2db454d77aad25b1660173445fb2fc/parser.go#L95).

The combination of two is accepted. K8s has a similar issue with CronJobs, and it doesn't accept TZ/CRON_TZ prefixes: see https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#unsupported-timezone-specification.

Should we handle it the same way k8s does? Should I raise this with the manager team?

tnozicka: I think so + stricter is always better start with the validation

rzetelskik: ok, done

rzetelskik:
@tnozicka I added cron and timezone validation. TZ and CRON_TZ are not accepted as part of cron for now. I'll raise an issue with the manager team.

rzetelskik:
manager flake and upgrade flake - can't possibly be deteriorated by this PR
https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2066105109
https://github.com/scylladb/scylla-operator/issues/1233#issuecomment-2066106629

/test images
/retest

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -221,6 +229,53 @@ func ValidateRepairTaskSpec(repairTaskSpec *scyllav1.RepairTaskSpec, fldPath *fi
...
+	if schedulerTaskSpec.Timezone != nil {
+		_, err := time.LoadLocation(*schedulerTaskSpec.Timezone)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("timezone"), schedulerTaskSpec.Timezone, err.Error()))
+		}
+	}
+
+	// We can only validate interval when cron is non-nil for backwards compatibility.
+	if schedulerTaskSpec.Interval != nil && schedulerTaskSpec.Cron != nil {
+		if intervalDuration, err := duration.ParseDuration(*schedulerTaskSpec.Interval); err != nil {
```
tnozicka, line 268: nit: please split the assignment to make it shorter

pkg/api/scylla/validation/cluster_validation_test.go
```diff
@@ -87,6 +87,425 @@ func TestValidateScyllaCluster(t *testing.T) {
...
+					},
+					Intensity: "1",
+				})
+				return cluster
+			}(),
+			expectedErrorList:   field.ErrorList{},
+			expectedErrorString: "",
+		},
+		{
+			name: "invalid timezone in manager backup task spec",
```
tnozicka, line 268: it's unclear whether the test should produce and error or not from it's name (in other places as well), especially given it says `invalid` and produces no error

tnozicka (commented):
thanks, the api and overall logic looks sound to me, had some technical comments

pkg/api/scylla/validation/cluster_validation_test.go
rzetelskik: my bad, it shouldn't say invalid here

rzetelskik: done

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -34,6 +38,8 @@ var (
 		scyllav1.BroadcastAddressTypeServiceClusterIP,
 		scyllav1.BroadcastAddressTypeServiceLoadBalancerIngress,
 	}
+
+	schedulerTaskSpecCronParser = cron.NewParser(cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor)
```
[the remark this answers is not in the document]
rzetelskik: done

rzetelskik:
> 
> ci/prow/e2e-gke-release-script-latest	[415c60b](https://github.com/scylladb/scylla-operator/commit/415c60b1b0dcf5fe7a85ec399e8ce47c4fd79bf2)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1851/pull-scylla-operator-master-e2e-gke-release-script-latest/1773008427048505344)	true	`/test e2e-gke-release-script-latest`
> ci/prow/e2e-gke-parallel-clusterip	[7ea90ba](https://github.com/scylladb/scylla-operator/commit/7ea90bae594da3591cffac1567476f6493c50ff2)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1851/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1781267441016377344)	true	`/test e2e-gke-parallel-clusterip`
> ci/prow/e2e-gke-parallel	[7ea90ba](https://github.com/scylladb/scylla-operator/commit/7ea90bae594da3591cffac1567476f6493c50ff2)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1851/pull-scylla-operator-master-e2e-gke-parallel/1781267440487895040)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1851). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

manager flake
/retest

rzetelskik:
same thing again
/retest

rzetelskik:
and again
/retest

tnozicka (approved):
thanks for the updates

/approve
high level lgtm but I only scanned the test
/assign @zimnx

rzetelskik:
aaand again

/test images
/retest

pkg/api/scylla/validation/cluster_validation.go
rzetelskik: Xref https://github.com/scylladb/scylla-manager/issues/3818

zimnx (approved):
thanks
/lgtm

PR #1855, opened 2024-03-20
Collect related objects for ScyllaClusters and ScyllaDBMonitorings in must-gather

### LINKED ISSUE #1568: Teach must-gather about related resources for scyllaclusters

It should collect all resources related to scyllacluster in the namespace. Probably collecting the namespace it self is good enough as the endpoints another objects may interfere without explicit ownerrefs.

observability

noticed in https://github.com/scylladb/scylla-operator/issues/1566#issuecomment-1810427688

tnozicka:
This PR adds support for collecting related resources for ScyllaClusters and ScyllaDBMonitorings. Because not all resources are connected through ownerReferences we dump the same namespace and don't filter the objects. Some examples of where ownerRefs wouldn't works are `Services`, `Endpoints` or other objects that don't have a direct relationship but affect say firewall rules (like NetworkPolicy). In the end, dumping all the objects in the same namespace seems like the best path.

Resolves #1568

### DIFF
--- a/pkg/gather/collect/collect.go
+++ b/pkg/gather/collect/collect.go
@@ -8,13 +8,14 @@ import (
 	"os"
 	"path/filepath"
 
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"github.com/scylladb/scylla-operator/pkg/kubeinterfaces"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/pointer"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


tnozicka:
https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2034003561
/retest

pkg/gather/collect/collect.go
```diff
@@ -446,6 +447,59 @@ func (c *Collector) collectNamespace(ctx context.Context, u *unstructured.Unstru
...
+	err := c.writeResource(ctx, u, resourceInfo)
+	if err != nil {
+		return fmt.Errorf("can't write resource %q: %w", resourceInfo, err)
+	}
+
+	if !c.relatedResources {
+		return nil
+	}
+
+	err = c.collectNamespaceForObject(ctx, u, resourceInfo)
```
rzetelskik, line 477: given the func is called `collectScyllaCluster`, I'd move conditional collection of related resources to the switch case

pkg/gather/collect/collect.go
```diff
@@ -446,6 +447,59 @@ func (c *Collector) collectNamespace(ctx context.Context, u *unstructured.Unstru
...
```
rzetelskik, line 495: ditto

rzetelskik (commented):
two comments, rest lgtm

pkg/gather/collect/collect.go
```diff
@@ -472,10 +532,6 @@ func (c *Collector) CollectObject(ctx context.Context, u *unstructured.Unstructu
 func (c *Collector) CollectResource(ctx context.Context, resourceInfo *ResourceInfo, namespace, name string) error {
 	obj, err := c.dynamicClient.Resource(resourceInfo.Resource).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
 	if err != nil {
-		if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
tnozicka: Yeah, essentially both options suck. This can either silently skip wrongly specified resource or have issue collecting clusters that have creation/deletion hot-loops on some resources. Putting that aside, in a working cluster the issue is only with ephemeral resources that may force you to run this again.

Let me see if I can come up with something smarter.

pkg/gather/collect/collect.go
tnozicka: It's not a free function but member of a struct that has a field called `relatedResources`, so the member functions are supposed to account for them. Also in line with existing https://github.com/scylladb/scylla-operator/blob/5c354cea3a521661116f9d5f80a1bd62a5724cf4/pkg/gather/collect/collect.go#L403-L413

pkg/gather/collect/collect.go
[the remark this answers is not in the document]
tnozicka: I've recalled. This is inside the collector, so this should get an error and the caller can decide. Some tools can translate it into a warning, some fail, and test should fail with it. See
https://github.com/scylladb/scylla-operator/blob/a6cb4e692855e2073212faa097917d0cef8b0c64/pkg/cmd/operator/mustgather.go#L315-L322

pkg/gather/collect/collect.go
rzetelskik: ok

rzetelskik (approved):
lgtm
/assign zimnx

zimnx (approved):
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1900
/retest

tnozicka:
master moved / images issue
/test images

tnozicka:
same manager flake
/retest

tnozicka:
and once again https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2066302006
/retest

Issue #1802, opened 2024-03-07
[Flake] ScyllaCluster [It] should rotate TLS certificates before they expire


tnozicka:
### Link to the job that flaked.

[should rotate TLS certificates before they expire](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1795/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1765714974254043136#1:test-build-log.txt%3A863)

### Snippet of what failed.

   << Timeline
  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc0004fcbe0>: 
      can't create keyspace: gocql: no response received from cassandra within timeout period
      {
          msg: "can't create keyspace: gocql: no response received from cassandra within timeout period",
          err: <*errors.errorString | 0x391a300>{
              s: "gocql: no response received from cassandra within timeout period",
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/verify.go:312 @ 03/07/24 12:46:44.1
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.insertAndVerifyCQLDataUsingDataInserter({0x2643940, 0xc0005d82a0}, 0x399b540?)
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/verify.go:312 +0x78
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.insertAndVerifyCQLData({0x2643940, 0xc0005d82a0}, {0xc000976430, 0x1, 0x1}, {0x0?, 0x7fa3a270bec0?, 0xc0005927e0?})
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/verify.go:297 +0x174
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.glob..func17.2(0xc000579350)
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_upgrades.go:74 +0xab9
    reflect.Value.call({0x1da61c0?, 0xc000426f90?, 0x13?}, {0x22c9d6b, 0x4}, {0xc0005f6300, 0x1, 0x1?})
    	reflect/value.go:596 +0xce7
[...]

PR #1857, opened 2024-03-20
Validate whether Scylla container is tuned before Pod is unblocked

### LINKED ISSUE #1856: Scylla Pod is unblocked even when tuning fails
Operator Sidecar waits until all matching NodeConfig's are done with tuning, but it only validates whether Node where given Pod is running is done. It doesn't wait until Scylla container is tuned.

This causes that E2E validating our node tuning gives us a false positive, as even when tuning fails, Scylla Pod is started.

zimnx:
Scylla Pod was unblocked even when container wasn't yet tuned. E2E validating node optimizations wasn't validating it either giving us a false positive.

Fixes #1856

### DIFF
--- a/pkg/controllerhelpers/scylla_test.go
+++ b/pkg/controllerhelpers/scylla_test.go
@@ -777,3 +777,102 @@ func TestGetRequiredScyllaHosts(t *testing.T) {
 		})
 	}
 }
+
+func TestIsNodeTunedForContainer(t *testing.T) {
+	t.Parallel()
+
+	tt := []struct {
+		name          string
+		nodeConfig    *scyllav1alpha1.NodeConfig
+		nodeName      string
+		containerID   string
+		expectedTuned bool
+	}{
+		{
+			name:          "empty nodeConfig status isn't considered to be tuned",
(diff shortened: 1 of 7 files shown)


pkg/controllerhelpers/scylla_test.go
```diff
@@ -781,3 +781,103 @@ func TestGetRequiredScyllaHosts(t *testing.T) {
...
+				},
+			},
+			nodeName:      "node1",
+			containerID:   "container-id",
+			expectedTuned: true,
+		},
+	}
+
+	for i := range tt {
+		tc := tt[i]
```
tnozicka, line 873: 1.22+ is now enforce so this is no longer needed

tnozicka (approved):
/approve
lgtm, modulo one nit

pkg/controllerhelpers/scylla_test.go
zimnx: done

tnozicka:
/lgtm
thanks

zimnx:
Provisioning failed + https://github.com/scylladb/scylla-operator/issues/1802#issuecomment-2025678115

/retest

PR #1873, opened 2024-04-02
Extract Operator binary path to variable

zimnx:
Operator binary path is hardcoded making it impossible to reuse controllers copying operator binary, as binary executing it may be different.
To fix this, binary path is extracted as controller parameter and kept as global variable allowing forks to specify different one, while still allowing for reusing entire operator command.

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -38,6 +38,11 @@ const (
 	cryptoKeyBufferSizeMaxFlagKey = "crypto-key-buffer-size-max"
 )
 
+var (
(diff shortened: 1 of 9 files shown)


tnozicka (commented):
I don't think this should be a adjustable with a flag. Both pieces are controlled from the same binary a global variable that can be changed in forks would seem to cover the cases I can think of.

zimnx:
> I don't think this should be a adjustable with a flag. Both pieces are controlled from the same binary a global variable that can be changed in forks would seem to cover the cases I can think of.

Global variables are an anti pattern I tried to avoid. Changed to it anyways.

PR #1881, opened 2024-04-04
Support multi-datacenter clusters in test runner and script

rzetelskik:
**Description of your changes:** This PR adds support for multi-datacenter clusters in our test runner and scripts.

Prerequisite for https://github.com/scylladb/scylla-operator/pull/1632.

/kind feature
/priority important-longterm
/cc

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -8,8 +8,23 @@ shopt -s inherit_errexit
 
 source "$( dirname "${BASH_SOURCE[0]}" )/../../lib/kube.sh"
 
+if [ -z "${KUBECONFIG_DIR+x}" ]; then
+  KUBECONFIGS=("${KUBECONFIG}")
+else
+  KUBECONFIGS=()
+  for f in $( find "$( realpath "${KUBECONFIG_DIR}" )" -maxdepth 1 -type f -name '*.kubeconfig' ); do
+    KUBECONFIGS+=("${f}")
+  done
+fi
+
 # gather-artifacts is a self sufficient function that collects artifacts without depending on any external objects.
+# $1- target directory
 function gather-artifacts {
+  if [ -z "${1+x}" ]; then
+    echo -e "Missing target directory.\nUsage: ${FUNCNAME[0]} target_directory" > /dev/stderr
+    exit 2
+  fi
+
   if [ -z "${SO_IMAGE+x}" ]; then
     echo "SO_IMAGE can't be empty" > /dev/stderr
     exit 2
@@ -57,8 +72,8 @@ EOF
 
   exit_code="$( wait-for-container-exit-with-logs gather-artifacts must-gather must-gather )"
 
-  kubectl -n=gather-artifacts cp --retries=42 -c=wait-for-artifacts must-gather:/tmp/artifacts "${ARTIFACTS}/must-gather"
-  ls -l "${ARTIFACTS}/must-gather"
+  kubectl -n=gather-artifacts cp --retries=42 -c=wait-for-artifacts must-gather:/tmp/artifacts "${1}"
+  ls -l "${1}"
 
   kubectl -n=gather-artifacts delete pod/must-gather --wait=false
 
@@ -68,8 +83,23 @@ EOF
   fi
 }
 
-function gather-artifact-on-exit {
-  gather-artifacts || "Error gathering artifacts" > /dev/stderr
+function gather-artifacts-on-exit {
+  for i in "${!KUBECONFIGS[@]}"; do
+    KUBECONFIG="${KUBECONFIGS[$i]}" gather-artifacts "${ARTIFACTS}/must-gather/${i}" &
+    gather_artifacts_bg_pids["${i}"]=$!
+  done
+
+  for pid in "${gather_artifacts_bg_pids[@]}"; do
+    wait "${pid}"
+  done
+}
+
+function unset-default-storageclass {
+  for f in "${KUBECONFIGS[@]}"; do
+    for r in $( KUBECONFIG="${f}" kubectl get storageclasses -o name ); do
+      KUBECONFIG="${f}" kubectl patch "${r}" -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
+    done
+  done
 }
 
 function apply-e2e-workarounds {
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -15,8 +15,7 @@ source "$( dirname "${BASH_SOURCE[0]}" )/../lib/kube.sh"
 source "$( dirname "${BASH_SOURCE[0]}" )/lib/e2e.sh"
 parent_dir="$( dirname "${BASH_SOURCE[0]}" )"
 
-trap gather-artifact-on-exit EXIT
-
+trap gather-artifacts-on-exit EXIT
 
 SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=./hack/.ci/manifests/cluster/nodeconfig.yaml}"
 export SO_NODECONFIG_PATH
@@ -29,13 +28,20 @@ if [[ "${SO_DISABLE_NODECONFIG:-false}" == "true" ]]; then
   SO_CSI_DRIVER_PATH=""
 else
   # Make sure there is no default storage class before we create our own.
-  for r in $( kubectl get storageclasses -o name ); do kubectl patch "${r}" -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'; done
+  unset-default-storageclass
 fi
 
 SCYLLA_OPERATOR_FEATURE_GATES="${SCYLLA_OPERATOR_FEATURE_GATES:-AllAlpha=true,AllBeta=true}"
 export SCYLLA_OPERATOR_FEATURE_GATES
 
-timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}"
+for i in "${!KUBECONFIGS[@]}"; do
+  KUBECONFIG="${KUBECONFIGS[$i]}" timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}" &
+  ci_deploy_bg_pids["${i}"]=$!
+done
 
-apply-e2e-workarounds
-run-e2e
+for pid in "${ci_deploy_bg_pids[@]}"; do
+  wait "${pid}"
+done
+
+KUBECONFIG="${KUBECONFIGS[0]}" apply-e2e-workarounds
+KUBECONFIG="${KUBECONFIGS[0]}" run-e2e
--- a/pkg/genericclioptions/genericclioptions.go
+++ b/pkg/genericclioptions/genericclioptions.go
@@ -9,8 +9,13 @@ import (
 
 	"github.com/scylladb/scylla-operator/pkg/version"
 	"github.com/spf13/cobra"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	defaultKubeconfig = ""
 )
 
 // IOStreams is a structure containing all standard streams.
@@ -23,13 +28,113 @@ type IOStreams struct {
 	ErrOut io.Writer
 }
 
-type ClientConfig struct {
-	Kubeconfig    string
+type ClientConfigBase struct {
 	QPS           float32
 	Burst         int
 	UserAgentName string
-	RestConfig    *restclient.Config
-	ProtoConfig   *restclient.Config
+}
+
+func NewDefaultClientConfigBase(userAgentName string) ClientConfigBase {
+	return ClientConfigBase{
+		QPS:           50,
+		Burst:         75,
+		UserAgentName: userAgentName,
+	}
+}
+
+func (ccb *ClientConfigBase) AddFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().Float32VarP(&ccb.QPS, "qps", "", ccb.QPS, "Maximum allowed number of queries per second.")
+	cmd.PersistentFlags().IntVarP(&ccb.Burst, "burst", "", ccb.Burst, "Allows extra queries to accumulate when a client is exceeding its rate.")
+}
+
+func (ccb *ClientConfigBase) Validate() error {
+	return nil
+}
+
+func (ccb *ClientConfigBase) Complete() error {
+	return nil
+}
+
+type ClientConfigSet struct {
+	ClientConfigBase
+	kubeconfigs   []string
+	ClientConfigs []ClientConfig
+}
+
+func NewClientConfigSet(userAgentName string) ClientConfigSet {
+	return ClientConfigSet{
+		ClientConfigBase: NewDefaultClientConfigBase(userAgentName),
+		kubeconfigs:      []string{defaultKubeconfig},
+	}
+}
+
+func (ccs *ClientConfigSet) AddFlags(cmd *cobra.Command) {
+	ccs.ClientConfigBase.AddFlags(cmd)
+
+	cmd.PersistentFlags().StringArrayVarP(&ccs.kubeconfigs, "kubeconfig", "", ccs.kubeconfigs, "Path to kubeconfig file(s).")
+}
+
+func (ccs *ClientConfigSet) Validate() error {
+	var errs []error
+	var err error
+
+	err = ccs.ClientConfigBase.Validate()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("invalid client config base: %w", err))
+	}
+
+	if len(ccs.kubeconfigs) == 0 {
+		errs = append(errs, fmt.Errorf("at least one kubeconfig must be provided"))
+	}
+
+	for _, kubeconfig := range ccs.kubeconfigs {
+		cc := NewClientConfig(ccs.UserAgentName)
+		cc.Kubeconfig = kubeconfig
+		cc.QPS = ccs.QPS
+		cc.Burst = ccs.Burst
+
+		err = cc.Validate()
+		if err != nil {
+			errs = append(errs, fmt.Errorf("invalid client config for kubeconfig %q: %w", kubeconfig, err))
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (ccs *ClientConfigSet) Complete() error {
+	var err error
+
+	err = ccs.ClientConfigBase.Complete()
+	if err != nil {
+		return fmt.Errorf("can't complete client config base: %w", err)
+	}
+
+	clientConfigs := make([]ClientConfig, 0, len(ccs.kubeconfigs))
+	for _, kubeconfig := range ccs.kubeconfigs {
+		cc := NewClientConfig(ccs.UserAgentName)
+		cc.Kubeconfig = kubeconfig
+		cc.QPS = ccs.QPS
+		cc.Burst = ccs.Burst
+
+		err = cc.Complete()
+		if err != nil {
+			return fmt.Errorf("can't complete client config for kubeconfig %q: %w", kubeconfig, err)
+		}
+
+		clientConfigs = append(clientConfigs, cc)
+	}
+
+	ccs.ClientConfigs = clientConfigs
+
+	return nil
+}
+
+type ClientConfig struct {
+	ClientConfigBase
+	Kubeconfig  string
+	RestConfig  *restclient.Config
+	ProtoConfig *restclient.Config
 }
 
 func MakeVersionedUserAgent(baseName string) string {
@@ -45,27 +150,38 @@ func MakeVersionedUserAgent(baseName string) string {
 
 func NewClientConfig(userAgentName string) ClientConfig {
 	return ClientConfig{
-		Kubeconfig:    "",
-		QPS:           50,
-		Burst:         75,
-		UserAgentName: userAgentName,
-		RestConfig:    nil,
-		ProtoConfig:   nil,
+		ClientConfigBase: NewDefaultClientConfigBase(userAgentName),
+		Kubeconfig:       defaultKubeconfig,
+		RestConfig:       nil,
+		ProtoConfig:      nil,
 	}
 }
 
 func (cc *ClientConfig) AddFlags(cmd *cobra.Command) {
+	cc.ClientConfigBase.AddFlags(cmd)
+
 	cmd.PersistentFlags().StringVarP(&cc.Kubeconfig, "kubeconfig", "", cc.Kubeconfig, "Path to the kubeconfig file.")
-	cmd.PersistentFlags().Float32VarP(&cc.QPS, "qps", "", cc.QPS, "Maximum allowed number of queries per second.")
-	cmd.PersistentFlags().IntVarP(&cc.Burst, "burst", "", cc.Burst, "Allows extra queries to accumulate when a client is exceeding its rate.")
 }
 
 func (cc *ClientConfig) Validate() error {
-	return nil
+	var errs []error
+	var err error
+
+	err = cc.ClientConfigBase.Validate()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("invalid client config base: %w", err))
+	}
+
+	return utilerrors.NewAggregate(errs)
 }
 
 func (cc *ClientConfig) Complete() error {
 	var err error
+
+	err = cc.ClientConfigBase.Complete()
+	if err != nil {
+		return fmt.Errorf("can't complete client config base: %w", err)
+	}
 
 	loader := clientcmd.NewDefaultClientConfigLoadingRules()
 	// Use explicit kubeconfig if set.
--- a/test/e2e/framework/cluster.go
+++ b/test/e2e/framework/cluster.go
@@ -1,0 +1,156 @@
+// Copyright (C) 2024 ScyllaDB
+
+package framework
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+)
+
+type ClusterInterface interface {
+	AdminClientInterface
+	DefaultNamespaceIfAny() (*corev1.Namespace, Client, bool)
+	CreateUserNamespace(ctx context.Context) (*corev1.Namespace, Client)
+}
+
+type createNamespaceFunc func(ctx context.Context, adminClient kubernetes.Interface, adminClientConfig *restclient.Config) (*corev1.Namespace, Client)
+
+type Cluster struct {
+	AdminClient
+
+	name string
+
+	createNamespace    createNamespaceFunc
+	defaultNamespace   *corev1.Namespace
+	defaultClient      Client
+	namespacesToDelete []*corev1.Namespace
+}
+
+var _ AdminClientInterface = &Cluster{}
+var _ ClusterInterface = &Cluster{}
+
+func NewCluster(name string, restConfig *restclient.Config, createNamespace createNamespaceFunc) *Cluster {
+	adminClientConfig := restclient.CopyConfig(restConfig)
+
+	return &Cluster{
+		AdminClient: AdminClient{
+			Config: adminClientConfig,
+		},
+
+		name: name,
+
+		createNamespace:  createNamespace,
+		defaultNamespace: nil,
+		defaultClient: Client{
+			Config: nil,
+		},
+		namespacesToDelete: []*corev1.Namespace{},
+	}
+}
+
+func (c *Cluster) DefaultNamespaceIfAny() (*corev1.Namespace, Client, bool) {
+	if c.defaultNamespace == nil {
+		return nil, Client{}, false
+	}
+
+	return c.defaultNamespace, c.defaultClient, true
+}
+
+func (c *Cluster) CreateUserNamespace(ctx context.Context) (*corev1.Namespace, Client) {
+	ns, nsClient := c.createNamespace(ctx, c.KubeAdminClient(), c.AdminClientConfig())
+
+	c.namespacesToDelete = append(c.namespacesToDelete, ns)
+
+	return ns, nsClient
+}
+
+func (c *Cluster) Cleanup(ctx context.Context) {
+	for _, ns := range c.namespacesToDelete {
+		collectAndDeleteNamespace(ctx, c.KubeAdminClient(), c.DynamicAdminClient(), ns)
+	}
+
+	c.namespacesToDelete = []*corev1.Namespace{}
+}
+
+func collectAndDeleteNamespace(ctx context.Context, adminClient kubernetes.Interface, dynamicAdminClient dynamic.Interface, ns *corev1.Namespace) {
+	defer func() {
+		keepNamespace := false
+		switch TestContext.DeleteTestingNSPolicy {
+		case DeleteTestingNSPolicyNever:
+			keepNamespace = true
+		case DeleteTestingNSPolicyOnSuccess:
+			if g.CurrentSpecReport().Failed() {
+				keepNamespace = true
+			}
+		case DeleteTestingNSPolicyAlways:
+		default:
+		}
+
+		if keepNamespace {
+			By("Keeping namespace %q for debugging", ns.Name)
+			return
+		}
+
+		deleteNamespace(ctx, adminClient, dynamicAdminClient, ns)
+
+	}()
+
+	// Print events if the test failed.
+	if g.CurrentSpecReport().Failed() {
+		By(fmt.Sprintf("Collecting events from namespace %q.", ns.Name))
+		DumpEventsInNamespace(ctx, adminClient, ns.Name)
+	}
+
+	// CI can't keep namespaces alive because it could get out of resources for the other tests
+	// so we need to collect the namespaced dump before destroying the namespace.
+	// Collecting artifacts even for successful runs helps to verify if it went
+	// as expected and the amount of data is bearable.
+	if len(TestContext.ArtifactsDir) != 0 {
+		By(fmt.Sprintf("Collecting dumps from namespace %q.", ns.Name))
+
+		d := path.Join(TestContext.ArtifactsDir, "e2e")
+		err := os.Mkdir(d, 0777)
+		if err != nil && !os.IsExist(err) {
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}
+
+		err = DumpNamespace(ctx, cacheddiscovery.NewMemCacheClient(adminClient.Discovery()), dynamicAdminClient, adminClient.CoreV1(), d, ns.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+	}
+}
+
+func deleteNamespace(ctx context.Context, adminClient kubernetes.Interface, dynamicAdminClient dynamic.Interface, ns *corev1.Namespace) {
+	By("Destroying namespace %q.", ns.Name)
+	var gracePeriod int64 = 0
+	var propagation = metav1.DeletePropagationForeground
+	err := adminClient.CoreV1().Namespaces().Delete(
+		ctx,
+		ns.Name,
+		metav1.DeleteOptions{
+			GracePeriodSeconds: &gracePeriod,
+			PropagationPolicy:  &propagation,
+			Preconditions: &metav1.Preconditions{
+				UID: &ns.UID,
+			},
+		},
+	)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	// We have deleted only the namespace object but it is still there with deletionTimestamp set.
+	By("Waiting for namespace %q to be removed.", ns.Name)
+	err = WaitForObjectDeletion(ctx, dynamicAdminClient, corev1.SchemeGroupVersion.WithResource("namespaces"), "", ns.Name, &ns.UID)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	klog.InfoS("Namespace removed.", "Namespace", ns.Name)
+}
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -4,18 +4,13 @@ package framework
 
 import (
 	"context"
-	"crypto/sha512"
-	"encoding/base64"
 	"fmt"
-	"os"
-	"path"
 	"strconv"
 	"time"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
-	scyllaclientset "github.com/scylladb/scylla-operator/pkg/client/scylla/clientset/versioned"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	scyllafixture "github.com/scylladb/scylla-operator/test/e2e/fixture/scylla"
 	corev1 "k8s.io/api/core/v1"
@@ -24,12 +19,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/storage/names"
-	"k8s.io/client-go/discovery"
-	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
-	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
-	"k8s.io/klog/v2"
 )
 
 const (
@@ -40,27 +31,37 @@ const (
 )
 
 type Framework struct {
-	name      string
-	namespace *corev1.Namespace
+	FullClient
 
-	adminClientConfig *restclient.Config
-	clientConfig      *restclient.Config
-	username          string
+	name string
+
+	clusters []*Cluster
 }
 
-func NewFramework(name string) *Framework {
-	uniqueName := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", name))
+var _ FullClientInterface = &Framework{}
+var _ ClusterInterface = &Framework{}
 
-	adminClientConfig := restclient.CopyConfig(TestContext.RestConfig)
-	adminClientConfig.UserAgent = "scylla-operator-e2e"
-	adminClientConfig.QPS = 20
-	adminClientConfig.Burst = 50
-
+func NewFramework(namePrefix string) *Framework {
 	f := &Framework{
-		name:              uniqueName,
-		username:          "admin",
-		adminClientConfig: adminClientConfig,
+		name:       names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", namePrefix)),
+		FullClient: FullClient{},
 	}
+
+	clusters := make([]*Cluster, 0, len(TestContext.RestConfigs))
+	for i, restConfig := range TestContext.RestConfigs {
+		clusterName := fmt.Sprintf("%s-%d", f.name, i)
+		c := NewCluster(
+			clusterName,
+			restConfig,
+			func(ctx context.Context, adminClient kubernetes.Interface, adminClientConfig *restclient.Config) (*corev1.Namespace, Client) {
+				return CreateUserNamespace(ctx, clusterName, f.CommonLabels(), adminClient, adminClientConfig)
+			},
+		)
+		clusters = append(clusters, c)
+	}
+	f.clusters = clusters
+
+	f.FullClient.AdminClient.Config = f.defaultCluster().AdminClientConfig()
 
 	g.BeforeEach(f.beforeEach)
 	g.AfterEach(f.afterEach)
@@ -68,12 +69,17 @@ func NewFramework(name string) *Framework {
 	return f
 }
 
-func (f *Framework) Namespace() string {
-	return f.namespace.Name
+func (f *Framework) Cluster(idx int) ClusterInterface {
+	o.Expect(idx).NotTo(o.BeNumerically(">=", len(f.clusters)))
+	return f.clusters[idx]
 }
 
-func (f *Framework) Username() string {
-	return f.username
+func (f *Framework) Namespace() string {
+	return f.defaultCluster().defaultNamespace.Name
+}
+
+func (f *Framework) DefaultNamespaceIfAny() (*corev1.Namespace, Client, bool) {
+	return f.defaultCluster().DefaultNamespaceIfAny()
 }
 
 func (f *Framework) GetIngressAddress(hostname string) string {
@@ -85,58 +91,7 @@ func (f *Framework) GetIngressAddress(hostname string) string {
 }
 
 func (f *Framework) FieldManager() string {
-	h := sha512.Sum512([]byte(fmt.Sprintf("scylla-operator-e2e-%s", f.Namespace())))
-	return base64.StdEncoding.EncodeToString(h[:])
-}
-
-func (f *Framework) ClientConfig() *restclient.Config {
-	return f.clientConfig
-}
-
-func (f *Framework) AdminClientConfig() *restclient.Config {
-	return f.adminClientConfig
-}
-
-func (f *Framework) DiscoveryClient() *discovery.DiscoveryClient {
-	client, err := discovery.NewDiscoveryClientForConfig(f.ClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
-}
-
-func (f *Framework) DynamicClient() dynamic.Interface {
-	client, err := dynamic.NewForConfig(f.ClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
-}
-
-func (f *Framework) DynamicAdminClient() dynamic.Interface {
-	client, err := dynamic.NewForConfig(f.AdminClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
-}
-
-func (f *Framework) KubeClient() *kubernetes.Clientset {
-	client, err := kubernetes.NewForConfig(f.ClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
-}
-
-func (f *Framework) KubeAdminClient() *kubernetes.Clientset {
-	client, err := kubernetes.NewForConfig(f.AdminClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
-}
-
-func (f *Framework) ScyllaClient() *scyllaclientset.Clientset {
-	client, err := scyllaclientset.NewForConfig(f.ClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
-}
-
-func (f *Framework) ScyllaAdminClient() *scyllaclientset.Clientset {
-	client, err := scyllaclientset.NewForConfig(f.AdminClientConfig())
-	o.Expect(err).NotTo(o.HaveOccurred())
-	return client
+	return FieldManager(f.ClientConfig().UserAgent, f.Namespace())
 }
 
 func (f *Framework) CommonLabels() map[string]string {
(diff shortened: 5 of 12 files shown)


rzetelskik:
https://github.com/scylladb/scylla-operator/issues/1525#issuecomment-2049733594

https://github.com/scylladb/scylla-operator/issues/1694#issuecomment-2049735439

I don't see how these changes could contribute to flakiness, since its just some changes in machinery, so I'm not investigating.

/test images
/retest

rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[9ff84af](https://github.com/scylladb/scylla-operator/commit/9ff84af2560fded69a3aad45043ca879029240a8)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-parallel/1778420481334972416)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1881). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

Cluster provisioning failed.
/test images
/retest

rzetelskik:
> 
> ci/prow/e2e-gke-serial	[61326f2](https://github.com/scylladb/scylla-operator/commit/61326f2d61271d2cec9a6d2cc258f2f24d07c40b)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-serial/1779463324602208256)	true	`/test e2e-gke-serial`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1881). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

Cluster provisioning failed.
/retest

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[d3b46ef](https://github.com/scylladb/scylla-operator/commit/d3b46ef3e09886880ae98d156a08ba9f7b97b431)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1780872762919751680)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1881). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

Known Scylla Manager  flake. Already working on it, this PR is unrelated.

/retest

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+      kubectl_create -f="${SO_NODECONFIG_PATH}"
+      kubectl_create -n=local-csi-driver -f=./hack/.ci/manifests/namespaces/local-csi-driver/
+    fi
+
+    kubectl create namespace e2e --dry-run=client -o=yaml | kubectl_create -f=-
+    kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o=yaml | kubectl_create -f=-
+  )
+}
+
+for i in ${!kubeconfigs[@]}; do
```
zimnx, line 195:
```suggestion
for i in "${!kubeconfigs[@]}"; do
```

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+  setup "${kubeconfigs[$i]}" &
+  setup_bg_pids["${i}"]=$!
+done
+
+setup_failed=0
+for pid in ${setup_bg_pids[*]}; do
```
zimnx, line 201:
```suggestion
for pid in "${setup_bg_pids[*]}"; do
```

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,109 @@ type IOStreams struct {
...
+}
+
+func (ccs *ClientConfigSet) AddFlags(cmd *cobra.Command) {
+	ccs.sharedClientConfig.AddFlags(cmd)
+
+	cmd.PersistentFlags().StringArrayVarP(&ccs.Kubeconfigs, "kubeconfig", "", ccs.Kubeconfigs, "Path to kubeconfig file(s).")
+}
+
+func (ccs *ClientConfigSet) Validate() error {
+	var err error
```
zimnx, line 78: This should be an array and each rule should append to it

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,109 @@ type IOStreams struct {
...
+	err = ccs.sharedClientConfig.Validate()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (ccs *ClientConfigSet) Complete() error {
+	var err error
```
zimnx, line 89: ditto

test/e2e/utils/exec.go
```diff
@@ -29,7 +29,7 @@ type ExecOptions struct {
 // returning stdout, stderr and error. `options` allowed for
 // additional parameters to be passed.
 func ExecWithOptions(client corev1client.CoreV1Interface, options ExecOptions) (string, string, error) {
-	config := framework.TestContext.RestConfig
+	config := framework.TestContext.RestConfigs[0]
```
zimnx, line 32: Why the first one? This limits executing on pods to first kube cluster. We should extract it to parameter.

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,109 @@ type IOStreams struct {
...
+}
+
+func (ccs *ClientConfigSet) Validate() error {
+	var err error
+
+	err = ccs.sharedClientConfig.Validate()
+	if err != nil {
+		return err
+	}
+
```
zimnx, line 84: We should validate there's at least one

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+      exit "${exit_code}"
+    fi
+
+    wait "${must_gather_bg_pid}"
+  )
 }
 
 function handle-exit {
-  gather-artifacts || "Error gathering artifacts" > /dev/stderr
+  for i in ${!kubeconfigs[@]}; do
```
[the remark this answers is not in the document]
rzetelskik: done

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+    gather-artifacts "${kubeconfigs[$i]}" &
+    gather_artifacts_bg_pids["${i}"]=$!
+  done
+
+  gather_artifacts_failed=0
+  for pid in ${gather_artifacts_bg_pids[*]}; do
```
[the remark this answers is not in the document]
rzetelskik: done

hack/.ci/run-e2e-gke.sh
```diff
@@ -256,15 +317,15 @@ kubectl -n=e2e wait --for=condition=Ready pod/e2e
   }
   trap unblock-e2e-pod EXIT
 
-  kubectl -n=e2e exec pod/e2e -c=wait-for-artifacts -- bash -euEo pipefail -O inherit_errexit -c "until [[ -f /tmp/done ]]; do sleep 1; done; ls -l /tmp/artifacts"
-  kubectl -n=e2e cp --retries=42 e2e:/tmp/artifacts -c=wait-for-artifacts "${ARTIFACTS}"
+  kubectl -n=e2e exec pod/e2e -c wait-for-artifacts -- bash -euEo pipefail -O inherit_errexit -c "until [[ -f /tmp/done ]]; do sleep 1; done; ls -l /tmp/artifacts"
```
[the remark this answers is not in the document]
rzetelskik: looks like rebasing went wrong, thanks

hack/.ci/run-e2e-gke.sh
rzetelskik: done

hack/.ci/run-e2e-gke.sh
```diff
@@ -240,12 +296,17 @@ spec:
       mountPath: /tmp/artifacts
     - name: gcs-service-account-credentials
       mountPath: /var/run/secrets/gcs-service-account-credentials
+    - name: kubeconfigs
+      mountPath: /tmp/kubeconfigs
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/utils/exec.go
rzetelskik: I wanted to avoid making any changes to existing tests, but ok, I guess we can pass a framework through parameters

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

pkg/genericclioptions/genericclioptions.go
rzetelskik: I agree for validation, but completion can be sequential - I'd leave it as is here

rzetelskik: moved the validation away from this function though, ptal

rzetelskik:
@zimnx thanks for the review, I replied to all of your comments

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
```diff
@@ -256,21 +257,21 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
...
+			stdout, stderr, err := executeInPod(f.AdminClientConfig(), f.KubeClient().CoreV1(), clientPod, "blkid", "--output=value", "--match-tag=TYPE", discoveredRaidDeviceOnHost)
 			g.Expect(err).NotTo(o.HaveOccurred(), stderr)
 
 			g.Expect(strings.TrimSpace(stdout)).To(o.Equal("xfs"))
 		}).WithPolling(1 * time.Second).WithTimeout(3 * time.Minute).Should(o.Succeed())
 
 		framework.By("Checking if RAID was mounted at the provided location with correct options")
 		o.Eventually(func(g o.Gomega) {
-			stdout, stderr, err := executeInPod(f.KubeClient().CoreV1(), clientPod, "mount")
+			stdout, stderr, err := executeInPod(f.AdminClientConfig(), f.KubeClient().CoreV1(), clientPod, "mount")
```
zimnx, line 274: Isn't f.ClientConfig enough? It should have all necessary access in the namespace where this is executed.

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
...
+	var errs []error
+	var err error
+
+	err = ccs.sharedClientConfig.Validate()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("invalid shared client config: %w", err))
+	}
+
+	if len(ccs.Kubeconfigs) == 0 {
+		errs = append(errs, fmt.Errorf("invalid kubeconfigs: at least one kubeconfig must be provided"))
```
zimnx, line 87: Parameter is called `kubeconfig`
```suggestion
		errs = append(errs, fmt.Errorf("invalid kubeconfig: at least one kubeconfig must be provided"))
```

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
rzetelskik: correct, done

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[66167d5](https://github.com/scylladb/scylla-operator/commit/66167d58b68af85924fe88a6596dbb3124d8520b)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-parallel/1784905794609745920)	true	`/test e2e-gke-parallel`
> ci/prow/e2e-gke-parallel-clusterip	[66167d5](https://github.com/scylladb/scylla-operator/commit/66167d58b68af85924fe88a6596dbb3124d8520b)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1784905795058536448)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1881). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

known manager flake for both, not investigating since it's unrelated and we're already working on fixing it
/retest

zimnx (approved):
thanks for updates, lgtm
/assign tnozicka

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+    fi
+
+    wait "${must_gather_bg_pid}"
+  )
 }
 
 function handle-exit {
-  gather-artifacts || "Error gathering artifacts" > /dev/stderr
+  for i in "${!kubeconfigs[@]}"; do
+    gather-artifacts "${kubeconfigs[$i]}" &
```
tnozicka, line 102:
```suggestion
  for kubeconfig in ${kubeconfigs[@]}; do
    KUBECONFIG="${kubeconfig}" gather-artifacts &
```

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
-    resources:
-    - pods/ephemeralcontainers
-    verbs:
-    - patch
-EOF
+# $1 - kubeconfig path
+function setup {
+  (
+    KUBECONFIG="${1}"
+    export KUBECONFIG
```
tnozicka, line 122: same here

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+      kubectl_create -n=local-csi-driver -f=./hack/.ci/manifests/namespaces/local-csi-driver/
+    fi
+
+    kubectl create namespace e2e --dry-run=client -o=yaml | kubectl_create -f=-
+    kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o=yaml | kubectl_create -f=-
+  )
+}
+
+for i in "${!kubeconfigs[@]}"; do
+  setup "${kubeconfigs[$i]}" &
```
tnozicka, line 196: and here

hack/.ci/run-e2e-gke.sh
```diff
@@ -174,6 +227,8 @@ ingress_controller_address="$( kubectl -n=haproxy-ingress get svc haproxy-ingres
 
 kubectl create -n=e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o=yaml | kubectl_create -f=-
 
+kubectl create -n=e2e configmap kubeconfigs $( IFS=' '; echo "${kubeconfigs[@]/#/--from-file=}" ) --dry-run=client -o yaml | kubectl_create -f -
```
tnozicka, line 230:
```suggestion
kubectl create -n=e2e configmap kubeconfigs "$( IFS=' '; echo "${kubeconfigs[@]/#/--from-file=}" )" --dry-run=client -o=yaml | kubectl_create -f=-
```

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
+
+if [ ${setup_failed} -eq 1 ]; then
+  echo "Error setting up clusters" > /dev/stderr
+  exit 2
 fi
 
-kubectl create namespace e2e --dry-run=client -o=yaml | kubectl_create -f=-
-kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o=yaml | kubectl_create -f=-
+KUBECONFIG="${kubeconfigs[0]}"
+export KUBECONFIG
```
tnozicka, line 212: why isn't the operator installed in the other DC as well?

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
 	ErrOut io.Writer
 }
 
-type ClientConfig struct {
-	Kubeconfig    string
+type sharedClientConfig struct {
```
tnozicka, line 31:
```suggestion
type ClientConfigBase struct {
```
I think this should be public so you can use it as an initializer in the struct that embeds it
(`shared` usually implies concurrent access)

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
...
 	QPS           float32
 	Burst         int
 	UserAgentName string
-	RestConfig    *restclient.Config
-	ProtoConfig   *restclient.Config
+}
+
+func newSharedClientConfig(userAgentName string) sharedClientConfig {
```
tnozicka, line 37:
```suggestion
func newDefaultClientConfigBase(userAgentName string) DefaultClientConfigBase {
```

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
...
+}
+
+func (scc *sharedClientConfig) Complete() error {
+	return nil
+}
+
+type ClientConfigSet struct {
+	sharedClientConfig
+	Kubeconfigs   []string
+	ClientConfigs []ClientConfig
```
tnozicka, line 61: ClientConfig  already contains a kubeconfig - this seems to be duplicit

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
...
+func (scc *sharedClientConfig) Validate() error {
+	return nil
+}
+
+func (scc *sharedClientConfig) Complete() error {
+	return nil
+}
+
+type ClientConfigSet struct {
+	sharedClientConfig
```
tnozicka, line 59: QPS (an other options) can differ for clusters, e.g. the first one can run extra operator loops with automated multidc

test/e2e/framework/framework.go
```diff
@@ -39,6 +39,19 @@ const (
...
 )
 
+type FrameworkSet []*Framework
+
+func NewFrameworkSet(generateName string) FrameworkSet {
+	fs := make(FrameworkSet, 0, len(TestContext.RestConfigs))
+
+	uniqueName := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", generateName))
+	for _, restConfig := range TestContext.RestConfigs {
+		fs = append(fs, NewFramework(uniqueName, WithRestConfig(restConfig)))
```
tnozicka, line 49: restConfig is a required arg, let's just put it as an arg to the function constructor so its type safe and nobody can forget to give it to the constructor

tnozicka (changes requested):

hack/.ci/run-e2e-gke.sh
```diff
@@ -36,6 +36,15 @@ if [ -z ${ARTIFACTS+x} ]; then
   exit 2
 fi
 
+if [ -z ${KUBECONFIG_DIR+x} ]; then
+  kubeconfigs=("${KUBECONFIG}")
+else
+  kubeconfigs=()
+  for f in $( find $(realpath "${KUBECONFIG_DIR}") -maxdepth 1 -type f ); do
```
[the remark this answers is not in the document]
rzetelskik: ok

hack/.ci/run-e2e-gke.sh
rzetelskik: done, although I still need the index

hack/.ci/run-e2e-gke.sh
```diff
@@ -53,109 +62,153 @@ function kubectl_create {
...
-  kubectl -n=e2e logs -f pod/must-gather
-  exit_code=$( kubectl -n=e2e get pods/must-gather --output='jsonpath={.status.containerStatuses[0].state.terminated.exitCode}' )
-  kubectl -n=e2e delete pod/must-gather --wait=false
-
-  if [[ "${exit_code}" != "0" ]]; then
-    echo "Collecting artifacts using must-gather failed"
-    exit "${exit_code}"
-  fi
+    KUBECONFIG="${1}"
+    export KUBECONFIG
```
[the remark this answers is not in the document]
rzetelskik: done

hack/.ci/run-e2e-gke.sh
rzetelskik: done

hack/.ci/run-e2e-gke.sh
rzetelskik: done

hack/.ci/run-e2e-gke.sh
rzetelskik: done

hack/.ci/run-e2e-gke.sh
rzetelskik: looks like the next couple of lines should be a part of `setup` func as well, thanks

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

pkg/genericclioptions/genericclioptions.go
rzetelskik: this is there to keep the flags as they are now - iirc it was a requirement that we don't need to change the existing CI jobs

pkg/genericclioptions/genericclioptions.go
rzetelskik: how do you make it configurable per cluster without changing the existing set of flags? Do you just allow for an array and assume the indexes are corresponding to kubeconfigs?

test/e2e/framework/framework.go
rzetelskik: the issue with this imo is that so far it was defaulted, so suddenly every existing test has to pass a rest config to the framework constructor, which not only requires changes to every single test, but sort of brakes the separation of concerns?

pkg/genericclioptions/genericclioptions.go
tnozicka: This struct isn't specific to test framework, I only ask for the struct to allow different settings at this point.

test/e2e/framework/framework.go
```diff
@@ -39,6 +39,19 @@ const (
 	serviceAccountTokenSecretWaitTimeout = 1 * time.Minute
 )
 
+type FrameworkSet []*Framework
```
tnozicka, line 42: Some of the methods on the framework, like logging or default resources are global and I don't see how forking it per cluster fits there. Maybe we should have just one global framework and use it like `f.Cluster(2).KubeClient()` and `f.KubeClient()` would be a wrapper to redirect it to cluster `0`.

pkg/genericclioptions/genericclioptions.go
rzetelskik: I see your point, but I think it does now? ClientConfigSet has a slice of ClientConfigs and each ClientConfig has configurable options (because it embeds ClientConfigBase) - so the ClientConfigBase allows to retain the flags as they are, but it could easily be removed from ClientConfigSet later on and configured per-client (or per-cluster)?

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

test/e2e/framework/framework.go
rzetelskik: I like it, let me try

test/e2e/framework/framework.go
rzetelskik: nevermind, works well with the cluster idea from the other comment

test/e2e/framework/framework.go
rzetelskik: @tnozicka please have a look if this makes sense, especially how the artifacts are structured: https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-parallel/1788683762628825088/artifacts/

tnozicka: roughly yes, I'll go into naming next week. I think it would be beneficial to make an interface for the "Cluster" and make sure the framework implements it fully (by 0 redirection).

artifacts wise, I'd expect cluster index to come before namespace name as it is more significant but a cluster level folder would be even nicer. You also have to run multiple must-gathers

rzetelskik: > I think it would be beneficial to make an interface for the "Cluster" and make sure the framework implements it fully (by 0 redirection) so the methods don't get out of sync.

good point, I'll add that

> artifacts wise, I'd expect cluster index to come before namespace name as it is more significant

that's a framework name + cluster idx now. I think the framework name is more important, since it corresponds to a test set?

>  but a cluster level folder would be even nicer

So <artifacts-dir>/e2e/<framework-name>/<cluster-name>/...?

tnozicka: Actually, thinking about this the second time - when you see a failure in e2e logs you are most likely want to search by the namespace name (statistically unique even across clusters and frameworks) so not nesting things would be easier for the main use case. 
(Instead of 2 jumps for searching the framework name + searching the namespace name in that result.)

`<artifacts-dir>/e2e/{namespaces,cluster-scoped}/<framework-name>-<cluster-name>-<resource-name>`

rzetelskik: > for the main use case.
> (Instead of 2 jumps for searching the framework name + searching the namespace name in that result.)
> 
> <artifacts-dir>/e2e/{namespaces,cluster-scoped}/<framework-name>-<cluster-name>-<resource-name>

wouldn't that make it quite difficult to find correlated namespaces in multi-datacenter runs? You'd have to run `find` I guess? Although you'd get the other namespaces from the logs, so maybe that's not necessary altogether.

Also I'm not a fan of the prefixes - seems like a lot of redundant information, since all resources in that namespaces are going to come from the same framework/cluster, aren't they? Maybe we could annotate/label the resources with these or just create metadata file(s) there?

tnozicka: > wouldn't that make it quite difficult to find correlated namespaces in multi-datacenter runs? 

won't they be sorted out for you by the common prefix?

> Also I'm not a fan of the prefixes - seems like a lot of redundant information

it's technically the same amount of information, you just change the `/` separator to `-`

> since all resources in that namespaces are going to come from the same framework/cluster, aren't they?

are you sure you've read the syntax right? there is a difference between `namespaces` and `<namespace>`

it's similar to what we do today (with cluster index added)

https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1901/pull-scylla-operator-master-e2e-gke-parallel/1788824474968133632/artifacts/e2e/namespaces/

rzetelskik: > won't they be sorted out for you by the common prefix?

discard my comment, over the weekend I managed to forget the namespaces were generated with framework/cluster names as prefix

> are you sure you've read the syntax right? there is a difference between namespaces and <namespace>

I first understood your syntax as such that every leaf resource would have this prefix. Nevermind - I updated it now, I suppose we're on the same page.

tnozicka: we are, thx

pkg/genericclioptions/genericclioptions.go
rzetelskik: @tnozicka ping

tnozicka: I think that technically works, but a similar point about flag-only options being private would apply here as well

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
...
+	var errs []error
+	var err error
+
+	err = ccs.ClientConfigBase.Validate()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("invalid client config base: %w", err))
+	}
+
+	if len(ccs.kubeconfigs) == 0 {
+		errs = append(errs, fmt.Errorf("kubeconfig(s) must be provided"))
```
tnozicka, line 87:
```suggestion
		errs = append(errs, fmt.Errorf("at least one kubeconfig must be provided"))
```

pkg/genericclioptions/genericclioptions.go
```diff
@@ -23,13 +28,120 @@ type IOStreams struct {
...
+	for _, kubeconfig := range ccs.kubeconfigs {
+		cc := NewClientConfig(ccs.UserAgentName)
+		cc.Kubeconfig = kubeconfig
+		cc.QPS = ccs.QPS
+		cc.Burst = ccs.Burst
+
+		err = cc.Complete()
+		if err != nil {
+			errs = append(errs, fmt.Errorf("can't complete client config for kubeconfig %q: %w", kubeconfig, err))
+			continue
```
tnozicka, line 124: complete conventionally exits on first error

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+func (c *Cluster) Namespace() string {
+	return c.namespace.Name
+}
+
+func (c *Cluster) Username() string {
+	return c.username
+}
+
+func (c *Cluster) FieldManager() string {
+	h := sha512.Sum512([]byte(fmt.Sprintf("scylla-operator-e2e-%s", c.Namespace())))
```
tnozicka, line 71: defaultUserAgent

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+	"k8s.io/client-go/discovery"
+	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+)
+
+const (
+	defaultUserAgent = "scylla-operator-e2e"
```
tnozicka, line 34: `default` kind of implies this can be changed but it's hardcoded at some places

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+func (c *Cluster) ScyllaAdminClient() *scyllaclientset.Clientset {
+	client, err := scyllaclientset.NewForConfig(c.AdminClientConfig())
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return client
+}
+
+func (c *Cluster) CommonLabels() map[string]string {
+	return map[string]string{
+		"e2e":               "scylla-operator",
+		"framework-cluster": c.name,
```
tnozicka, line 128: What's the point of labeling objects by cluster within the cluster?
Maybe we should just plumb in the framework name.

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// We have deleted only the namespace object but it is still there with deletionTimestamp set.
+
+	By("Waiting for namespace %q to be removed.", ns.Name)
+	err = WaitForObjectDeletion(ctx, c.DynamicAdminClient(), corev1.SchemeGroupVersion.WithResource("namespaces"), "", ns.Name, &ns.UID)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	klog.InfoS("Namespace removed.", "Namespace", ns.Name)
+}
+
+func (c *Cluster) beforeEach(ctx context.Context) {
```
tnozicka, line 276: I'd expect these on the framework level - make a namespace per `It` on the cluster 0, the rest is on demand

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,330 @@
+// Copyright (C) 2024 ScyllaDB
```
tnozicka, line 1: Where is the `ClusterInterface` and the assertion that both the `Cluster` and `Framework` implement it?

test/e2e/framework/framework.go
```diff
@@ -68,12 +54,20 @@ func NewFramework(name string) *Framework {
 	return f
 }
 
+func (f *Framework) Cluster(idx int) *Cluster {
```
tnozicka, line 57:
```suggestion
func (f *Framework) Cluster(idx int) *ClusterInterface {
```

test/e2e/framework/framework.go
```diff
@@ -68,12 +54,20 @@ func NewFramework(name string) *Framework {
...
+	return f.clusters[idx]
+}
+
+func (f *Framework) Clusters() []*Cluster {
```
tnozicka, line 61:
```suggestion
func (f *Framework) Clusters() []*ClusterInterface {
```

test/e2e/framework/framework.go
```diff
@@ -68,12 +54,20 @@ func NewFramework(name string) *Framework {
...
+	return f.clusters
+}
+
 func (f *Framework) Namespace() string {
-	return f.namespace.Name
+	return f.Cluster(0).Namespace()
```
tnozicka, line 66: nit: use `DefaultCluster` wrapper

test/e2e/framework/framework.go
```diff
@@ -159,205 +135,38 @@ func (f *Framework) GetDefaultScyllaCluster() *scyllav1.ScyllaCluster {
...
-				UID: &ns.UID,
-			},
-		},
-	)
-	o.Expect(err).NotTo(o.HaveOccurred())
+	var wg sync.WaitGroup
+	defer wg.Wait()
 
-	// We have deleted only the namespace object but it is still there with deletionTimestamp set.
+	for _, c := range f.clusters {
```
tnozicka, line 145: auto namespace only in the default one

tnozicka (commented):
framework and the scripts needs some tweaks but this looks close, thanks!

hack/.ci/run-e2e-gke.sh
```diff
@@ -11,12 +11,36 @@ if [ -z "${ARTIFACTS+x}" ]; then
...
 
+  gather_artifacts_failed=0
+  for pid in "${gather_artifacts_bg_pids[*]}"; do
+    wait "${pid}" || gather_artifacts_failed=1
+  done
+
+  if [ ${gather_artifacts_failed} -eq 1 ]; then
+    echo "Error gathering artifacts" > /dev/stderr
+  fi
+}
```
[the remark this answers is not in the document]
rzetelskik: > this is not specific to the GKE so it should be reusable in the lib/e2e

it's not, but having multiple kubeconfigs is specific to this script at this point. Are you saying we should have a single sourced function handling both KUBECONFIG and KUBECONFIG_DIR?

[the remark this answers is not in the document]
tnozicka: > but having multiple kubeconfigs is specific to this script at this point

https://github.com/scylladb/scylla-operator/pull/1837 will need it as well

the future EKS runner would need it too

> Are you saying we should have a single sourced function handling both KUBECONFIG and KUBECONFIG_DIR?

yes, we need a new a function that can collect artifacts in multiple clusters. Unfortunately, I don't think bash is good with function pointer so you likely need dedicated versions of those function instead of a generic runner.

hack/.ci/lib/e2e.sh
```diff
@@ -57,8 +58,8 @@ EOF
 
   exit_code="$( wait-for-container-exit-with-logs gather-artifacts must-gather must-gather )"
 
-  kubectl -n=gather-artifacts cp --retries=42 -c=wait-for-artifacts must-gather:/tmp/artifacts "${ARTIFACTS}/must-gather"
-  ls -l "${ARTIFACTS}/must-gather"
+  kubectl -n=gather-artifacts cp --retries=42 -c=wait-for-artifacts must-gather:/tmp/artifacts "${1}"
```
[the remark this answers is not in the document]
rzetelskik: done, also added a check for directory

hack/.ci/run-e2e-gke.sh
```diff
@@ -11,12 +11,36 @@ if [ -z "${ARTIFACTS+x}" ]; then
   exit 2
 fi
 
+if [ -z ${KUBECONFIG_DIR+x} ]; then
```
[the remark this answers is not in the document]
rzetelskik: why the double brackets? The entire script uses single ones. Added double quotes.

hack/.ci/run-e2e-gke.sh
```diff
@@ -11,12 +11,36 @@ if [ -z "${ARTIFACTS+x}" ]; then
...
+  kubeconfigs=("${KUBECONFIG}")
+else
+  kubeconfigs=()
+  for f in $( find "$(realpath "${KUBECONFIG_DIR}")" -maxdepth 1 -type f -name '*.kubeconfig' ); do
```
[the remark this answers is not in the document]
rzetelskik: done

hack/.ci/run-e2e-gke.sh
```diff
@@ -28,14 +52,45 @@ if [[ "${SO_DISABLE_NODECONFIG:-false}" == "true" ]]; then
   SO_NODECONFIG_PATH=""
   SO_CSI_DRIVER_PATH=""
 else
+  function unset-default-storageclass {
```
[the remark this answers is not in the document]
rzetelskik: done

hack/.ci/run-e2e-gke.sh
```diff
@@ -28,14 +52,45 @@ if [[ "${SO_DISABLE_NODECONFIG:-false}" == "true" ]]; then
...
+    for r in $( kubectl get storageclasses -o name ); do
+      kubectl patch "${r}" -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
+    done
+  }
+
   # Make sure there is no default storage class before we create our own.
-  for r in $( kubectl get storageclasses -o name ); do kubectl patch "${r}" -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'; done
+  for i in "${!kubeconfigs[@]}"; do
+    KUBECONFIG="${kubeconfigs[$i]}" unset-default-storageclass &
```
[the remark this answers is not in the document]
rzetelskik: ok, done

hack/.ci/run-e2e-gke.sh
```diff
@@ -28,14 +52,45 @@ if [[ "${SO_DISABLE_NODECONFIG:-false}" == "true" ]]; then
...
 
-timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}"
+for i in "${!kubeconfigs[@]}"; do
+  KUBECONFIG="${kubeconfigs[$i]}" timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}" &
+  ci_deploy_bg_pids["${i}"]=$!
+done
+
+ci_deploy_failed=0
+for pid in "${ci_deploy_bg_pids[*]}"; do
+  wait "${pid}" || ci_deploy_failed=1
```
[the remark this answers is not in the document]
rzetelskik: hmm ok, the functions already log errors so it should be fine

pkg/genericclioptions/genericclioptions.go
rzetelskik: done

pkg/genericclioptions/genericclioptions.go
rzetelskik: ok, done

test/e2e/framework/cluster.go
rzetelskik: done

hack/.ci/run-e2e-gke.sh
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/framework/cluster.go
rzetelskik: I think this should actually be coming from the propagated rest config. Looks like atm we propagate them but then override with the default values anyway, so the flags don't actually change anything?

test/e2e/framework/cluster.go
rzetelskik: done

test/e2e/framework/cluster.go
tnozicka: /shrug

test/e2e/framework/framework.go
rzetelskik: done

test/e2e/framework/cluster.go
rzetelskik: must have missed it, although I could swear I was adding it, maybe lost it during rebasing somehow... anyway added

test/e2e/framework/cluster.go
rzetelskik: that's not going to sit well with the cluster interface - there's a `func Namespace() string`, which returns the auto namespace and adding an error to it would require changing a lot of existing code. I don't really see this as part of framework struct. `setupNamespace` also configures the clientConfig, so `KubeClient` and similar functions would have to become optional as well? We could wrap them for the framework, but then having the framework implement the cluster interface doesn't really work without changes to the existing test. Why not have it set up for every cluster anyway?

test/e2e/framework/framework.go
rzetelskik: done

test/e2e/framework/framework.go
rzetelskik: done

test/e2e/framework/framework.go
rzetelskik: https://github.com/scylladb/scylla-operator/pull/1881#discussion_r1606594825

test/e2e/framework/cluster.go
rzetelskik: removed the defaults, let me know if this should be a part of a separate PR

rzetelskik:
I'll test the changes against a multi-datacenter e2e after we'll have gone through all review iterations, I takes too long to do this every time
/hold

test/e2e/framework/cluster.go
tnozicka: I am fine with having the `Namespace()` there for manual call, just not the beforEach in all of them. that seems easily doable by having the before each set only on the cluster and calling it on the cluster index 0.

rzetelskik: Having the namespace there is not the issue, it's just that it can't return an error without changing the code of all our tests. I added a safeguard though - best I can do here. 
I changed the framework to only set up and destroy namespace for the default cluster.

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,352 @@
...
+	"k8s.io/client-go/discovery"
+	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+)
+
+type Cluster interface {
+	Namespace() string
```
tnozicka, line 34: I think this belong to the framework only (and is implemented by the cluster 0, to mimic the it block. After that when you want a namespace in a different cluster, you ask for it. But we shouldn't assume you need a namespace in all the clusters, or that it will be present by default. Only the framework gets the default namespace guaranteed (unless you disable it but I don't think that implemented today).

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,352 @@
...
+	AdminClientConfig() *restclient.Config
+	DiscoveryClient() *discovery.DiscoveryClient
+	DynamicClient() dynamic.Interface
+	DynamicAdminClient() dynamic.Interface
+	KubeClient() *kubernetes.Clientset
+	KubeAdminClient() *kubernetes.Clientset
+	ScyllaClient() *scyllaclientset.Clientset
+	ScyllaAdminClient() *scyllaclientset.Clientset
+	CommonLabels() map[string]string
+	SetUp(ctx context.Context)
```
tnozicka, line 47: CreateNamespace might be more useful. You may also need more then one.

rzetelskik: This isn't just creating the namespace but also setting up a service account, rbac and creating a restricted client. So the following methods are all depending on it:
```
	ClientConfig() *restclient.Config
	DiscoveryClient() *discovery.DiscoveryClient
	DynamicClient() dynamic.Interface
	KubeClient() *kubernetes.Clientset
	ScyllaClient() *scyllaclientset.Clientset
```
So I don't know what exactly you have in mind. If you want me to implement it exactly the way you'd do it, please describe it in more detail, I don't want to do n more iterations on this PR playing a guessing game.

test/e2e/framework/cluster.go
rzetelskik: https://github.com/scylladb/scylla-operator/pull/1881#discussion_r1611217425

test/e2e/framework/cluster.go
rzetelskik: Discussed offline to sync - we'll most likely need something like a Namespace abstraction, but we have to maintain backwards compatibility with Framework to not have to change the existing tests. I think we could move the above methods away from Cluster interface to a Namespace interface, and have Framework implement this interface as well.

tnozicka: I took a stab at it at 
https://github.com/tnozicka/scylla-operator/commits/kacper-multi-cluster-framework/ 
feel free to pick that as a starting point and make changes to make it work if needed - I only took it to the point where it compiled to see the abstraction make sense

rzetelskik: @tnozicka I don't get the `Interface` suffix in interface types and changing the `cluster` struct to public `Cluster`, can you explain?

rzetelskik: also why would `Cluster` interface still have the insecure Namespace method
```
func (c *Cluster) Namespace() string {
	return c.namespace.Name
}
```
if only the default one should have the default namespace?
You also added the `BeforeEach` and `AfterEach` funcs to Cluster implementation, which would imply they should be run, but then I only see them run for the default cluster - imo that's counter-intuitive. If you only want them run for the default cluster, they shouldn't be `Cluster`'s methods, but `Framework`'s.

tnozicka: > @tnozicka I don't get the Interface suffix in interface types and changing the cluster struct to public Cluster, can you explain?

I made them public, so I needed to distinguish them. The `client` was also too often colliding and shadowing with variable names.

> also why would Cluster interface still have the insecure Namespace method

It goes into the frawework.FieldManager for a similar point as bellow

> if only the default one should have the default namespace?
You also added the BeforeEach and AfterEach funcs to Cluster implementation, which would imply they should be run, but then I only see them run for the default cluster - imo that's counter-intuitive. If you only want them run for the default cluster, they shouldn't be Cluster's methods, but Framework's.

Some of this could be on the Framework but at that point you loose the `ClientInterface` on the `Cluster`, so it's a choice to make.

Happy to chat about this tomorrow and see your point of view.

rzetelskik: > > also why would Cluster interface still have the insecure Namespace method
> 
> It goes into the frawework.FieldManager for a similar point as bellow

ok, but why make it insecure? I just don't get the point of these two methods
```
func (c *Cluster) DefaultNamespaceIfAny() string {
	o.Expect(c.namespace).NotTo(o.BeNil())
	return c.namespace.Name
}

func (c *Cluster) Namespace() string {
	return c.namespace.Name
}
```

if there's a reason for keeping both I'd expect something like
```
func (c *Cluster) DefaultNamespaceIfAny() (string, bool) {
	if c.Namespace == nil {
	  return "", false
	}
	
	return c.Namespace.Name, true
}

func (c *Cluster) Namespace() string {
	o.Expect(c.namespace).NotTo(o.BeNil())
	return c.Namespace.Name
}
```

rzetelskik: @tnozicka I applied the changes. Changed some things here and there but the overall approach stayed the same. Notable changes that I remember wrt your suggestion:
- cluster doesn't have a Namespace method - didn't make sense if only the default cluster has a default namespace and it's enforced by framework
- cluster only has AdminClient, not Client - same point as above
- moved fieldManager away from cluster - same point as above
- cluster doesn't have beforeEach/afterEach method - that only made sense for the default cluster, which is only a concept in the context of framework, so imo framework should be doing that work

You also left a FIXME in namespace deletion func - was that referring to movinf artifacts dir creation away from it? Because imo it's ok to do it there, even if you end up doing it multiple times unnecessarily.

tnozicka: > ok, but why make it insecure? I just don't get the point of these two methods

It's all to have the `FullClientInterface` on the `Cluster`.

I found it a bit confusing for the `Framework` to implement `FullClientInterface` and when you do `Framework.DefaultCluster` you reduce it to `AdminClientInterface`

essentially all the rest is a consequence of this choice

tnozicka: ok, let's try it with just the admin client, it's a valid option

hack/.ci/lib/e2e.sh
```diff
@@ -69,7 +75,31 @@ EOF
 }
 
 function gather-artifact-on-exit {
-  gather-artifacts || "Error gathering artifacts" > /dev/stderr
+  gather-artifacts "${ARTIFACTS}/must-gather" || "Error gathering artifacts" > /dev/stderr
+}
+
+# $1- kubeconfig ...
```
tnozicka, line 81:
```suggestion
# $...- kubeconfig(s)
```

tnozicka: `$1` the items are not really all passed in the first arg but each of the items is an independent arg

hack/.ci/lib/e2e.sh
```diff
@@ -69,7 +75,31 @@ EOF
 }
 
 function gather-artifact-on-exit {
-  gather-artifacts || "Error gathering artifacts" > /dev/stderr
+  gather-artifacts "${ARTIFACTS}/must-gather" || "Error gathering artifacts" > /dev/stderr
```
tnozicka, line 78: Why isn't gathering the artifacts in every cluster hooked up by default here? if there is more then one config you gather more if there is only a kubeconfig you still do the same thing we did so far.

hack/.ci/run-e2e-gke.sh
```diff
@@ -11,12 +11,20 @@ if [ -z "${ARTIFACTS+x}" ]; then
...
 fi
 
+if [ -z "${KUBECONFIG_DIR+x}" ]; then
+  kubeconfigs=("${KUBECONFIG}")
+else
+  kubeconfigs=()
+  for f in $( find "$( realpath "${KUBECONFIG_DIR}" )" -maxdepth 1 -type f -name '*.kubeconfig' ); do
+    kubeconfigs+=("${f}")
+  done
+fi
```
tnozicka, line 21: I think this would be a good function in the library to get unified access in the other functions so they can always usu the kubeconfigs array even for single KUBECONFIG

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,156 @@
...
+		createNamespace:  createNamespace,
+		defaultNamespace: nil,
+		defaultClient: Client{
+			Config: nil,
+		},
+		namespacesToDelete: []*corev1.Namespace{},
+	}
+}
+
+func (c *Cluster) DefaultNamespaceIfAny() (*corev1.Namespace, Client, bool) {
```
tnozicka, line 63: I though you said you want it only on the framework - is this still pending changes or did I misinterpret your answer?

rzetelskik: `Namespace` func is only on the framework. `DefaultNamespaceIfAny` is there so that clusters can have a somewhat similar workflow, i.e. so that you can get namespace and its client of the default cluster from the test (for other clusters you'd just do `CreateUserNamespace`). Otherwise the only way to get it would be through framework, so it wouldn't align with the other clusters' workflow. It's a tradeoff at this point, but it sucks a bit one way or another.

hack/.ci/lib/e2e.sh
rzetelskik: yeah I just didn't know how to put it there, thanks

rzetelskik: fixed

hack/.ci/run-e2e-gke.sh
rzetelskik: I moved it to a func but I'm not convinced about using it in other functions by default, you don't always want to call them for all clusters, so you loose some level of control.

hack/.ci/lib/e2e.sh
rzetelskik: Because the other script (release) doesn't support it for now (at least this PR doesn't introduce support there), and imo at this point we would have to start doing the same thing for other functions, so we start losing a degree of control. https://github.com/scylladb/scylla-operator/pull/1881#discussion_r1622273827

hack/.ci/run-e2e-gke.sh
rzetelskik: Ok, added this to two other functions. I'm still not convinced this is the right way to go though, it feels like it doesn't belong as part of a lib.

hack/.ci/lib/e2e.sh
tnozicka: I meant it more like to auto-detect the kubeconfigs variables but this is a good start.

tnozicka (commented):
I think this is a good start and it bumps into multiple design issues at once. It brings a lot of value and I think we can iterate on the scripts or framework interfaces in the future, so it's not set in stone. It will be easier to address each of those nits individually, if needed.

/approve
lgtm, but you need to fix the CI failure

hack/.ci/lib/e2e.sh
rzetelskik: I wanted to do that at first but found this more readable. It also allows to pass a subset if needed.

rzetelskik:
> lgtm, but you need to fix the CI failure

https://github.com/scylladb/scylla-operator-release/pull/206 sent a PR to fix the typo on the CI side

tnozicka:
https://github.com/scylladb/scylla-operator-release/pull/206 landed
/retest

rzetelskik:
```
Waiting for cluster to be provisioned...
Cluster provisioning failed. Exiting.
Missing kubeconfigs.
Usage: /usr/bin/bash kubeconfig [kubeconfig ...]
```

> 
> ci/prow/e2e-gke-release-script-latest	[d03fb8e](https://github.com/scylladb/scylla-operator/commit/d03fb8e0f1a08759392832d780ff22c781e17705)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1881/pull-scylla-operator-master-e2e-gke-release-script-latest/1797534974547595264)	unknown	`/test e2e-gke-release-script-latest`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1881). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

```
Waiting for cluster to be provisioned...
Cluster provisioning failed. Exiting.
Missing kubeconfigs.
Usage: /usr/bin/bash kubeconfig [kubeconfig ...]
```

@tnozicka should I fix this in scylla-operator-release (pass kubeconfig to funcs) or make them discover kubeconfigs here? Also it seems like $0 in funcs doesn't exactly behave as expected

tnozicka:
I'd use the env vars for `KUBECONFIGS`  to match how everything else handles `KUBECONFIG`. If you get `KUBECONFIGS` it wins over a `KUBECONFIG`, if you get `KUBECONFIG`, translate it to `KUBECONFIGS[0]` so you can use `KUBECONFIGS` consistently.

rzetelskik:
> I'd use the env vars for KUBECONFIGS to match how everything else handles KUBECONFIG. If you get KUBECONFIGS it wins over a KUBECONFIG, if you get KUBECONFIG, translate it to KUBECONFIGS[0] so you can use KUBECONFIGS consistently, if needed.

You can't really pass/test arrays as env vars that way, so best I could do here is to do this for `KUBECONFIG_DIR` on sourcing e2e lib. Unless you know a reasonable workaround for this.

tnozicka:
> You can't really pass/test arrays as env vars that way, so best I could do here is to do this for KUBECONFIG_DIR on sourcing e2e lib.

sounds good

rzetelskik:
@tnozicka I realised I haven't passed kubeconfigs to the e2e pod in this PR. Should we land this as a starting point regardless? As I'm trying to run a multi-dc e2e test I'll probably bump into some other issues, but I think the baseline for framework etc is solid.

tnozicka:
I am fine with followups

/approve
/lgtm

PR #1918, opened 2024-05-13
Aggregate NodeConfig status conditions

### LINKED ISSUE #1557: NodeConfig needs to aggregate status conditions so they can be queried

The needs to be a way to asses whether NodeConfig has succeded but we only have conditions like `RaidControllerNodeubuntu-2204Degraded` where the node is variable and impractical/impossible to query.

Visibility and debugability. When say a RAID setup or mounts fail, it need to fail at the correct place where we wait for the NodeConfig to apply and not 5 command later when manager or scyllacluster breaks.

_No response_

rzetelskik:
**Description of your changes:** Currently, node setup controller only sets up conditions in the form of `Node%sAvailable` and the like. They are not practical and can't be easily used to query the NodeConfig status. This PR makes it so that NodeConfig conditions are aggregated to the generic `Available`, `Progressing` and `Degraded`, by:
- ~Replacing the proprietary `NodeConfigCondition` with metav1.Condition in NodeConfig API. What follows is that the helpers for NodeConfigCondition are no longer needed, so any related code is removed.~ Edit: we can't do in this API version due to metav1.Condition having a tighter validation. Left TODOs about making this change in the next API version.
- Adding conditions for NodeConfig controller
- Extending NodeConfig controller with conditions aggregation. Node setup controller still aggregates conditions per-node, but NodeConfig controller performs the top level aggregation. In case any of the nodes' aggregated conditions are missing, it assumes Available=False and Progressing=True for the given node. 
- Deprecating the existing condition of type `NodeConfigReconciledConditionType`. The standard workload conditions are now also used to calculate the deprecated condition.
- Reacting to changes in existing E2E tests.

Resolves #1557 

/kind feature
/priority important-soon

/cc


rzetelskik:
@tnozicka you mention here https://github.com/scylladb/scylla-operator/issues/1557#issuecomment-1806240002 that the tests should be extended with trying out broken raid/mount configurations, but after thinking this through I don't think they should be a part of the e2e suite - condition aggregation is covered by unit tests, and invalid configurations shouldn't go past validation. Any scenarios with corrupted devices require some hacks on the host, are prone to be flaky and are difficult to even come up with - e.g. a scenario working on my local setup wouldn't reproduce in CI.

tnozicka:
> invalid configurations shouldn't go past validation

I don't think validation has a chance to assess this. Say something is already mounted at the target path - it's not something to be assessed on the API level.

> Any scenarios with corrupted devices require some hacks on the host, are prone to be flaky and are difficult to even come up with - e.g. a scenario working on my local setup wouldn't reproduce in CI.

I think it should be possible to come up with at least one without node changes, like trying to mount something over `/dev/zero` or similar but I don't feel strong.

rzetelskik:
> I think it should be possible to come up with at least one without node changes, like trying to mount something over /dev/zero or similar but I don't feel strong.

Ok, I'll try adding one to test degraded condition propagation between controllers.

rzetelskik:
> 
> ci/prow/e2e-gke-serial	[46488f8](https://github.com/scylladb/scylla-operator/commit/46488f815beb3e9429abe9e6b8e0af50fdb630fa)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1918/pull-scylla-operator-master-e2e-gke-serial/1791022420711182336)	true	`/test e2e-gke-serial`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1918). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1918/pull-scylla-operator-master-e2e-gke-serial/1791022420711182336#1:test-build-log.txt%3A637
```
• [FAILED] [34.617 seconds]
Node Setup should make RAID0 array out of loop devices, format it to XFS, and mount at desired location [It] out of three loop devices [Serial]
github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_disksetup.go:305
  [FAILED] %!(EXTRA string=mdadm: Cannot get exclusive access to /host/dev/md127:Perhaps a running process, mounted filesystem or active volume group?
  )
  Unexpected error:
      <exec.CodeExitError>: 
      command terminated with exit code 1
      {
          Err: <*errors.errorString | 0xc000b62500>{
              s: "command terminated with exit code 1",
          },
          Code: 1,
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_disksetup.go:263 @ 05/16/24 08:44:10.709
```
/hold 
I have to investigate, looks like removing the additional wait for node statuses made it flaky.

rzetelskik:
> 
> ci/prow/e2e-gke-serial	[a20b428](https://github.com/scylladb/scylla-operator/commit/a20b42845a7903d4d3d141a9359a516dd738dd2c)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1918/pull-scylla-operator-master-e2e-gke-serial/1791040685504204800)	true	`/test e2e-gke-serial`
> ci/prow/e2e-gke-parallel-clusterip	[a20b428](https://github.com/scylladb/scylla-operator/commit/a20b42845a7903d4d3d141a9359a516dd738dd2c)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1918/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1791040684799561728)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1918). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

https://github.com/scylladb/scylla-operator/issues/1802#issuecomment-2114880001

pkg/controller/nodesetup/sync.go
```diff
@@ -93,12 +97,63 @@ func (nsc *Controller) sync(ctx context.Context) error {
...
+			Message:            "",
+			ObservedGeneration: nc.Generation,
+		},
+	)
+	if err != nil {
+		aggregationErrs = append(aggregationErrs, fmt.Errorf("can't aggregate available node status conditions: %w", err))
+	}
+
+	nodeProgressingConditionType := fmt.Sprintf(internalapi.NodeProgressingConditionFormat, sanitizedNodeName)
+	progressingCondition, err := controllerhelpers.AggregateStatusConditions(
```
zimnx, line 118: ditto

pkg/controller/nodesetup/sync.go
```diff
@@ -93,12 +97,63 @@ func (nsc *Controller) sync(ctx context.Context) error {
...
+			Message:            "",
+			ObservedGeneration: nc.Generation,
+		},
+	)
+	if err != nil {
+		aggregationErrs = append(aggregationErrs, fmt.Errorf("can't aggregate progressing node status conditions: %w", err))
+	}
+
+	nodeDegradedConditionType := fmt.Sprintf(internalapi.NodeDegradedConditionFormat, sanitizedNodeName)
+	degradedCondition, err := controllerhelpers.AggregateStatusConditions(
```
zimnx, line 133: ditto

pkg/controller/nodeconfig/status.go
```diff
@@ -4,53 +4,17 @@ package nodeconfig
...
-
-	reconciled, err := controllerhelpers.IsDaemonSetRolledOut(ds)
-	if err != nil {
-		return nil, fmt.Errorf("can't determine is a daemonset %q is reconiled: %w", naming.ObjRef(ds), err)
-	}
-
-	cond := scyllav1alpha1.NodeConfigCondition{
-		Type:               scyllav1alpha1.NodeConfigReconciledConditionType,
-		ObservedGeneration: nc.Generation,
-	}
```
zimnx, line 40: Where it is being set now? I don't see it in PR.

pkg/controller/nodesetup/sync.go
```diff
@@ -93,12 +97,63 @@ func (nsc *Controller) sync(ctx context.Context) error {
 		errs = append(errs, fmt.Errorf("can't sync mounts: %w", err))
 	}
 
-	err = controllerhelpers.SetAggregatedNodeConditions(nsc.nodeName, &conditions, nc.Generation)
+	// Aggregate node conditions.
+	var aggregationErrs []error
+	nodeAvailableConditionType := fmt.Sprintf(internalapi.NodeAvailableConditionFormat, sanitizedNodeName)
+	availableCondition, err := controllerhelpers.AggregateStatusConditions(
```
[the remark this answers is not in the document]
rzetelskik: this is the result of the aggregation, which is not node scoped (the aggregated components are)

pkg/controller/nodeconfig/sync.go
```diff
@@ -103,34 +111,160 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
...
+	for _, n := range matchingNodes {
+		nodeAvailableConditionType := fmt.Sprintf(internalapi.NodeAvailableConditionFormat, controllerhelpers.DNS1123SubdomainToValidStatusConditionReason(n.Name))
+		nodeAvailableStatusCondition := apimeta.FindStatusCondition(availableConditions, nodeAvailableConditionType)
+		if nodeAvailableStatusCondition == nil || nodeAvailableStatusCondition.ObservedGeneration < nc.Generation {
+			availableConditions = append(availableConditions, metav1.Condition{
+				Type:               nodeAvailableConditionType,
+				Status:             metav1.ConditionFalse,
+				ObservedGeneration: nc.Generation,
+				Reason:             internalapi.AwaitingConditionReason,
+				Message:            fmt.Sprintf("Awaiting available condition of node %q to be set.", naming.ObjRef(n)),
```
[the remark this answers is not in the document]
rzetelskik: done

pkg/controller/nodeconfig/sync.go
```diff
@@ -46,6 +49,11 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("can't get ScyllaOperatorConfig: %w", err)
 	}
 
+	matchingNodes, err := ncc.getMatchingNodes(nc)
```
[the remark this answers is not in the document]
rzetelskik: done

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
```diff
@@ -293,13 +297,79 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
 			nc.Name,
 			controllerhelpers.WaitForStateOptions{TolerateDelete: false},
 			utils.IsNodeConfigRolledOut,
-			utils.IsNodeConfigDoneWithNodes(matchingNodes),
+			utils.IsNodeConfigDoneWithNodeTuningFunc(matchingNodes),
```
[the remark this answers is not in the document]
rzetelskik: this func is only checking the tuning part though - there is no need for `IsNodeConfigDoneWithNodes` now as the conditions are aggregated so the corresponding logic is checked in `IsNodeConfigRolledOut`

pkg/controller/nodeconfig/status.go
rzetelskik: it isn't - progressing and available conditions are used instead: https://github.com/scylladb/scylla-operator/pull/1918/files?diff=split&w=0#diff-fe70114ecaf255f5bde24d205c100d57e5b6417be893421752f55453b2cb4bedR60-R72

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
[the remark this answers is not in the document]
rzetelskik: the other func should also go away with https://github.com/scylladb/scylla-operator/issues/1920

pkg/controller/nodeconfig/sync.go
```diff
@@ -103,34 +111,160 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
...
+		errs = append(errs, fmt.Errorf("can't sync DaemonSet(s): %w", err))
 	}
 
-	err = ncc.syncDaemonSet(ctx, nc, soc, daemonSets)
+	// Aggregate conditions.
+	var aggregationErrs []error
+
+	availableConditions := controllerhelpers.FindStatusConditionsWithSuffix(status.Conditions, scyllav1alpha1.AvailableCondition)
+
+	// Use a default available node condition for nodes which haven't propagated their conditions to status yet.
```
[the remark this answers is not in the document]
rzetelskik: I extracted a func creating default node conditions. I could probably go further but imo it would make it more convoluted than it's worth it. Let me know what you think.

pkg/api/scylla/v1alpha1/types_nodeconfig.go
```diff
@@ -22,37 +22,6 @@ import (
...
 )
 
-type NodeConfigConditionType string
-
-const (
-	// Reconciled indicates that the NodeConfig is fully deployed and available.
-	NodeConfigReconciledConditionType NodeConfigConditionType = "Reconciled"
-)
-
-type NodeConfigCondition struct {
```
tnozicka, line 32: the kubebuilder validation for metav1.Condition differs - I am afraid we can't do that in the same API version

rzetelskik: does the rule about not tightening the validation also apply for statuses? Because this only changes the status.

tnozicka: I thinks so - on a storage migration you need to update your object. New status is dropped and the old one stays - then the object is validated - which I assume fails. You can try it out though.

rzetelskik: It may fail currently because the constructed aggregated node conditions can fail validation (although this also applies to the regular aggregated conditions, see: https://github.com/scylladb/scylla-operator/issues/2102). Imo we could move to Conditions, but not in one release, as the NodeCondition created now wouldn't pass Condition's validation. Imo we could merge this (without NodeCondition->Condition switch) and fix Condition and NodeCondition aggregation to make sure they pass validation and then make a switch in following releases.

tnozicka: You can't be sure the sync loop for every single object in a previous release got to store the new status successfully. I'd rather do validation changes  with v1alpha2.

This isn't also required for condition aggregations - you can internally convert the conditions at the start of the sync loop and back at the end
```golang
// TODO: Move NodeConfigCondition to metav1.Condition in the next API version.

type NodeConfigCondition struct {
	// type is the type of the NodeConfig condition.
	Type NodeConfigConditionType `json:"type"`
	// status represents the state of the condition, one of True, False, or Unknown.
	Status corev1.ConditionStatus `json:"status"`
	// observedGeneration represents the .metadata.generation that the condition was set based upon.
	// For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
	// with respect to the current state of the instance.
	// +optional
	// +kubebuilder:validation:Minimum=0
	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
	// lastTransitionTime is last time the condition transitioned from one status to another.
	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
	// reason is the reason for condition's last transition.
	Reason string `json:"reason"`
	// message is a human-readable message indicating details about the transition.
	Message string `json:"message"`
}

func (c *NodeConfigCondition) ToMetaV1Condition() metav1.Condition {
	return metav1.Condition{
		Type:               string(c.Type),
		Status:             metav1.ConditionStatus(c.Status),
		ObservedGeneration: c.ObservedGeneration,
		LastTransitionTime: c.LastTransitionTime,
		Reason:             c.Reason,
		Message:            c.Message,
	}
}

func NewNodeConfigCondition(c metav1.Condition) NodeConfigCondition {
	return NodeConfigCondition{
		Type:               NodeConfigConditionType(c.Type),
		Status:             corev1.ConditionStatus(c.Status),
		ObservedGeneration: c.ObservedGeneration,
		LastTransitionTime: c.LastTransitionTime,
		Reason:             c.Reason,
		Message:            c.Message,
	}
}

type NodeConfigConditions []NodeConfigCondition

func (c NodeConfigConditions) ToMetaV1Conditions() []metav1.Condition {
	return slices.ConvertSlice(c, func(from NodeConfigCondition) metav1.Condition {
		return from.ToMetaV1Condition()
	})
}

func NewNodeConfigConditions(cs []metav1.Condition) NodeConfigConditions {
	return slices.ConvertSlice(cs, func(from metav1.Condition) NodeConfigCondition {
		return NewNodeConfigCondition(from)
	})
}
```
```golang
func (ncc *Controller) sync(ctx context.Context, key string) error {
// ...
    statusConditions := status.Conditions.ToMetaV1Conditions()
    if nc.DeletionTimestamp != nil {
        return ncc.updateStatus(ctx, nc, status, statusConditions)
    }
// ...
    err = ncc.updateStatus(ctx, nc, status, statusConditions)
// ...
}
```

rzetelskik: > This isn't also required for condition aggregations - you can internally convert the conditions at the start of the sync loop and back at the end

Agreed, I wanted to do this get rid of NodeCondition helpers without changing the schema

pkg/controller/nodesetup/sync.go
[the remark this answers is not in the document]
rzetelskik: done

pkg/controller/nodesetup/sync.go
rzetelskik: done

pkg/controller/nodesetup/sync.go
rzetelskik: done

pkg/api/scylla/v1alpha1/types_nodeconfig.go
rzetelskik: done

rzetelskik:
rebased and applied changes, @zimnx @tnozicka ptal

pkg/controller/nodeconfig/status.go
zimnx: then your should raise the apiversion to v1alpha2

rzetelskik:
> 
> ci/prow/e2e-gke-release-script-latest	[a20b428](https://github.com/scylladb/scylla-operator/commit/a20b42845a7903d4d3d141a9359a516dd738dd2c)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1918/pull-scylla-operator-master-e2e-gke-release-script-latest/1791073418175582208)	true	`/test e2e-gke-release-script-latest`
> ci/prow/e2e-gke-parallel	[77c0be6](https://github.com/scylladb/scylla-operator/commit/77c0be6ac733c5d7bf90442e73a23a4ccc3d96ee)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1918/pull-scylla-operator-master-e2e-gke-parallel/1830909329918660608)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=1918). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

known manager flake
https://github.com/scylladb/scylla-operator/issues/2061#issuecomment-2328690143

pkg/controller/nodeconfig/status.go
rzetelskik: I'll add it back, although the tests won't be depending on it

rzetelskik: done

pkg/controller/nodeconfig/sync.go
```diff
@@ -250,3 +419,67 @@ func (ncc *Controller) getServiceAccounts() (map[string]*corev1.ServiceAccount,
 
 	return sasMap, nil
 }
+
+func (ncc *Controller) getMatchingNodes(nc *scyllav1alpha1.NodeConfig) ([]*corev1.Node, error) {
```
tnozicka, line 423: how does this deal with affinities, anti affinities,and the rest of scheduling?

rzetelskik: I'm not sure I understand - it uses an existing func `IsNodeConfigSelectingNode` which is being used by existing controllers and is covered with unit tests. It covers label selectors, affinities and taints and tolerations that can be set in a NodeConfig spec. Are there any cases not covered by this? If a nodeconfig pod is not scheduled to a node for some reason despite matching these conditions, we should assume unavailable and progressing, no?

pkg/controller/nodeconfig/status.go
rzetelskik: @tnozicka is there any benefit in doing this over what I did with computing this from available/progressing/degraded conditions?

rzetelskik:
@zimnx ping

zimnx (approved):
lgtm
/assign rzetelskik

pkg/controller/nodeconfig/status.go
tnozicka: I haven't seen the new projection at a time I wrote this - looking at it, it changes the semantics of it but I guess that's ok

tnozicka (approved):
/approve
/lgtm

PR #1923, opened 2024-05-17
Add version command

tnozicka:
This PR add `version` command to `scylla-operator` and `scylla-operator-tests` binary so it's easy to figure out where the code was build from.

### DIFF
--- a/pkg/cmd/tests/tests.go
+++ b/pkg/cmd/tests/tests.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"fmt"
 
+	versioncmd "github.com/scylladb/scylla-operator/pkg/cmd/version"
 	"github.com/scylladb/scylla-operator/pkg/cmdutil"
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
 	ginkgotest "github.com/scylladb/scylla-operator/pkg/test/ginkgo"
@@ -77,6 +78,7 @@ func NewTestsCommand(streams genericclioptions.IOStreams) *cobra.Command {
 	}
 
 	userAgent := "scylla-operator-e2e"
+	cmd.AddCommand(versioncmd.NewCmd(streams))
 	cmd.AddCommand(NewRunCommand(streams, Suites, userAgent))
 
 	// TODO: wrap help func for the root command and every subcommand to add a line about automatic env vars and the prefix.
--- a/pkg/cmd/version/version.go
+++ b/pkg/cmd/version/version.go
@@ -1,0 +1,75 @@
+// Copyright (C) 2024 ScyllaDB
+
+package version
+
+import (
+	"fmt"
+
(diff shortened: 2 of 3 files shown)


pkg/cmd/version/version.go
```diff
@@ -0,0 +1,75 @@
...
+		SilenceUsage:  true,
+	}
+
+	return cmd
+}
+
+func (o *Options) Validate() error {
+	var errs []error
+
+	return utilerrors.NewAggregate(errs)
```
rzetelskik, line 65: return nil?

pkg/cmd/version/version.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type Options struct {
+}
+
+func NewOptions(streams genericclioptions.IOStreams) *Options {
+	return &Options{}
+}
+
+func NewVersionCmd(streams genericclioptions.IOStreams) *cobra.Command {
```
rzetelskik, line 22: nit: NewCommand
since it has its own package

pkg/cmd/tests/tests.go
```diff
@@ -77,6 +78,7 @@ func NewTestsCommand(streams genericclioptions.IOStreams) *cobra.Command {
 	}
 
 	userAgent := "scylla-operator-e2e"
+	cmd.AddCommand(versioncmd.NewVersionCmd(streams))
```
zimnx, line 81: how about we set cmd.Version to have `--version` argument on root command? This PR would be `+4/-0`.

tnozicka: That would mean the command without any args would have to tell you that you need to specify either `--version` flag or a sub-command which is inconsistent, so I very much prefer the separate sub-command. This is also an analogy to `kubectl version`.

pkg/cmd/version/version.go
tnozicka: it effectively does (and doesn't do any allocation) while still sets the patter we use in every one of these validate commands

pkg/cmd/version/version.go
tnozicka: done

pkg/cmd/tests/tests.go
tnozicka: also the sub-command is more fitting to doing additional flags like `--json`

pkg/cmd/version/version.go
rzetelskik: I don't see any reason to do this preemptively but ok, let's not waste time on discussing this

pkg/cmd/tests/tests.go
rzetelskik: I agree with Tomas, `$command version` is more idiomatic in the entire go-cli landscape and I'd expect our executables to do that rather than the flag

zimnx (approved):
lgtm
/assign rzetelskik

PR #1926, opened 2024-05-17
Split ScyllaDB API status probes into a dedicated container

tnozicka:
**Description of your changes:**
At this point, [ScyllaDB doesn't provide](https://github.com/scylladb/scylladb/issues/8275) standardized `/readyz` and `/healthz` probes, so we serve a wrapper around CQL status (with some tweak) from an operator binary. This is unfortunately running in the same container as ScyllaDB and shares the cpu quota with it and has a negative impact on latency. There are others but this is a first step towards fixing and probes are really the really problematic ones as they serve continuously.

This PR move ScyllaDB probes to run in a separate container so it run in a separate cpu slice and doesn't affect ScyllaDB latency. This is also more architecturally clean and is a step forward to stop coping the operator binary into ScyllaDB container. It also abstracts generic probe patterns and related objects, so we can add other probes in the future, like for the Alternator or other APIs.

This also puts the probe logs at a correct place and no longer mix them with ScyllaDB logs.

**Which issue is resolved by this Pull Request:**
Resolves #620

PR #1934, opened 2024-05-22
Split startup logic out of ScyllaDB container

### LINKED ISSUE #1941: Split startup logic out of ScyllaDB container
Startup logic should run in a separate container. See https://github.com/scylladb/scylla-operator/issues/1940 for more context.

Preferably anything that doesn't need runtime args, should be done in the operator.

tnozicka:
Startup logic should run in a separate container. See https://github.com/scylladb/scylla-operator/issues/1940 for more context.

This PR adds the `scylla-operator run-ignition` command, that runs in its own container and signals ScyllaDB container by creating a file that unblocks its start up. Unfortunately, we can't do this with an init container because we need ScyllaDB container running so tuning can be run and its result can unblock the startup.

Contrary to the old logic that was running in the same container where scylla container ID didn't change (it was restarted with it), the new logic has to deal with all the externalities like living through a scylla container restart and its container id changing.

(It also raises log level for checking the tuning ConfigMap so it's clearer what's being waited on.)

Resolves #1941 

### Requires
- [x] #1926
- [x] #2058

### DIFF
--- a/pkg/cmd/operator/ignition.go
+++ b/pkg/cmd/operator/ignition.go
@@ -1,0 +1,259 @@
+package operator
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/api/scylla/validation"
+	"github.com/scylladb/scylla-operator/pkg/cmd/operator/probeserver"
+	"github.com/scylladb/scylla-operator/pkg/controller/ignition"
+	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/signals"
+	"github.com/scylladb/scylla-operator/pkg/version"
+	"github.com/spf13/cobra"
+	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	cliflag "k8s.io/component-base/cli/flag"
+	"k8s.io/klog/v2"
+)
+
+type IgnitionOptions struct {
+	genericclioptions.ClientConfig
+	genericclioptions.InClusterReflection
+	probeserver.ServeProbesOptions
+
+	ServiceName                       string
+	NodesBroadcastAddressTypeString   string
+	ClientsBroadcastAddressTypeString string
+
+	kubeClient                  kubernetes.Interface
+	mux                         *http.ServeMux
+	nodesBroadcastAddressType   scyllav1.BroadcastAddressType
+	clientsBroadcastAddressType scyllav1.BroadcastAddressType
+}
+
+func NewIgnitionOptions(streams genericclioptions.IOStreams) *IgnitionOptions {
+	mux := http.NewServeMux()
+
+	return &IgnitionOptions{
+		ServeProbesOptions: *probeserver.NewServeProbesOptions(streams, naming.ScyllaDBIgnitionProbePort, mux),
+		ClientConfig:       genericclioptions.NewClientConfig("scylla-operator-ignition"),
+		mux:                mux,
+	}
+}
+
+func (o *IgnitionOptions) AddFlags(cmd *cobra.Command) {
+	o.ClientConfig.AddFlags(cmd)
+	o.InClusterReflection.AddFlags(cmd)
+	o.ServeProbesOptions.AddFlags(cmd)
+
+	cmd.Flags().StringVarP(&o.ServiceName, "service-name", "", o.ServiceName, "Name of the service corresponding to the managed node.")
+	cmd.Flags().StringVarP(&o.NodesBroadcastAddressTypeString, "nodes-broadcast-address-type", "", o.NodesBroadcastAddressTypeString, "Address type that is broadcasted for communication with other nodes.")
+	cmd.Flags().StringVarP(&o.ClientsBroadcastAddressTypeString, "clients-broadcast-address-type", "", o.ClientsBroadcastAddressTypeString, "Address type that is broadcasted for communication with clients.")
+
+}
+
+func NewIgnitionCmd(streams genericclioptions.IOStreams) *cobra.Command {
+	o := NewIgnitionOptions(streams)
+
+	cmd := &cobra.Command{
+		Use:   "run-ignition",
+		Short: "Bootstraps necessary configurations and waits for tuning to finish.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := o.Validate(args)
+			if err != nil {
+				return err
+			}
+
+			err = o.Complete(args)
+			if err != nil {
+				return err
+			}
+
+			err = o.Run(streams, cmd)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+		Hidden:        true,
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	o.AddFlags(cmd)
+
+	return cmd
+}
+
+func (o *IgnitionOptions) Validate(args []string) error {
+	var errs []error
+
+	errs = append(errs, o.ClientConfig.Validate())
+	errs = append(errs, o.InClusterReflection.Validate())
+	errs = append(errs, o.ServeProbesOptions.Validate(args))
+
+	if len(o.ServiceName) == 0 {
+		errs = append(errs, fmt.Errorf("service-name can't be empty"))
+	} else {
+		serviceNameValidationErrs := apimachineryvalidation.NameIsDNS1035Label(o.ServiceName, false)
+		if len(serviceNameValidationErrs) != 0 {
+			errs = append(errs, fmt.Errorf("invalid service name %q: %v", o.ServiceName, serviceNameValidationErrs))
+		}
+	}
+
+	if !slices.ContainsItem(validation.SupportedBroadcastAddressTypes, scyllav1.BroadcastAddressType(o.NodesBroadcastAddressTypeString)) {
+		errs = append(errs, fmt.Errorf("unsupported value of nodes-broadcast-address-type %q, supported ones are: %v", o.NodesBroadcastAddressTypeString, validation.SupportedBroadcastAddressTypes))
+	}
+
+	if !slices.ContainsItem(validation.SupportedBroadcastAddressTypes, scyllav1.BroadcastAddressType(o.ClientsBroadcastAddressTypeString)) {
+		errs = append(errs, fmt.Errorf("unsupported value of clients-broadcast-address-type %q, supported ones are: %v", o.ClientsBroadcastAddressTypeString, validation.SupportedBroadcastAddressTypes))
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (o *IgnitionOptions) Complete(args []string) error {
+	err := o.ClientConfig.Complete()
+	if err != nil {
+		return err
+	}
+
+	err = o.InClusterReflection.Complete()
+	if err != nil {
+		return err
+	}
+
+	err = o.ServeProbesOptions.Complete(args)
+	if err != nil {
+		return err
+	}
+
+	o.kubeClient, err = kubernetes.NewForConfig(o.ProtoConfig)
+	if err != nil {
+		return fmt.Errorf("can't build kubernetes clientset: %w", err)
+	}
+
+	o.nodesBroadcastAddressType = scyllav1.BroadcastAddressType(o.NodesBroadcastAddressTypeString)
+	o.clientsBroadcastAddressType = scyllav1.BroadcastAddressType(o.ClientsBroadcastAddressTypeString)
+
+	return nil
+}
+
+func (o *IgnitionOptions) Run(originalStreams genericclioptions.IOStreams, cmd *cobra.Command) (returnErr error) {
+	klog.Infof("%s version %s", cmd.Name(), version.Get())
+	cliflag.PrintFlags(cmd.Flags())
+
+	stopCh := signals.StopChannel()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		<-stopCh
+		cancel()
+	}()
+
+	return o.Execute(ctx, originalStreams, cmd)
+}
+
+func (o *IgnitionOptions) Execute(cmdCtx context.Context, originalStreams genericclioptions.IOStreams, cmd *cobra.Command) error {
+	identityKubeInformers := informers.NewSharedInformerFactoryWithOptions(
+		o.kubeClient,
+		12*time.Hour,
+		informers.WithNamespace(o.Namespace),
+		informers.WithTweakListOptions(
+			func(options *metav1.ListOptions) {
+				options.FieldSelector = fields.OneTermEqualSelector("metadata.name", o.ServiceName).String()
+			},
+		),
+	)
+	nodeconfigDataCMKubeInformers := informers.NewSharedInformerFactoryWithOptions(
+		o.kubeClient,
+		12*time.Hour,
+		informers.WithNamespace(o.Namespace),
+		informers.WithTweakListOptions(
+			func(options *metav1.ListOptions) {
+				options.LabelSelector = labels.Set{
+					naming.ConfigMapTypeLabel: string(naming.NodeConfigDataConfigMapType),
+				}.String()
+			},
+		),
+	)
+
+	ignitionController, err := ignition.NewController(
+		o.Namespace,
+		o.ServiceName,
+		o.nodesBroadcastAddressType,
+		o.nodesBroadcastAddressType,
+		o.kubeClient,
+		nodeconfigDataCMKubeInformers.Core().V1().ConfigMaps(),
+		identityKubeInformers.Core().V1().Services(),
+		identityKubeInformers.Core().V1().Pods(),
+	)
+	if err != nil {
+		return fmt.Errorf("can't create ignition controller: %w", err)
+	}
+
+	readyzFunc := func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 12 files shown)


rzetelskik:
> ping @rzetelskik @zimnx

Sorry, this one slipped away. I'll send a review today.

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+	healthzFunc := func(w http.ResponseWriter, r *http.Request) {
+		ignited := o.ignited.Load()
+		klog.V(3).InfoS("Probe call", "URI", r.RequestURI, "Ignited", ignited)
+		if !ignited {
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+		return
+	}
+	o.mux.HandleFunc(naming.ReadinessProbePath, healthzFunc)
+	o.mux.HandleFunc(naming.LivenessProbePath, healthzFunc)
```
rzetelskik, line 196: is this used? The container doesn't have a liveness probe set up

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+
+func (o *IgnitionOptions) RunCtx(ctx context.Context, originalStreams genericclioptions.IOStreams, cmd *cobra.Command) (returnErr error) {
+	startTime := time.Now()
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	healthzFunc := func(w http.ResponseWriter, r *http.Request) {
+		ignited := o.ignited.Load()
+		klog.V(3).InfoS("Probe call", "URI", r.RequestURI, "Ignited", ignited)
```
rzetelskik, line 189: nit: I'd lower the level, it generates a lot of noise

pkg/cmd/operator/sidecar.go
```diff
@@ -209,142 +198,21 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
...
-			default:
-				return true, fmt.Errorf("unexpected event type %v", t)
-			}
-		},
-	)
-	if err != nil {
-		return fmt.Errorf("can't wait for optimization: %w", err)
-	}
-
 	klog.V(2).InfoS("Starting scylla")
```
rzetelskik, line 216: the PR description says `[ignition command] signals ScyllaDB container by creating a file that unblocks its start up` - where does that actually happen? I can't seem to find this piece of code. To me it looks like sidecar doesn't wait for anything now?

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+	defer cancel()
+	go func() {
+		<-stopCh
+		cancel()
+	}()
+
+	return o.RunCtx(ctx, originalStreams, cmd)
+}
+
+func (o *IgnitionOptions) RunCtx(ctx context.Context, originalStreams genericclioptions.IOStreams, cmd *cobra.Command) (returnErr error) {
```
zimnx, line 181: In other commands it's called Execute

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+	startTime := time.Now()
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
zimnx, line 185: nit: could be moved closer to where it's used

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+
+	healthzFunc := func(w http.ResponseWriter, r *http.Request) {
+		ignited := o.ignited.Load()
+		klog.V(3).InfoS("Probe call", "URI", r.RequestURI, "Ignited", ignited)
+		if !ignited {
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+		return
+	}
+	o.mux.HandleFunc(naming.ReadinessProbePath, healthzFunc)
```
zimnx, line 195: healtzFunc used in readiness probe, if it's desired, a comment would be great, as it looks like a bug

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+		err := o.ServeProbesOptions.Execute(ctx, originalStreams, cmd)
+		if err != nil {
+			klog.ErrorS(err, "Error running probe server")
+		}
+	}()
+
+	// Wait for the service that holds identity for this scylla node.
+	klog.InfoS("Waiting for Service availability and IP address",
+		"Service", naming.ManualRef(o.Namespace, o.ServiceName),
+	)
```
zimnx, line 211: nit: inconsistent with codebase and unnecessary line break, line isn't very long

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+
+	ignited atomic.Bool
+}
+
+func NewIgnitionOptions(streams genericclioptions.IOStreams) *IgnitionOptions {
+	mux := http.NewServeMux()
+
+	return &IgnitionOptions{
+		ServeProbesOptions: *probeserver.NewServeProbesOptions(streams, naming.ScyllaDBIgnitionProbePort, mux),
+		ClientConfig:       genericclioptions.NewClientConfig("scylla-operator-api-status-probe"),
```
zimnx, line 61:
```suggestion
		ClientConfig:       genericclioptions.NewClientConfig("scylla-operator-ignition"),
```

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("can't wait for optimization: %w", err)
+	}
+
+	cm, ok := cmEvent.Object.(*corev1.ConfigMap)
+	if !ok {
+		return fmt.Errorf("internall error: unexpected object type %T", cmEvent.Object)
```
zimnx, line 356:
```suggestion
		return fmt.Errorf("internal error: unexpected object type %T", cmEvent.Object)
```

pkg/cmd/operator/ignition.go
```diff
@@ -0,0 +1,375 @@
...
+	startTime := time.Now()
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	healthzFunc := func(w http.ResponseWriter, r *http.Request) {
+		ignited := o.ignited.Load()
+		klog.V(3).InfoS("Probe call", "URI", r.RequestURI, "Ignited", ignited)
+		if !ignited {
+			w.WriteHeader(http.StatusServiceUnavailable)
```
zimnx, line 191: it's fatal to the container but since current version of PR doesn't have liveness probe configured at all, we need to consider if values defined there make sense, leaving as a thought for next iteration.

pkg/cmd/operator/sidecar.go
tnozicka: It's been a few months, not sure where it got lost. Added the file wait into the resource defs. Thanks!

tnozicka:
/retest 
https://github.com/scylladb/scylla-operator-release/issues/281

pkg/cmd/operator/ignition.go
rzetelskik: We usually set it to 4 but you're right that this command won't really do much else so I'm fine with leaving as is, feel free to resolve either way

pkg/controller/scyllacluster/resource.go
```diff
@@ -543,37 +543,53 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
...
+--feature-gates=` + func() string {
 										features := utilfeature.DefaultMutableFeatureGate.GetAll()
 										res := make([]string, 0, len(features))
 										for name := range features {
 											res = append(res, fmt.Sprintf("%s=%t", name, utilfeature.DefaultMutableFeatureGate.Enabled(name)))
 										}
 										sort.Strings(res)
 										return strings.Join(res, ",")
-									}()),
-									fmt.Sprintf("--nodes-broadcast-address-type=%s", func() scyllav1.BroadcastAddressType {
```
rzetelskik, line 557: why move away from Sprintf?

tnozicka: its no longer an array of strings but one string so it's consistent with the fixed flags

tnozicka: (ideally all of this would be a go template but that's a change to be discussed separately for sure)

pkg/cmd/operator/ignition.go
tnozicka: I'd generally state that 2 is the default. That said, we have used 4 as a default in some nested containers for the lack of configuration and more complex logic.

tnozicka:
@zimnx @rzetelskik please take another look. I had to add an extra controller instead of the serial (and non-repeated) check to evaluate the ignition / tuning conditions, like it used to,  because a scylla container can now restart independently and tuning was getting stuck on the old containerID not being tuned. (The independent restart always happens in our NodeConfig `should correctly project state for each scylla pod` test because a startup probe times out when the tuning is blocked for 10m30s. The container restart, combined with preStopHook and cri-o bug has also uncovered the other issues around contexts and retries with NodeConfigs handled in the other PRs like #2058 #2073)

pkg/controller/ignition/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	}()
+
+	ignited := true
+
+	svc, err := c.serviceLister.Services(c.namespace).Get(c.serviceName)
+	if err != nil {
+		return fmt.Errorf("can't get service %q: %w", c.serviceName, err)
+	}
+
+	// TODO: This isn't bound to the lifecycle of the ScyllaDB Pod and should be evaluated in the controller.
```
rzetelskik, line 106: leftover? It's part of the controller now?

pkg/controller/ignition/controller.go
```diff
@@ -0,0 +1,216 @@
...
+		return fmt.Errorf("mutiple tuning configmaps are present for pod %q with UID %q", naming.ObjRef(pod), pod.UID)
+	}
+
+	if ignited {
+		klog.V(2).InfoS("Ignition successful", "SignalFile", naming.ScyllaDBIgnitionDonePath)
+	} else {
+		klog.V(2).InfoS("Waiting for ignition to complete.", "SignalFile", naming.ScyllaDBIgnitionDonePath)
+	}
+
+	klog.V(2).InfoS("Updating ignition", "Ignited", ignited, "SignalFile", naming.ScyllaDBIgnitionDonePath)
```
rzetelskik, line 202: nit: Updating ignition state

pkg/controller/ignition/controller.go
```diff
@@ -0,0 +1,216 @@
...
+
+	containerID, err := controllerhelpers.GetScyllaContainerID(pod)
+	if err != nil {
+		return controllertools.NonRetriable(
+			fmt.Errorf("can't get scylla container id in pod %q: %v", naming.ObjRef(pod), err),
+		)
+	}
+
+	if len(containerID) == 0 {
+		klog.V(2).InfoS("ContainerID is not yet set", "Pod", klog.KObj(pod), "UID", pod.UID)
```
rzetelskik, line 144: nit: Scylla ContainerID

pkg/controller/ignition/controller.go
```diff
@@ -0,0 +1,216 @@
...
+		ignited = false
+	} else {
+		klog.V(2).InfoS("Pod has ContainerID set", "Pod", klog.KObj(pod), "UID", pod.UID, "ContainerID", containerID)
```
rzetelskik, line 147: ditto

pkg/controller/scyllacluster/resource.go
```diff
@@ -916,13 +992,26 @@ func agentContainer(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster) corev1.Conta
...
+			"-O",
+			"inherit_errexit",
 			"-c",
-			path.Join(naming.ScyllaAgentConfigDirName, naming.ScyllaAgentAuthTokenFileName),
+			strings.TrimSpace(`
+printf '{"L":"INFO","T":"%s","M":"Waiting for /mnt/shared/ignition.done"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+until [[ -f "/mnt/shared/ignition.done" ]]; do
+  sleep 1;
+done
+printf '{"L":"INFO","T":"%s","M":"Ignited. Starting ScyllaDB Manager"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
```
rzetelskik, line 1008:
```suggestion
printf '{"L":"INFO","T":"%s","M":"Ignited. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
```

It's also still called Scylla Manager, but that's a nit

rzetelskik (commented):
a couple of naming comments
rest lgtm

/assign zimnx

pkg/controller/scyllacluster/resource.go
```diff
@@ -526,37 +526,52 @@ func StatefulSetForRack(r scyllav1.RackSpec, c *scyllav1.ScyllaCluster, existing
...
-									"sidecar",
-									fmt.Sprintf("--feature-gates=%s", func() string {
+									"/usr/bin/bash",
+									"-euEo",
+									"pipefail",
+									"-O",
+									"inherit_errexit",
+									"-c",
+									strings.TrimSpace(`
+printf 'INFO %s ignition - Waiting for /mnt/shared/ignition.done\n' "$( date '+%Y-%m-%d %H:%M:%S,%3N' )" > /dev/stderr
```
zimnx, line 536: What if scylla is ignited, starts and then crashes? This file won't be removed so container starts right away, but Scylla container changed it's ID and might not be tuned anymore. Am I right or I'm missing something?

pkg/controller/ignition/controller.go
tnozicka: nope, but I lost the reference to the issue https://github.com/scylladb/scylla-operator/issues/604

pkg/controller/scyllacluster/resource.go
tnozicka: Do we want to wait for tuning in that case or rather bring ScyllaDB back asap?

pkg/controller/ignition/controller.go
tnozicka: Not state per same, the state can be identical mostly, it touches just the timestamp in most cases. There is an independent log regarding the state.

pkg/controller/ignition/controller.go
tnozicka: done

pkg/controller/scyllacluster/resource.go
tnozicka: done

pkg/controller/ignition/controller.go
tnozicka: done

pkg/controller/scyllacluster/resource.go
tnozicka: probably half way but their docs says ScyllaDB Manager https://manager.docs.scylladb.com/branch-3.3/index.html
fixed the agent part

pkg/controller/scyllacluster/resource.go
zimnx: Restarted container may get different CPU and tuning needs to update CPU mask for IRQ pinning. Since Scylla startup script runs the benchmark at the startup I guess we want to wait for tuning to complete before we evaluate resources.

zimnx: We want to wait not only for tuning, but for values we wait in ignition (IPs), as these are still used in sidecar to generate scylla parameters without checking if they are available.

tnozicka: >  tuning needs to update CPU mask for IRQ pinning

I does. My point was that ScyllaDB coming **back** on 90% of performance immediately and after 2 minutes getting to a 100% is better then waiting 2 minutes to start ScyllaDB at 100% of performance. IOW a degraded node is better then none and IRQ pinning is done asynchronously. Also the independent ignition probe will still show the pod as not ready so it will block the rollout or evictions until re-tuned.

> Since Scylla startup script runs the benchmark at the startup

The volume / disk is still the same, already tuned and the results are cached.
https://github.com/scylladb/scylla-operator/blob/efaff1235eb0ac468324bb3f1ca2f376146bd61a/pkg/sidecar/config/config.go#L302

What benchmark do you have in mind?

> for values we wait in ignition (IPs), as these are still used in sidecar to generate scylla parameters without checking if they are available.

It has run before so they were there already. Checking the service LB IP is and *has been* always racy, if someone messes up with the services in the meantime. 

Is there something you think the Scylla entrypoint wouldn't get through or something that wouldn't be re-tuned while ScyllaDB can already be restarting/serving?

tnozicka: (guess I can remove the file in a preStop hook or check the dates but restarting faster on failures seemed like a good thing)

zimnx: It all depends when crash happens. If it's before the benchmark, then we want to wait until tuning is completed, as we don't want to have skewed results stored in the cache and then used as source of truth by Scylla. 
If it's past the benchmark, then it make sense to start Scylla without waiting for tuning to be completed. 

But we also wait for Service and Pod IPs - for a reason - Pod IP is ephemeral and is not available in Status immediately after Pod starts.  Not having it, would result in empty parameter which would cause another crash. 

IMO we shouldn't differentiate whether benchmark was completed or not. The right way is to wait for all preconditions to be completed before starting Scylla. Removing ignition file in the preStop makes sense.

tnozicka: > But we also wait for Service and Pod IPs - for a reason - Pod IP is ephemeral and is not available in Status immediately after Pod starts.

That's not the case of a container restart though. The IP is still the same. If a pod is deleted the ignition file goes away with the ephemeral storage.

> If it's past the benchmark, then it make sense to start Scylla without waiting for tuning to be completed.

You are correct that this would have to check the presence of io_tune results (if tuning is enabled) as it can fail during the generation. Let's leave it out for now and I'll update the preStop hook.

tnozicka: added the rm

pkg/controller/ignition/controller.go
tnozicka: added the issue link

zimnx (approved):
lgtm
/assign rzetelskik

rzetelskik (approved):
/lgtm

rzetelskik:
@tnozicka there's still a WIP label, is this on purpose?

tnozicka:
> @tnozicka there's still a WIP label, is this on purpose?

nope, probably some race, thx

PR #1949, opened 2024-05-30
Expose scylladb as service

Siva-Saminathan:
Have added new sample yaml to expose the scylla db cluster as a service, So that people can connect to it via host and port.
Updated the docs to add instruction for it and how to connect to scylla db with host and port

### DIFF
--- a/examples/generic/cluster_expose.yaml
+++ b/examples/generic/cluster_expose.yaml
@@ -1,0 +1,61 @@
+# Namespace where the Scylla Cluster will be created
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: scylla
+
+---
+
+# Simple Scylla Cluster
+apiVersion: scylla.scylladb.com/v1
+kind: ScyllaCluster
+metadata:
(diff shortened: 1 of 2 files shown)


examples/generic/cluster_expose.yaml
```diff
@@ -0,0 +1,61 @@
...
+---
+
+# ScyllaDB Service to expose the cluster
+apiVersion: v1
+kind: Service
+metadata:
+  name: scylla-service
+  namespace: scylla
+spec:
+  type: LoadBalancer  # Use LoadBalancer if you need to expose the service externally
```
tnozicka, line 54: I don't think this is needed. Are you familiar with https://operator.docs.scylladb.com/stable/exposing.html ?

tnozicka (changes requested):
I don't think this is needed. Are you familiar with https://operator.docs.scylladb.com/stable/exposing.html ?

zimnx:
Closing as provided example is already available in documentation page.

PR #1952, opened 2024-06-04
Support overriding deploy dir in deploy script

rzetelskik:
**Description of your changes:** Currently, ci-deploy script hardcodes a subpath for its deploy manifests as `"${ARTIFACTS}/deploy"`. To make it possible to align the file structure with other existing workflows (i.e. `<artifacts>/<deploy>/<cluster-idx>`, this PR allows for overriding the deploy target path.

Release testing workflow will need other tweaks so I'll send it in a separate PR.

Resolves #

/kind feature
/priority important-soon
/cc tnozicka

### DIFF
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -35,7 +35,7 @@ SCYLLA_OPERATOR_FEATURE_GATES="${SCYLLA_OPERATOR_FEATURE_GATES:-AllAlpha=true,Al
 export SCYLLA_OPERATOR_FEATURE_GATES
 
 for i in "${!KUBECONFIGS[@]}"; do
-  KUBECONFIG="${KUBECONFIGS[$i]}" timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}" &
+  KUBECONFIG="${KUBECONFIGS[$i]}" DEPLOY_DIR="${ARTIFACTS}/deploy/${i}" timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}" &
   ci_deploy_bg_pids["${i}"]=$!
 done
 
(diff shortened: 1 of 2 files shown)


hack/ci-deploy.sh
```diff
@@ -15,32 +15,35 @@ if [[ -z ${1+x} ]]; then
     exit 1
 fi
 
-ARTIFACTS=${ARTIFACTS:-$( mktemp -d )}
+if [ -z "${DEPLOY_DIR+x}" ]; then
+  ARTIFACTS=${ARTIFACTS:-$( mktemp -d )}
```
tnozicka, line 19: this shouldn't be conditioned on the deploy_dir

hack/.ci/run-e2e-gke.sh
```diff
@@ -35,7 +35,7 @@ SCYLLA_OPERATOR_FEATURE_GATES="${SCYLLA_OPERATOR_FEATURE_GATES:-AllAlpha=true,Al
 export SCYLLA_OPERATOR_FEATURE_GATES
 
 for i in "${!KUBECONFIGS[@]}"; do
-  KUBECONFIG="${KUBECONFIGS[$i]}" timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}" &
+  KUBECONFIG="${KUBECONFIGS[$i]}" DEPLOY_DIR="${ARTIFACTS}/deploy/${i}" timeout --foreground -v 10m "${parent_dir}/../ci-deploy.sh" "${SO_IMAGE}" &
```
tnozicka, line 38: shouldn't we do a deploy dir per cluster? I see cases where what we will deploy to each cluster will differ at some point

rzetelskik: I don't understand, this is doing exactly that? I.e. a deploy dir per cluster

hack/ci-deploy.sh
rzetelskik: agreed, fixed

PR #1957, opened 2024-06-06
Evaluate ScyllaDB 6.0.0-rc2 in CI

zimnx:
This bumps ScyllaDB version used in CI to 6.0.0-rc2 to evaluate whether it doesn't break our tests.

/cc

### DIFF
--- a/pkg/controller/scyllacluster/resource.go
+++ b/pkg/controller/scyllacluster/resource.go
@@ -15,6 +15,7 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	"github.com/scylladb/scylla-operator/pkg/pointer"
+	"github.com/scylladb/scylla-operator/pkg/scyllafeatures"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


dorlaor:
Cheers for the tests!

zimnx:
> hm, I was expecting the scaling tests to fail on https://github.com/scylladb/scylladb/issues/16739 - we should look into why they didn't fail...

Me too, they were downscaling the cluster so I added such test as well, but going from 2->1 core didn't hit it. They were going from 5 to 2.

zimnx:
There were two failures:
1. Restore - manager rejects 6.0.0-rc2 saying it's not supported version.
2. Alternator test - from the test we were reading `system_auth.roles` table which was moved to `system.roles`

We can align alternator test (I did and it passed), but restore failure is something I still need to look into.

rzetelskik:
@zimnx restore won't work until https://github.com/scylladb/scylla-manager/pull/3876 is merged. So we'll need a test table for 5.4.x and 6.0.0 with and without the workaround, but the latter will only work with the next manager release.

zimnx:
Looks like resharding works, at least in our test suite. Segmentation fault, reported in https://github.com/scylladb/scylladb/issues/16739 cannot be reproduced, even when we are scaling by the same amount of CPUs.
[Logs of a node](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/1957/pull-scylla-operator-master-e2e-gke-parallel/1799017452449828864/artifacts/e2e/namespaces/e2e-test-scyllacluster-x7h9v-0-k9pkm/pods/basic-8lwr8-us-east-1-us-east-1a-0/scylla.current) confirms node was resharding and successfully went UN afterwards. 
The weird thing, it that only two tables (system.truncated, system.compaction_history) were resharded, while there're many more system tables, and one keyspace created keyspace. 

Manager doesn't support restore in 6.0.0, it returns an error instead. As Kacper pointed out, we need to wait until https://github.com/scylladb/scylla-manager/pull/3876 is merged and released.

PR #1960, opened 2024-06-10
Add Execute() to OperatorOptions

zimnx:
Allows for reusing Operator logic without context signal registration and print of command version which is unwanted when custom command reuses OperatorOptions.

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -206,6 +206,10 @@ func (o *OperatorOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Co
 		cancel()
 	}()
 
+	return o.Execute(ctx, streams, cmd)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/operator/operator.go
```diff
@@ -206,6 +206,10 @@ func (o *OperatorOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Co
 		cancel()
 	}()
 
+	return o.Execute(ctx, streams, cmd)
```
rzetelskik, line 209: we could just move the above to RunE instead?

zimnx: We had similar discussion [here](https://github.com/scylladb/scylla-operator/pull/1926/files#diff-a1f618d37d8c9d3b43efd39ea637e339593f9f1aee3a4a87ee514a0c0f6f3acb) and I think one with Execute method is clearer.

rzetelskik (commented):
lgtm
/assign tnozicka

tnozicka (approved):
/approve
/lgtm

PR #1961, opened 2024-06-10
[v1.13] Pin tags to v1.13

zimnx:
In preparation for v1.13 release pin image tags to v1.13 rolling tag.


tnozicka:
this should target v1.13, not master

zimnx:
> this should target v1.13, not master

good catch, fixed.

PR #1970, opened 2024-06-14
Add a NodeConfig example for generic environments

### LINKED ISSUE #1556: Provide a NodeConfig that can setup storage on any node without manual actions

We should be able to configure the nodes and local-csi driver without the need to manually mess with the node (e.g.) loopback devices or making folders.

Today it's seamless only if you run on real disks, otherwise you need to start tweaking with loop devices or additional virtual disks.

This is very important for us so we can recommend the local-csi-driver as our primary path for every one and have consistent docs and demos.

One of the tasks would be https://github.com/scylladb/scylla-operator/issues/1397 but it may not be the only one.

Consistent setup experience, docs and demos.

_No response_

rzetelskik:
**Description of your changes:** This PR adds a NodeConfig example dedicated for generic environments. It sets up local storage  on loop devices.

Resolves #1556 

/kind feature
/priority important-soon

### DIFF
--- a/examples/generic/nodeconfig-alpha.yaml
+++ b/examples/generic/nodeconfig-alpha.yaml
@@ -1,0 +1,26 @@
+apiVersion: scylla.scylladb.com/v1alpha1
+kind: NodeConfig
+metadata:
+  name: cluster
+spec:
+  localDiskSetup:
+    loopDevices:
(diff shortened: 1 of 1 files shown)


zimnx (approved):
/lgtm

PR #1971, opened 2024-06-14
Add dedicated ServiceAccount for perftune jobs

### LINKED ISSUE #1975: Perftune jobs should use dedicated serviceaccount
At this point, perftune jobs that run in `scylla-operator-node-tuning` namespace use the default ServiceAccount which makes it impossible to change the individual permissions of perftune jobs that need to run privileged operations. We should wire up a dedicated service account for them which will also unblock running them on OpenShift where they need to be granted additional RBAC rules.

tnozicka:
This PR adds a dedicated ServiceAccount to use for perftune jobs which enhances security isolation and allows some platforms (like OpenShift) to grant it extra permissions.

Resolves #1975

### Requires:
- [x] https://github.com/scylladb/scylla-operator/pull/2000

### DIFF
--- a/pkg/controller/nodeconfig/sync.go
+++ b/pkg/controller/nodeconfig/sync.go
@@ -63,12 +63,22 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 		objectErrs = append(objectErrs, err)
 	}
 
+	roles, err := ncc.getRoles()
+	if err != nil {
+		objectErrs = append(objectErrs, err)
+	}
+
 	serviceAccounts, err := ncc.getServiceAccounts()
 	if err != nil {
 		objectErrs = append(objectErrs, err)
 	}
 
 	clusterRoleBindings, err := ncc.getClusterRoleBindings()
+	if err != nil {
+		objectErrs = append(objectErrs, err)
+	}
+
+	roleBindings, err := ncc.getRoleBindings()
 	if err != nil {
 		objectErrs = append(objectErrs, err)
 	}
@@ -114,6 +124,11 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 		errs = append(errs, fmt.Errorf("sync ClusterRole(s): %w", err))
 	}
 
+	err = ncc.syncRoles(ctx, nc, roles)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("sync Role(s): %w", err))
+	}
+
 	err = ncc.syncServiceAccounts(ctx, serviceAccounts)
 	if err != nil {
 		errs = append(errs, fmt.Errorf("sync ServiceAccount(s): %w", err))
@@ -122,6 +137,11 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 	err = ncc.syncClusterRoleBindings(ctx, clusterRoleBindings)
 	if err != nil {
 		errs = append(errs, fmt.Errorf("sync ClusterRoleBinding(s): %w", err))
+	}
+
+	err = ncc.syncRoleBindings(ctx, nc, roleBindings)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("sync RoleBinding(s): %w", err))
 	}
 
 	err = ncc.syncDaemonSet(ctx, nc, soc, daemonSets)
@@ -181,6 +201,38 @@ func (ncc *Controller) getClusterRoleBindings() (map[string]*rbacv1.ClusterRoleB
 	}
 
 	return crbMap, nil
+}
+
+func (ncc *Controller) getRoles() (map[string]*rbacv1.Role, error) {
+	roles, err := ncc.roleLister.List(labels.SelectorFromSet(map[string]string{
+		naming.NodeConfigNameLabel: naming.NodeConfigAppName,
+	}))
+	if err != nil {
+		return nil, fmt.Errorf("list roles: %w", err)
+	}
+
+	res := map[string]*rbacv1.Role{}
+	for i := range roles {
+		res[roles[i].Name] = roles[i]
+	}
+
+	return res, nil
+}
+
+func (ncc *Controller) getRoleBindings() (map[string]*rbacv1.RoleBinding, error) {
+	roleBindings, err := ncc.roleBindingLister.List(labels.SelectorFromSet(map[string]string{
+		naming.NodeConfigNameLabel: naming.NodeConfigAppName,
+	}))
+	if err != nil {
+		return nil, fmt.Errorf("list rolebindings: %w", err)
+	}
+
+	res := map[string]*rbacv1.RoleBinding{}
+	for i := range roleBindings {
+		res[roleBindings[i].Name] = roleBindings[i]
+	}
+
+	return res, nil
 }
 
 func (ncc *Controller) getServiceAccounts() (map[string]*corev1.ServiceAccount, error) {
--- a/pkg/controller/nodeconfig/sync_rolebindings.go
+++ b/pkg/controller/nodeconfig/sync_rolebindings.go
@@ -1,0 +1,50 @@
+// Copyright (C) 2024 ScyllaDB
+
+package nodeconfig
+
+import (
+	"context"
+	"fmt"
+
+	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
+	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
+	"github.com/scylladb/scylla-operator/pkg/resourceapply"
+	rbacv1 "k8s.io/api/rbac/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+func (ncc *Controller) syncRoleBindings(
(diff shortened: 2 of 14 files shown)


tnozicka:
@zimnx @rzetelskik ptal

pkg/controller/nodeconfig/sync.go
```diff
@@ -73,6 +78,11 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 		objectErrs = append(objectErrs, err)
 	}
 
+	roleBindings, err := ncc.getRoleBindings()
```
zimnx, line 81: This file already has a mixed way of getting dependent objects, lets use new approach for new code.

pkg/controller/nodeconfig/sync_rolebindings.go
```diff
@@ -0,0 +1,79 @@
...
+	return utilerrors.NewAggregate(errs)
+}
+
+func (ncc *Controller) syncRoleBindings(
+	ctx context.Context,
+	roleBindings map[string]*rbacv1.RoleBinding,
+) error {
+	requiredRoleBindings := ncc.makeRoleBindings()
+
+	// Delete any excessive RoleBindings.
```
rzetelskik, line 62: nit: redundant comment imo

pkg/controller/nodeconfig/sync.go
rzetelskik: doesn't `controllerhelpers.GetObjects` only apply to controlled objects, while these are non-namespaced? Unless you mean something else by new approach

zimnx: They are originating from NodeConfig so there should be a controllerRef on them, and these are namespaced RoleBindings and Roles.

pkg/controller/nodeconfig/sync_rolebindings.go
```diff
@@ -0,0 +1,79 @@
...
+	// Delete has to be the first action to avoid getting stuck on quota.
+	if err := ncc.pruneRoleBindings(ctx, requiredRoleBindings, roleBindings); err != nil {
+		return fmt.Errorf("can't delete RoleBinding(s): %w", err)
+	}
```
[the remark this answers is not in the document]
tnozicka: copied it from the clusterrole part, but lets do this

pkg/controller/nodeconfig/sync.go
tnozicka: the existing SA doesn't use controller ref either

https://github.com/scylladb/scylla-operator/blob/11394625d288542782cd23fee50a3d9ff6957ebe/pkg/controller/nodeconfig/resource.go#L29-L39
These are shared by multiple NodeConfig and a for the lack of a better place we reconcile them with each of them.

pkg/controller/nodeconfig/sync_rolebindings.go
tnozicka: it's in all the other functions, it this one worth the inconsistency?

pkg/controller/nodeconfig/sync_rolebindings.go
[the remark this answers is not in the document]
tnozicka: done

pkg/controller/nodeconfig/sync_rolebindings.go
rzetelskik: doesn't matter that much, hence a nit. Feel free to ignore if you prefer to keep it consistent

pkg/controller/nodeconfig/sync_roles.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (ncc *Controller) syncRoles(ctx context.Context, Roles map[string]*rbacv1.Role) error {
+	requiredRoles := ncc.makeRoles()
+
+	// Delete any excessive Roles.
+	// Delete has to be the first action to avoid getting stuck on quota.
+	if err := ncc.pruneRoles(ctx, requiredRoles, Roles); err != nil {
+		return fmt.Errorf("can't delete Role(s): %w", err)
+	}
```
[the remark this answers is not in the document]
tnozicka: done

pkg/controller/nodeconfig/sync_roles.go
```diff
@@ -0,0 +1,77 @@
...
+	"context"
+	"fmt"
+
+	"github.com/scylladb/scylla-operator/pkg/resourceapply"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+func (ncc *Controller) makeRoles() []*rbacv1.Role {
```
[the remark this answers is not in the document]
tnozicka: inlined

pkg/controller/nodeconfig/sync_roles.go
```diff
@@ -0,0 +1,77 @@
...
+	Roles := []*rbacv1.Role{
```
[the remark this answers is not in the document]
tnozicka: fixed

pkg/resourceapply/rbac_test.go
```diff
@@ -1565,3 +1565,492 @@ func TestApplyRoleBinding(t *testing.T) {
...
+
+						for i := range roleList.Items {
+							err := roleCache.Add(&roleList.Items[i])
+							if err != nil {
+								t.Fatal(err)
+							}
+						}
+					}
+
+					gotSts, gotChanged, gotErr := ApplyRole(ctx, client.RbacV1(), roleLister, recorder, tc.required, ApplyOptions{})
```
zimnx, line 2008: s/Sts/Role here and below

pkg/controller/nodeconfig/sync.go
zimnx: Then enqueueing owner in event handlers must not work and your PR does exactly that

pkg/controller/nodeconfig/sync_roles.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func (ncc *Controller) syncRoles(ctx context.Context, nc *scyllav1alpha1.NodeConfig, roles map[string]*rbacv1.Role) error {
+	requiredRoles := []*rbacv1.Role{
+		makePerftuneRole(),
+	}
+
+	// Delete any excessive Roles.
+	// Delete has to be the first action to avoid getting stuck on quota.
+	err := controllerhelpers.Prune(
```
zimnx, line 23: shouldn't Prune complain about missing controllerref on deleted objects? We are not allowed to delete them. There should be an option to disable it on justified cases similar to apply functions.

pkg/controller/nodeconfig/sync.go
tnozicka: yeah, that should enqueueAll but it must be broken in all the other places already there that I copied it from

pkg/controller/nodeconfig/sync_roles.go
tnozicka: You already give it a set of claimed and pre-existing objects so it's assumed you own them by how you've claimed them in the fist place. If you were to do a patch or update instead you'd not have rechecked again. For the case of the helper it's cheap but I'd not couple it to this PR.

pkg/resourceapply/rbac_test.go
tnozicka: fixed, thx

pkg/controller/nodeconfig/sync.go
tnozicka: updated based on #2000

pkg/controller/nodeconfig/sync_roles.go
```diff
@@ -0,0 +1,47 @@
...
+		ctx,
+		requiredRoles,
+		roles,
+		&controllerhelpers.PruneControlFuncs{
+			DeleteFunc: ncc.kubeClient.RbacV1().Roles(nc.Namespace).Delete,
+		},
+		ncc.eventRecorder)
+	if err != nil {
+		return fmt.Errorf("can't prune RoleBinding(s): %w", err)
```
zimnx, line 32:
```suggestion
		return fmt.Errorf("can't prune Role(s): %w", err)
```

pkg/controller/nodeconfig/sync_roles.go
zimnx: I meant to add a safety check so we won't delete objects we weren't meant to delete. Filing an issue for future improvement is fine.

tnozicka: https://github.com/scylladb/scylla-operator/issues/2010

pkg/controller/nodeconfig/sync_roles.go
tnozicka: fixed, thx

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1996#issuecomment-2216892629
/retest

tnozicka:
https://github.com/scylladb/scylla-operator/issues/1996#issuecomment-2217136797
/retest

Issue #1206, opened 2023-04-05
Add support for running ScyllaClusters on IPv6


tnozicka:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
Allow exposing ScyllaClusters on IPv6 and add a testing coverage.

**What is use case behind this feature:**
UX, scalability.

**Additional Information:**
From the top of my head I assume this requires us:
- to [move away from ClusterIPs to PodIPs]()
- tweak a few places like the code that serves probes
- setup a CI with Kubernetes cluster on IPv6

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2435.

PR #1973, opened 2024-06-15
Update config.go

Mokto:
# Description of your changes:
Supports ipv6 addresses. I assume that there is more to be done on this PR (make it conditional maybe?) so we can merge it. I would appreciate you guide me in the right direction.

It did work on my ipv6 cluster though.

Thanks!

# Which issue is resolved by this Pull Request:
Resolves https://github.com/scylladb/scylla-operator/issues/1206


pkg/sidecar/config/config.go
```diff
@@ -208,7 +208,7 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
 	}
 
 	// Listen on all interfaces so users or a service mesh can use localhost.
-	listenAddress := "0.0.0.0"
+	listenAddress := "::"
 	prometheusAddress := "0.0.0.0"
```
tnozicka, line 212: just one line bellow your change you have left an explicit IPv4 address
/shrug

tnozicka (commented):
I am fine landing some ipv6 fixes (without having a CI support) but it should be done in a structured and coherent way throughout the codebase.

Mokto:
Thanks for the feedback. I'm afraid that my change would break some ipv4 environments maybe ?

tnozicka:
I think dual stack is viable but it needs to be wired consistently.
/ok-to-test

tnozicka:
closing as stale, feel free to reopen if you want to get back to it

PR #1974, opened 2024-06-15
Extend Handlers' EnqueueAll with enqueing func parameter

rzetelskik:
**Description of your changes:** Conditionally enqueueing a subset of controlled objects seems to be a repeating use case in our codebase. This PR extends the EnqueueAll handler method parameters with an enqueueing function, making it possible to use function composition for such use cases.

Xref: https://github.com/scylladb/scylla-operator-enterprise/pull/25#discussion_r1638032987

Resolves #

/kind feature
/priority important-longterm

/cc zimnx
/assign zimnx

### DIFF
--- a/pkg/controllerhelpers/handlers.go
+++ b/pkg/controllerhelpers/handlers.go
@@ -90,14 +90,20 @@ func (h *Handlers[T]) Enqueue(depth int, untypedObj kubeinterfaces.ObjectInterfa
 func (h *Handlers[T]) EnqueueAll(depth int, untypedObj kubeinterfaces.ObjectInterface, op HandlerOperationType) {
 	klog.V(4).InfoSDepth(depth, "Enqueuing all controller objects", getObjectLogContext(untypedObj, nil)...)
 
-	controllerObjs, err := h.getterLister.List(untypedObj.GetNamespace(), labels.Everything())
-	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("couldn't list all controller objects for %T: %w", untypedObj, err))
-		return
-	}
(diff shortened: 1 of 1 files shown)


pkg/controllerhelpers/handlers.go
```diff
@@ -90,14 +90,20 @@ func (h *Handlers[T]) Enqueue(depth int, untypedObj kubeinterfaces.ObjectInterfa
...
 
-	controllerObjs, err := h.getterLister.List(untypedObj.GetNamespace(), labels.Everything())
-	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("couldn't list all controller objects for %T: %w", untypedObj, err))
-		return
-	}
+	h.EnqueueAllFunc(h.Enqueue)(depth+1, untypedObj, op)
+}
+
+func (h *Handlers[T]) EnqueueAllFunc(enqueueFunc EnqueueFuncType) EnqueueFuncType {
```
tnozicka, line 96: I think we always want to call the `.Enqueue` function. The goal from  https://github.com/scylladb/scylla-operator-enterprise/pull/25#discussion_r1638032987 seems to be more like `EnqueueIf`

/lgtm cancel

rzetelskik: > I think we always want to call the .Enqueue function. 

Why?

This is aligned with existing `EnqueueOwner` and `EnqueueOwnerFunc` and imo allows for greater flexibility than introducing something like `EnqueueAllWithFilter` (I guess this is what you had in mind by EnqueueIf?), since you can achieve that with function composition then.

tnozicka: I haven't actually expanded all the code around, you are right that this is in line with the other patterns there.
/lgtm

PR #1991, opened 2024-06-24
Bump default ScyllaDB version used in E2E's to 6.0.1

### LINKED ISSUE #1954: Update ScyllaDB version in e2e to 6.0.x
We should update our CI to use the 6.0.x, if not out yet with the latest rc and replacing it before landing it.

We are likely to find out that resharding with tablets doesn't work and we need to disable them by default https://github.com/scylladb/scylladb/issues/16739

zimnx:
Bumps default ScyllaDB version used in E2E's to 6.0.1, and align tests to 6.0 changes.

Scaling E2E was aligned to make sure it doesn't break minimal required quorum on scaling changes.
Existing test scaled below keyspace RF which is no longer possible, as Scylla rejects decommision when there's a keyspace having RF higher than node count.
Test step checking decommission of drained node was moved earlier to fix the same quorum breakage.

Alternator E2E required a table name change from which we are gettingpassword. Table name was renamed in 6.0.

Restore E2E was parametrized to make sure we test the procedure for both default ScyllaDB version and 2024.1 where a workaround explained in the documentation is required.

Fixes #1954

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_alternator.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_alternator.go
@@ -128,7 +128,7 @@ authorizer: CassandraAuthorizer
 		}
 
 		q := cqlSession.Query(
-			`SELECT salted_hash FROM system_auth.roles WHERE role = ?`,
+			`SELECT salted_hash FROM system.roles WHERE role = ?`,
 			awsCredentials.AccessKeyID,
 		).WithContext(ctx)
 		err = q.Scan(&awsCredentials.SecretAccessKey)
--- a/test/e2e/set/scyllacluster/scyllamanager_object_storage.go
+++ b/test/e2e/set/scyllacluster/scyllamanager_object_storage.go
@@ -32,12 +32,26 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
 
 	f := framework.NewFramework("scyllacluster")
 
-	g.It("should register cluster, sync backup tasks and support manual restore procedure", func() {
+	type entry struct {
+		scyllaRepository           string
+		scyllaVersion              string
+		preTargetClusterCreateHook func(cluster *scyllav1.ScyllaCluster)
+		postSchemaRestoreHook      func(context.Context, *framework.Framework, *scyllav1.ScyllaCluster)
+	}
+
+	g.DescribeTable("should register cluster, sync backup tasks and support manual restore procedure", func(e entry) {
 		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
 		defer cancel()
 
 		sourceSC := f.GetDefaultScyllaCluster()
 		sourceSC.Spec.Datacenter.Racks[0].Members = 1
+
+		if len(e.scyllaRepository) != 0 {
+			sourceSC.Spec.Repository = e.scyllaRepository
+		}
+		if len(e.scyllaVersion) != 0 {
+			sourceSC.Spec.Version = e.scyllaVersion
+		}
 
 		objectStorageType := f.GetObjectStorageType()
 		switch objectStorageType {
@@ -266,8 +280,12 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
 
 		targetSC := f.GetDefaultScyllaCluster()
 		targetSC.Spec.Datacenter.Racks[0].Members = sourceSC.Spec.Datacenter.Racks[0].Members
-		// Restoring schema with ScyllaDB OS 5.4.X or ScyllaDB Enterprise 2024.1.X and consistent_cluster_management isn’t supported.
-		targetSC.Spec.ScyllaArgs = "--consistent-cluster-management=false"
+		targetSC.Spec.Repository = sourceSC.Spec.Repository
+		targetSC.Spec.Version = sourceSC.Spec.Version
+
+		if e.preTargetClusterCreateHook != nil {
+			e.preTargetClusterCreateHook(targetSC)
+		}
 
 		switch objectStorageType {
 		case framework.ObjectStorageTypeGCS:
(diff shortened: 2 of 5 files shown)


tnozicka (commented):
can you try with https://github.com/scylladb/scylla-manager/releases/tag/3.3.0-0.20240624.9dd98ab35 ?

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -442,7 +442,8 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
 
 		verifyCQLData(ctx, di)
 	},
-		g.Entry("using default ScyllaDB version", entry{}),
+		// FIXME: Test should be enabled when 3.5.x Scylla Manager is available and bumped in CI.
+		g.PEntry("using default ScyllaDB version", entry{}),
```
tnozicka, line 446: shouldn't the PR be marked as WIP?
/hold

zimnx:
New issue with our tests:
https://github.com/scylladb/scylladb/issues/16677#issuecomment-1894071518

> > Do we now require the nr of nodes in the cluster must bigger than the RF of any tablet tables?

> Yes, this is working as expected.


Our tests doesn't lower RF upon scaling down.

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
zimnx: We need to decide one way or the other. We either proceed with having coverage for 6.0.x with 30'ish passing tests, or not having any coverage because one is failing due to known issue.

tnozicka: Is it a matter of just making a manager release? I though the PR landed there or is there another fix needed?

ginkgo wise that would be a skip, instead of pending test IMO
`--fail-on-pending will fail the suite if it contains any pending specs. These are generally only used while developing the suite and should not be committed`
https://onsi.github.io/ginkgo/

tnozicka:
> Do we now require the nr of nodes in the cluster must bigger than the RF of any tablet tables?

I am fine bumping the default replicas to maintain the RF where needed.

> Our tests doesn't lower RF upon scaling down.

I am not sure what you mean by that? Is there something else needed then just raising the default replicas?

zimnx:
> I am not sure what you mean by that? Is there something else needed then just raising the default replicas?

It's not matter of default replicas. In scaling test where we scaled cluster from 3->2->1->3 we didn't lower RF of keyspaces. It was a violation which started to be validated within Scylla.

tnozicka:
> It's not matter of default replicas. In scaling test where we scaled cluster from 3->2->1->3 we didn't lower RF of keyspaces. It was a violation which started to be validated within Scylla.

guess my point was to go say 5->4->3->4->5 with RF=3

zimnx:
> > It's not matter of default replicas. In scaling test where we scaled cluster from 3->2->1->3 we didn't lower RF of keyspaces. It was a violation which started to be validated within Scylla.
> 
> guess my point was to go say 5->4->3->4->5 with RF=3

This would lower case coverage compared to scenario with rf change and scaling down. Although core doesn't recommend running less then 3 nodes, so we may enter path where issues are expected. I'll amend node count then.

tnozicka:
> This would lower case coverage compared to scenario with rf change and scaling down. But it's not what we have now, we can always add it. I'll amend node count then.

Got it, I find the case you describe valuable as well. Please make an issue and send a PR to cover RF change as well.

zimnx:
> can you try with https://github.com/scylladb/scylla-manager/releases/tag/3.3.0-0.20240624.9dd98ab35 ?

It went well. I asked Manager team about next release, it should be out very soon.

zimnx:
Manager was released and bumped on master so this is no longer WIP.

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -128,7 +128,7 @@ authorizer: CassandraAuthorizer
 		}
 
 		q := cqlSession.Query(
-			`SELECT salted_hash FROM system_auth.roles WHERE role = ?`,
+			`SELECT salted_hash FROM system.roles WHERE role = ?`,
```
rzetelskik, line 131: Do we want to test >=6.0.0 only?

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -438,7 +441,38 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		verifyCQLData(ctx, di)
-	})
+	},
+		g.Entry("using default ScyllaDB version", entry{}),
+		// Restoring schema with ScyllaDB OS 5.4.X or ScyllaDB Enterprise 2024.1.X and consistent_cluster_management isn’t supported.
+		// This test validates a workaround explained in the docs.
```
rzetelskik, line 447: nit: ref to docs page?

rzetelskik:
> Manager was released and bumped on master so this is no longer WIP.

fwiw manager client might need a bump

zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/1525#issuecomment-2197238093
/retest

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
zimnx: added

zimnx:
> fwiw manager client might need a bump as well

API wasn't changed, so it's not required.

rzetelskik (approved):
lgtm
/assign tnozicka

zimnx:
Possible flake - https://github.com/scylladb/scylla-operator/issues/1996

/retest

tnozicka (approved):
/approve

Issue #1987, opened 2024-06-24
ci-deploy-release script should only deploy local-csi-driver if `SO_CSI_DRIVER_PATH` env var is not empty


rzetelskik:
### What happened?

ci-deploy-release script times out waiting for local-csi-driver to roll out, despite `SO_CSI_DRIVER_PATH` being empty.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-latest-e2e-gke-arm64-serial/1805169160804634624#1:test-build-log.txt%3A306

/kind bug
/priority important-soon
/assign rzetelskik

### What did you expect to happen?

local-csi-driver shouldn't be deployed and the script shouldn't wait for it to roll out.

### How can we reproduce it (as minimally and precisely as possible)?

n/a

### Scylla Operator version

master

### Kubernetes platform name and version

n/a

### Please attach the must-gather archive.

n/a

rzetelskik:
Alternatively, if we don't want to use `SO_CSI_DRIVER_PATH` there, it should skip local-csi-driver deployment if `SO_NODECONFIG_PATH` is not provided.

tnozicka:
empty string is a valid intention to skip and allows you to distinguish between forgetting to set the value and explicitly asking not to do it

rzetelskik:
> empty string is a valid intention to skip and allows you to distinguish between forgetting to set the value and explicitly asking not to do it

Looking at `ci-deploy-release` and `run-e2e-gke-release` there is no `SO_CSI_DRIVER_PATH` whatsoever so it looks like the intention was to always run it with local-csi-driver. This is wrong imo as it's not always desired, e.g. when running our serial suite.

tnozicka:
I agree, for serial suite you have to delete the nodeconfig before starting the e2e, or make this conditional

PR #1992, opened 2024-06-25
Skip CSI driver creation for empty SO_CSI_DRIVER_PATH in deploy release script

rzetelskik:
Currently ci-deploy-release script times out waiting for local-csi-driver to roll out, despite `SO_CSI_DRIVER_PATH` being empty.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-latest-e2e-gke-arm64-serial/1805169160804634624#1:test-build-log.txt%3A306

~From my understanding the intention was to always run the given revision of local-csi-driver example manifests, since, contrary to the regular deploy script, the release deploy script doesn't account for `SO_CSI_DRIVER_PATH` whatsoever.~

~Hence this PR skips local-csi-driver creation together with NodeConfig creation when empty `SO_NODECONFIG_PATH` is provided. We can always make it more granular if we need to deploy local-csi-driver without configuring a NodeConfig, but for now that seems good enough.~

Edit: local-csi-driver path can now be overridden by setting SO_CSI_DRIVER_PATH. In case the env var is empty, local-csi-driver deployment is skipped.

Resolves https://github.com/scylladb/scylla-operator/issues/1987

/kind bug
/priority important-soon
/cc tnozicka

### DIFF
--- a/hack/ci-deploy-release.sh
+++ b/hack/ci-deploy-release.sh
@@ -78,8 +78,15 @@ else
  kubectl_create -f="${SO_NODECONFIG_PATH}"
 fi
 
-kubectl_create -n=local-csi-driver -f="${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/"{00_namespace.yaml,00_scylladb-local-xfs.storageclass.yaml,10_csidriver.yaml,10_driver.serviceaccount.yaml,10_provisioner_clusterrole.yaml,20_provisioner_clusterrolebinding.yaml,50_daemonset.yaml}
-kubectl -n=local-csi-driver rollout status --timeout=5m daemonset.apps/local-csi-driver
+if [[ -z "${SO_CSI_DRIVER_PATH+x}" ]]; then
+  kubectl_create -n=local-csi-driver -f="${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/"{00_namespace.yaml,00_scylladb-local-xfs.storageclass.yaml,10_csidriver.yaml,10_driver.serviceaccount.yaml,10_provisioner_clusterrole.yaml,20_provisioner_clusterrolebinding.yaml,50_daemonset.yaml}
+  kubectl -n=local-csi-driver rollout status --timeout=5m daemonset.apps/local-csi-driver
+elif [[ -n "${SO_CSI_DRIVER_PATH}" ]]; then
+  kubectl_create -n=local-csi-driver -f="${SO_CSI_DRIVER_PATH}"
+  kubectl -n=local-csi-driver rollout status --timeout=5m daemonset.apps/local-csi-driver
+else
+  echo "Skipping CSI driver creation"
+fi
 
 mkdir "${ARTIFACTS}/manager"
 cat > "${ARTIFACTS}/manager/kustomization.yaml" << EOF


hack/ci-deploy-release.sh
```diff
@@ -73,13 +73,13 @@ kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/scylla-op
 kubectl -n=scylla-operator rollout status --timeout=5m deployment.apps/webhook-server
 
 if [[ -z "${SO_NODECONFIG_PATH:-}" ]]; then
- echo "Skipping NodeConfig creation"
+ echo "Skipping NodeConfig and CSI driver creation"
 else
  kubectl_create -f="${SO_NODECONFIG_PATH}"
-fi
```
tnozicka, line 79: I don't think SO_NODECONFIG_PATH should affect what happens with local-csi-driver. Please set up a dedicated var.

rzetelskik: Should it allow for overriding the local-csi-driver path though (so `SO_CSI_DRIVER_PATH` like in the regular deploy script)? Or just for skipping it?

tnozicka: `SO_CSI_DRIVER_PATH` defaulting to the URL seems reasonable

rzetelskik: done

hack/ci-deploy-release.sh
```diff
@@ -78,8 +78,13 @@ else
  kubectl_create -f="${SO_NODECONFIG_PATH}"
 fi
 
-kubectl_create -n=local-csi-driver -f="${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/"{00_namespace.yaml,00_scylladb-local-xfs.storageclass.yaml,10_csidriver.yaml,10_driver.serviceaccount.yaml,10_provisioner_clusterrole.yaml,20_provisioner_clusterrolebinding.yaml,50_daemonset.yaml}
-kubectl -n=local-csi-driver rollout status --timeout=5m daemonset.apps/local-csi-driver
+SO_CSI_DRIVER_PATH="${SO_CSI_DRIVER_PATH=${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/}"
```
tnozicka, line 81: URLs can't list directories

tnozicka (changes requested):

hack/ci-deploy-release.sh
tnozicka, line 81:
```suggestion
SO_CSI_DRIVER_PATH="${SO_CSI_DRIVER_PATH=${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/}"{00_namespace.yaml,00_scylladb-local-xfs.storageclass.yaml,10_csidriver.yaml,10_driver.serviceaccount.yaml,10_provisioner_clusterrole.yaml,20_provisioner_clusterrolebinding.yaml,50_daemonset.yaml}
```

rzetelskik: fixed, although with a use of `eval`. Let me know if this is fine

rzetelskik: > even 2 kubectl apply statements are fine

Done, I couldn't find a different way to do this (other than eval).

hack/ci-deploy-release.sh
```diff
@@ -78,8 +78,17 @@ else
  kubectl_create -f="${SO_NODECONFIG_PATH}"
 fi
 
-kubectl_create -n=local-csi-driver -f="${source_url}/${revision}/examples/common/local-volume-provisioner/local-csi-driver/"{00_namespace.yaml,00_scylladb-local-xfs.storageclass.yaml,10_csidriver.yaml,10_driver.serviceaccount.yaml,10_provisioner_clusterrole.yaml,20_provisioner_clusterrolebinding.yaml,50_daemonset.yaml}
-kubectl -n=local-csi-driver rollout status --timeout=5m daemonset.apps/local-csi-driver
+if [[ -z "${SO_CSI_DRIVER_PATH:-}" && -n "${SO_CSI_DRIVER_PATH+x}" ]]; then
+  echo "Skipping CSI driver creation"
+else
+  if [[ -z "${SO_CSI_DRIVER_PATH+x}" ]]; then
```
tnozicka, line 84: i think you actually need just 2 comparisons
```
if [[ -z "${SO_CSI_DRIVER_PATH+x}" ]]; then
...
else if [[ -n "${SO_CSI_DRIVER_PATH}" ]]; then
...
else
...
fi
```

rzetelskik: that's just a tradeoff between the nested check and duplicating the waiting for rollout line

tnozicka: it avoiding testing the same condition `"${SO_CSI_DRIVER_PATH+x}"` twice in the same if block, so the branching is easier to read and evaluate

rzetelskik: I don't mind either way, changed

tnozicka (approved):
/approve
/lgtm
thanks

PR #1993, opened 2024-06-25
Mark v1.13 as stable version in documentation

zimnx:
After v1.13.0 is called GA, we should switch `stable` endpoint of documentation to point to v1.13 docs.


tnozicka (approved):
/approve
/lgtm

PR #1994, opened 2024-06-26
Add proposal for new API `v1alpha1.ScyllaDBDatacenter`

### LINKED ISSUE #1978: Create proposal for ScyllaDBDatacenter/v1alpha1 API
Over the years there was several things we could have done better while designing our ScyllaCluster API, some even contradict kubernetes API conventions. Even the API itself truly represents only a single datacenter, not the whole cluster. This bring in some confusion we should fix as well.

Edited:

We should create new API called `ScyllaDBDatacenter/v1alpha1` in the same API group (`scylla.scylladb.com`) that can take over the implementation logic for handling a data center and the old `ScyllaCluster/v1` can create this resource on the background as an implementation detail. This means there will a a conversion path for `ScyllaCluster/v1` into `ScyllaDBDatacenter/v1alpha1`.

The new API must comply with https://github.com/kubernetes/community/blob/ecb8918/contributors/devel/sig-architecture/api-conventions.md

zimnx:
This proposal aims to introduce new API called `v1alpha1.ScyllaDBDatacenter` that will replace existing `v1.ScyllaCluster`. The new API is going to be backward compatible and will contain fixes for mistakes we did in the past.

Resolves #1978


zimnx:
Marking as WIP after offline discussion to prepare API for new changes.

rzetelskik (commented):
Left a couple of comments.
Please review the code snippets for inconsistent indentations, there's a couple of places where they're off.

Also I haven't noticed any mentions of a deprecation plan. Is the existing CRD going to be deprecated at some point? If so, how long are we going to keep it, when are we going to get rid of the conversion controller etc. Otherwise it should be mentioned in the doc either way imo.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+
+- Introduce new features alongside new API.
+
+## Proposal
+
+I propose to add new `v1alpha1.ScyllaDBDatacenter` to the existing `scylla.scylladb.com` API group.
+Existing controllers reconciling `v1.ScyllaCluster` are going to switch to reconcile new API, 
+while new controller is going to convert every existing `v1.ScyllaCluster` and all changes  into `v1alpha1.ScyllaDBDatacenter`.
+This will allow existing users to continue using API they are used to. 
+By changing every mention of `v1.ScyllaCluster` in our documentation and examples, we will recommend using new API.
```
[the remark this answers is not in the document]
zimnx: After we are backward compatible. The only missing piece is Manager integration which requires proposal for defining Repair/Backup/Restore Task CRDs and handling of them.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+When rolled back, changes to `v1alpha1.ScyllaDBDatacenters` are not going to be reconciled, users have to continue using `v1.ScyllaCluster`.
+
+## Implementation History
+
+26.06.204 - Initial version
+
+## Drawbacks
+
+Such change is not something we couldn't live with, but it helps to reduce the ambiguity of naming mistakes, 
+as well as reduce the technical dept.
```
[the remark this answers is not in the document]
zimnx: Being unnecessary is a drawback, but there's a motivation behind it.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+
+Not known.
+
+## Design Details
+
+### Minor changes
+The following list of minor issues is going to be fixed in the new API. Major ones are described in separate paragraphs.
+- `v1.ScyllaCluster` provides a single datacenter spec, while ScyllaDB can be deployed in multiple datacenters. 
+  Having `Cluster` in name is wrong, as this resource represents only a single DC. 
+  Higher level abstractions on top, will allow for deploying multiple datacenters.
```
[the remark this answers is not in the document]
zimnx: There's nothing to cross reference. This is an idea, that such single DC API can be composed into higher level abstraction, instead transforming ScyllaCluster into multi datacenter one.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+
+```
+
+### Uniform datacenter
+
+The current flexibility of having non-uniform racks will be preserved.
+
+ScyllaDB recommendation is to have uniform racks.
+Current API required from users to provide resources on every rack.
+There are two characteristics that make racks uniform: number of nodes and allocated resources.
```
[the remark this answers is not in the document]
zimnx: It's described it consists of two properties, then these properties are discussed and after a separator there's an example of such API. What you would like to improve?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+Rack level settings override ones set on datacenter level.
+
+#### Node count
+
+New field `nodes` will be added on datacenter level. Scylla Operator will balance this number between all defined racks.
+Having single place where desired number of nodes is specified, enables us for having a support for scale subresource,
+which brings users a handy scaling operations using `kubectl scale` and pave a path towards available kubernetes autoscalers.
+The downside of it that these won't be able to add more racks, but provides a nice way of scaling for static topology clusters.
+
+Datacenter level `nodes` cannot be used together with `nodes` specified per rack. Webhook will prevent it.
```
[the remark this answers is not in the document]
zimnx: I think place for explaining any ambiguity is description which is also part of the API. What do you suggest?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+
+To prevent misuse of this simplification keys of provided label selector will be validated against `topology.kubernetes.io/*` rule, 
+and usage of both `nodeAffinity` and `topologyLabelSelector` will be forbidden by the webhook.
+Users wanting to construct more complex placement rules will have to do it using `placement` spec.
+
+### Resource model
+
+Storage `capacity` is going to be merged into `resources` field under `scyllaDB`, but other settings like metadata
+and storage class name, will stay in place. 
+This makes it to be in sync on how PVC template is defined.
```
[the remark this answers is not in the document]
zimnx: Most of the changes done in this proposal are about simplifying and making UX better. Simple case == simple CRD representation. If user wants to have more advanced setup, then there're fields for these. StorageClass is not required because user may want to use the default one which is not something uncommon, while capacity is required, the same as cpu and memory.
We can't support full volumeClaimTemplate (name, access mode, selector, volumeName, volumeMode...) so I don't think we should go that way.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
rzetelskik: > What you would like to improve?

It just doesn't really say anywhere that this proposal introduces a API for unified racks (other than the name of this section). It just says that the current API will be preserved and that ScyllaDB recommends to have unified racks and what makes racks uniform, so I was reading it waiting for the actual proposed changed but it was never there - unless I missed it.

So I'd say adding even something as simple as "ScyllaDB recommendation is to have uniform racks.. I propose to extend the API so that the racks can be defined uniformly in a simplified manner, while preserving the existing flexibility of having non-uniform racks by overriding dedicated specifications" would make this clearer.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+and what's in the CRD, hence it cannot be simply mapped one-to-one.
+New helm chart called `scylladbdatacenter` will simply copy `spec` from `values` to a new CRD solving this issue making it 
+easily extensible, while keeping schema of CRD and helm schema in sync.
+
+### Scylla Manager integration
+
+Scylla Manager tasks do not fit into single datacenter, as Manager tasks work on the entire cluster. 
+These are removed from the spec and status of the new API until a better alternative for these is available.
+Existing controller reconciling these will continue using `v1.ScyllaCluster`, users wanting to use these, 
+won't be able to transition to new API until an extension is available.
```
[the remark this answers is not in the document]
zimnx: added a mention about it.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+  This is unnecessary when there is documentation about performance tuning containing these recommendations.
+- `hostNetworking` support is discontinued and setting it becomes deprecated and discouraged by not having a field to set it in API. 
+  To preserve backward compatibility controllers will support enabling it by setting special annotation 
+  on `v1alpha1.ScyllaDBDatacenter`. 
+  Controller migrating from `v1.ScyllaCluster` to `v1alpha1.ScyllaDBDatacenter` will set this annotation when required.
+
+### Helm chart
+
+Existing `scylla` helm chart is going to use existing `v1.ScyllaCluster`. It has several differences between what's in CRD
+and what's in the CRD, hence it cannot be simply mapped one-to-one.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+- `v1.ScyllaCluster` provides a single datacenter spec, while ScyllaDB can be deployed in multiple datacenters. 
+  Having `Cluster` in name is wrong, as this resource represents only a single DC. 
+  Higher level abstractions on top, will allow for deploying multiple datacenters.
+- Some time ago, the name of a database we are managing was changed to ScyllaDB. Resources should reflect that.
+  Similarly, references to Scylla Manager Agent are renamed to ScyllaDB Manager Agent, even though naming change isn't completed.
+- Instead, having a single field specifying container image of used components,
+  there are two fields, one for the repository and the second for version. 
+  This unnecessary split is error-prone.
+- Kubernetes API Convention suggests using arrays instead of maps. The status of racks doesn't follow that rule.
+- `cpuSet` setting is removed, as it only controls stricter validation of allocated resources.
```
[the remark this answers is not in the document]
zimnx: Aligned

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,741 @@
...
+## Design Details
+
+### Minor changes
+The following list of minor issues is going to be fixed in the new API. Major ones are described in separate paragraphs.
+- `v1.ScyllaCluster` provides a single datacenter spec, while ScyllaDB can be deployed in multiple datacenters. 
+  Having `Cluster` in name is wrong, as this resource represents only a single DC. 
+  Higher level abstractions on top, will allow for deploying multiple datacenters.
+- Some time ago, the name of a database we are managing was changed to ScyllaDB. Resources should reflect that.
+  Similarly, references to Scylla Manager Agent are renamed to ScyllaDB Manager Agent, even though naming change isn't completed.
+- Instead, having a single field specifying container image of used components,
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: removed

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
rzetelskik: Maybe we could separate the general fields and fields uniformly applying to racks by moving the latter to a shared `template`/`rackTemplate` struct?

```
template:
  scylladb:
    image: docker.io/scylladb/scylla:6.0.1
    resources:
      requests:
        cpu: 1
        memory: 4Gi
        storage: 100Gi
  scyllaDBManagerAgent:
    image: docker.io/scylladb/scylla-manager-agent:3.3.0
    resources:
      requests:
        cpu: 100m
        memory: 100Mi
nodes: 9
racks:
- ...
```

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
tnozicka: > while capacity is required, the same as cpu and memory.

I don't think that's on us to decide, if regular Pods run without it we shouldn't add more obstacles, unless it breaks without it.

E.g. You can set defaults for these with LimitRanges - https://kubernetes.io/docs/concepts/policy/limit-range/

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: I see what you mean now. Added statement you suggested.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,721 @@
...
+A new helm chart called `scylladbdatacenter` will simply copy `spec` from `values` to a new CRD solving this issue and making it
+easily extensible, while keeping the schema of CRD and helm schema in sync.
+
+### Scylla Manager integration
+
+Scylla Manager tasks do not fit into a single datacenter, as Manager tasks work on the entire cluster.
+These are removed from the spec and status of the new API until a better alternative for these is available.
+The existing controller reconciling these will continue using `v1.ScyllaCluster`, users wanting to use these,
+won't be able to transition to the new API until an extension is available.
+It's going to be fixed in another proposal rethinking the whole integration.
```
tnozicka, line 82:
```suggestion
Extending the manager integration may be addressed in a future proposal that rethinks the overall approach in more detail.
```

tnozicka (commented):
Thanks for sending the proposal. A few comments to start with, I'll continue tomorrow.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,744 @@
+# Replace ScyllaCluster API with ScyllaDBDatacenter API
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,744 @@
...
+
+## Summary
+
+This proposal aims to introduce a new API called `v1alpha1.ScyllaDBDatacenter` that will replace the existing `v1.ScyllaCluster`.
+The new API is going to be backward compatible and will contain fixes for mistakes we made in the past.
```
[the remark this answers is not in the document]
zimnx: I meant feature compatibility. Mentioned it.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Non-Goals
+
+- Introduce new features alongside new API.
+
+## Proposal
+
+I propose to add a new `v1alpha1.ScyllaDBDatacenter` to the existing `scylla.scylladb.com` API group.
+Existing controllers reconciling `v1.ScyllaCluster` are going to switch to reconcile the new API,
+while the new controller is going to convert every existing `v1.ScyllaCluster` and all changes into `v1alpha1.ScyllaDBDatacenter`.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,721 @@
...
+This will allow existing users to continue using the API they are used to.
+By changing every mention of `v1.ScyllaCluster` in our documentation and examples, we will recommend using a new API.
```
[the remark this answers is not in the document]
zimnx: added mention about additional documentation pages and examples

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,721 @@
...
+
+### User Stories
+
+#### Existing users of `v1.ScyllaCluster`
+
+As a user of `v1.ScyllaCluster` I want to continue using it. This change must not interrupt anyhow.
+
+#### New users
+
+As a new user, I should learn about `v1alpha1.ScyllaDBDatacenter` from examples and documentation and have support for
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,721 @@
...
+- Some time ago, the name of a database we are managing was changed to ScyllaDB. Resources should reflect that.
+  Similarly, references to Scylla Manager Agent are renamed to ScyllaDB Manager Agent, even though the naming change isn't completed.
+- Instead of having a single field specifying the container image of used components,
+  there are two fields, one for the repository and the second for the version.
+  This unnecessary split is error-prone.
+- Kubernetes API Convention suggests using arrays instead of maps. The status of racks doesn't follow that rule.
+- `cpuSet` is confusing, as it doesn't change anything related to cpuset, it only controls stricter validation of allocated resources.
+  This is unnecessary when there is documentation about performance tuning containing these recommendations.
+  This field will be removed and validation will be relaxed.
+- We initially thought that `hostNetworking` helps with performance, but later we discovered we were wrong.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,721 @@
...
+  As we only see the downsides of using it, `hostNetworking` support will be discontinued, and setting it becomes deprecated
+  and discouraged by not having a field to set it in API.
+  To preserve backward compatibility controllers will support enabling it by setting special annotation
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,721 @@
...
+
+The existing `scylla` helm chart will use the existing `v1.ScyllaCluster`. It has several differences between what's in CRD
+and what's in the chart templates, hence it cannot be simply mapped one-to-one.
+A new helm chart called `scylladbdatacenter` will simply copy `spec` from `values` to a new CRD solving this issue and making it
+easily extensible, while keeping the schema of CRD and helm schema in sync.
+
+### Scylla Manager integration
+
+Scylla Manager tasks do not fit into a single datacenter, as Manager tasks work on the entire cluster.
+These are removed from the spec and status of the new API until a better alternative for these is available.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+    // +optional
+    Labels map[string]string `json:"labels,omitempty"`
+
+    // annotations is a custom key value map that gets merged with managed object annotations.
+    // +optional
+    Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// ScyllaDBOverrides allows to override a subset of ScyllaDB settings.
+type ScyllaDBOverrides struct {
```
tnozicka, line 284: This struct kind of requires us to define the racks and partially options on ScyllaDBDatacenterSpec again and let only certain fields be adjusted per rack. Maybe even the image and others could be adjusted per rack for blue-green updates or testing / debugging.

I am wondering whether we could somehow define this once and use it with the overrides again. Like by defining a RackBase struct that would be embedded at the spec level and also be able to optionally set in for each rack. Something like "defaultRack" and then defining the array of names and topology label selectors unless you want to adjust it.

Slightly related idea that has been lingering in my head for quite a while is to create a `ScyllaDBRack` to split the API responsibilities and controllers, allow API based adjustments and possibly manage Pods directly at some point, if needed. It may be a step `#2` but it would also give us `RackTemplateSpec` to use.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+    Image string `json:"image"`
+
+    // alternatorOptions designates this cluster an Alternator cluster.
+    // +optional
+    AlternatorOptions *AlternatorOptions `json:"alternatorOptions,omitempty"`
+
+    // scyllaArgs will be appended to the ScyllaDB binary during startup.
+    // When set, ScyllaDB may behave unexpectedly, and every such setup is considered unsupported.
+    // +optional
+    ScyllaArgs []string `json:"scyllaArgs,omitempty"`
```
tnozicka, line 358:
```suggestion
    AdditionalScyllaArgs []string `json:"additionalScyllaArgs,omitempty"`
```

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+    // +optional
+    Racks []RackSpec `json:"racks"`
+
+    // automaticOrphanedNodeCleanup controls if automatic orphan node cleanup should be performed.
+    AutomaticOrphanedNodeCleanup bool `json:"automaticOrphanedNodeCleanup,omitempty"`
+
+    // sysctls holds the sysctl properties to be applied during initialization given as a list of key=value pairs.
+    // Example: fs.aio-max-nr=232323
+    // +optional
+    Sysctls []string `json:"sysctls,omitempty"`
```
tnozicka, line 243: racks/zones can have different machine types - maybe this should be on the rack level - i guess a similar point can be made for some other fields on this level

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+}
+```
+
+New Condition will be added to the `status` of `v1alpha1.ScyllaDBDatacenter`, providing a simple way of noticing rack imbalance.
+Its type is going to be `UnbalancedRacks` and the value of it will be reconciled accordingly to the state of rack imbalance.
+
+--- 
+
+An example of nine node cluster having three racks with uniform sizes and resources.
+```yaml
```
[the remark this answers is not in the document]
tnozicka: would be good to show it with topology selector here, or somewhere else

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: to make the naming in the above more meaningfull, how about something like
```yaml
spec:
  rackTemplate:
    scylladb:
      resources:    
     ...
  racks:
  - name: foo
  - name: bar
    rackTemplate:
      scylladb:
        resources:
```

rzetelskik: I had similar remarks and we discussed them with @zimnx yesterday.  In my opinion there shouldn't be a split between the template and per-rack level (in other words the overrides shouldn't be narrower) - seems like we agree on this one. To add to this comment though, you'd still need global values like the proposed `nodes` (https://github.com/scylladb/scylla-operator/pull/1994#discussion_r1662331391) which don't fall into the `rackTemplate`.
So to cover the datacenter-level fields you could extend it with something like:

```
spec:
  scylladb:
    nodes:
    enableDeveloperMode:
    ...
  rackTemplate:
    scylladb:
      resources:    
     ...
  racks:
  - name: foo
  - name: bar
    rackTemplate:
      scylladb:
        resources:
```

tnozicka: I mostly agree but `nodes` shouldn't be under scylladb (the apply to manager agent / the whole Pod equally

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
rzetelskik: let's discuss this under https://github.com/scylladb/scylla-operator/pull/1994#discussion_r1664573792

enhancements/proposals/1978-scylladbdatacenter-api/README.md
rzetelskik: > the apply to manager agent / the whole Pod equally

yes, but only because its implied by the number of Scylla nodes. I'm not saying it has to be under `scylladb`, but that's not a reason not to imo

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+)
+
+// ScyllaDBDatacenterSpec defines the desired state of ScyllaDBDatacenter.
+type ScyllaDBDatacenterSpec struct {
+    // metadata controls shared metadata for all pods created based on this spec.
+    // +optional
+    Metadata *ObjectTemplateMetadata `json:"metadata,omitempty"`
+
+    // datacenterName is the name of the scylla datacenter. Used as datacenter name in GossipingPropertyFileSnitch.
+    DatacenterName string `json:"datacenterName"`
```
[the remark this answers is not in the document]
zimnx: Imo yes, otherwise it would be confusing as datacenter name is propagated into scylla and visible to the user. In such case, you couldn't have two `us-east1` datacenters in the same namespace, even though they could be part of different cluster.

But it raised an important issue we need to solve as well. There's also a `cluster_name` property of Scylla, which should be in sync between datacenters in order for two dcs to join each other. So far, we used `scyllacluster.metadata.name` for it, but since we changed the name of CRD, it doesn't make sense anymore. 

Maybe we should have a dedicated field for all three (cluster, dc, rack), wdyt?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+The existing controller reconciling these will continue using `v1.ScyllaCluster`, users wanting to use these,
+won't be able to transition to the new API until an extension is available.
+Extending the manager integration may be addressed in a future proposal that rethinks the overall approach in more detail.
+
+### Topology placement
+
+Currently, to provide placement rules for a rack, users have to provide a very lengthy `nodeAffinity`.
+Even the simplest rules require providing 10'ish lines with 4'ish indentation levels.
+This makes the entire resource huge and UX isn't great.
+To improve it, the new field `toplogyLabelSelector` added on the datacenter and rack level,
```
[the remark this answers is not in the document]
zimnx: I planned to have two fields for topology label selector, one on DC level and second on rack level. Having a prefix for both of these might be confusing, as they are going to be merged together which is mentioned below.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+
+    // datacenterName is the name of the scylla datacenter. Used as datacenter name in GossipingPropertyFileSnitch.
+    DatacenterName string `json:"datacenterName"`
+
+    // scyllaDB holds a specification of ScyllaDB.
+    ScyllaDB ScyllaDB `json:"scyllaDB"`
+
+    // scyllaManagerAgent holds a specification of ScyllaDB Manager Agent.
+    // +optional
+    ScyllaDBManagerAgent *ScyllaDBManagerAgent `json:"scyllaDBManagerAgent,omitempty"`
```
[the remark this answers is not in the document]
zimnx: I haven't, but this need to be in sync with manager version. We could prepare API and mark agent image as optional and required validation in the webhook until we implement default image.  I'll do it.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
tnozicka: I think this makes sense. Making them optional with defaults based on the metadata.name for the simple cases would make a nice UX.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: I haven't look deep enough to see whether some sysctls make sense per container, but I see that the majority should be per node. So maybe this would go away with https://github.com/scylladb/scylla-operator/issues/749. I agree that internal annotation let's us defer the decision. (Btw. it would be good to add all those reasons to the annotation const definitions, so we know when we can drop it.)

tnozicka: I suppose we should file issues to track deprecation of fields that you are moving to annotation, unless they already have one like hostNetworking.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
tnozicka: This also depends under which key this ends up being, https://github.com/scylladb/scylla-operator/pull/1994#discussion_r1664573792 so let's defer this

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+  This unnecessary split is error-prone.
+- Kubernetes API Convention suggests using arrays instead of maps. The status of racks doesn't follow that rule.
+- `cpuSet` is confusing, as it doesn't change anything related to cpuset, it only controls stricter validation of allocated resources.
+  This is unnecessary when there is documentation about performance tuning containing these recommendations.
+  This field will be removed and validation will be relaxed.
+- We initially thought that `hostNetworking` helps with performance, but later we discovered we were wrong.
+  As we only see the downsides of using it, `hostNetworking` support will be discontinued, and setting it becomes deprecated
+  and discouraged by not having a field to set it in API. [#1378](https://github.com/scylladb/scylla-operator/issues/1378)
+  To preserve backward compatibility controllers will support enabling it by setting special 
+  annotation `scylla-operator.scylladb.com/enable-host-networking` to `true` on `v1alpha1.ScyllaDBDatacenter`.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: done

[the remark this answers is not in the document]
zimnx: Since it has different behavior than everything specified in rackTemplate (it's not overwriten but merged), I propose to keep it separate from the template.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+Extending the manager integration may be addressed in a future proposal that rethinks the overall approach in more detail.
+
+### Topology placement
+
+Currently, to provide placement rules for a rack, users have to provide a very lengthy `nodeAffinity`.
+Even the simplest rules require providing 10'ish lines with 4'ish indentation levels.
+This makes the entire resource huge and UX isn't great.
+To improve it, the new field `toplogyLabelSelector` added on the datacenter and rack level,
+provides a label selector which will be translated into placement nodeAffinity rules selecting nodes matching these labels.
+Label selectors from the datacenter and rack level will be merged into the final node affinity.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+
+To prevent misuse of this simplification keys of the provided label selector will be validated against the `topology.kubernetes.io/*` rule,
```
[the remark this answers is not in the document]
zimnx: Huge benefit of standardization are nice integrations. This field is meant for topology constraints, kuberetes has these [reserved and documented](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion), so everyone should follow. But I don't mind relaxing it.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+and usage of both `nodeAffinity` and `topologyLabelSelector` will be forbidden by the webhook.
+Users wanting to construct more complex placement rules will have to do it using `placement` spec.
+
+### Uniform datacenter
+
+ScyllaDB recommendation is to have uniform racks. 
+I propose to extend the API so that the racks can be defined uniformly in a simplified manner, 
+while preserving the existing flexibility of having non-uniform racks by overriding dedicated specifications.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+Two characteristics make racks uniform: the number of nodes and allocated resources.
```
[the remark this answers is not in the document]
zimnx: Made all of them to be shared within the template

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+    AvailableCondition   = "Available"
+    ProgressingCondition = "Progressing"
+    DegradedCondition    = "Degraded"
+)
+
+// ScyllaDBDatacenterSpec defines the desired state of ScyllaDBDatacenter.
+type ScyllaDBDatacenterSpec struct {
+    // metadata controls shared metadata for all pods created based on this spec.
+    // +optional
+    Metadata *ObjectTemplateMetadata `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,724 @@
...
+    ScyllaDBManagerAgent *ScyllaDBManagerAgent `json:"scyllaDBManagerAgent,omitempty"`
+
+    // imagePullSecrets is an optional list of references to secrets in the same namespace
+    // used for pulling any images used by this spec.
+    // +optional
+    ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+    // network holds network configuration.
+    // +optional
+    Network *Network `json:"network,omitempty"`
```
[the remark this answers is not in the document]
zimnx: removed nesting

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: Once we have everything finalized I will create issues for missing parts

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+To prevent misuse of this simplification keys of the provided label selector will be validated against the `topology.kubernetes.io/*` rule,
+and usage of both `nodeAffinity` and `topologyLabelSelector` will be forbidden by the webhook.
+Users wanting to construct more complex placement rules will have to do it using `placement` spec.
+
+Given following spec:
+```yaml
+spec:
+  topologyLabelSelector:
+    topology.kubernetes.io/region: us-east1
```
tnozicka, line 108: nit: region is in most cases identical for all nodes in the cluster and is already selected by the zone. At this level I'd primarily set node pool labels selector as an example.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+Having a single place where the desired number of nodes is specified, enables us to have support for scale subresource,
+which brings users handy scaling operations using `kubectl scale` and paves a path towards available Kubernetes autoscalers.
+The downside of it is that these won't be able to add more racks, but provides a nice way of scaling for static topology clusters.
+
+Datacenter level `nodes` cannot be used together with `nodes` specified per rack. Webhook will prevent it.
+
+The balancing algorithm is iterative, adding/removing one node at a time in each iteration to ensure that racks stay as balanced as possible.
+
+```go
+func balanceNodesAcrossRacks(racks []*appsv1.StatefulSet, totalNodes int) []*appsv1.StatefulSet {
```
tnozicka, line 154: Why do we need this?

Making this depend on the current state kind of beats the point of indempotent desired configs. The racks are ordered and they should always have their scale grown so that `nodes(N)>=nodes(N+1)` which will produce the same desired sub-scales, no matter the state.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+    }
+    
+    // Remove a node from biggest rack.
+    maxRackIndex := findBiggestStatefulSet(racks)
+    racks[maxRackIndex].Replicas--
+    return balanceNodesAcrossRacks(racks, totalNodes)
+}
+```
+
+New Condition will be added to the `status` of `v1alpha1.ScyllaDBDatacenter`, providing a simple way of noticing rack imbalance.
```
tnozicka, line 180: what does imbalanced mean here? I assume even in a normal state when `numNodes%numRacks!=0` there will be up to `-1` difference, is that also to be considered to be imbalanced?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+// ScyllaDBDatacenterSpec defines the desired state of ScyllaDBDatacenter.
+type ScyllaDBDatacenterSpec struct {
+	// metadata controls shared metadata for all pods created based on this spec.
+	// +optional
+	Metadata ObjectTemplateMetadata `json:"metadata,omitempty"`
+
+	// clusterName specifies the name of the scylla cluster.
+	// When joining two DCs, their cluster name must match.
+	ClusterName string `json:"clusterName"`
```
tnozicka, line 238: please describe where the default comes from

tnozicka: optional+omitEmpty

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+	// datacenterName specifies the name of the scylla datacenter. Used as datacenter name in GossipingPropertyFileSnitch.
+	// If empty, it's taken from the 'scylladbdatacenter.metadata.name'.
+	DatacenterName string `json:"datacenterName"`
```
tnozicka, line 242: optional+omitEmpty

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+	ExposeOptions *ExposeOptions `json:"exposeOptions,omitempty"`
+
+	// nodes specify desired total number of nodes in this datacenter.
+	// This number will be balanced out across all specified racks.
+	// Cannot be used together with `nodes` specified within any rack spec.
+	// +optional
+	Nodes *int32 `json:"nodes"`
+
+	// topologyLabelSelector specifies a label selector which will be used to target nodes at specified topology constraints.
+	// Datacenter topologyLabelSelector is merged with rack topologyLabelSelector and then converted into nodeAffinity
```
tnozicka, line 280: this is now also in rackTemplate which means 4 fields would have to be merged to produce the affinities which may make it a bit complex

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+	// annotations is a custom key value map that gets merged with managed object annotations.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+type RackTemplate struct {
+	// nodes specifies desired number of nodes in rack.
+	// +optional
+	Nodes *int32 `json:"nodes"`
```
tnozicka, line 334: it would be good to describe somewhere how `.nodes`, `rackTemplate.nodes` and `racks[].nodes` merged and which combinations are forbidden

enhancements/proposals/1978-scylladbdatacenter-api/README.md
rzetelskik: > they should always have their scale grown so that nodes(N)>=nodes(N+1)

I don't think that's a sufficient condition. If you want to distribute 7 nodes over 3 racks both [3,3,1] and [3,2,2] satisfy it, but the latter would be desired, I suppose. I.e. you also want to minimise the pair-wise difference of rack sizes imo.
Having said that, this algorithm seems way over-complicated. Isn't it just `racks[i] = N/len(racks) + (i < N mod len(racks))`?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
rzetelskik: my understanding is it just means state != desiredState (so in terms of the above func `sts != balanceNodesAcrossRacks(sts.racks, sts.nodes)` . But isn't that just progressing either way?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+- Introduce new features alongside new API.
+
+## Proposal
+
+I propose to add a new `v1alpha1.ScyllaDBDatacenter` to the existing `scylla.scylladb.com` API group.
+Existing controllers reconciling `v1.ScyllaCluster` are going to switch to reconcile the new API,
+and we'll add a controller that creates a backing `v1alpha1.ScyllaDBDatacneter` for every existing `v1.ScyllaCluster`.
+This will allow existing users to continue using the API they are used to.
+Examples and additional documentation pages will mention new API
+which will bring these improvements to more adventurous users.
```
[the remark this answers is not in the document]
tnozicka: I don't want to duplicate our docs, so my vote is for a single page and then a hard switch when it gets promoted

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: after talking with Maciek on a call this was meant as node count or resources don't equal, even with n-1 difference - we may need to extend the nomenclature, but let's fist agreen on the distribution algorithm

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: you are right that I've left the definition a bit ambiguous (I have talked about at most 1 difference somewhere else), it should have been `nodes(N+1) <= nodes(N) <= nodes(N+1)+1` - effectively you'd distribute the node count to each rack  by 1 node, in order, cyclically

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: change to additionalScyllaDBArguments

enhancements/proposals/1978-scylladbdatacenter-api/README.md
rzetelskik: here's the code in go https://go.dev/play/p/w0lZMWN_w-N since that notation wasn't clear

tnozicka: to merge the ideas above into an algorithm, how about this?
```golang
for i := range sd.spec.racks {
    sd.spec.racks[i].nodes = len(sd.spec.nodes) / len(sd.spec.racks)
    if i < len(sd.spec.nodes) % len(sd.spec.racks) {
        sd.spec.racks[i].nodes++
    }
}
```

zimnx: > Why do we need this?
> Making this depend on the current state kind of beats the point of indempotent desired configs. The racks are ordered and they should always have their scale grown so that nodes(N)>=nodes(N+1) which will produce the same desired sub-scales, no matter the state.

Relying on existing state is important because when recommended NetworkTopologyStrategy is used by the keyspace, Scylla tries to put a copy of entire dataset on every rack. Which means, scaling down the rack when cluster is growing can cause that rack won't have enough capacity to hold the copy of the data. 

> here's the code in go https://go.dev/play/p/w0lZMWN_w-N since that notation wasn't clear

Your algorithm produce `[3,2,2]` for nodes=7, racks=3, while `[2,2,2,1]` for nodes=7, racks=4 which is not what we would want, as when cluster is growing, it shouldn't scale down any rack due to above.

What my algorithm would produce is `[3,2,2,0]` which is what I think is expected in this case. 

---

Another important topic we haven't though about is decommissioning a single rack. Right now we forbid removing a rack until it's first scaled down to 0. With balancing algorithm and inability to override desired node count for particular rack this seems impractical.

Given that we disagree about the desired behavior for scaling up, and we haven't though well entire topology change topic I propose to push this feature out of this proposal and prepare next one focusing on this particular change. 
We should consult Core team about what should we do in these cases, describe and understand every scenario.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: As discussed on the meeting, I was afraid it's immutable and indeed it is.
Scylla refuses to boot when cluster name is changed:
```
ERROR 2024-07-11 15:37:31,392 [shard 0:main] init - Startup failed: exceptions::configuration_exception (Saved cluster name a2x2j8q69 != configured name a6hxxm6c8)
```

So this has to be required.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+- `v1.ScyllaCluster` provides a single datacenter spec, while ScyllaDB can be deployed in multiple datacenters.
+  Having `Cluster` in the name is wrong, as this resource represents only a single DC.
+- Some time ago, the name of a database we are managing was changed to ScyllaDB. Resources should reflect that.
+  Similarly, references to Scylla Manager Agent are renamed to ScyllaDB Manager Agent, even though the naming change isn't completed.
+- Instead of having a single field specifying the container image of used components,
+  there are two fields, one for the repository and the second for the version, which prevents from using 
+  full format of container images [#821](https://github.com/scylladb/scylla-operator/issues/821).
+  This unnecessary split is error-prone.
+- Kubernetes API Convention suggests using arrays instead of maps. The status of racks doesn't follow that rule.
+- `cpuSet` is confusing, as it doesn't change anything related to cpuset, it only controls stricter validation of allocated resources.
```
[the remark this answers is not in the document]
zimnx: added mention

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+  This field will be removed and validation will be relaxed.
+- We initially thought that `hostNetworking` helps with performance, but later we discovered we were wrong.
+  As we only see the downsides of using it, `hostNetworking` support will be discontinued, and setting it becomes deprecated
+  and discouraged by not having a field to set it in API. [#1378](https://github.com/scylladb/scylla-operator/issues/1378)
+  To preserve backward compatibility controllers will support enabling it by setting special 
+  annotation `internal.scylla-operator.scylladb.com/enable-host-networking` to `true` on `v1alpha1.ScyllaDBDatacenter`.
+  The controller migrating from `v1.ScyllaCluster` to `v1alpha1.ScyllaDBDatacenter` will set this annotation when required.
+- `sysctls` doesn't belong to ScyllaDB cluster nor datacenter, these are configuration options for nodes. 
+  A better place to have them is NodeConfig. Until it's available the NodeConfig, we will support setting them 
+  via `internal.scylla-operator.scylladb.com/set-sysctls` annotation, in a format `{key}={value},{key}={value}`.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: Moved to rackTemplate

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+  This is unnecessary when there is documentation about performance tuning containing these recommendations.
+  This field will be removed and validation will be relaxed.
+- We initially thought that `hostNetworking` helps with performance, but later we discovered we were wrong.
+  As we only see the downsides of using it, `hostNetworking` support will be discontinued, and setting it becomes deprecated
+  and discouraged by not having a field to set it in API. [#1378](https://github.com/scylladb/scylla-operator/issues/1378)
+  To preserve backward compatibility controllers will support enabling it by setting special 
+  annotation `internal.scylla-operator.scylladb.com/enable-host-networking` to `true` on `v1alpha1.ScyllaDBDatacenter`.
+  The controller migrating from `v1.ScyllaCluster` to `v1alpha1.ScyllaDBDatacenter` will set this annotation when required.
+- `sysctls` doesn't belong to ScyllaDB cluster nor datacenter, these are configuration options for nodes. 
+  A better place to have them is NodeConfig. Until it's available the NodeConfig, we will support setting them
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: > Which means, scaling down the rack when cluster is growing can cause that rack won't have enough capacity to hold the copy of the data.

The algorithm only computes the desired config, it doesn't say how you apply it. We may first scale the racks that need it up, wait and then scale down, right?

> Your algorithm produce [3,2,2] for nodes=7, racks=3, while [2,2,2,1] for nodes=7, racks=4 which is not what we would want, as when cluster is growing, it shouldn't scale down any rack due to above.

> What my algorithm would produce when you already have [3,2,2] cluster is [3,2,2,0] which is what I think is expected in this case.

How would you get into the desired state if you want to say grow your RF from 3 to 5 by declaring 2 extra racks. If they become of size 0, it can't server RF=5.

> Another important topic we haven't though about is decommissioning a single rack. Right now we forbid removing a rack until it's first scaled down to 0. With balancing algorithm and inability to override desired node count for particular rack this seems impractical.

That's a good point. Since we know the existing racks and the desired ones, I guess we could scale it on behind, before pruning it out. A validation may be an easier start - if you use global "nodes", you can only add racks. (You can switch it to manual nodes and then remove it.)

> Given that we disagree about the desired behavior for scaling up

well, we now know where the difference is - I'll setup a call today and we can talk about the approach or deffer it, if we still have doubts

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: I agree, we shouldn't let users be surprised about a default they can't change later.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: > here's the code in go https://go.dev/play/p/w0lZMWN_w-N since that notation wasn't clear

I now see you were faster :)

tnozicka: This about this a second time, this may well deserve a separate proposal with a discussion limited for that topic. We know the implication for the initial API - the `nodes` counts have to be pointers but required for now, so we can later loosen this.

rzetelskik: > Your algorithm produce [3,2,2] for nodes=7, racks=3, while [2,2,2,1] for nodes=7, racks=4 which is not what we would want, as when cluster is growing, it shouldn't scale down any rack due to above.
What my algorithm would produce when you already have [3,2,2] cluster is [3,2,2,0] which is what I think is expected in this case.

From the perspective of just this simple API, i.e. only declaring the number of nodes and number of racks, I think you'd expect that increasing the number of racks would result in redistributing the nodes. Whether and how that plays with Scylla is an entirely different topic, and maybe a good example for the `nodes` field being too much of a simplification?

> Given that we disagree about the desired behavior for scaling up, and we haven't though well entire topology change topic I propose to push this feature out of this proposal and prepare next one focusing on this particular change.
We should consult Core team about what should we do in these cases, describe and understand every scenario.

+1 for moving this to a separate proposal, there's more to discuss here. I'd also remove the `.spec.nodes` from this proposal entirely in case we end up with something more complicated.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+type RackSpec struct {
+	RackTemplate
+
+	// name is the name of the ScyllaDB Rack. Used as rack name in GossipingPropertyFileSnitch.
+	Name string `json:"name"`
+}
+
+// ScyllaDB holds configuration options related to ScyllaDB.
+type ScyllaDB struct {
+	ScyllaDBOverrides
```
tnozicka, line 399: I'd merged the 2 into a generic ScyllaDBTemplate which will make it uniform with the `rackTemplate` and picking particular overrides at the racks level

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+	// servingCertificate references a TLS certificate for serving secure traffic.
+	// +kubebuilder:default:={type:"OperatorManaged"}
+	// +optional
+	ServingCertificate *TLSCertificate `json:"servingCertificate,omitempty"`
+}
+
+// ScyllaDBManagerAgent holds configuration options related to Scylla Manager Agent.
+type ScyllaDBManagerAgent struct {
+	ScyllaDBManagerAgentOverrides
```
tnozicka, line 486: and here

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+	// +optional
+	UpdatedNodes *int32 `json:"updatedNodes,omitempty"`
+
+	// readyNodes is the total number of ready nodes in datacenter.
+	// +optional
+	ReadyNodes *int32 `json:"readyNodes,omitempty"`
+
+	// rackCount is the number of ScyllaDB racks in this cluster.
+	// +optional
+	RackCount *int32 `json:"rackCount,omitempty"`
```
tnozicka, line 708: what's the use case for a projection of len(racks)?

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+Additional E2E to validate converting controller.
+
+### Upgrade / Downgrade Strategy
+
+New CRD needs to be installed upon Scylla Operator update. Our documentation and every release notes cover that.
+No specific actions are required for the downgrade.
+
+### Version Skew Strategy
+
+Existing `v1.ScyllaCluster` is going to be converted to `v1alpha.ScyllaDBDatacenter` and controllers are going to work on both of these resources.
```
tnozicka, line 751: We are not doing "conversion"of resources.

Is this just stale? I thought the plan is to change scyllacluster controller to create ScyllaDBDatacenter for each ScyllaCluster in it's sync loop (with a controllerRef set) and then a new controller (with the modified old sync loops from the old scyllacluster controller) will reconcile ScyllaDBDatacenters, no matter if controller or user created.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+When rolled back, changes to `v1alpha1.ScyllaDBDatacenters` are not going to be reconciled, users have to continue using `v1.ScyllaCluster`.
```
tnozicka, line 752:
```suggestion
When rolled back, changes to `v1alpha1.ScyllaDBDatacenters` are not going to be reconciled, users have to go back to using `v1.ScyllaCluster`, if they started using `v1alpha1.ScyllaDBDatacenters` instead.
```

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+## Implementation History
+
+26.06.204 - Initial version
+
+## Drawbacks
+
+Such change is not something we couldn't live with, but it helps to reduce the ambiguity of naming mistakes,
+as well as reduce the technical dept.
```
tnozicka, line 761:
```suggestion
as well as reduce the technical debt.
```

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: i don't understand what you mean. Can you provide a code change?
Keep in mind that all properties of non-embedded part of ScyllaDB struct are global to datacenter and cannot be specified on rack level.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: removed

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: I am proposing to be able to template+override the rest bellow, like `AdditionalScyllaDBArguments`

`AdditionalScyllaDBArguments` can differ by racks, if you want to test a limited change
`Image` can differ if you want to test it in a particular racks.
Similarly for the rest. All of them can technically  differ by the nature of a rolling update. Unifying the approach would make this consistent, even though some cases are not mainstream.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+	ScyllaDBManagerAgent *ScyllaDBManagerAgentOverrides `json:"scyllaDBManagerAgent"`
+
+	// topologyLabelSelector specifies a label selector which will be used to target nodes at specified topology constraints.
+	// Datacenter topologyLabelSelector is merged with rack topologyLabelSelector and then converted into nodeAffinity
+	// targeting nodes having specified topology.
+	TopologyLabelSelector *metav1.LabelSelector `json:"topologyLabelSelector,omitempty"`
+}
+
+// ScyllaDBOverrides allows overriding a subset of ScyllaDB settings.
+type ScyllaDBOverrides struct {
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+	ScyllaDB *ScyllaDBOverrides `json:"scyllaDB"`
+
+	// scyllaDBManagerAgent provides overrides for properties of ScyllaDB Manager Agent for this rack.
+	// They override the settings set on Datacenter level.
+	ScyllaDBManagerAgent *ScyllaDBManagerAgentOverrides `json:"scyllaDBManagerAgent"`
+
+	// topologyLabelSelector specifies a label selector which will be used to target nodes at specified topology constraints.
+	// Datacenter topologyLabelSelector is merged with rack topologyLabelSelector and then converted into nodeAffinity
+	// targeting nodes having specified topology.
+	TopologyLabelSelector *metav1.LabelSelector `json:"topologyLabelSelector,omitempty"`
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+type RackTemplate struct {
+	// nodes specifies desired number of nodes in rack.
+	// +optional
+	Nodes *int32 `json:"nodes"`
+
+	// placement describes restrictions for the nodes ScyllaDB is scheduled on.
+	// +optional
+	Placement *Placement `json:"placement,omitempty"`
+
+	// scyllaDB provides overrides for properties of ScyllaDB for this rack.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+	// +optional
+	Nodes *int32 `json:"nodes"`
+
+	// rackTemplate provides a template for every rack.
+	// Every rack inherits properties specified in the template, unless it's overwritten on the rack level.
+	// +optional
+	RackTemplate *RackTemplate `json:"rackTemplate,omitempty"`
+
+	// racks specify the racks in the datacenter.
+	// +optional
```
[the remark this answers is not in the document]
zimnx: removed optional

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+	// exposeOptions specifies parameters related to exposing ScyllaCluster backends.
+	// +optional
+	ExposeOptions *ExposeOptions `json:"exposeOptions,omitempty"`
+
+	// nodes specify desired total number of nodes in this datacenter.
+	// This number will be balanced out across all specified racks.
+	// Cannot be used together with `nodes` specified within any rack spec.
+	// +optional
+	Nodes *int32 `json:"nodes"`
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+	// clusterName specifies the name of the scylla cluster.
+	// When joining two DCs, their cluster name must match.
+	// This field is immutable.
+	ClusterName string `json:"clusterName"`
+
+	// datacenterName specifies the name of the scylla datacenter. Used as datacenter name in GossipingPropertyFileSnitch.
+	// If empty, it's taken from the 'scylladbdatacenter.metadata.name'.
+	// +optional
+	DatacenterName string `json:"datacenterName,omitempty"`
```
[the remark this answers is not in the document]
zimnx: Datacenter name can be updated live

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+  This field will be removed and validation will be relaxed. [#508](https://github.com/scylladb/scylla-operator/issues/508)
+- We initially thought that `hostNetworking` helps with performance, but later we discovered we were wrong.
+  As we only see the downsides of using it, `hostNetworking` support will be discontinued, and setting it becomes deprecated
+  and discouraged by not having a field to set it in API. [#1378](https://github.com/scylladb/scylla-operator/issues/1378)
+  To preserve backward compatibility controllers will support enabling it by setting special 
+  annotation `internal.scylla-operator.scylladb.com/enable-host-networking` to `true` on `v1alpha1.ScyllaDBDatacenter`.
+  The controller migrating from `v1.ScyllaCluster` to `v1alpha1.ScyllaDBDatacenter` will set this annotation when required.
+- `sysctls` doesn't belong to ScyllaDB cluster nor datacenter, these are configuration options for nodes. 
+  A better place to have them is NodeConfig. Until it's available in NodeConfig, we will support setting them 
+  via `internal.scylla-operator.scylladb.com/set-sysctls` annotation, in a JSON encoded array of `"key": "value"` format.
```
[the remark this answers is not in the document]
zimnx: added

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+- Introduce new features alongside new API.
+
+## Proposal
+
+I propose to add a new `v1alpha1.ScyllaDBDatacenter` to the existing `scylla.scylladb.com` API group.
+Existing controllers reconciling `v1.ScyllaCluster` are going to switch to reconcile the new API,
+and we'll add a controller that creates a backing `v1alpha1.ScyllaDBDatacneter` for every existing `v1.ScyllaCluster`.
+This will allow existing users to continue using the API they are used to.
+Until API is promoted beyond alpha, it will be mentioned in autogenerated API Reference in the Operator documentation.
```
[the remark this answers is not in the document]
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: Removed

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,761 @@
...
+	// Every rack inherits properties specified in the template, unless it's overwritten on the rack level.
+	// +optional
+	RackTemplate *RackTemplate `json:"rackTemplate,omitempty"`
+
+	// racks specify the racks in the datacenter.
+	// +optional
+	Racks []RackSpec `json:"racks"`
+
+	// automaticOrphanedNodeCleanup controls if automatic orphan node cleanup should be performed.
+	AutomaticOrphanedNodeCleanup bool `json:"automaticOrphanedNodeCleanup,omitempty"`
```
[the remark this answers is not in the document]
zimnx: Changed to `DisableAutomaticOrhpanedNodeReplace` as "cleanup" is ambiguous. We do the node replacement when node is orphaned.

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,764 @@
...
+  Similarly, references to Scylla Manager Agent are renamed to ScyllaDB Manager Agent, even though the naming change isn't completed.
+- Instead of having a single field specifying the container image of used components,
+  there are two fields, one for the repository and the second for the version, which prevents from using 
+  full format of container images [#821](https://github.com/scylladb/scylla-operator/issues/821).
+  This unnecessary split is error-prone.
+- Kubernetes API Convention suggests using arrays instead of maps. The status of racks doesn't follow that rule.
+- `cpuSet` is confusing, as it doesn't change anything related to cpuset, it only controls stricter validation of allocated resources.
+  This is unnecessary when there is documentation about performance tuning containing these recommendations.
+  This field will be removed and validation will be relaxed.
+- We initially thought that `hostNetworking` helps with performance, but later we discovered we were wrong.
```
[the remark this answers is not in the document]
rzetelskik: discussed on a call, resolved

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,711 @@
...
+    nodeSelectorTerms:
+      - matchExpressions:
+        - key: scylla.scylladb.com/node-type
+          operator: In
+          values:
+            - scylla
+        - key: topology.kubernetes.io/zone
+          operator: In
+          values:
+          - us-east1a
```
tnozicka, line 126: nit: remove the extra indent for yaml arrays for consistency

enhancements/proposals/1978-scylladbdatacenter-api/README.md
```diff
@@ -0,0 +1,711 @@
...
+	// rackTemplate provides a template for every rack.
+	// Every rack inherits properties specified in the template, unless it's overwritten on the rack level.
+	// +optional
+	RackTemplate *RackTemplate `json:"rackTemplate,omitempty"`
+
+	// racks specify the racks in the datacenter.
+	Racks []RackSpec `json:"racks"`
+
+	// disableAutomaticOrphanedNodeReplace controls if automatic orphan node replace should be disabled.
+	DisableAutomaticOrhpanedNodeReplace bool `json:"disableAutomaticOrphanedNodeReplace,omitempty"`
```
tnozicka, line 239:
```suggestion
	// disableAutomaticOrphanedNodeReplacement controls if automatic orphan node replacement should be disabled.
	DisableAutomaticOrphanedNodeReplacement bool `json:"disableAutomaticOrphanedNodeReplacement,omitempty"`
```
@rzetelskik wdyt?

tnozicka: + `s/Orhpaned/Orphaned/`

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: ok, let's leave it like this for now. Since all have effectively a default value, we can talk individual points later.

tnozicka (approved):
last nits / typos, rest lgtm

thanks for the updates

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

enhancements/proposals/1978-scylladbdatacenter-api/README.md
tnozicka: the whole thing under `nodeSelectorTerms` as well

tnozicka:
lgtm
/assign rzetelski

enhancements/proposals/1978-scylladbdatacenter-api/README.md
zimnx: done

rzetelskik (approved):
/lgtm
Thanks!

PR #1999, opened 2024-07-02
Fix flakiness of cleanup E2E test

### LINKED ISSUE #1525: Flake: ScyllaCluster [It] nodes are cleaned up after horizontal scaling

zimnx:
Cleanup token hash annotation is reconciled from the sidecar via polling method. If ring annotation is updated slower than new node joins/decommisions then cleanup job can be triggered twice. This is expected, so e2e test shouldn't fail in such case.

Test logic was aligned to accept extra cleanup job for stale token ring hash, by filtering observed jobs to get only those related to current token ring.

Fixes #1525

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_cleanup.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_cleanup.go
@@ -96,6 +96,10 @@ var _ = g.Describe("ScyllaCluster", func() {
 
 		o.Expect(jobEvents).NotTo(o.BeEmpty())
 
+		tokenRingHash, err := utils.GetCurrentTokenRingHash(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		framework.Infof("Current token ring hash is %q", tokenRingHash)
+
 		nodeJobMatcher := func(nodeName string) func(utils.ObserverEvent[*batchv1.Job]) bool {
 			return func(e utils.ObserverEvent[*batchv1.Job]) bool {
 				return e.Obj.Labels[naming.NodeJobLabel] == nodeName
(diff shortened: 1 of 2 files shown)


test/e2e/set/scyllacluster/scyllacluster_cleanup.go
```diff
@@ -96,14 +96,20 @@ var _ = g.Describe("ScyllaCluster", func() {
 
 		o.Expect(jobEvents).NotTo(o.BeEmpty())
 
+		tokenRingHash, err := utils.GetCurrentTokenRingHash(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		framework.Infof("Current token ring hash is %q", tokenRingHash)
```
rzetelskik, line 101: Why is this a valuable information? Should I be able to compare it with something? Or know if it's not the expected one?

rzetelskik (commented):
one comment, otherwise lgtm

test/e2e/set/scyllacluster/scyllacluster_cleanup.go
zimnx: Each cleanup Job has ring hash annotation, the one Operator noticed and created a Job for. Annotations in artifacts should be the same as one printed here. In case of a failure, we can verify if job was stale or not.

tnozicka (approved):
/approve
lgtm

rzetelskik (approved):
/lgtm

PR #2000, opened 2024-07-03
Fix NodeConfig controller handlers to react on related object updates

### LINKED ISSUE #2001: ServiceAccount and other objects managed by NodeConfig are not reconciled on update
I've tried to delete our managed hash from `scylla-node-config` ServiceAccount and it never made it back.

```
apiVersion: v1
kind: ServiceAccount
metadata:
  creationTimestamp: "2024-06-20T12:35:52Z"
  labels:
    foo: bar2
    scylla-operator.scylladb.com/node-config-name: scylla-node-config
  name: scylla-node-config
  namespace: scylla-operator-node-tuning
  resourceVersion: "554112"
  uid: cea11e93-fb23-468c-a092-ceb6c45ab8e5
```

The issue is that the handlers use `EnqueueOwner` but the objects don't have a controllerRef because they are singletons shared by multiple NodeConfigs.

Stems from https://github.com/scylladb/scylla-operator/pull/1971#discussion_r1662510514

tnozicka:
This PR fixes the handlers so the reconciled objects that don't have ownerRefs on them requeue the NodeConfigs. Given they are shared, they need to requeue all of them but the number of NodeConfigs is usually low single digit number anyway.

Resolves #2001

### DIFF
--- a/pkg/controller/nodeconfig/controller.go
+++ b/pkg/controller/nodeconfig/controller.go
@@ -14,6 +14,7 @@ import (
 	scyllav1alpha1listers "github.com/scylladb/scylla-operator/pkg/client/scylla/listers/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/kubeinterfaces"
+	"github.com/scylladb/scylla-operator/pkg/naming"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -71,6 +72,10 @@ type Controller struct {
 	handlers *controllerhelpers.Handlers[*scyllav1alpha1.NodeConfig]
 
 	operatorImage string
+}
+
+func isManagedByNodeConfigController(obj kubeinterfaces.ObjectInterface) bool {
+	return obj.GetLabels()[naming.NodeConfigNameLabel] == naming.NodeConfigAppName
 }
 
 func NewController(
@@ -182,6 +187,12 @@ func NewController(
 		DeleteFunc: ncc.deleteDaemonSet,
 	})
 
+	namespaceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    ncc.addNamespace,
+		UpdateFunc: ncc.updateNamespace,
+		DeleteFunc: ncc.deleteNamespace,
+	})
+
 	return ncc, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controllerhelpers/handlers.go
```diff
@@ -93,6 +93,16 @@ func (h *Handlers[T]) EnqueueAll(depth int, untypedObj kubeinterfaces.ObjectInte
 	h.EnqueueAllFunc(h.Enqueue)(depth+1, untypedObj, op)
 }
 
+func (h *Handlers[T]) EnqueueAllWithUntypedFilterFunc(filterFunc func(kubeinterfaces.ObjectInterface) bool) EnqueueFuncType {
```
rzetelskik, line 96: you'd cover more use cases with

```
func (h *Handlers[T]) EnqueueWithUntypedFilterFunc(filterFunc func(kubeinterfaces.ObjectInterface) bool) EnqueueFuncType {
	return func(depth int, untypedObj kubeinterfaces.ObjectInterface, op HandlerOperationType) {
		if !filterFunc(untypedObj) {
			return
		}

		h.Enqueue(depth+1, untypedObj, op)
	}
}
```

which you can compose to have the same functionality

```
ncc.handlers.EnqueueAllFunc(ncc.handlers.EnqueueWithUntypedFilterFunc(isManagedByNodeConfigController))
```

pkg/controller/nodeconfig/controller.go
```diff
@@ -73,6 +74,15 @@ type Controller struct {
 	operatorImage string
 }
 
+func isManagedByNodeConfigController(obj kubeinterfaces.ObjectInterface) bool {
+	labels := obj.GetLabels()
+	if labels == nil {
+		return false
+	}
```
rzetelskik, line 81: nit: unnecessary

rzetelskik (commented):
one comment and one nit, rest lgtm

pkg/controllerhelpers/handlers.go
tnozicka: notice the different argument types, those do not compose

pkg/controller/nodeconfig/controller.go
tnozicka: many object definitions may not have the labels map initialized, this feels safe to avoid panics

pkg/controllerhelpers/handlers.go
rzetelskik: copy-pasted from my comment https://github.com/rzetelskik/scylla-operator/commit/8c8f4b4f4fa616d3dbdeed26e869e7b980a8c5f2, compiles for me

pkg/controller/nodeconfig/controller.go
rzetelskik: > many object definitions may not have the labels map initialized

yes, but you can safely read in that case anyway.
[`A nil map is equivalent to an empty map except that no elements may be added.`](https://go.dev/ref/spec#Map_types)

You can still prefer to do this imo, hence a nit

tnozicka: not a nit, you are correct :+1:

zimnx (approved):
lgtm
/assign rzetelskik

pkg/controller/nodeconfig/controller.go
tnozicka: fixed

pkg/controllerhelpers/handlers.go
rzetelskik: ok, my bad, these are not equivalent

rzetelskik (approved):
/lgtm

tnozicka:
/retest
(likely slow clusterip again, need to find the link later)

PR #2009, opened 2024-07-08
Stop depending on default StorageClass' presence for ScyllaClusters in our manifests, examples, scripts and tests

### LINKED ISSUE #1990: Use explicit storageclass in all of our manifests
Using a "wrong" storageclass happens too often. Also in production, users would specify the storage class explicitly, instead of making our storageclass to be the default for everything. (Cost-wise it's not needed for other apps.)

Let's wire all of our manifests and examples to use `scylladb-local-xfs` storageclass which makes it less error prone for us and our users.

CI wise, for cases where want to test a setup with different storageclass and alias may be a viable option without the need to edit all the configs in every CI job.

rzetelskik:
**Description of your changes:** This PR addresses the issue of ScyllaClusters in our manifests, examples, scripts and tests often depending on the presence of a default StorageClass. This is problematic because there can only be one, indicated with an annotation, so setting/unsetting a default tends to break immutability.

Simultaneously I've tried not to break existing workflows and not have to change all of our CI jobs by expecting a StorageClass to be set explicitly. Therefore:
- test options are extended with an optional `scyllacluster-storageclass-name` flag, which defaults to empty string, which is then interpreted as falling back to default storageclass (not setting storageClassName),
- all ScyllaClusters in our manifests and examples have `storageClassName` set to `scylladb-local-xfs` by default
- e2e/deploy scripts take `SO_SCYLLACLUSTER_STORAGECLASS_NAME` which propagates to Scylla Manager's and tests' clusters; if `SO_SCYLLACLUSTER_STORAGECLASS_NAME` is explicitly set and empty, `storageClassName` is not set, falling back to default storageclass
 
Resolves #1990 

/kind feature
/priority important-soon

/cc

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -94,14 +94,6 @@ function gather-artifacts-on-exit {
   done
 }
 
-function unset-default-storageclass {
-  for f in "${KUBECONFIGS[@]}"; do
-    for r in $( KUBECONFIG="${f}" kubectl get storageclasses -o name ); do
-      KUBECONFIG="${f}" kubectl patch "${r}" -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
-    done
-  done
-}
-
 function apply-e2e-workarounds {
   if [ -z "${SO_IMAGE+x}" ]; then
     echo "SO_IMAGE can't be empty" > /dev/stderr
@@ -191,6 +183,11 @@ function run-e2e {
     exit 2
   fi
 
+  if [ -z "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]; then
+    echo "SO_SCYLLACLUSTER_STORAGECLASS_NAME can't be empty" > /dev/stderr
+    exit 2
+  fi
+
   if [ -z "${ARTIFACTS+x}" ]; then
     echo "ARTIFACTS can't be empty" > /dev/stderr
     exit 2
@@ -262,6 +259,7 @@ spec:
     - "--scyllacluster-node-service-type=${SO_SCYLLACLUSTER_NODE_SERVICE_TYPE}"
     - "--scyllacluster-nodes-broadcast-address-type=${SO_SCYLLACLUSTER_NODES_BROADCAST_ADDRESS_TYPE}"
     - "--scyllacluster-clients-broadcast-address-type=${SO_SCYLLACLUSTER_CLIENTS_BROADCAST_ADDRESS_TYPE}"
+    - "--scyllacluster-storageclass-name=${SO_SCYLLACLUSTER_STORAGECLASS_NAME}"
     - "--object-storage-bucket=${SO_BUCKET_NAME}"
     - "--gcs-service-account-key-path=${gcs_sa_in_container_path}"
     image: "${SO_IMAGE}"
--- a/hack/.ci/run-e2e-gke-release.sh
+++ b/hack/.ci/run-e2e-gke-release.sh
@@ -20,8 +20,8 @@ trap gather-artifacts-on-exit EXIT
 SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=${parent_dir}/manifests/cluster/nodeconfig.yaml}"
 export SO_NODECONFIG_PATH
 
-# Make sure there is no default storage class before we create our own so we always use our own provisioner from the release.
-unset-default-storageclass
+SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME=scylladb-local-xfs}"
+export SO_SCYLLACLUSTER_STORAGECLASS_NAME
 
 SCYLLA_OPERATOR_FEATURE_GATES="${SCYLLA_OPERATOR_FEATURE_GATES:-AllAlpha=true,AllBeta=true}"
 export SCYLLA_OPERATOR_FEATURE_GATES
--- a/hack/.ci/run-e2e-gke.sh
+++ b/hack/.ci/run-e2e-gke.sh
@@ -21,14 +21,13 @@ SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=./hack/.ci/manifests/cluster/nodeconfig
 export SO_NODECONFIG_PATH
 SO_CSI_DRIVER_PATH="${parent_dir}/manifests/namespaces/local-csi-driver/"
 export SO_CSI_DRIVER_PATH
+SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME=scylladb-local-xfs}"
+export SO_SCYLLACLUSTER_STORAGECLASS_NAME
 
 # Backwards compatibility. Remove when release repo stops using SO_DISABLE_NODECONFIG.
 if [[ "${SO_DISABLE_NODECONFIG:-false}" == "true" ]]; then
   SO_NODECONFIG_PATH=""
   SO_CSI_DRIVER_PATH=""
-else
-  # Make sure there is no default storage class before we create our own.
-  unset-default-storageclass
 fi
 
 SCYLLA_OPERATOR_FEATURE_GATES="${SCYLLA_OPERATOR_FEATURE_GATES:-AllAlpha=true,AllBeta=true}"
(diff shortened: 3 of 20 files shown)


rzetelskik:
This is going to require explicitly setting `SO_SCYLLACLUSTER_STORAGECLASS_NAME` in `e2e-gke-serial` presubmit to pass. Before I send a PR for the prow job, let's verify the approach.

/cc zimnx tnozicka

pkg/cmd/tests/options.go
```diff
@@ -63,6 +64,7 @@ func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent stri
 			NodeServiceType:             string(scyllav1.NodeServiceTypeHeadless),
 			NodesBroadcastAddressType:   string(scyllav1.BroadcastAddressTypePodIP),
 			ClientsBroadcastAddressType: string(scyllav1.BroadcastAddressTypePodIP),
+			StorageClassName:            "scylladb-local-xfs",
```
tnozicka, line 67: nit: I think empty is a good default given how default storageclasses work, I don't think this binary is the place to be opinionated about naming. We'd always set the flag explicitly in our tests thought.

hack/.ci/run-e2e-gke-release.sh
```diff
@@ -20,8 +20,11 @@ trap gather-artifacts-on-exit EXIT
 SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=${parent_dir}/manifests/cluster/nodeconfig.yaml}"
 export SO_NODECONFIG_PATH
 
-# Make sure there is no default storage class before we create our own so we always use our own provisioner from the release.
-unset-default-storageclass
+SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME=}"
+export SO_SCYLLACLUSTER_STORAGECLASS_NAME
+if [[ -z "${SO_CSI_DRIVER_PATH+x}" || -n "${SO_CSI_DRIVER_PATH}" ]]; then
+  SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-scylladb-local-xfs}"
```
tnozicka, line 26: won't this variable be always set given the defaulting above? structuring this as if+else might be less bug prone

hack/.ci/run-e2e-gke.sh
```diff
@@ -21,14 +21,15 @@ SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=./hack/.ci/manifests/cluster/nodeconfig
...
+export SO_SCYLLACLUSTER_STORAGECLASS_NAME
 
 # Backwards compatibility. Remove when release repo stops using SO_DISABLE_NODECONFIG.
 if [[ "${SO_DISABLE_NODECONFIG:-false}" == "true" ]]; then
   SO_NODECONFIG_PATH=""
   SO_CSI_DRIVER_PATH=""
 else
-  # Make sure there is no default storage class before we create our own.
-  unset-default-storageclass
+  SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-scylladb-local-xfs}"
```
tnozicka, line 32: same here

tnozicka: I think generally just defaulting this to `scylladb-local-xfs` would be fine and the caller can always choose to set it explicitly

pkg/cmd/tests/options.go
rzetelskik: > I think empty is a good default given how default storageclasses work

can you elaborate? I don't understand why that'd make it a good default. Empty storageClassName means you won't get any storageClass, it's not equivalent to not setting it at all.

hack/.ci/run-e2e-gke.sh
rzetelskik: > I think generally just defaulting this to scylladb-local-xfs would be fine and the caller can always choose to set it explicitly

I ended up setting it to an empty string if SO_DISABLE_NODECONFIG is true for ease of debugging - in cases like the serial job you have to set `SO_SCYLLACLUSTER_STORAGECLASS_NAME` explicitly, otherwise you end up with an empty storageClassName, which is easier to spot (https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2009/pull-scylla-operator-master-e2e-gke-serial/1810962794783182848/artifacts/deploy/0/manager/50_scyllacluster.yaml). It's a minor improvement though, so if you think simplifying the default more valuable, I can change it.

hack/.ci/run-e2e-gke-release.sh
rzetelskik: no longer applicable - I applied changes we discussed internally

hack/.ci/run-e2e-gke.sh
rzetelskik: done

zimnx (approved):
lgtm 
/assign tnozicka

hack/ci-deploy.sh
```diff
@@ -71,6 +71,11 @@ else
   kubectl_create -n=local-csi-driver -f="${SO_CSI_DRIVER_PATH}"
 fi
 
+if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME}" ]]; then
+  yq e --inplace ".spec.datacenter.racks[0].storage.storageClassName = \"${SO_SCYLLACLUSTER_STORAGECLASS_NAME}\"" "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
```
tnozicka, line 75:
```suggestion
  yq e --inplace '.spec.datacenter.racks[0].storage.storageClassName = env(SO_SCYLLACLUSTER_STORAGECLASS_NAME)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
```

hack/ci-deploy.sh
```diff
@@ -71,6 +71,11 @@ else
...
+elif [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]]; then
+    yq e --inplace "del(.spec.datacenter.racks[0].storage.storageClassName)" "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
```
tnozicka, line 77:
```suggestion
    yq e --inplace 'del(.spec.datacenter.racks[0].storage.storageClassName)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
```
tiny nit, but can be coupled to the edit above

hack/.ci/lib/e2e.sh
```diff
@@ -191,6 +183,11 @@ function run-e2e {
     exit 2
   fi
 
+  if [ -z "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]; then
+    echo "SO_SCYLLACLUSTER_STORAGECLASS_NAME can't be empty" > /dev/stderr
+    exit 2
+  fi
```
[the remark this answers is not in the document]
rzetelskik: ci-deploy/ci-deploy-release scripts are standalone and can be used independently of the run-e2e-gke/run-e2e-gke-release scripts - that's why I wanted them to have a reasonable default behaviour. The run-e2e function is only uses in  run-e2e-gke/run-e2e-gke-release scripts, and the env var is defined there. I left it here as a failsafe

hack/ci-deploy.sh
rzetelskik: I'd rather leave it as is, I think it's complicated enough already :sweat_smile:

hack/ci-deploy.sh
rzetelskik: why do you think that's better?

hack/.ci/lib/e2e.sh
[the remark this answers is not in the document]
tnozicka: ok, wost case we'll do a follow up ;)

hack/ci-deploy.sh
tnozicka: it's in line with the other ones but I don't mind

hack/ci-deploy.sh
tnozicka: it's native and properly quotes the value (e.g. nested quotes)

rzetelskik: ok, done

tnozicka:
/lgtm
thanks

/hold cancel
https://github.com/scylladb/scylla-operator-release/pull/274 has landed

/retest

PR #2033, opened 2024-07-23
Use identical file system for node tuning and keep symlinks working

### LINKED ISSUE #2029: NodeConfig ignores /var which also break other folders when symlinked
On CoreOS like fs, all data is efectivelly stored in `/var/` and other folders point to it:
```
lrwxrwxrwx.   2 root root    7 Nov 20  2023 bin -> usr/bin
drwxr-xr-x.   8 root root 1024 Jul 19 09:14 boot
drwxr-xr-x.  17 root root 3720 Jul 19 14:23 dev
drwxr-xr-x.  96 root root 8192 Jul 19 14:23 etc
lrwxrwxrwx.   2 root root    8 Nov 20  2023 home -> var/home
lrwxrwxrwx.   2 root root    7 Nov 20  2023 lib -> usr/lib
lrwxrwxrwx.   2 root root    9 Nov 20  2023 lib64 -> usr/lib64
lrwxrwxrwx.   2 root root    9 Nov 20  2023 media -> run/media
lrwxrwxrwx.   2 root root    7 Nov 20  2023 mnt -> var/mnt
lrwxrwxrwx.   2 root root    7 Nov 20  2023 opt -> var/opt
lrwxrwxrwx.   2 root root   14 Nov 20  2023 ostree -> sysroot/ostree
dr-xr-xr-x. 237 root root    0 Jul 19 14:22 proc
lrwxrwxrwx.   2 root root   12 Nov 20  2023 root -> var/roothome
drwxr-xr-x.  56 root root 1480 Jul 19 14:23 run
lrwxrwxrwx.   2 root root    8 Nov 20  2023 sbin -> usr/sbin
lrwxrwxrwx.   2 root root    7 Nov 20  2023 srv -> var/srv
dr-xr-xr-x.  13 root root    0 Jul 19 14:22 sys
drwxr-xr-x.   4 root root   66 Nov 20  2023 sysroot
drwxrwxrwt.  11 root root  220 Jul 19 17:41 tmp
drwxr-xr-x.  12 root root  155 Jan  1  1970 usr
drwxr-xr-x.  24 root root 4096 Jul 19 09:13 var
```
but NodeConfig doesn't mount say `/var` at all
https://github.com/scylladb/scylla-operator/blob/d0ccf5553e72d5ecf1a8a60a66c66b14cc9de8c0/pkg/controller/nodeconfig/resource.go#L251
So while we can [now resolve symlinks](#2014), the path actually doesn't exist and can't be resolved, neither used.

tnozicka:
When the Kubernetes node file system contains top level symlinks our node setup fails to resolve them in the chroot. This is because we skip bind mounting some directories, like `/var` and other top level dirs can be symlinks pointing there. (This is a majority of top level dirs on Core OS and immutable filesystems.)

To keep top level symlinks for node setup working we need to create an identical filesystem structure. This will also clean up the overrides we have.

Resolves #2029

### DIFF
--- a/pkg/controller/nodeconfig/resource.go
+++ b/pkg/controller/nodeconfig/resource.go
@@ -193,7 +193,8 @@ func makeNodeSetupDaemonSet(nc *scyllav1alpha1.NodeConfig, operatorImage, scylla
 					Labels: labels,
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: naming.NodeConfigAppName,
+					ServiceAccountName:           naming.NodeConfigAppName,
+					AutomountServiceAccountToken: pointer.Ptr(false),
 					// Required for getting the right iface name to tune
 					HostNetwork:  true,
 					NodeSelector: nc.Spec.Placement.NodeSelector,
@@ -206,6 +207,35 @@ func makeNodeSetupDaemonSet(nc *scyllav1alpha1.NodeConfig, operatorImage, scylla
 								HostPath: &corev1.HostPathVolumeSource{
 									Path: "/",
 									Type: pointer.Ptr(corev1.HostPathDirectory),
+								},
+							},
+						},
+						{
+							Name: "kube-api-access",
+							VolumeSource: corev1.VolumeSource{
+								Projected: &corev1.ProjectedVolumeSource{
+									DefaultMode: pointer.Ptr[int32](420),
+									Sources: []corev1.VolumeProjection{
+										{
+											ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
+												Path: "token",
+											},
+										},
+										{
+											ConfigMap: &corev1.ConfigMapProjection{
+												LocalObjectReference: corev1.LocalObjectReference{
+													Name: "kube-root-ca.crt",
+												},
+												Items: []corev1.KeyToPath{
+													{
+
+														Key:  corev1.ServiceAccountRootCAKey,
+														Path: corev1.ServiceAccountRootCAKey,
+													},
+												},
+											},
+										},
+									},
 								},
 							},
 						},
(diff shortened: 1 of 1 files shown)


pkg/controller/nodeconfig/resource.go
```diff
@@ -226,11 +256,12 @@ func makeNodeSetupDaemonSet(nc *scyllav1alpha1.NodeConfig, operatorImage, scylla
 							Args: []string{
 								`
 # Create a temporary root that will represent the host file system and devices.
+# It shall contain identical file tree as the host, so all symlinks keep working,
+# and we'll add additional "virtual" data into /scylla-operator folder.
+# This will avoid polluting the node with the extra data and avoids the need to clean up. 
 cd "$( mktemp -d )"
```
rzetelskik, line 262: Trying to understand the context, so more of a question to the existing implementation, rather than this PR, but why do this instead of an emtpyDir + an emptyDir subpath for scylla-operator dir? You could then mount the kube-api-access volume directly, instead of binding it. Or am I missing something here?

pkg/controller/nodeconfig/resource.go
```diff
@@ -242,52 +273,46 @@ done
...
-	touch "./run/secrets/kubernetes.io/serviceaccount/${f}"
-	mount --bind "/run/secrets/kubernetes.io/serviceaccount/${f}" "./run/secrets/kubernetes.io/serviceaccount/${f}"
-done
+# Mount container run to propagate secrets and configmaps.
+mkdir './scylla-operator/run'
+mount --rbind {,./scylla-operator}/run
+							
+# Mount container tmp.
+mkdir './scylla-operator/tmp'
+mount --rbind {,./scylla-operator}/tmp
```
rzetelskik, line 290: nit: is this necessary? ./scylla-operator is already in tmp, so wouldn't it actually be cleaner to not bind this?

pkg/controller/nodeconfig/resource.go
tnozicka: I guess an emptyDir + workingDir would do the same

pkg/controller/nodeconfig/resource.go
tnozicka: in hosts's tmp, we don't want to write to the host fs

zimnx (approved):
lgtm
/assign rzetelskik

pkg/controller/nodeconfig/resource.go
rzetelskik: ok, I don't mind either approach

rzetelskik (approved):
/lgtm

PR #2039, opened 2024-07-29
Set an explicit key for gcs-service-account-credentials secret in e2e scripts

rzetelskik:
**Description of your changes:** This PR modifies the gcs-service-account-credentials secret creation kubectl command by giving it an explicit key so that the credentials file passed through SO_GCS_SERVICE_ACCOUNT_CREDENTIALS_PATH doesn't have to have a matching name. This is helpful for local development.

Resolves #

/kind machinery
/priority important-longterm


tnozicka (approved):
/approve
/lgtm

Issue #1997, opened 2024-07-01
Multiple E2E tests time out waiting for Scylla Cluster to roll out when running on arm64 machines


rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-latest-e2e-gke-arm64-parallel/1807655512678862848#1:test-build-log.txt%3A2566

```
Summarizing 8 Failures:
  [FAIL] ScyllaCluster [It] should set up Alternator API when enabled
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_alternator.go:106
  [FAIL] ScyllaCluster webhook [It] should forbid invalid requests
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_webhooks.go:91
  [FAIL] ScyllaCluster should be exposed [It] using ClusterIP Service, and broadcast it to both client and nodes
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_expose.go:184
  [FAIL] ScyllaCluster [It] should connect to cluster via Ingresses
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_expose.go:81
  [FAIL] ScyllaCluster [It] should setup and maintain up to date TLS certificates
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_tls.go:68
  [FAIL] ScyllaCluster should be exposed [It] using LoadBalancer Service, and broadcast external IP to client and PodIP to nodes
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_expose.go:184
  [FAIL] ScyllaCluster [It] should rotate TLS certificates before they expire
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_tls.go:283
  [FAIL] ScyllaCluster [It] should create ingress objects when ingress exposeOptions are provided
  github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_expose.go:425
```

When running on arm64 machines, multiple tests in our e2e parallel suite fail waiting for Scylla Cluster to roll out.
We should determine whether it's caused by slower disks (Tau T2A machines do not support local SSDs) or if it's infrastructure specific.

/kind failing-test

rzetelskik:
```
Summarizing 4 Failures:
[...]

PR #2040, opened 2024-07-29
Add duration to certificate creation logs

rzetelskik:
**Description of your changes:** This PR adds duration to certificate creation logs for ease of debugging in scenarios similar to https://github.com/scylladb/scylla-operator/issues/1997. At the moment the pre and post-creation logs have to be matched to asses the time it took to create the certificate, which can be a long running action.

```
I0725 14:09:06.446762       1 kubecrypto/certs.go:201] "Creating certificate" Secret="e2e-test-scyllacluster-2fws8-0-cs67t/basic-7xzt6-local-client-ca" Reason="needs new cert"
I0725 14:09:48.664868       1 kubecrypto/certs.go:206] "Certificate created" Secret="e2e-test-scyllacluster-2fws8-0-cs67t/basic-7xzt6-local-client-ca"
```

Resolves #

/kind feature
/priority important-longterm

### DIFF
--- a/pkg/kubecrypto/certs.go
+++ b/pkg/kubecrypto/certs.go
@@ -198,12 +198,13 @@ func makeCertificate(ctx context.Context, name string, certCreator ocrypto.CertC
 	}
 
 	if len(refreshReason) != 0 {
+		startTime := time.Now()
 		klog.V(2).InfoS("Creating certificate", "Secret", naming.ObjRef(tlsSecret.GetSecret()), "Reason", refreshReason)
 		cert, key, err := certCreator.MakeCertificate(ctx, keyGetter, signer, validity)
 		if err != nil {
 			return nil, fmt.Errorf("can't create certificate: %w", err)
 		}
-		klog.V(2).InfoS("Certificate created", "Secret", naming.ObjRef(tlsSecret.GetSecret()))
+		klog.V(2).InfoS("Certificate created", "Secret", naming.ObjRef(tlsSecret.GetSecret()), "ElapsedTime", time.Now().Sub(startTime))
 
 		certBytes, err := ocrypto.EncodeCertificates(cert)
 		if err != nil {


pkg/kubecrypto/certs.go
```diff
@@ -198,12 +198,13 @@ func makeCertificate(ctx context.Context, name string, certCreator ocrypto.CertC
...
 
 	if len(refreshReason) != 0 {
+		startTime := time.Now()
 		klog.V(2).InfoS("Creating certificate", "Secret", naming.ObjRef(tlsSecret.GetSecret()), "Reason", refreshReason)
 		cert, key, err := certCreator.MakeCertificate(ctx, keyGetter, signer, validity)
 		if err != nil {
 			return nil, fmt.Errorf("can't create certificate: %w", err)
 		}
-		klog.V(2).InfoS("Certificate created", "Secret", naming.ObjRef(tlsSecret.GetSecret()))
+		klog.V(2).InfoS("Certificate created", "Secret", naming.ObjRef(tlsSecret.GetSecret()), "Duration", time.Now().Sub(startTime))
```
tnozicka, line 207: `duration` is a bit ambiguous here (with cert lifetime)

tnozicka (commented):
Technically, these times will be mangled by the cache / buffer underneath. Shouldn't we rather trace the actual key generation which is what takes the time?

tnozicka:
> Technically, these times will be mangled by the cache / buffer underneath. Shouldn't we rather trace the actual key generation which is what takes the time?

but these are also misplaced with regards to the sync loop, so this may just be good enough

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[11d6b50](https://github.com/scylladb/scylla-operator/commit/11d6b5091dd1b618afbe88a46a1a406dda887c81)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2040/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1817850879848484864)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2040). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

New flake https://github.com/scylladb/scylla-operator/issues/2042

pkg/kubecrypto/certs.go
rzetelskik: right, changed to `ElapsedTime`

rzetelskik:
> > Technically, these times will be mangled by the cache / buffer underneath. Shouldn't we rather trace the actual key generation which is what takes the time?
> 
> but these are also misplaced with regards to the sync loop, so this may just be good enough

right, key generation already has the duration and you can look there, but my reasoning was that if you start debugging from the reconciliation loop this is easier to notice

PR #2055, opened 2024-08-02
Add context on which condition failed in WaitForObjectState

### LINKED ISSUE #2054: WaitForObjectState doesn't identify which condition has failed
[WaitForObjectState](https://github.com/scylladb/scylla-operator/blob/110ee353a7412ea55e4d1795913e8870ea789c29/pkg/controllerhelpers/wait.go#L104) (and all its wrappers) take multiple conditions as arguments but don't indicate which of the conditions have caused it to fail. It's possibly to manually guess it from the artifacts but that's cumbersome and racy.

tnozicka:
When multiple conditions are used with `WaitForObjectState` and it times out, it will now print information about which conditions were evaluated  and their state.

Resolves #2054

### DIFF
--- a/pkg/controllerhelpers/wait.go
+++ b/pkg/controllerhelpers/wait.go
@@ -3,6 +3,8 @@ package controllerhelpers
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
@@ -15,6 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllerhelpers/wait_test.go
```diff
@@ -0,0 +1,75 @@
...
+			expectedString: "[true,false]",
+			expectedErr:    nil,
+		},
+		{
+			name:           "done+undone+done condition is undone",
+			ac:             NewAggregatedConditions[*corev1.ConfigMap](hasKeyFunc("alpha"), hasKeyFunc("doesnotexist"), hasKeyFunc("beta")),
+			expectedDone:   false,
+			expectedString: "[true,false,true]",
+			expectedErr:    nil,
+		},
```
rzetelskik, line 54: I'd also add a test for missing state (`<nil>`)

tnozicka: added

zimnx (approved):
lgtm
/assign rzetelskik

tnozicka:
https://github.com/scylladb/scylla-operator/issues/2061#issue-2450222847
/retest

pkg/controllerhelpers/wait.go
```diff
@@ -31,6 +34,73 @@ type WaitForStateOptions struct {
...
+
+func (ac AggregatedConditions[Obj]) Condition(obj Obj) (bool, error) {
+	ac.clearState()
+
+	allDone := true
+	var err error
+	for i, cond := range ac.conditions {
+		var done bool
+		done, err = cond(obj)
+		ac.state[i] = &done
```
rzetelskik, line 69: shouldn't the err be checked before assigning the value? I think it would be more intuitive that you'd end up with a `<nil>` string for the condition that couldn't be evaluated, instead of a default value

pkg/controllerhelpers/wait.go
```diff
@@ -31,6 +34,73 @@ type WaitForStateOptions struct {
...
+		if err != nil {
+			return done, err
```
rzetelskik, line 71: why return early instead of aggregating the errors?

pkg/controllerhelpers/wait.go
tnozicka: > for the condition that couldn't be evaluated

it **was** evaluated, ended with an error and this is its return value

nil is for the conditions that were not evaluated

rzetelskik: > and this is its return value

which, using this error pattern, you wouldn't normally take into account, since it returned an error

> nil is for the conditions that were not evaluated

I know what it is for now, but I'm saying imo it would make more sense to use it for the ones that returned an error, continue evaluation of other conditions and aggregate the errors, but I can live with the argument that it maintains the current behaviour

pkg/controllerhelpers/wait.go
rzetelskik: It wouldn't change the overall result as you would get an error either way - user shouldn't depend on the value in that case. That is unless you take any side effects into account.

Anyway I'm fine with maintaining the existing behaviour argument.

rzetelskik (approved):
left more comments but I don't see much point in delaying this
/lgtm

pkg/controllerhelpers/wait.go
tnozicka: >  you wouldn't normally take into account

You may or may not take it into account depending on function behaviour / docs, but it's the value that was returned which is my reasoning for including it. If the function meant to return no value, it would return `*bool`. Also with the done, the function return value is valid - anything that returns an error needs to return false.

PR #2062, opened 2024-08-07
Substitute unbound SO_SCYLLACLUSTER_STORAGECLASS_NAME var in ci-deploy script

rzetelskik:
**Description of your changes:** In cases where SO_SCYLLACLUSTER_STORAGECLASS_NAME is unbound, hack/ci-deploy.sh script will now fail with an error:
```
./hack/ci-deploy.sh: line 74: SO_SCYLLACLUSTER_STORAGECLASS_NAME: unbound variable
```

This is not the intended behaviour and it now requires providing the value explicitly, which is not the intended behaviour.

This PR substitutes that variable, so that the default intended behaviour is executed for an unset var.

Resolves #

/kind bug
/priority important-longterm

### DIFF
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -71,7 +71,7 @@ else
   kubectl_create -n=local-csi-driver -f="${SO_CSI_DRIVER_PATH}"
 fi
 
-if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME}" ]]; then
+if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-}" ]]; then
   yq e --inplace '.spec.datacenter.racks[0].storage.storageClassName = env(SO_SCYLLACLUSTER_STORAGECLASS_NAME)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
 elif [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]]; then
   yq e --inplace 'del(.spec.datacenter.racks[0].storage.storageClassName)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"


hack/ci-deploy.sh
```diff
@@ -71,7 +71,7 @@ else
   kubectl_create -n=local-csi-driver -f="${SO_CSI_DRIVER_PATH}"
 fi
 
-if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME}" ]]; then
+if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-}" ]]; then
   yq e --inplace '.spec.datacenter.racks[0].storage.storageClassName = env(SO_SCYLLACLUSTER_STORAGECLASS_NAME)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
 elif [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]]; then
```
tnozicka, line 76: this checks if it's undefined - shouldn't reversing the order fix it?

rzetelskik: The logic behind this is roughly
```
if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-}" ]]; then
  echo "set and nonempty, replace storageClassName"
elif [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]]; then
  echo "set and empty, delete explicit storageClassName (use default)"
else
  echo "not set, do not modify the manifest"
fi
```

if you reverse the order without adding the substitution you still end up with unbound variable error in case it's not set I believe

tnozicka: you are correct, I though +x gets passed this

tnozicka (approved):
/approve
/lgtm

rzetelskik:
> [APPROVALNOTIFIER] This PR is **APPROVED**
> 
> This pull-request has been approved by: _[rzetelskik](https://github.com/scylladb/scylla-operator/pull/2062#)_, _[tnozicka](https://github.com/scylladb/scylla-operator/pull/2062#pullrequestreview-2233262172)_
> 
> The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=scylladb%2Fscylla-operator).
> 
> The pull request process is described [here](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process)

https://github.com/scylladb/scylla-operator/issues/1426#issuecomment-2288682820
/test images
/retest

PR #2066, opened 2024-08-07
[WIP] Migrate controllers to v1alpha1.ScyllaDBDatacenter

zimnx:
WIP

/cc


pkg/controller/scyllaclustermigration/sync.go
```diff
@@ -0,0 +1,226 @@
...
+		},
+	)
+	if err != nil {
+		objectErrs = append(objectErrs, err)
+	}
+
+	objectErr := utilerrors.NewAggregate(objectErrs)
+	if objectErr != nil {
+		return objectErr
+	}
```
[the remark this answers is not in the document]
tnozicka: nit: it might be nice to refactor all the releases into a function

pkg/controller/scyllaclustermigration/sync.go
```diff
@@ -0,0 +1,226 @@
...
+
+	sdc, err := scmc.scyllaDBDatacenterLister.ScyllaDBDatacenters(namespace).Get(name)
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
```
tnozicka, line 199: this needs to check the controller ref before counting it into the status - generally the `controllerhelpers.GetObjects` pattern and propagate it further

pkg/controller/scyllaclustermigration/migrate_scyllacluster.go
```diff
@@ -0,0 +1,72 @@
...
+
+	sdc, upgradeContext, err := migrateV1ScyllaClusterToV1Alpha1ScyllaDBDatacenter(sc)
+	if err != nil {
+		return progressingConditions, fmt.Errorf("can't migrate scyllav1.ScyllaCluster to v1alpha1.ScyllaDBDatacenter: %w", err)
+	}
+
+	existingSDc, err := scmc.scyllaDBDatacenterLister.ScyllaDBDatacenters(sc.Namespace).Get(sc.Name)
+	if err != nil && !errors.IsNotFound(err) {
+		return progressingConditions, fmt.Errorf("can't get v1alpha1.ScyllaDBDatacenter %q: %w", naming.ManualRef(sc.Namespace, sc.Name), err)
+	}
```
tnozicka, line 30: this should read the map of owned objects

pkg/controller/scyllaclustermigration/migrate_scyllacluster.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Non-nil upgrade context means Operator was updated during upgrade of ScyllaCluster.
+	// Before we create the migrated ScyllaDBDatacenter, we have to make sure the ongoing upgrade state is migrated to new ConfigMap which holds it.
```
tnozicka, line 33: What if we have already migrated the ConfigMap, then the operator was rolled back, the state has changed, then a new operator version was redeployed and the state on ScyllaCluster.status and the ConfigMap differ?

tnozicka: Don't we need to migrate the state but block the new state machine, delete the old state and enable the new state machine?

pkg/controller/scyllaclustermigration/migrate_scyllacluster.go
```diff
@@ -0,0 +1,72 @@
...
+	// Once ScyllaDBDatacenter is created, it becomes an authoritative upgrade state.
```
tnozicka, line 34: Why don't we just generate the desired ScyllaDBDatacenter and desired ConfigMap and apply it?

- if ScyllaCluster has an upgrade state, it should be projected into the CM (but paused)
- if CM apply passes, delete the ScyllaCluster upgrade state
- if ScyllaCluster doesn't have an upgrade state, unpause the CM state machine, retain the existing state there (similar to say svcClusterIP)

tnozicka: or we can pause it on the ScyllaDBDatacenter which would allow creating it before the CM but it feels like the split between the 2 resources complicates things

pkg/controller/scyllaclustermigration/status.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		klog.ErrorS(err, "can't list configmaps", "Namespace", sdc.Namespace, "Selector", naming.ClusterSelector(sdc))
+		return scyllav1.ScyllaClusterStatus{}
+	}
+
+	services, err := scmc.serviceLister.Services(sdc.Namespace).List(naming.ClusterSelector(sdc))
+	if err != nil {
+		klog.ErrorS(err, "can't list services", "Namespace", sdc.Namespace, "Selector", naming.ClusterSelector(sdc))
+		return scyllav1.ScyllaClusterStatus{}
+	}
```
tnozicka, line 52: why don't we check the ownerRefs on these?

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -213,17 +239,34 @@ func servicePorts(cluster *scyllav1.ScyllaCluster) []corev1.ServicePort {
...
+				return nil, fmt.Errorf("can't parse alternator port annotation %q: %w", alternatorPortAnnotation, err)
+			}
+		}
+		enableHTTPAnnotation, ok := sdc.Annotations[naming.TransformScyllaClusterToScyllaDBDatacenterInsecureEnableHTTPAnnotation]
+		if ok {
+			enableHTTP, err = strconv.ParseBool(enableHTTPAnnotation)
+			if err != nil {
+				return nil, fmt.Errorf("can't parse enable http annotation %q: %w", enableHTTPAnnotation, err)
+			}
+		}
```
tnozicka, line 265: nit: would be nice to make these into helper functions so they can be reused and test them

rzetelskik:
[Proposal's test plan](https://github.com/scylladb/scylla-operator/blob/master/enhancements/proposals/1978-scylladbdatacenter-api/README.md#test-plan) mentions an `Additional E2E to validate converting controller` - shouldn't it be a part of this PR?

pkg/controller/scyllacluster/status.go
```diff
@@ -1,165 +0,0 @@
...
-	if desiredRackVersion != actualRackVersion {
-		controllerhelpers.SetRackCondition(status, scyllav1.RackConditionTypeUpgrading)
-	}
-
-	// Update Scaling Down condition
-	for _, svc := range serviceMap {
-		// Check if there is a decommission in progress
-		if _, ok := svc.Labels[naming.DecommissionedLabel]; ok {
-			// Add MemberLeaving Condition to rack status
-			controllerhelpers.SetRackCondition(status, scyllav1.RackConditionTypeMemberLeaving)
```
rzetelskik, line 113: I don't see a counterpart of this in the updated controller (only in the migration) - was that left out on purpose?

rzetelskik (commented):
first batch

pkg/controller/scylladbdatacenter/status.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return version, nil
+}
+
+// calculateRackStatus calculates a status for the rack.
+// sts and old status may be nil.
+func (scc *Controller) calculateRackStatus(sdc *scyllav1alpha1.ScyllaDBDatacenter, sts *appsv1.StatefulSet) *scyllav1alpha1.RackStatus {
```
rzetelskik, line 61: Leaving a note here that rack status conditions are dropped from the new API as it wasn't described in the proposal so it confused me for a bit. @zimnx explained it was on purpose as it can still be concluded from the status. I'm fine with the change personally.
fyi @tnozicka

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -394,7 +385,12 @@ func (scc *Controller) pruneStatefulSets(
 			continue
 		}
 
-		delete(status.Racks, rackName)
+		_, idx, ok := slices.Find(status.Racks, func(rackStatus scyllav1alpha1.RackStatus) bool {
+			return rackStatus.Name == rackName
+		})
+		if ok {
+			status.Racks = append(status.Racks[:idx], status.Racks[idx+1:]...)
+		}
```
rzetelskik, line 393: nit: FilterOut?

pkg/controller/scylladbdatacenter/status.go
```diff
@@ -0,0 +1,138 @@
...
+
+	status.Name = sts.Labels[naming.RackNameLabel]
+	status.Nodes = pointer.Ptr(*sts.Spec.Replicas)
+	status.ReadyNodes = pointer.Ptr(sts.Status.ReadyReplicas)
+	status.AvailableNodes = pointer.Ptr(sts.Status.AvailableReplicas)
+	status.UpdatedNodes = pointer.Ptr(sts.Status.UpdatedReplicas)
+	status.CurrentNodes = pointer.Ptr(sts.Status.CurrentReplicas)
+	status.Stale = pointer.Ptr(sts.Status.ObservedGeneration < sts.Generation)
+
+	scyllaDBImageVersion, err := naming.ImageToVersion(sdc.Spec.ScyllaDB.Image)
```
rzetelskik, line 83: this is going to fail for non-tagged image references, or even tagged ones with a digest - wasn't the new api supposed to support any image reference format?

pkg/controller/scyllaclustermigration/controller.go
```diff
@@ -0,0 +1,570 @@
...
+}
+
+func (scmc *Controller) deleteScyllaCluster(obj interface{}) {
+	scmc.handlers.HandleDelete(
+		obj,
+		scmc.handlers.Enqueue,
+	)
+}
+
+func (scmc *Controller) enqueueIfScyllaClusterIsOwner() controllerhelpers.EnqueueFuncType {
```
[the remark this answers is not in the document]
zimnx: EnqueueOwner enqueues owner of the object, here we want to do it only if it's owned by ScyllaCluster Kind (not yet migrated resources)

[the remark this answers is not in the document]
rzetelskik: > EnqueueOwner enqueues owner of the object

only if it matches the controller's kind, and ScyllaCluster is your controller's kind, no?

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// externalSeeds specifies the external seeds to propagate to ScyllaDB binary on startup as "seeds" parameter of seed-provider.
+	ExternalSeeds []string `json:"externalSeeds,omitempty"`
+
+	// alternatorOptions designates this cluster an Alternator cluster.
+	// +optional
+	AlternatorOptions *AlternatorOptions `json:"alternatorOptions,omitempty"`
+
+	// additionalScyllaDBArguments will be appended to the ScyllaDB binary during startup.
+	// When set, ScyllaDB may behave unexpectedly, and every such setup is considered unsupported.
```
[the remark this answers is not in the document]
zimnx: It's useful, for example changing log level wouldn't make entire cluster unsupported, it's just a message to discourage users from using it if they don't know consequences.

pkg/controller/scylladbdatacenter/status.go
zimnx: This PR doesn't extend supported image types. It prepares API to allow it.

pkg/controller/scyllaclustermigration/sync.go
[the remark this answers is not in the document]
zimnx: changed to releaseErrs. imo relasing part is consistent with other controllers how sync functions are structured so leaving as is

pkg/controller/scyllaclustermigration/sync.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (scmc *Controller) sync(ctx context.Context, key string) error {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		klog.ErrorS(err, "Failed to split meta namespace cache key", "cacheKey", key)
+		return err
+	}
+
+	startTime := time.Now()
+	klog.V(4).InfoS("Started migrating ScyllaCluster", "ScyllaCluster", klog.KRef(namespace, name), "startTime", startTime)
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllaclustermigration/controller.go
```diff
@@ -0,0 +1,570 @@
+// Copyright (c) 2024 ScyllaDB.
+
+package scyllaclustermigration
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllaclustermigration/sync.go
zimnx: done

pkg/controller/scyllaclustermigration/migrate_scyllacluster.go
zimnx: done

pkg/controller/scyllaclustermigration/status.go
zimnx: added

pkg/controller/scylladbdatacenter/resource.go
zimnx: there are unit tests targeted at these branches. I don't see any benefits of extracting it, these won't be used elsewhere

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ScyllaDBDatacenterSpec defines the desired state of ScyllaDBDatacenter.
+type ScyllaDBDatacenterSpec struct {
+	// metadata controls shared metadata for all pods created based on this spec.
+	// +optional
+	Metadata ObjectTemplateMetadata `json:"metadata,omitempty"`
+
+	// clusterName specifies the name of the scylla cluster.
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// scyllaDBManagerAgent specifies ScyllaDB Manager Agent properties for this rack.
+	// These override the settings set on Datacenter level.
+	// +optional
+	ScyllaDBManagerAgent *ScyllaDBManagerAgentTemplate `json:"scyllaDBManagerAgent"`
+}
+
+// ScyllaDBTemplate allows overriding a subset of ScyllaDB settings.
+type ScyllaDBTemplate struct {
+	// resources requirements for the ScyllaDB container
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// NodeServiceTypeLoadBalancer means nodes will be exposed via LoadBalancer Service.
+	NodeServiceTypeLoadBalancer NodeServiceType = "LoadBalancer"
+)
+
+type NodeServiceTemplate struct {
+	ObjectTemplateMetadata `json:",inline"`
+
+	// type is the Kubernetes Service type.
+	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scylladbdatacenter/controller.go
```diff
@@ -138,29 +138,29 @@ func NewController(
...
 
-		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "scyllacluster"),
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "scylladbdatacenter"),
 
 		keyGetter: keyGetter,
 	}
 
 	var err error
-	scc.handlers, err = controllerhelpers.NewHandlers[*scyllav1.ScyllaCluster](
+	scc.handlers, err = controllerhelpers.NewHandlers[*scyllav1alpha1.ScyllaDBDatacenter](
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scylladbdatacenter/sync_statefulsets.go
zimnx: done

zimnx:
> I didn't find ValidateScyllaDBDatacenter - did I miss it?

No, it was missing. Thanks for noticing. Added it

> [Proposal's test plan](https://github.com/scylladb/scylla-operator/blob/master/enhancements/proposals/1978-scylladbdatacenter-api/README.md#test-plan) mentions an Additional E2E to validate converting controller - shouldn't it be a part of this PR?

Conversion is tested in unit tests, e2e's were extended to verify ownership chain. Imo when existing e2e's logic weren't touched and there's unit test coverage for conversion, there's nothing to test in e2e.

pkg/controller/scyllaclustermigration/migrate_scyllacluster.go
zimnx: > What if we have already migrated the ConfigMap, then the operator was rolled back, the state has changed, then a new operator version was redeployed and the state on ScyllaCluster.status and the ConfigMap differ?

This scenario assumes there're no additional steps required during rollback. Since state machine authoritative source changes when Operator is rolled back, users would need to manually migrate it back to ScyllaCluster status. 

> Don't we need to migrate the state but block the new state machine, delete the old state and enable the new state machine?

We can't delete the old state, as currently it's visible to the user and it needs to stay like that.
Delaying of ScyllaDBDatacenter creation is essentially a pause of the new state machine. As part of rollback procedure, users have to orphan delete ScyllaDBDatacenter to allow dependants to be adopted by the ScyllaCluster, so blocking would prevent state skew in case of redeployment.

zimnx:
Flake - https://github.com/scylladb/scylla-operator/issues/2096
/retest

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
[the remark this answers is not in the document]
tnozicka: I agree with @zimnx - the purpose of the field is to get some extra flexibility, maybe for simple things or things that are not implemented yet. The "unsupported" means you can do crazy things, conflict with the operator but we won't solve your issue if it doesn't work.

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -32,90 +33,30 @@ var (
...
-		}
-	}
-
-	for _, ip := range opts.AdditionalIPAddresses {
-		for _, msg := range apimachineryutilvalidation.IsValidIP(ip) {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("additionalIPAddresses"), opts.AdditionalIPAddresses, msg))
-		}
+	_, err := scyllacluster.MigrateV1ScyllaClusterSpecToV1Alpha1ScyllaDBDatacenterSpec(c.Name, c.Spec)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(field.NewPath("spec"), fmt.Errorf("can't migrate v1.ScyllaCluster.Spec to v1alpha1.ScyllaDBDatacenter.Spec: %w", err)))
```
tnozicka, line 56: this looks prone to tightening the validation which is strictly forbidden - does it need to retun the error?

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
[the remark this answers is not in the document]
rzetelskik: Ok either way, but maybe the comment could be extended with this explanation (i.e. saying that it might conflict with the operator etc.) and saying something along the lines of consider using customConfigMapRef instead, if possible?

pkg/controller/scylladbdatacenter/status.go
rzetelskik: It could use at least a todo then

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -160,7 +117,9 @@ func ValidateAlternatorSpec(alternator *scyllav1.AlternatorSpec, fldPath *field.
 func ValidateScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	rackNames := sets.NewString()
+	allErrs = append(allErrs, ValidateStructSliceFieldUniqueness(spec.Datacenter.Racks, func(rackSpec scyllav1.RackSpec) string {
```
rzetelskik, line 120: can we have the changes not strictly related to API migration extracted to a separate PR? (common funcs, refactoring etc) It would make it easier to review the validation of the new object

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+	// rackTemplate provides a template for every rack.
+	// Every rack inherits properties specified in the template, unless it's overwritten on the rack level.
+	// +optional
+	RackTemplate *RackTemplate `json:"rackTemplate,omitempty"`
+
+	// racks specify the racks in the datacenter.
+	Racks []RackSpec `json:"racks"`
+
+	// disableAutomaticOrphanedNodeReplacement controls if automatic orphan node replacement should be disabled.
+	DisableAutomaticOrphanedNodeReplacement bool `json:"disableAutomaticOrphanedNodeReplacement,omitempty"`
```
rzetelskik, line 65: shouldn't it be marked as optional? From api convention: `Note that for backward compatibility, any field that has the omitempty struct tag will be considered to be optional, but this may change in the future and having the +optional comment tag is highly recommended.`

pkg/api/scylla/validation/cluster_validation.go
rzetelskik: discussed offline, resolving

helm/scylla-operator/templates/validatingwebhook.yaml
```diff
@@ -34,3 +34,4 @@ webhooks:
     - UPDATE
     resources:
     - nodeconfigs
+    - scylladbdatacenters
```
tnozicka, line 37: which e2e test makes sure the webhook is actually correctly setup? I don't see any webhook tests for ScyllaDBDatacenter (at least in the folder structure)

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// placement describes restrictions for the nodes ScyllaDB is scheduled on.
+	// +optional
+	Placement *Placement `json:"placement,omitempty"`
+
+	// topologyLabelSelector specifies a label selector which will be used to target nodes at specified topology constraints.
+	// Datacenter topologyLabelSelector is merged with rack topologyLabelSelector and then converted into nodeAffinity
+	// targeting nodes having specified topology.
+	// +optional
+	TopologyLabelSelector map[string]string `json:"topologyLabelSelector,omitempty"`
```
tnozicka, line 115: [In the proposal](https://github.com/scylladb/scylla-operator/blame/2870781/enhancements/proposals/1978-scylladbdatacenter-api/README.md#L288) and say in DaemonSets this is `*metav1.LabelSelector` - why the change?

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+	OperatorManagedOptions *OperatorManagedTLSCertificateOptions `json:"operatorManagedOptions,omitempty"`
+}
+
+// AlternatorOptions holds Alternator settings.
+type AlternatorOptions struct {
+	// writeIsolation specifies the isolation level.
+	WriteIsolation string `json:"writeIsolation,omitempty"`
+
+	// servingCertificate references a TLS certificate for serving secure traffic.
+	// +kubebuilder:default:={type:"OperatorManaged"}
```
tnozicka, line 260: let's drop these in favour of webhook validation (globally)
We have agreed to proceed with webhooks for the future and mixing the 2 kind of sucks.

Also this validation style can't be unit tested - you'd have to do the testing matrix in e2e.

pkg/controller/scyllacluster/migrate.go
```diff
@@ -0,0 +1,431 @@
...
+				Labels:      scSpec.PodMetadata.Labels,
+				Annotations: scSpec.PodMetadata.Annotations,
+			}
+		}(),
+		ClusterName:    scName,
+		DatacenterName: scSpec.Datacenter.Name,
+		ScyllaDB: scyllav1alpha1.ScyllaDB{
+			// v1.ScyllaCluster doesn't provide API to specify cluster/datacenter wide properties of scylladb.
+			ScyllaDBTemplate: scyllav1alpha1.ScyllaDBTemplate{},
+			Image:            fmt.Sprintf("%s:%s", scSpec.Repository, scSpec.Version),
```
tnozicka, line 43: version can be empty string which will make this and invalid reference ending with `:`

pkg/controller/scyllacluster/migrate.go
```diff
@@ -0,0 +1,431 @@
...
+					}(),
+				}
+			}(),
+			AdditionalScyllaDBArguments: strings.Split(scSpec.ScyllaArgs, " "),
+			EnableDeveloperMode:         pointer.Ptr(scSpec.DeveloperMode),
+		},
+		ScyllaDBManagerAgent: &scyllav1alpha1.ScyllaDBManagerAgent{
+			// v1.ScyllaCluster doesn't provide API to specify cluster/datacenter wide properties of scylladb manager agent.
+			ScyllaDBManagerAgentTemplate: scyllav1alpha1.ScyllaDBManagerAgentTemplate{},
+			Image:                        pointer.Ptr(fmt.Sprintf("%s:%s", scSpec.AgentRepository, scSpec.AgentVersion)),
```
tnozicka, line 85: and here

pkg/controller/scyllacluster/migrate.go
```diff
@@ -0,0 +1,431 @@
...
+								}
+							}(),
+						},
+					}
+				}(),
+			}
+		}(),
+		// v1.ScyllaCluster doesn't provide API to specify cluster/datacenter wide properties of scylladb.
+		RackTemplate: nil,
+		Racks: func() []scyllav1alpha1.RackSpec {
```
tnozicka, line 166: nit: maybe some of these big conversion sub-functions could be extracted and uni-tested separately

pkg/controller/scyllacluster/status.go
```diff
@@ -38,128 +37,60 @@ func (scc *Controller) updateStatus(ctx context.Context, currentSC *scyllav1.Scy
...
-		ReadyMembers:            0,
-		AvailableMembers:        pointer.Ptr(int32(0)),
-		UpdatedMembers:          pointer.Ptr(int32(0)),
-		Stale:                   pointer.Ptr(true),
-	}
-
-	// Persist ReplaceAddressFirstBoot.
-	if oldRackStatus != nil && oldRackStatus.ReplaceAddressFirstBoot != nil {
-		status.ReplaceAddressFirstBoot = oldRackStatus.DeepCopy().ReplaceAddressFirstBoot
+		klog.ErrorS(err, "can't list configmaps", "Namespace", sdc.Namespace, "Selector", naming.ClusterSelector(sdc))
```
tnozicka, line 53: why isn't this error returned?

pkg/controller/scyllacluster/sync.go
```diff
@@ -44,304 +47,193 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
...
+
 	scSelector := labels.SelectorFromSet(labels.Set{
 		naming.ClusterNameLabel: sc.Name,
 	})
 
-	type CT = *scyllav1.ScyllaCluster
-	var objectErrs []error
-
-	statefulSetMap, err := controllerhelpers.GetObjects[CT, *appsv1.StatefulSet](
+	err = controllerhelpers.ReleaseObjects[CT, *appsv1.StatefulSet](
```
tnozicka, line 57: I wonder, if you were to create ScyllaDBDatacenter first and than release the objects - would that avoid the case where the objects can be left orphaned around or are there issues with that approach? The Three rules of controllers should prevent any unwanted action with overlapping selectors.

pkg/controller/scylladbdatacenter/status.go
zimnx: It's not the only place requiring changing and which are is not obvious.

pkg/api/scylla/validation/cluster_validation.go
zimnx: indeed, removed.

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
zimnx: changed to pointer and marked optional

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// scyllaDB specifies ScyllaDB properties for this rack.
+	// These override the settings set on Datacenter level.
+	// +optional
+	ScyllaDB *ScyllaDBTemplate `json:"scyllaDB"`
+
+	// scyllaDBManagerAgent specifies ScyllaDB Manager Agent properties for this rack.
+	// These override the settings set on Datacenter level.
+	// +optional
+	ScyllaDBManagerAgent *ScyllaDBManagerAgentTemplate `json:"scyllaDBManagerAgent"`
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/validation/scylladbdatacenter_validation.go
```diff
@@ -0,0 +1,322 @@
...
+
+func ValidateScyllaDBDatacenter(sdc *scyllav1alpha1.ScyllaDBDatacenter) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, ValidateScyllaDBDatacenterSpec(&sdc.Spec, field.NewPath("spec"))...)
+
+	return allErrs
+}
+
+func ValidateScyllaDBDatacenterSpec(spec *scyllav1alpha1.ScyllaDBDatacenterSpec, fldPath *field.Path) field.ErrorList {
```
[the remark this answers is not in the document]
zimnx: There's nothing to validate

helm/scylla-operator/templates/validatingwebhook.yaml
zimnx: added e2e

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
zimnx: `metav1.LabelSelector` is a complex field, I had `map[string]string` in mind. Examples presented in proposals are using correct type.

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
zimnx: This is default, we don't have mutating webhook so defaulting is done in the OpenAPI schema.

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ScyllaDBDatacenterSpec defines the desired state of ScyllaDBDatacenter.
+type ScyllaDBDatacenterSpec struct {
+	// metadata controls shared metadata for all pods created based on this spec.
+	// +optional
+	Metadata ObjectTemplateMetadata `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
zimnx: done

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
```diff
@@ -0,0 +1,528 @@
...
+	DNSPolicy *corev1.DNSPolicy `json:"dnsPolicy,omitempty"`
+
+	// dnsDomains specifies a list of DNS domains this cluster is reachable by.
+	// These domains are used when setting up the infrastructure, like certificates.
+	// +optional
+	DNSDomains []string `json:"dnsDomains,omitempty"`
+
+	// forceRedeploymentReason specifies the latest redeployment reason.
+	// Can be used to force a rolling restart of all racks in this DC by providing a unique string.
+	// +optional
```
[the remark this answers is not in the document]
zimnx: changed to pointer

pkg/cmd/operator/operator.go
```diff
@@ -267,21 +268,40 @@ func (o *OperatorOptions) run(ctx context.Context, streams genericclioptions.IOS
...
 		o.OperatorImage,
 		o.CQLSIngressPort,
 		rsaKeyGenerator,
 	)
 	if err != nil {
-		return fmt.Errorf("can't create scyllacluster controller: %w", err)
+		return fmt.Errorf("can't create scylladbdatacenter controller: %w", err)
+	}
+
+	scmc, err := scyllacluster.NewController(
```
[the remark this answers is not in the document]
zimnx: done

pkg/cmd/operator/operator.go
```diff
@@ -267,21 +268,40 @@ func (o *OperatorOptions) run(ctx context.Context, streams genericclioptions.IOS
...
+		kubeInformers.Rbac().V1().RoleBindings(),
+		kubeInformers.Apps().V1().StatefulSets(),
+		kubeInformers.Policy().V1().PodDisruptionBudgets(),
+		kubeInformers.Networking().V1().Ingresses(),
+		kubeInformers.Batch().V1().Jobs(),
+		scyllaInformers.Scylla().V1().ScyllaClusters(),
+		scyllaInformers.Scylla().V1alpha1().ScyllaDBDatacenters(),
+	)
+	if err != nil {
+		return fmt.Errorf("can't create scyllacluster migration controller: %w", err)
```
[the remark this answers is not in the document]
zimnx: done

pkg/controller/scyllacluster/migrate.go
```diff
@@ -0,0 +1,431 @@
...
+	"github.com/scylladb/scylla-operator/pkg/internalapi"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/pointer"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+)
+
+func MigrateV1ScyllaClusterSpecToV1Alpha1ScyllaDBDatacenterSpec(scName string, scSpec scyllav1.ScyllaClusterSpec) (scyllav1alpha1.ScyllaDBDatacenterSpec, error) {
```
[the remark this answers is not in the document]
zimnx: I recall you had objections for calling it `converting`. I'm sticking with migrate.

pkg/controller/scyllacluster/migrate.go
zimnx: added error handling

pkg/controller/scyllacluster/migrate.go
zimnx: added error handling

pkg/controller/scyllacluster/status.go
zimnx: This is status, it can't return errors

pkg/controller/scyllacluster/sync.go
zimnx: I guess it would be the same, PR version is cleaner in my opinion.

pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -32,90 +33,30 @@ var (
...
+		scyllav1alpha1.BroadcastAddressTypePodIP,
+		scyllav1alpha1.BroadcastAddressTypeServiceClusterIP,
+		scyllav1alpha1.BroadcastAddressTypeServiceLoadBalancerIngress,
+	}
+
 	schedulerTaskSpecCronParseOptions = cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor
 )
 
 func ValidateScyllaCluster(c *scyllav1.ScyllaCluster) field.ErrorList {
-	return ValidateScyllaClusterSpec(&c.Spec, field.NewPath("spec"))
```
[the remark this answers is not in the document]
zimnx: it's a tradeoff between copy pasting the same logic in multiple places. I tried not to add any extra rules to common functions, if you spot one please point it out.

zimnx:
@rzetelskik @tnozicka  I think I addressed all your comments, ptal

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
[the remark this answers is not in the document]
zimnx: added mention about customConfigMapRef

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
rzetelskik: I don't see it as a pointer still - @zimnx are you sure you pushed the changes?

pkg/api/scylla/validation/scylladbdatacenter_validation.go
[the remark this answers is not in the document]
rzetelskik: ok

pkg/controller/scylladbdatacenter/status.go
rzetelskik: ok

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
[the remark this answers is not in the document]
rzetelskik: all good now, thx

pkg/controller/scyllacluster/status.go
tnozicka: it can, the workque doesn't care who it is

rzetelskik (approved):
lgtm
/assign tnozicka

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
tnozicka: ok, lets keep it

pkg/api/scylla/validation/cluster_validation.go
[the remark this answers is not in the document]
tnozicka: > if you spot one please point it out.

It's hard to say from the diff, that's kind of the point. I get the common logic but if we risk breaking backwards compatibility vs. duplicating the code, I'd lean to duplicating it. @rzetelskik WDYT?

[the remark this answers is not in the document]
rzetelskik: I agree it's hard to review which makes it easy to miss a subtle change - I haven't seen any, but that's hardly enough of a proof there are none. I don't care about code duplication much, especially that we'll slowly start moving away from the old API either way.

pkg/controller/scyllacluster/status.go
zimnx: Then what you had in mind when you added this comment? It seems right to always suceed for status calculation, and reflect errors within.
https://github.com/scylladb/scylla-operator/blame/master/pkg/controller/scyllacluster/status.go#L145-L147

pkg/api/scylla/validation/cluster_validation.go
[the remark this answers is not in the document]
zimnx: Removed changes from cluster validation

pkg/controller/scyllacluster/status.go
tnozicka: There is a difference between objects missing and a generic List failure - if we have broken listers / clients / REST calls - it should surface an error in the controller logs and also retry.

Effectively, you can always fail before the status calculation when listing / adopting objects. Maybe it would be better if you'd list the objects outside this function and supply it as an argument, similar to controlled objects elsewhere?

pkg/controller/scylladbdatacenter/status.go
tnozicka: They seem to be deduced when mapping the status from ScyllaDBDatacenter to ScyllaCluster so compatibility wise that should be ok, I am fine not having them on the new resource at this point.

pkg/api/scylla/validation/scylladbdatacenter_validation.go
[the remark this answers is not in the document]
tnozicka: not sure I follow - there seems to be a plenty of stuff to validate in RackTemplate like:
- non-negative number of nodes
- all the types under Placement
- valid topologyLabelSelector
- resources (similar to https://github.com/kubernetes/kubernetes/blob/8db07446f199387d644212770d0e04e7271f661f/pkg/apis/core/validation/validation.go#L6700)
- dnsLabels for customConfigMapRef, .....

pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go
tnozicka: don't we want to support set based requirements?
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

zimnx: Whole point of topologyLabelSelector was simplicity, as most usecases for topology is setting a single label requirement. Users wanting to use expressions and other complex rules have to use what's available in placement.

pkg/controller/scyllacluster/status.go
zimnx: moved it to caller

pkg/api/scylla/validation/scylladbdatacenter_validation.go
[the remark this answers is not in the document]
zimnx: added validation to couple of fields from rackTemplate, for resources and placement validating functions are only available within kubernetes/kubernetes, not in staging repositories so i didn't validate them. 
We can't validate rack's as we cannot restrict existing validation, and scylladbdatacenter racks are projected from ScyllaCluster.

PR #2074, opened 2024-08-13
Add scylla-manager-agent readiness probe

### LINKED ISSUE #2060: scylla-manager-agent container misses a readiness probe
scylla-manager-agent container misses a readiness probe which makes it go ready immediately and doesn't provide any signal about the container state.

https://github.com/scylladb/scylla-operator/blob/e929ed/pkg/controller/scyllacluster/resource.go#L914-L959

We should add at least a basic readiness probe to check it really started which helps with observability and debugging.

tnozicka:
This PR adds a readiness probe for scylla-manager-agent to better project the container state.

Resolves #2060


zimnx (approved):
lgtm
/assign rzetelskik

rzetelskik (approved):
/lgtm

Issue #2071, opened 2024-08-13
Allow overriding embedded image references for testing 


tnozicka:
We should allow overriding the embedded image references to e.g. Grafana or Prometheus for testing purposes. ScyllaOperatorConfig seems like a fitting place.

For production purposes, we need to rely on exactly the same content of the referenced images to be present to avoid incompatibilities breakages so this should only be adjusted for testing purposes. All the fields should be behind `unsupported` prefix.

Given the components used in the stack can change, all such fields should be optional.

tnozicka:
done in #2081

Issue #2061, opened 2024-08-06
[Flake] Scylla Manager integration should register cluster, sync backup tasks and support manual restore procedure [It] using default ScyllaDB version [RequiresObjectStorage]


tnozicka:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2055/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1820487766287323136#1:test-build-log.txt%3A1002

### Snippet of what failed.
```
   [FAILED] %!(EXTRA string=Error: create restore target, units and views: init target: look for schema file: get schema file from gcs:so-e490a80d-2b84-4d9d-a784-d0d9d7c27ee3/backup/schema: many schema files with sm_20240805161410UTC snapshot tag: [cluster/6f019377-acd2-4f23-9aad-d6cb21a05008/task_a7b7283e-c7eb-4830-a975-4fd5d16f19af_tag_sm_20240805161410UTC_schema_with_internals.json.gz cluster/82d8dd25-42fc-4c83-87a9-1e76a0e6a15e/task_d24f0a86-f287-4e04-87f8-46220bddda35_tag_sm_20240805161410UTC_schema_with_internals.json.gz]
  Trace ID: EmDCY15VQlG_ct5tWy6W_A (grep in scylla-manager logs)
  )
  Unexpected error:
      <exec.CodeExitError>: 
      command terminated with exit code 1
      {
          Err: <*errors.errorString | 0xc000c7c180>{
              s: "command terminated with exit code 1",
          },
          Code: 1,
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager_object_storage.go:361 @ 08/05/24 16:15:57.416
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.init.func21.1({{0x0, 0x0}, {0x0, 0x0}, 0x0, 0x0})
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager_object_storage.go:361 +0x4b47
    reflect.Value.call({0x25ea820?, 0xc000b821e0?, 0x13?}, {0x2d32c3a, 0x4}, {0xc000b2be18, 0x1, 0x1?})
    	reflect/value.go:596 +0xca6
[...]

PR #2081, opened 2024-08-16
Manage global configs using ScyllaOperatorConfig, add unsupported overrides and extract defaults to a common config

### LINKED ISSUE #1589: Add configuration file to specify versions and images so it's easy to update
We need a config file in the repo that defines :
- default scylladb version in our manifests
- manager version
- any fixed images the operator deploys
- scylla-operator version reference (we half do it in the makefile already)
- possibly others
and uses generation to put it into the correct places (or assets that read it directly)

This will allow us to have a single palce to review it and possibly add a periodic that checks against latest available versions for master and notifies us so we don't forget to bump.

tnozicka:
This PR extends the ScyllaOperatorConfig controller to properly report status and the images that are used and wires the rest of the controllers to consume its status. It also wires *unsupported* overrides for the images used which can be used for testing or temporarily to help disconnected installs. (Users are responsible to make sure it's the same image version when touching the settings.)

It also extracts all image and version settings to a common place in `assets/config/config.yaml`.

The only place left to manually update versions is the documentation. Maybe in the future we can let it include real examples and sync the versions there automatically as well.

Resolves #1589 #2071

#### Requires
- [x] #2091


tnozicka:
@zimnx @rzetelskik rebased to pick up the prerequisites and ready

examples/eks/cluster.yaml
```diff
@@ -16,88 +14,88 @@ spec:
   agentVersion: 3.3.0
   version: 6.1.0
   sysctls:
-  - "fs.aio-max-nr=2097152"
+    - fs.aio-max-nr=2097152
```
rzetelskik, line 17: Iirc yq doesn't support configuring this but could we maybe use a workaround to avoid all these indentations changes in arrays? Especially that it goes against our supposed formatting rule

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
```diff
@@ -0,0 +1,140 @@
...
+					Type:     field.ErrorTypeRequired,
+					Field:    "spec.unsupportedPrometheusVersionOverride",
+					BadValue: "",
+					Detail:   "version can't be empty",
+				},
+			},
+			expectedErrorString: `spec.unsupportedPrometheusVersionOverride: Required value: version can't be empty`,
+		},
+		{
+			name: "UnsupportedPrometheusImageOverride can't contain only spaces",
```
rzetelskik, line 93: s/UnsupportedPrometheusImageOverride/UnsupportedPrometheusVersionOverride

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
```diff
@@ -0,0 +1,140 @@
...
+					Type:     field.ErrorTypeInvalid,
+					Field:    "spec.unsupportedPrometheusVersionOverride",
+					BadValue: " ",
+					Detail:   "version contains only spaces",
+				},
+			},
+			expectedErrorString: `spec.unsupportedPrometheusVersionOverride: Invalid value: " ": version contains only spaces`,
+		},
+		{
+			name: "UnsupportedPrometheusImageOverride accepts valid version",
```
rzetelskik, line 110: ditto

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
```diff
@@ -0,0 +1,140 @@
...
+					Detail:   "unable to parse image: invalid reference format",
+				},
+			},
+			expectedErrorString: `spec.unsupportedGrafanaImageOverride: Invalid value: " ": unable to parse image: invalid reference format`,
+		},
+		{
+			name: "UnsupportedGrafanaImageOverride accepts valid image",
+			ScyllaOperatorConfig: &scyllav1alpha1.ScyllaOperatorConfig{
+				Spec: scyllav1alpha1.ScyllaOperatorConfigSpec{
+					UnsupportedGrafanaImageOverride: pointer.Ptr("docker.io/scylladb/scylla-operator:latest"),
```
rzetelskik, line 69: nit: add a test for non-tagged image reference as well?

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/scylladb/scylla-operator/pkg/api/scylla/validation"
+	"github.com/scylladb/scylla-operator/pkg/pointer"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestValidateScyllaOperatorConfig(t *testing.T) {
+	t.Parallel()
+
+	tt := []struct {
```
rzetelskik, line 19: it could also use a set of test cases for UnsupportedBashToolsImageOverride

pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -5,12 +5,49 @@ package v1alpha1
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 type ScyllaOperatorConfigSpec struct {
-	// scyllaUtilsImage is a Scylla image used for running scylla utilities.
-	// +kubebuilder:validation:MinLength=1
+	// scyllaUtilsImage is a ScyllaDB image used for running ScyllaDB utilities.
 	ScyllaUtilsImage string `json:"scyllaUtilsImage"`
```
rzetelskik, line 9: shouldn't this be marked as optional now? Or does that not work without being able to switch to a pointer?

examples/eks/cluster.yaml
tnozicka: I am not aware of a way to avoid it. It sucks but all the manifests edited with yq have it already.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
tnozicka: I suppose you might convert a non optional field to a optional pointer that's omit empty but we don't need to make a change on the wire to distinguish not set from empty. This just retains the existing behaviour and loosens the validation.

rzetelskik: I think that if it's possible it'd be good to do it for the sake of consistency alone

rzetelskik:
This place should probably also take the version from config: https://github.com/scylladb/scylla-operator/blob/80c490c5ea47070014f60b3998bc175e7e826102/test/e2e/set/scyllacluster/scyllacluster_replace.go#L173.

We also have two (that I found) tests that are using the enterprise image:
1. https://github.com/scylladb/scylla-operator/blob/80c490c5ea47070014f60b3998bc175e7e826102/test/e2e/set/scyllacluster/scyllacluster_replace.go#L178
2. https://github.com/scylladb/scylla-operator/blob/80c490c5ea47070014f60b3998bc175e7e826102/test/e2e/set/scyllacluster/scyllamanager_object_storage.go#L458

2 is somewhat specific because it needs a particular minor version, but did you consider adding a config field for a general enterprise version (I suppose in `operatorTests`)?

tnozicka:
> "5.2.6"

Looks like that was being left out in the past quite often, I'll wire it and update

> enterprise

I think our goal should be to have a second version of the suite and run all the tests with enterprise (using different / replaced config.yaml)

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
tnozicka: added

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
tnozicka: added

pkg/api/scylla/validation/scyllaoperatorconfig_test.go
tnozicka: fixed, thx

pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -5,12 +5,49 @@ package v1alpha1
...
+
+	// unsupportedBashToolsImageOverride allows to adjust a generic Bash image with extra tools used by the operator
+	// for auxiliary purposes.
+	// Setting this field renders your cluster unsupported. Use at your own risk.
+	// +optional
+	UnsupportedBashToolsImageOverride *string `json:"unsupportedBashToolsImageOverride,omitempty"`
```
zimnx, line 15: `Unsupported` prefix is confusing. In the past, users of fields prefixed with it (unsupportedOptions in NodeConfig mounts) didn't understand whether this field works, what it does, because of this prefix. I think it's a mistake to use it. 

I think it's enough to make it explicit in the field description, that using it should be restricted to advanced users knowing that they are doing. We can explain consequences of changing it there.
Looking at Kubernetes resources, there're none having this confusing prefix, yet there're field which users shouldn't set (Job selector for example).

tnozicka: users don't read the docs and the field name is the only thing that they see so I'd stick with it. it's very much the same like UnsupportedScyllaDBArgs

zimnx: Then why we are getting complains about quality of documentation? They do, and it's the right place to write longer descriptions. 

We don't have UnsupportedScyllaDBArgs in new API, we have AdditionalScyllaDBArgs + description.

pkg/controller/scylladbmonitoring/controller.go
```diff
@@ -261,6 +271,28 @@ func (smc *Controller) deleteScyllaDBMonitoring(obj interface{}) {
 	)
 }
 
+func (smc *Controller) addScyllaOperatorConfig(obj interface{}) {
+	smc.handlers.HandleAdd(
+		obj.(*scyllav1alpha1.ScyllaOperatorConfig),
+		smc.handlers.EnqueueOwner,
```
zimnx, line 277: It's a singleton object it doesn't have any owner, we should just enqueue all ScyllaDBMonitorings
https://github.com/scylladb/scylla-operator/blob/69375cff1653b9e40805e68372513b8e106beb40/pkg/controller/scyllaoperatorconfig/resource.go#L11-L20

pkg/api/scylla/v1alpha1/types_operatorconfig.go
tnozicka: Why? The desired state is vice versa but that can't be done for the sake of compatibility. We only have the 2 state logic here.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
tnozicka: > Then why we are getting complains about quality of documentation? They do

This is a probabilistic case. Just a few users/customers, who don't, are enough and manage to build something on top of the unsupported options and you'll get all the fires when it finally breaks no matter the docs said unsupported. At OpenShift, we have been there, see e.g.
https://github.com/openshift/api/blob/01b3675ba7b364e312ac5da6e632251fbaefdda0/operator/v1/types_ingress.go#L254
https://github.com/search?q=repo%3Aopenshift%2Fapi+unsupported&type=code&p=2
It's better to be safe then sorry. I don't intent to get paged over that.

pkg/controller/scylladbmonitoring/controller.go
tnozicka: fixed, thx

pkg/api/scylla/v1alpha1/types_operatorconfig.go
zimnx: I'm sure they will build on top of it anyway. Having confusing prefix or not is not going to stop them. That's why we should improve documentation and user understanding. Having confusing prefix doesn't help with that.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
rzetelskik: > > I think that if it's possible it'd be good to do it for the sake of consistency alone
> Why?

Why what? Why is it good to have consistency? It's an optional field now being indistinguishable from a required field by just looking at the API, it just doesn't have any validation

pkg/api/scylla/v1alpha1/types_operatorconfig.go
rzetelskik: > I'm sure they will build on top of it anyway. Having confusing prefix or not is not going to stop them.

I think this is more about having a stronger case for not carting about such cases than it is about preventing it entirely. I do however agree with @zimnx that `Unsupported` prefix is confusing, I myself didn't understand what `unsupportedOptions` in NodeConfig were supposed to stand for at first, but that's something that actually encourages you to go and check, instead of just using it. Perhaps we could find a better prefix, but I don't have a better idea myself. Between `unsupported` and dropping the prefix altogether, I'd go with the former.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
tnozicka: It used to be a required field.

> It's an optional field

No, it is still a required field. It has a special value `""` which says use the same version as the operator. It is required to still be present on the wire, as it used to be, even if empty. 

>  it just doesn't have any validation

It does have validation. It only allows an empty string as a valid value.

> Why is it good to have consistency? 

You are asking to make a wire change. Old fields stay how they were and new fields are optional by design. That's how APIs evolve. In a next API version these can be reconsidered freely.

> now being indistinguishable from a required field

This is not just a hint - it changes the wire representation. Old decoders won't be able to decode the type anymore and fail on a missing the required field, if strict. There is a high bar to do wire changes. It may be necessary some days for a higher good and limiting the focus on the major decoders and our code and the kube-apiserver and accepting that the most of the system keeps working.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
tnozicka: > That's why we should improve documentation and user understanding. Having confusing prefix doesn't help with that.

It does, the prefix is meant for people who don't reach the docs. I have a hard time undestanding how enhancing the docs helps the group of people who don't read it.

tnozicka:
https://github.com/scylladb/scylla-operator/issues/2061
/retest

rzetelskik:
> > We also have two (that I found) tests that are using the enterprise image:
> > 
> > 2. https://github.com/scylladb/scylla-operator/blob/80c490c5ea47070014f60b3998bc175e7e826102/test/e2e/set/scyllacluster/scyllamanager_object_storage.go#L458
> 
> I think our goal should be to have a second version of the suite and run all the tests with enterprise (using different / replaced config.yaml)

So what about this one? It's a special case that won't be dependent on the general version being tested - should it stay defined in the test itself? Imo leaving it there, as opposed to moving to the test config, makes it more prone to being forgotten about, but I don't feel strong about this.

Other than that lgtm, no further comments.
/assign zimnx

tnozicka:
>>  So what about this one? It's a special case that won't be dependent on the general version being tested - should it stay defined in the test itself? Imo leaving it there, as opposed to moving to the test config, makes it more prone to being forgotten about, but I don't feel strong about this.

> I think this sucks both ways as this shouldn't be in the config, if we want to switch all regular versions in the enterprise but I guess to address the concerns for the interim, I can put the there with a notion of removing it soon

done

pkg/api/scylla/v1alpha1/types_operatorconfig.go
zimnx: We shouldn't optimize for slackers, it's entirely their fault they don't read the documentation.
I can imagine someone trying to deploy 1.0 Scylla without checking supported versions - part of the documentation. Following your logic, shouldn't we add confusing prefix to `version` field to encourage them to read the docs? 

Sounds like a security by obscurity but for API.

pkg/controller/nodeconfig/controller.go
```diff
@@ -403,19 +404,26 @@ func (ncc *Controller) deleteNodeConfig(obj interface{}) {
...
 	)
 }
 
 func (ncc *Controller) updateScyllaOperatorConfig(old, cur interface{}) {
 	ncc.handlers.HandleUpdate(
 		old.(*scyllav1alpha1.ScyllaOperatorConfig),
 		cur.(*scyllav1alpha1.ScyllaOperatorConfig),
-		ncc.handlers.EnqueueAllWithUntypedFilterFunc(isManagedByNodeConfigController),
+		ncc.handlers.EnqueueAll,
 		nil,
```
zimnx, line 416: this nil is a delete function, since it's added in this PR, lets supply this one.

assets/config/config.go
```diff
@@ -0,0 +1,46 @@
...
+		return c, err
+	}())
+)
+
+type OperatorConfig struct {
+	ScyllaDBVersion string `json:"scyllaDBVersion"`
+	// scyllaDBEnterpriseVersionNeedingConsistentClusterManagementOverride sets enterprise version
+	// that requires consistent_cluster_management workaround for restore.
+	// In the future, enterprise versions should be run as a different config instance in its own run.
+	ScyllaDBEnterpriseVersionNeedingConsistentClusterManagementOverride string `json:"scyllaDBEnterpriseVersionNeedingConsistentClusterManagementOverride"`
```
rzetelskik, line 25: I would change this to `scyllaDBVersionNeedingConsistentClusterManagementOverride` - this way we'll be able to test this with 5.4.Z for OS suite and 2024.1.Z for enterprise suite

tnozicka: It uses enterprise repo
https://github.com/scylladb/scylla-operator/blob/f27f72f4273a6ef2e60b460b66987c8a5e59dae1/test/e2e/set/scyllacluster/scyllamanager_object_storage.go#L458

rzetelskik: We could configure both. It can be done once we actually have a separate suite for enterprise though.

pkg/controller/nodeconfig/controller.go
tnozicka: this will call EnqueueAll twice in a row but I guess it doesn't matter with the queue

tnozicka: done

pkg/controller/nodeconfig/sync_daemonsets.go
```diff
@@ -19,11 +21,14 @@ func (ncc *Controller) syncDaemonSet(
 	soc *scyllav1alpha1.ScyllaOperatorConfig,
 	daemonSets map[string]*appsv1.DaemonSet,
 ) error {
-	scyllaUtilsImage := soc.Spec.ScyllaUtilsImage
-	// FIXME: check that its not empty, emit event
-	// FIXME: add webhook validation for the format
+	if soc.Status.ScyllaDBUtilsImage == nil || len(*soc.Status.ScyllaDBUtilsImage) == 0 {
+		ncc.eventRecorder.Event(nc, corev1.EventTypeNormal, "MissingScyllaUtilsImage", "ScyllaOperatorConfig doesn't yet have scyllaUtilsImage available in the status.")
```
zimnx, line 25: shouldn't this result in Progressing condition instead? It's expected to be available eventually.

pkg/controller/nodeconfig/sync_daemonsets.go
```diff
@@ -19,11 +21,14 @@ func (ncc *Controller) syncDaemonSet(
...
+		return controllertools.NewNonRetriable("scylla operator config doesn't yet have scyllaUtilsImage available in the status")
+	}
+	scyllaDBUtilsImage := *soc.Status.ScyllaDBUtilsImage
```
zimnx, line 28: shouldn't this validate if generation == observedGeneration first?

test/e2e/set/scyllaoperatorconfig/config.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+)
+
+func GetDefaultSOC() *scyllav1alpha1.ScyllaOperatorConfig {
+	return &scyllav1alpha1.ScyllaOperatorConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: naming.SingletonName,
+		},
+	}
+}
```
zimnx, line 35: Test resources are kept in fixture package

pkg/controller/nodeconfig/sync_daemonsets.go
tnozicka: I think for a different controller (NCC vs. SOC) this is a degraded state - it doesn't know whether the other controller can eventually make it, or is failing.
Effectively, it should be there from the operator start time, so emitting an event seems adequate.

pkg/controller/nodeconfig/sync_daemonsets.go
tnozicka: Not really, it is a different resource and not dependent on the input + it reacts to updates. (It also avoid the possibly infinite loop by waiting for something to catch up. It's better to deploy what we have and  reconcile the updates.)

test/e2e/set/scyllaoperatorconfig/config.go
tnozicka: it's an empty definition but done

pkg/api/scylla/v1alpha1/types_operatorconfig.go
tnozicka: Without any new input I am afraid we have to agree to disagree on this one.

pkg/controller/nodeconfig/sync_daemonsets.go
zimnx: I added comment to wrong line, I meant to add Progressing condition instead returning an error which results in Degraded state.
I wouldn't call it degraded because you don't know if anything errored out, maybe it's just stale cache. Is stale cache a cause of object being degraded, I don't think so. In similar cases, our existing controllers are waiting with Progressing conditiong having set so this is inconsistent behavior.

tnozicka: > maybe it's just stale cache. Is stale cache a cause of object being degraded, I don't think so

Right after you install the operator, this field should get set on the first sync loop of SOC. Or this one upgrade. If it's missing on other cases, something is broken - not a stale cache.

> In similar cases, our existing controllers are waiting with Progressing conditiong having set so this is inconsistent behavior.

That's not exactly the same - the case you provide w.r.t. our other resources is within a single controller and dependent objects while these are 2 independent controllers (also without any parent / child relationship).

A comparable case would be to say wanting to read a `kube-root-ca.crt` provisioned by Kubernetes controllers. If it's not there, you are Degraded, not Progressing as you are not making any progress to fixing that state (either directly or indirectly). (And you don't know whether the Kubernetes controllers are progressing toward the state, or Degraded.)

tnozicka: I guess it's also a thin boundary so I can see a bit of point going both ways here.

zimnx: > That's not exactly the same - the case you provide w.r.t. our other resources is within a single controller and dependent objects while these are 2 independent controllers

What I meant for example, were cleanup Jobs waiting for HostID annotation reconciled by sidecar controller. These are two independent controllers.

> (also without any parent / child relationship).

I don't see how relationship matters. Nevertheless, ScyllaOperatorConfig is indirect parent of every resource reconciled by Operator, so they are related. Especially NodeConfigs, as since this PR SOC is part of its outputs.

> A comparable case would be to say wanting to read a kube-root-ca.crt provisioned by Kubernetes controllers. If it's not there, you are Degraded, not Progressing as you are not making any progress to fixing that state (either directly or indirectly). (And you don't know whether the Kubernetes controllers are Progressing toward the state, or Degraded.)

We don't do anything when tuning is running, nor we create cleanup jobs when HostID annotation is missing, yet we don't consider ScyllaCluster degraded but progressing.

There's artifical delay between when objects are created/updated when they need to be synchronized. That is expected and it's an error. In your case ConfigMap has a Namespace dependency so not having it immediately after Namespace creation is expected.

zimnx:
Comment threads where we don't agree are not that important to block the release. 
/lgtm

PR #2088, opened 2024-08-20
Deprecate cpuset field in v1.ScyllaCluster

### LINKED ISSUE #508: CpuSet research
Despite following comment around enabling cpuset for Scylla I tried setting in on environment with non-static CPU policy and Scylla did not crash even when cpuset was counting CPUs not allocated by the kubelet. 
https://github.com/scylladb/scylla-operator/blob/74caca42dd9d27650e9fe2ce8079b815f2835928/pkg/controllers/sidecar/config/config.go#L235-L236

Maybe we should enable it by default and depracate switch in CRD. 
Before doing that, we should check whether setting it always affect anyhow performance or stability of the cluster.

zimnx:

This deprecates `scyllacluster.spec.cpuset` field and removes mentions of it from examples and documentation.

Resolves #508

### DIFF
--- a/pkg/api/scylla/v1/types_cluster.go
+++ b/pkg/api/scylla/v1/types_cluster.go
@@ -63,7 +63,8 @@ type ScyllaClusterSpec struct {
 	// +optional
 	DeveloperMode bool `json:"developerMode,omitempty"`
 
-	// cpuset determines if the cluster will use cpu-pinning for max performance.
+	// cpuset determines if the cluster will use cpu-pinning.
+	// Deprecated: `cpuset` is deprecated and may be ignored in the future.
 	// +optional
 	CpuSet bool `json:"cpuset,omitempty"`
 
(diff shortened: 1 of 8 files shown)


pkg/api/scylla/v1/types_cluster.go
```diff
@@ -64,6 +64,7 @@ type ScyllaClusterSpec struct {
 	DeveloperMode bool `json:"developerMode,omitempty"`
 
 	// cpuset determines if the cluster will use cpu-pinning for max performance.
```
tnozicka, line 66:
```suggestion
	// cpuset determines if the cluster will use cpu-pinning.
```
if it doesn't matter for performance I'd reword it as well

zimnx: done

tnozicka (commented):
/approve
/lgtm
thansk

PR #2091, opened 2024-08-23
Collect, snapshot and gracefully restore global objects in e2e test

### LINKED ISSUE #2045: Serial e2e needs storage but requires the nodeconfig that sets it up to be removed
Our `serial` suite requires running ScyllaClusters which require `scylladb-local-xfs` CSI driver to get PV for its PVC and the storage provisioner requires NodeConfig mounts and filesystem but our serial suite requires `nodeconfig/cluster` not to be present, which meant the mounts and the fs setup and transitively the local CSI driver get broken in this cycle.

Depending on other variables you'll end up with errors like this:

```
I0731 12:35:36.456812       1 systemd/unit.go:132] "Disabling and stopping unit because it's no longer required" Name="mnt-persistent\\x2dvolumes.mount"
I0731 12:35:36.462615       1 systemd/unit.go:142] "Removing unit because it's no longer required" Name="mnt-persistent\\x2dvolumes.mount"
I0731 12:35:36.462930       1 record/event.go:376] "Event occurred" object="cluster" fieldPath="" kind="NodeConfig" apiVersion="scylla.scylladb.com/v1alpha1" type="Normal" reason="MountDeleted" message="Mount unit mnt-persistent\\x2dvolumes.mount has been deleted"
...
E0731 12:38:36.038155       1 nodetune/controller.go:175] syncing key 'key' failed: [can't sync jobs: can't make Jobs for containers: can't find data dir host path: can't inspect container "6ef99e094920f1b22a0ca395f31130071f7f5409142b62f931eac2c891fdafa9": rpc error: code = DeadlineExceeded desc = context deadline exceeded, can't update status: can't update node config status "cluster": client rate limiter Wait returned an error: context deadline exceeded]
```

```
0s          Warning   FailedMount                       pod/scylla-manager-cluster-manager-dc-manager-rack-0                MountVolume.MountDevice failed for volume "pvc-bfd48e41-0696-4432-a768-f186384452b9" : kubernetes.io/csi: attacher.MountDevice failed to create newCsiDriverClient: driver name local.csi.scylladb.com not found in the list of registered CSI drivers
```

```
$ kubectl -n local-csi-driver logs -f ds/local-csi-driver
Defaulted container "local-csi-driver" out of: local-csi-driver, csi-driver-registrar, liveness-probe, csi-provisioner
I0731 10:10:12.753700       1 local-csi-driver/driver.go:119] "Driver started" command="local-csi-driver" version="\"v0.5.0-alpha.0-2-g587547c\""
I0731 10:10:12.754464       1 flag/flags.go:64] FLAG: --driver-name="local.csi.scylladb.com"
I0731 10:10:12.754573       1 flag/flags.go:64] FLAG: --help="false"
I0731 10:10:12.754648       1 flag/flags.go:64] FLAG: --listen="/csi/csi.sock"
I0731 10:10:12.754718       1 flag/flags.go:64] FLAG: --loglevel="2"
I0731 10:10:12.754849       1 flag/flags.go:64] FLAG: --node-name="ubuntu-2204"
I0731 10:10:12.754924       1 flag/flags.go:64] FLAG: --v="2"
I0731 10:10:12.754990       1 flag/flags.go:64] FLAG: --volumes-dir="/mnt/persistent-volumes"
Error: unsupported volumes dir filesystem "ext2/ext3"
```

tnozicka:
This PR extends our e2e framework to be able to snapshot, collect and restore global resources like NodeConfigs or ScyllaOperator configs. This is very important to be able to debug the disruptive serial suites that use the same object in multiple tests and the final dump just shows the last test.

This also no longer requires the NodeConfig or Resource quotas to be missing before running the e2e suite an depending on the cleanup policy and whether the object has existed before, it will be restored after the test. This also help not to mess up local setups too much but more importantly it allows you to specify a cleanup policy to avoid deleting e.g. the NodeConfig used during the test.

Also this hooks extra dumps for the artifacts dir, e.g. `e2e/cluster-0/cluster-scoped-per-ns/e2e-test-nodesetup-q6bhx-0-d7s4w/cluster-scoped/nodeconfigs.scylla.scylladb.com/cluster.yaml`

Resolves #2045

### Requires
- [x] #2101

### DIFF
--- a/pkg/cmd/tests/options.go
+++ b/pkg/cmd/tests/options.go
@@ -13,6 +13,7 @@ import (
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/spf13/cobra"
 	apierrors "k8s.io/apimachinery/pkg/util/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/rest"
 )
 
@@ -42,26 +43,26 @@ var supportedBroadcastAddressTypes = []scyllav1.BroadcastAddressType{
 type TestFrameworkOptions struct {
 	genericclioptions.ClientConfigSet
 
-	ArtifactsDir                 string
-	DeleteTestingNSPolicyUntyped string
-	DeleteTestingNSPolicy        framework.DeleteTestingNSPolicyType
-	IngressController            *IngressControllerOptions
-	ScyllaClusterOptionsUntyped  *ScyllaClusterOptions
-	scyllaClusterOptions         *framework.ScyllaClusterOptions
-	ObjectStorageBucket          string
-	GCSServiceAccountKeyPath     string
-	S3CredentialsFilePath        string
-	objectStorageType            framework.ObjectStorageType
-	gcsServiceAccountKey         []byte
-	s3CredentialsFile            []byte
+	ArtifactsDir                string
+	CleanupPolicyUntyped        string
+	CleanupPolicy               framework.CleanupPolicyType
+	IngressController           *IngressControllerOptions
+	ScyllaClusterOptionsUntyped *ScyllaClusterOptions
+	scyllaClusterOptions        *framework.ScyllaClusterOptions
+	ObjectStorageBucket         string
+	GCSServiceAccountKeyPath    string
+	S3CredentialsFilePath       string
+	objectStorageType           framework.ObjectStorageType
+	gcsServiceAccountKey        []byte
+	s3CredentialsFile           []byte
 }
 
 func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent string) *TestFrameworkOptions {
 	return &TestFrameworkOptions{
-		ClientConfigSet:              genericclioptions.NewClientConfigSet(userAgent),
-		ArtifactsDir:                 "",
-		DeleteTestingNSPolicyUntyped: string(framework.DeleteTestingNSPolicyAlways),
-		IngressController:            &IngressControllerOptions{},
+		ClientConfigSet:      genericclioptions.NewClientConfigSet(userAgent),
+		ArtifactsDir:         "",
+		CleanupPolicyUntyped: string(framework.CleanupPolicyAlways),
+		IngressController:    &IngressControllerOptions{},
 		ScyllaClusterOptionsUntyped: &ScyllaClusterOptions{
 			NodeServiceType:             string(scyllav1.NodeServiceTypeHeadless),
 			NodesBroadcastAddressType:   string(scyllav1.BroadcastAddressTypePodIP),
@@ -81,11 +82,20 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
 	o.ClientConfigSet.AddFlags(cmd)
 
 	cmd.PersistentFlags().StringVarP(&o.ArtifactsDir, "artifacts-dir", "", o.ArtifactsDir, "A directory for storing test artifacts. No data is collected until set.")
-	cmd.PersistentFlags().StringVarP(&o.DeleteTestingNSPolicyUntyped, "delete-namespace-policy", "", o.DeleteTestingNSPolicyUntyped, fmt.Sprintf("Namespace deletion policy. Allowed values are [%s].", strings.Join(
+	cmd.PersistentFlags().StringVarP(&o.CleanupPolicyUntyped, "delete-namespace-policy", "", o.CleanupPolicyUntyped, fmt.Sprintf("Namespace deletion policy. Allowed values are [%s].", strings.Join(
 		[]string{
-			string(framework.DeleteTestingNSPolicyAlways),
-			string(framework.DeleteTestingNSPolicyNever),
-			string(framework.DeleteTestingNSPolicyOnSuccess),
+			string(framework.CleanupPolicyAlways),
+			string(framework.CleanupPolicyNever),
+			string(framework.CleanupPolicyOnSuccess),
+		},
+		", ",
+	)))
+	utilruntime.Must(cmd.PersistentFlags().MarkDeprecated("delete-namespace-policy", "--delete-namespace-policy is deprecated - please use --cleanup-policy instead"))
+	cmd.PersistentFlags().StringVarP(&o.CleanupPolicyUntyped, "cleanup-policy", "", o.CleanupPolicyUntyped, fmt.Sprintf("Cleanup policy. Allowed values are [%s].", strings.Join(
+		[]string{
+			string(framework.CleanupPolicyAlways),
+			string(framework.CleanupPolicyNever),
+			string(framework.CleanupPolicyOnSuccess),
 		},
 		", ",
 	)))
@@ -118,10 +128,10 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
 		errors = append(errors, err)
 	}
 
-	switch p := framework.DeleteTestingNSPolicyType(o.DeleteTestingNSPolicyUntyped); p {
-	case framework.DeleteTestingNSPolicyAlways,
-		framework.DeleteTestingNSPolicyOnSuccess,
-		framework.DeleteTestingNSPolicyNever:
+	switch p := framework.CleanupPolicyType(o.CleanupPolicyUntyped); p {
+	case framework.CleanupPolicyAlways,
+		framework.CleanupPolicyOnSuccess,
+		framework.CleanupPolicyNever:
 	default:
 		errors = append(errors, fmt.Errorf("invalid DeleteTestingNSPolicy: %q", p))
 	}
@@ -174,7 +184,7 @@ func (o *TestFrameworkOptions) Complete(args []string) error {
 		return err
 	}
 
-	o.DeleteTestingNSPolicy = framework.DeleteTestingNSPolicyType(o.DeleteTestingNSPolicyUntyped)
+	o.CleanupPolicy = framework.CleanupPolicyType(o.CleanupPolicyUntyped)
 
 	// Trim spaces so we can reason later if the dir is set or not
 	o.ArtifactsDir = strings.TrimSpace(o.ArtifactsDir)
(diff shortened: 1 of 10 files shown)


tnozicka:
/hold
for #2101

test/e2e/framework/framework.go
```diff
@@ -161,28 +191,51 @@ func (f *Framework) GetS3CredentialsFile() []byte {
 	return TestContext.S3CredentialsFile
 }
 
-func (f *Framework) defaultCluster() *Cluster {
+func (f *Framework) GetDefaultCluster() *Cluster {
```
rzetelskik, line 194: why make this public? Iirc the idea was that the default cluster was only there for backwards compatibility and the users of the framework were supposed to interact with clusters explicitly (or through framework as is)

test/e2e/framework/framework.go
```diff
@@ -42,28 +45,51 @@ var _ FullClientInterface = &Framework{}
...
-		clusters = append(clusters, c)
+		f.clusters = append(f.clusters, c)
 	}
-	f.clusters = clusters
 
-	f.FullClient.AdminClient.Config = f.defaultCluster().AdminClientConfig()
+	f.FullClient.AdminClient.Config = f.GetDefaultCluster().AdminClientConfig()
 
 	g.BeforeEach(f.beforeEach)
+	g.JustAfterEach(f.justAfterEach)
```
rzetelskik, line 92: I don't think we should further couple the framework initialisation with the ginkgo nodes as any of these prevents you from ensuring an order in the tests itself. Is there any benefit to doing this in JustAfterEach instead of the first action in AfterEach?

test/e2e/framework/cleanup.go
```diff
@@ -0,0 +1,228 @@
...
+			GracePeriodSeconds: pointer.Ptr[int64](0),
+			PropagationPolicy:  pointer.Ptr(metav1.DeletePropagationForeground),
+			Preconditions: &metav1.Preconditions{
+				UID: &nd.NS.UID,
+			},
+		},
+	)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	// We have deleted only the namespace object, but it can still there with deletionTimestamp set.
```
rzetelskik, line 69:
```suggestion
	// We have deleted only the namespace object, but it can still be there with deletionTimestamp set.
```

test/e2e/framework/framework.go
tnozicka: if you want to collect in the middle of a run you need to find the appropriate artifacts dir

test/e2e/framework/framework.go
tnozicka: you need an order in which everything is first collected and then separately cleaned up with independent failure domains.

```
/*
JustAfterEach nodes are similar to AfterEach nodes, however they are guaranteed to run *before* all AfterEach node closures - just after the It node closure. This can allow you to separate diagnostics collection from teardown for a spec.
```

test/e2e/framework/framework.go
tnozicka: see
https://github.com/scylladb/scylla-operator/pull/2091/files#diff-edab1e2b6ade28c3d773100f49b24e2bba8385435ec181b0b7857f0a5d790dacR260

pkg/cmd/tests/options.go
```diff
@@ -42,26 +43,26 @@ var supportedBroadcastAddressTypes = []scyllav1.BroadcastAddressType{
...
-	scyllaClusterOptions         *framework.ScyllaClusterOptions
-	ObjectStorageBucket          string
-	GCSServiceAccountKeyPath     string
-	S3CredentialsFilePath        string
-	objectStorageType            framework.ObjectStorageType
-	gcsServiceAccountKey         []byte
-	s3CredentialsFile            []byte
+	ArtifactsDir                string
+	CleanupNSPolicyUntyped      string
+	CleanupPolicy               framework.CleanupPolicyType
```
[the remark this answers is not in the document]
tnozicka: a bug, thx

test/e2e/framework/framework.go
rzetelskik: Ok, but you can already do this with `f.Cluster(0)`, The concept of a 'default cluster' was something we wanted to get away from, not further solidify it.

test/e2e/framework/cleanup.go
tnozicka: fixed

pkg/cmd/tests/options.go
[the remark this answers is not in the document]
tnozicka: fixed

test/e2e/framework/framework.go
tnozicka: This issue is that one of those clusters is still bound to the ginkgo which is the reason it still has a special meaning. I am not a fan of using a magic index when you don't choose from an array.

tnozicka: Also the object were created with the "default" client - `f.ScyllaAdminClient()` delegates through `GetDefaultCluster` and user didn't choose the index on that call, there should be a corresponding way to get the artifacts dir for it, also without knowing the magic index.

rzetelskik: My point is mainly to discourage the use of a "default cluster" in the actual test specs. Maybe calling it a `GinkgoCluster` would make sense then? Or anything that would make it clear it's related to test infra, and you're already using a `ginkgoNamespace` across this PR.

tnozicka:
/hold cancel
https://github.com/scylladb/scylla-operator/pull/2101 landed

test/e2e/framework/framework.go
tnozicka: it has crossed my mind as well but didn't pass a test of uniformity on the naming of how the objects get created. I've wired a special method instead, similar to the namespace.

rzetelskik: I think that's better, thanks

test/e2e/framework/cleanup.go
```diff
@@ -0,0 +1,228 @@
...
+
+func NewRestoringCleaner(ctx context.Context, client kubernetes.Interface, dynamicClient dynamic.Interface, resourceInfo collect.ResourceInfo, namespace string, name string, strategy RestoreStrategy) *RestoringCleaner {
+	g.By(fmt.Sprintf("Snapshotting object %s %q", resourceInfo.Resource, naming.ManualRef(namespace, name)))
+
+	if resourceInfo.Scope.Name() == meta.RESTScopeNameNamespace {
+		o.Expect(namespace).NotTo(o.BeEmpty())
+	}
+
+	obj, err := dynamicClient.Resource(resourceInfo.Resource).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
tnozicka: for the global objects there may or ma not be an object to snapshot before you start, say a nodecofig/cluster of your own

test/e2e/framework/cleanup.go
```diff
@@ -0,0 +1,228 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+)
+
+type CleanupInterface interface {
+	Print(ctx context.Context)
+	Collect(ctx context.Context, artifactsDir string, ginkgoNamespace string)
+	Cleanup(ctx context.Context)
+}
```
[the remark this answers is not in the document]
tnozicka: yeah, but you still need one interface to have an array of objects - how about I rename the `Print` to `CollectToLog`?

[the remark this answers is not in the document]
zimnx: yep, lets rename Print

test/e2e/framework/cleanup.go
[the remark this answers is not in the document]
zimnx: the second question still holds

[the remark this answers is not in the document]
tnozicka: nope it check the UID

test/e2e/framework/cleanup.go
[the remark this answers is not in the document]
tnozicka: renamed

test/e2e/framework/cleanup.go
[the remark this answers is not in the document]
tnozicka: https://github.com/scylladb/scylla-operator/blob/38923dadcfd8d752c1666eae3dea6ba0da4080d7/test/e2e/framework/cleanup.go#L222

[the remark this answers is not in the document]
tnozicka: I see, if you call the 2 manually, that might happen - the framework call only Cleanup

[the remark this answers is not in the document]
tnozicka: Converted those to private and added an assertion

zimnx (approved):
/assign rzetelskik

test/e2e/framework/framework.go
rzetelskik: Ok, from correctness perspective it makes sense. I guess it won't affect the amount of effort required to decouple those much either.

rzetelskik (approved):
/lgtm

rzetelskik:
> @tnozicka: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> ci/prow/e2e-gke-parallel-clusterip	[00a1223](https://github.com/scylladb/scylla-operator/commit/00a1223006975f755f489945368a39a4cdd584d7)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2091/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1829168862705225728)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2091). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:tnozicka).

https://github.com/scylladb/scylla-operator/issues/2061#issuecomment-2318163828
known manager flake
/retest

PR #2097, opened 2024-08-27
Add a note to documentation about exposeOptions being immutable 

### LINKED ISSUE #1966: Document exposeOptions can't be changed on an existing cluster
### What happened?

User needs to change how the cluster is exposed (leveraging `exposeOptions`)

User should find on the documentation changing `exposeOptions` is not currently supported

Deploy a cluster as private then switch `exposeOptions` to VPC-exposed

### Scylla Operator version

latest

N/A

N/A

Please document changing `exposeOptions` is not currently supported. Add link to GH issue that fixes this condition (if applicable) so users can track progress by themselves.

zimnx:
It's worth pointing out immutability in documentation and API spec.

Fixes  #1966

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -2932,7 +2932,7 @@ spec:
                     type: string
                   type: array
                 exposeOptions:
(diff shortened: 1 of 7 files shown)


rzetelskik (approved):
lgtm
/assign tnozicka

tnozicka (approved):
/approve
/lgtm

Issue #2098, opened 2024-08-27
When a test framework setup fails, no test are run but the run succeed


tnozicka:
When any assertion in the `Framework` fails, no test are executed but the run succeeds. It can be simulated by minting the framework like
```
func NewFramework(namePrefix string) *Framework {
	o.Expect(true).To(o.BeFalse())
```
this go through ginkgo.Recover and calls the fail handler correctly but my suspicion is that when no test gets registered is forgets to check the fail handler. I've tried failing the individual tests before the framework creation directly and that has correctly failed the run.

```
go run ./cmd/scylla-operator-tests run all --ingress-controller-address=$( kubectl -n haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' ) --loglevel=2 --parallelism=1 --progress --delete-namespace-policy=OnSuccess --feature-gates='AutomaticTLSCertificates=true' --artifacts-dir=/tmp/so-artifacts --fail-fast --loglevel=5
```
```I0827 19:39:14.798189  632799 featuregate/feature_gate.go:249] feature gates: &{map[AutomaticTLSCertificates:true]}
Flag --delete-namespace-policy has been deprecated, --delete-namespace-policy is deprecated - please use --cleanup-policy instead
I0827 19:39:14.798450  632799 tests/tests.go:74] maxprocs: Leaving GOMAXPROCS=[16]: CPU quota undefined
I0827 19:39:14.799273  632799 tests/tests_run.go:213] "scylla-operator-tests run" version "unknown"
I0827 19:39:14.799288  632799 flag/flags.go:64] FLAG: --artifacts-dir="/tmp/so-artifacts"
I0827 19:39:14.799292  632799 flag/flags.go:64] FLAG: --burst="75"
I0827 19:39:14.799297  632799 flag/flags.go:64] FLAG: --cleanup-policy="OnSuccess"
I0827 19:39:14.799300  632799 flag/flags.go:64] FLAG: --color="true"
I0827 19:39:14.799303  632799 flag/flags.go:64] FLAG: --delete-namespace-policy="OnSuccess"
I0827 19:39:14.799304  632799 flag/flags.go:64] FLAG: --dry-run="false"
I0827 19:39:14.799306  632799 flag/flags.go:64] FLAG: --fail-fast="true"
I0827 19:39:14.799307  632799 flag/flags.go:64] FLAG: --feature-gates="AutomaticTLSCertificates=true"
[...]

PR #2099, opened 2024-08-27
Validate that artifacts dir exists

tnozicka:
e2e tests expect the artifacts dir to exists and this PR make sure we check it early (especially given #2098).


rzetelskik (approved):
lgtm
/assign zimnx

PR #2100, opened 2024-08-28
Drop GinkgoRecover from ginkgo's container nodes

rzetelskik:
**Description of your changes:** GinkgoRecover should only be called when we're spawning goroutines from within a run spec. We are currently calling it from almost every container node, which, in case of a failed assertion in a container node (which shouldn't happen either way, see discussion in https://github.com/scylladb/scylla-operator/issues/2098), results in silencing an error from run spec tree construction: see e.g. https://github.com/onsi/ginkgo/issues/931#issuecomment-1040512693. It only doesn't happen now on broken framework initialization because we have two container nodes which don't call GinkgoRecover.
This PR drops the GinkgoRecover calls from the remaining container nodes.

Resolves #

/kind bug
/priority important-soon
/cc

### DIFF
--- a/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
+++ b/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
@@ -32,7 +32,6 @@ import (
 )
 
 var _ = g.Describe("Node Setup", framework.Serial, func() {
-	defer g.GinkgoRecover()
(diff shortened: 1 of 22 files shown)


zimnx (approved):
lgtm
/assign tnozicka

tnozicka (approved):
/approve
/lgtm
thanks

PR #2104, opened 2024-09-02
TMP: check SM regression fix

Michal-Leszczynski:
This tmp commit job is to validate that the fix from
https://github.com/scylladb/scylla-manager/pull/4000
indeed fixes the 3.3.1 SM regression reported in
https://github.com/scylladb/scylla-manager/issues/3989.

This PR is not supposed to be merged, it's here just to run the CI.

### DIFF
--- a/test/e2e/fixture/scylla/scyllacluster.yaml.tmpl
+++ b/test/e2e/fixture/scylla/scyllacluster.yaml.tmpl
@@ -7,7 +7,7 @@ metadata:
   annotations:
    bar: foo
 spec:
-  agentVersion: 3.3.0
+  agentVersion: 3.4.0-dev-0.20240902.c3ca05a61
   version: 6.1.0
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
@rzetelskik it looks like the fix passed the CI. It will be a part of 3.3.2 and I hope we will release it soon.
I'm closing the PR as it fulfilled its purpose.

zimnx:
It passed the CI, but you haven't changed Scylla Manager **Server** version. Agent is not involved in backup task validation that got broken.

zimnx:
To bump Scylla Manager Server version, you should change https://github.com/scylladb/scylla-operator/blob/acc67a786163a4ce5644db2c9a0222619be47bff/helm/scylla-manager/values.yaml#L10
and run `make update` in repository root to update all autogenerated files.

PR #2107, opened 2024-09-03
Synchronize ScyllaDB Manager version with config.yaml

tnozicka:
This fixes version synchronization for ScyllaDB Manager deployment with the `config.yaml`.

(@rzetelskik found out this is not getting replaced)


zimnx (approved):
lgtm
/assign rzetelskik

tnozicka:
well, we can't use the same sha for manager agent and manager images as those are different images (same version but not sha)
`message: Back-off pulling image "docker.io/scylladb/scylla-manager-agent:3.3.0@sha256:e8c5b62c9330f91dfca24f109b033df78113d3ffaac306edf6d3c4346e1fa0bf"`
@rzetelskik @zimnx can we get rid of the sha with a newer manager version?

zimnx:
> can we get rid of the sha with a newer manager version?

We can, although I think it would be better to have two configuration values with different sha's but the same semantic version for each image.

rzetelskik (approved):
/lgtm

PR #2114, opened 2024-09-06
Parametrize webhook command

zimnx:
Webhook command accepts a map of validators making this command be reusable and easily extensible.

### DIFF
--- a/pkg/cmd/operator/webhooks.go
+++ b/pkg/cmd/operator/webhooks.go
@@ -19,6 +19,7 @@ import (
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/api/scylla/validation"
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
+	"github.com/scylladb/scylla-operator/pkg/kubeinterfaces"
 	"github.com/scylladb/scylla-operator/pkg/signals"
 	"github.com/scylladb/scylla-operator/pkg/version"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


pkg/cmd/operator/webhooks.go
```diff
@@ -269,50 +329,21 @@ func validate(ar *admissionv1.AdmissionReview) error {
...
-			return apierrors.NewInvalid(obj.(*scyllav1alpha1.ScyllaOperatorConfig).GroupVersionKind().GroupKind(), obj.(*scyllav1alpha1.ScyllaOperatorConfig).Name, errList)
-		}
-		return nil
+	var errList field.ErrorList
+	switch ar.Request.Operation {
+	case admissionv1.Create:
+		errList = validator.ValidateCreate(obj)
+	case admissionv1.Update:
+		errList = validator.ValidateUpdate(obj, oldObj)
+	}
```
tnozicka, line 343: handle the default in case of a miss

pkg/cmd/operator/webhooks.go
```diff
@@ -105,6 +133,10 @@ func (o *WebhookOptions) Validate() error {
 		return errors.New("port can't be zero")
 	}
 
+	if len(o.Validators) == 0 {
```
tnozicka, line 136: Validate is meant to validate the options coming from the flags. This seems like it should fail in `NewWebhookCmd` where it is supplied.

pkg/cmd/operator/webhooks.go
tnozicka: given it's preexisting take it as a nit

zimnx: done

pkg/cmd/operator/webhooks.go
zimnx: Added a default instead of an error in this case.

tnozicka: I think this should fail. If you accidentally give it an empty array and it will silently use a global variable instead - that feels wrong IMO.

zimnx: Erroring would result in carrying it up to a main. None of our command initialization returns error so far. 
We could panic, as this is clearly a developer fault, but I think default is reasonable.

rzetelskik (approved):
lgtm, but +1 for @tnozicka's comment https://github.com/scylladb/scylla-operator/pull/2114#discussion_r1750118305

pkg/cmd/operator/webhooks.go
tnozicka: I can live with a panic but it shouldn't silently use something else.

zimnx: changed to panic

zimnx:
clean rebase

PR #2117, opened 2024-09-10
Fix systemd skip in unit tests

tnozicka:
When we switched to UBI the part that skipped the broken systemd unit test stopped working with 
`daemon_test.go:39: Process org.freedesktop.systemd1 exited with status 1` which is not very descriptive.
```
dbus.Error{Name:"org.freedesktop.DBus.Error.Spawn.ChildExited", Body:[]interface {}{"Process org.freedesktop.systemd1 exited with status 1"}}
```
There is a bunch of things that are not there and we try to half use them so this PR simplifies the check similar to [sd_booted check](https://www.freedesktop.org/software/systemd/man/latest/sd_booted.html) a looks whether `/run/systemd/system/` directory exists.

This currently blocks the whole CI on unit tests.

Resolves https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2116/pull-scylla-operator-master-unit/1833408548160147456#1:build-log.txt%3A116

### DIFF
--- a/pkg/systemd/daemon_test.go
+++ b/pkg/systemd/daemon_test.go
@@ -4,13 +4,36 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"testing"
 
-	godbus "github.com/godbus/dbus/v5"
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
+func hasSystemdRunning() (bool, error) {
+	const systemdRuntimeDir = "/run/systemd/system"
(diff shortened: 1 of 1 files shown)


zimnx (approved):
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/2118#issue-2516124430
/retest

PR #2129, opened 2024-09-25
Fix ScyllaDBDatacenter controller selector and resource naming

zimnx:
Controller selector should be in sync with selector labels added to managed resources.
It wasn't catched before because ScyllaCluster migrating controller used ScyllaCluster name as ScyllaDBDatacenter.spec.clusterName.
In case when ScyllaDBDatacenter would be created with different name than clusterName specified in it's spec, resources would've been orphaned immediately and managed objects couldn't progress.
To prevent two ScyllaDBDatacenters having the same clusterName in their spec colliding with each other, managed resources names were changed to use object name instead clusterName.

### DIFF
--- a/pkg/controller/scylladbdatacenter/resource_test.go
+++ b/pkg/controller/scylladbdatacenter/resource_test.go
@@ -2699,7 +2699,7 @@ func Test_MakeManagedScyllaDBConfig(t *testing.T) {
 					},
 				},
 				Spec: scyllav1alpha1.ScyllaDBDatacenterSpec{
-					ClusterName: "foo",
+					ClusterName: "foo-cluster",
 				},
 			},
 			enableTLSFeatureGate: false,
@@ -2712,8 +2712,11 @@ func Test_MakeManagedScyllaDBConfig(t *testing.T) {
 					Namespace: "foo-ns",
 					Name:      "foo-managed-config",
 					Labels: map[string]string{
-						"scylla/cluster": "foo",
-						"user-label":     "user-label-value",
+						"app":                          "scylla",
+						"app.kubernetes.io/managed-by": "scylla-operator",
+						"app.kubernetes.io/name":       "scylla",
+						"scylla/cluster":               "foo",
+						"user-label":                   "user-label-value",
 					},
 					OwnerReferences: []metav1.OwnerReference{
 						{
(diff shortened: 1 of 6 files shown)


zimnx:
No longer WIP

pkg/naming/labels.go
```diff
@@ -11,7 +11,7 @@ import (
 // for the given Cluster.
 func ClusterLabels(sdc *scyllav1alpha1.ScyllaDBDatacenter) map[string]string {
 	labels := ScyllaLabels()
-	labels[ClusterNameLabel] = sdc.Spec.ClusterName
+	labels[ClusterNameLabel] = sdc.Name
```
rzetelskik, line 14: the name of the constant is misleading now, should we change it (or have a separate one)?

pkg/controller/scylladbdatacenter/resource_test.go
```diff
@@ -2712,8 +2712,11 @@ func Test_MakeManagedScyllaDBConfig(t *testing.T) {
 					Namespace: "foo-ns",
 					Name:      "foo-managed-config",
 					Labels: map[string]string{
-						"scylla/cluster": "foo",
-						"user-label":     "user-label-value",
+						"app":                          "scylla",
+						"app.kubernetes.io/managed-by": "scylla-operator",
+						"app.kubernetes.io/name":       "scylla",
+						"scylla/cluster":               "foo",
```
rzetelskik, line 2718: the name of the resource and ClusterName is the same in the test, wouldn't it be better to make them different to make sure the right one is propagated to the label?

pkg/naming/labels.go
zimnx: It will be misleading anyway, as value of this constant is `scylla/cluster` which we cannot change.

pkg/controller/scylladbdatacenter/resource_test.go
zimnx: good point, changed

pkg/naming/labels.go
zimnx: Added a comment about which value should be provided.

rzetelskik (approved):
lgtm
/assign tnozicka

pkg/naming/constants.go
```diff
@@ -70,6 +70,9 @@ const (
 
 // Generic Labels used on objects created by the operator.
 const (
+	// ClusterNameLabel specifies ScyllaCluster name. It's used as a selector by both ScyllaCluster and ScyllaDBDatacenter controllers.
+	// As controller selector cannot be changed, this may be misleading in the context of ScyllaDBDatacenter. It should
+	// specify the ScyllaCluster object name ScyllaDBDatacenter originated from or object name in case it wasn't migrated.
```
tnozicka, line 75: > It should specify the ScyllaCluster object name ScyllaDBDatacenter originated from or object name in case it wasn't migrated

not sure I get that, it always takes `sdc.Name`

zimnx: sdc.Name comes from sc.Name.

tnozicka (approved):
/approve
/lgtm

Issue #1559, opened 2023-11-10
Retry NodeConfig mounts


tnozicka:
### What should the feature do?

When a mount fails, say because the actual filesystem needed a repair, we won't try again. Well systemd doesn't :(

Say when the admin runs the notorious xfs_repair he has no idea there is a unit `'mnt-persistent\x2dvolumes.mount'` that needs to be restarted. With some luck he will reboot the node...

### What is the use case behind this feature?

Stability

### Anything else we need to know?

```
root@ubuntu-2204:~# systemctl status 'mnt-persistent\x2dvolumes.mount'
× mnt-persistent\x2dvolumes.mount - Managed mount by Scylla Operator
     Loaded: loaded (/etc/systemd/system/mnt-persistent\x2dvolumes.mount; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Fri 2023-11-10 18:50:29 UTC; 10min ago
      Where: /mnt/persistent-volumes
       What: /dev/md127
        CPU: 5ms

Nov 10 18:50:29 ubuntu-2204 systemd[1]: Mounting Managed mount by Scylla Operator...
Nov 10 18:50:29 ubuntu-2204 systemd[1]: mnt-persistent\x2dvolumes.mount: Mount process exited, code=exited, status=32/n/a
Nov 10 18:50:30 ubuntu-2204 mount[3469]: mount: /mnt/persistent-volumes: mount(2) system call failed: Structure needs cleaning.
Nov 10 18:50:29 ubuntu-2204 systemd[1]: mnt-persistent\x2dvolumes.mount: Failed with result 'exit-code'.
Nov 10 18:50:29 ubuntu-2204 systemd[1]: Failed to mount Managed mount by Scylla Operator.
root@ubuntu-2204:~# systemctl restart 'mnt-persistent\x2dvolumes.mount'
root@ubuntu-2204:~# systemctl status 'mnt-persistent\x2dvolumes.mount'
● mnt-persistent\x2dvolumes.mount - Managed mount by Scylla Operator
     Loaded: loaded (/etc/systemd/system/mnt-persistent\x2dvolumes.mount; enabled; vendor preset: enabled)
     Active: active (mounted) since Fri 2023-11-10 19:01:32 UTC; 1s ago
      Where: /mnt/persistent-volumes
       What: /dev/md127
      Tasks: 0 (limit: 4535)
     Memory: 12.0K
        CPU: 8ms
[...]

PR #2134, opened 2024-09-30
Propagate mount unit failures and stop reconciling preexisting units unmanaged by NodeConfig

### LINKED ISSUE #1334: Operator should check for mount units failures and propagate them
NodeConfig controller sets up systemd units on the host but doesn't check their status. This makes it very hard to debug potential issues. We should check the unit state and propagate it to the controlling CR status as a condition.

https://github.com/tnozicka/scylla-operator/blob/873959d/pkg/controller/nodesetup/sync_mounts.go#L50-L53

### LINKED ISSUE #1909: NodeConfig should be in a degraded state when mount already exists and is not identical
### What happened?

If you define a NodeConfig such that it should mount the raid array over an existing mount, it won't complain, and the NodeConfig will not end up in a degraded state.

```sh
$ systemctl status mnt-test.mount
● mnt-test.mount - Managed mount by Scylla Operator
     Loaded: loaded (/proc/self/mountinfo; enabled; vendor preset: enabled)
     Active: active (mounted) since Wed 2024-05-01 20:04:50 UTC; 5min ago
      Where: /mnt/test
       What: /dev/mapper/ubuntu--vg-ubuntu--lv
      Tasks: 0 (limit: 4557)
     Memory: 48.0K
        CPU: 5ms
     CGroup: /system.slice/mnt-test.mount

May 01 20:04:50 ubuntu systemd[1]: Mounting Managed mount by Scylla Operator...
May 01 20:04:50 ubuntu systemd[1]: Mounted Managed mount by Scylla Operator.
```

```
$ kubectl logs -n scylla-operator-node-tuning daemonsets/cluster-node-setup
...
I0501 20:15:08.529961       1 nodesetup/sync_raids.go:86] "RAID0 array has been created" RAIDName="r6bc9qjw" Devices="/dev/loop10"
I0501 20:15:08.530029       1 record/event.go:376] "Event occurred" object="cluster" fieldPath="" kind="NodeConfig" apiVersion="scylla.scylladb.com/v1alpha1" type="Normal" reason="RAIDCreated" message="RAID0 array \"r6bc9qjw\" using /dev/loop10 devices has been created"
I0501 20:15:08.541078       1 nodesetup/sync_filesystems.go:56] "Filesystem has been created" Device="r6bc9qjw" Filesystem="xfs"
I0501 20:15:08.541188       1 nodesetup/sync_mounts.go:46] "Mount unit has been generated and queued for apply." Name="mnt-test.mount" Device="r6bc9qjw" MountPoint="/mnt/test"
I0501 20:15:08.541259       1 systemd/unit.go:108] "Checking if units need pruning" Existing=1 Desired=1
I0501 20:15:08.541294       1 record/event.go:376] "Event occurred" object="cluster" fieldPath="" kind="NodeConfig" apiVersion="scylla.scylladb.com/v1alpha1" type="Normal" reason="FilesystemCreated" message="xfs filesystem has been created device r6bc9qjw"
I0501 20:15:09.037629       1 systemd/unit.go:169] "Ensuring unit" Name="mnt-test.mount"
I0501 20:15:09.467564       1 systemd/unit.go:206] "Enabling and starting unit" Name="mnt-test.mount"
I0501 20:15:09.471249       1 nodesetup/status.go:44] "Updating status" NodeConfig="cluster" Node="ubuntu"
```

/kind bug

NodeConfig should end up in a degraded state and the controller should log an error.

```sh
$ systemctl status mnt-test.mount
● mnt-test.mount - Managed mount by Scylla Operator
     Loaded: loaded (/proc/self/mountinfo; enabled; vendor preset: enabled)
     Active: active (mounted) since Wed 2024-05-01 20:04:50 UTC; 22s ago
      Where: /mnt/test
       What: /dev/md124
      Tasks: 0 (limit: 4557)
     Memory: 1.7M
        CPU: 5ms
     CGroup: /system.slice/mnt-test.mount

May 01 20:04:50 ubuntu systemd[1]: Mounting Managed mount by Scylla Operator...
May 01 20:04:50 ubuntu systemd[1]: Mounted Managed mount by Scylla Operator.
```

On host:
```sh
$ mount --bind $( mktemp -d ) /mnt/test/
```

Create a NodeConfig:
```
kind: NodeConfig
metadata:
  name: cluster
spec:
  disableOptimizations: false
  localDiskSetup:
    filesystems:
    - device: lhwnjl5q
      type: xfs
    loopDevices:
    - imagePath: /mnt/disk.img
      name: disk
      size: 32M
    mounts:
    - device: lhwnjl5q
      fsType: xfs
      mountPoint: /mnt/test
      unsupportedOptions:
      - prjquota
    raids:
    - RAID0:
        devices:
          modelRegex: .*
          nameRegex: ^/dev/loops/disk$
      name: lhwnjl5q
      type: RAID0
  placement:
    nodeSelector:
      scylla.scylladb.com/node-type: scylla
```

### Scylla Operator version

master

n/a

n/a

_No response_

rzetelskik:
**Description of your changes:** This PR adds mount unit failure propagation as well as it prevents the mount controller from reconciling preexisting mount units that are not managed by NodeConfig, i.e. mount units that existed before they were added to a list of managed units in the status file. It also extends the existing workflow with progressing conditions on active states other than "active", so that the controller doesn't race with unit failures.

Initially I wanted this PR to also include restarting failed units on configuration changes (as discussed [here](https://github.com/scylladb/scylla-operator/issues/1909)). However, with how many different active state/substates there are, I was afraid of introducing regressions compared to the current behaviour, i.e. attempt to enable and start units every time regardless of the load/active state. It also wasn't clear what exactly should be considered a configuration change, and so what should trigger a potential restart, e.g. whether a change to a disk/filesystem configuration in NodeConfig, related to a given mount unit, should be considered a configuration change. 
I believe this requires more discussions and a well though-through approach (under https://github.com/scylladb/scylla-operator/issues/1559 ?), and so, it shouldn't delay these particular changes.

Hence this PR only introduces the minimal amount of changes necessary to handle these two particular cases and extends the unit and e2e test suites with the described scenarios.

Resolves #1334 
Resolves #1909

/priority important-soon
/kind feature

### DIFF
--- a/pkg/systemd/unit.go
+++ b/pkg/systemd/unit.go
@@ -13,6 +13,8 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 )
@@ -68,7 +70,7 @@ func (m *UnitManager) ReadStatus() (*unitManagerStatus, error) {
 		return nil, fmt.Errorf("can't open status file %q: %w", statusFile, err)
 	}
 
-	status := &unitManagerStatus{}
+	status := newUnitManagerStatus()
 	err = yaml.Unmarshal(data, status)
 	if err != nil {
 		return nil, fmt.Errorf("can't decode unit manager status: %w", err)
@@ -92,17 +94,34 @@ func (m *UnitManager) WriteStatus(status *unitManagerStatus) error {
 	return nil
 }
 
+const (
+	loadStateNotFound = "not-found"
+
+	activeStateActive = "active"
+	activeStateFailed = "failed"
+)
+
+type UnitStatus struct {
+	Name        string
+	LoadState   string
+	ActiveState string
+}
+
 type EnsureControlInterface interface {
 	DaemonReload(ctx context.Context) error
-	EnableAndStartUnit(ctx context.Context, unitFile string) error
+	EnableUnit(ctx context.Context, unitFile string) error
+	StartUnit(ctx context.Context, unitFile string) error
 	DisableAndStopUnit(ctx context.Context, unitFile string) error
+	GetUnitStatuses(ctx context.Context, unitFiles []string) ([]UnitStatus, error)
 }
 
 // EnsureUnits will make sure to remove any unit that is no longer desired and create/update those that are.
-func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeConfig, recorder record.EventRecorder, requiredUnits []*NamedUnit, control EnsureControlInterface) error {
+func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeConfig, recorder record.EventRecorder, requiredUnits []*NamedUnit, control EnsureControlInterface) ([]string, error) {
+	var progressingMessages []string
+
 	status, err := m.ReadStatus()
 	if err != nil {
-		return fmt.Errorf("can't list managed units: %w", err)
+		return progressingMessages, fmt.Errorf("can't list managed units: %w", err)
 	}
 
 	klog.V(4).InfoS(
@@ -114,7 +133,7 @@ func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeCo
 	// Reload unit definitions for pruning in case that we didn't make it to daemon reload when writing them.
 	err = control.DaemonReload(ctx)
 	if err != nil {
-		return fmt.Errorf("can't reload systemd: %w", err)
+		return progressingMessages, fmt.Errorf("can't reload systemd: %w", err)
 	}
 
 	for _, existingUnitName := range status.ManagedUnits {
@@ -135,7 +154,7 @@ func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeCo
 			if errors.Is(err, ErrNotExist) {
 				klog.V(2).InfoS("Skipped disabling and stopping unit that doesn't exist", "Name", existingUnitName)
 			} else {
-				return fmt.Errorf("can't disable unit %q: %w", existingUnitName, err)
+				return progressingMessages, fmt.Errorf("can't disable unit %q: %w", existingUnitName, err)
 			}
 		}
 
@@ -143,7 +162,7 @@ func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeCo
 		existingUnitPath := m.GetUnitPath(existingUnitName)
 		err = os.Remove(existingUnitPath)
 		if err != nil && !errors.Is(err, fs.ErrNotExist) {
-			return fmt.Errorf("can't prune unit %q: %w", existingUnitName, err)
+			return progressingMessages, fmt.Errorf("can't prune unit %q: %w", existingUnitName, err)
 		}
 		recorder.Eventf(
 			nc,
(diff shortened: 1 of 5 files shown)


pkg/systemd/unit.go
```diff
@@ -154,61 +173,120 @@ func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeCo
...
+		}
+		if len(unitStatuses) == 0 {
+			errs = append(errs, fmt.Errorf("missing status for unit %q", managedRequiredUnit.FileName))
+			continue
+		}
+		unitStatus := unitStatuses[0]
+
+		switch unitStatus.ActiveState {
+		case activeStateActive:
+
```
zimnx, line 273: Adding a comment here could explain nothing is expected to do here. Without it may suggest author forgot to implement this branch.

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
```diff
@@ -283,23 +241,34 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
...
-				{
-					Device:             fmt.Sprintf("/dev/%s", f.Namespace()),
-					MountPoint:         fmt.Sprintf("/mnt/%s", f.Namespace()),
-					FSType:             string(scyllav1alpha1.XFSFilesystem),
-					UnsupportedOptions: []string{"prjquota"},
-				},
-			},
+		if e.preNodeConfigCreationFunc != nil {
+			cleanupFunc := e.preNodeConfigCreationFunc(ctx, nc)
+			defer cleanupFunc(ctx)
```
zimnx, line 257: use g.DeferCleanup. Context may be already cancelled during defer (if timeout happens) so it should use context.Background with optional additional timeout.

pkg/systemd/unit.go
tnozicka: or a log

pkg/systemd/unit.go
```diff
@@ -154,61 +173,120 @@ func (m *UnitManager) EnsureUnits(ctx context.Context, nc *scyllav1alpha1.NodeCo
...
+			// Append error to propagate a degraded condition but do not break early to retry starting the unit.
+			errs = append(errs, fmt.Errorf("unit %q is in a failed state", managedRequiredUnit.FileName))
+			fallthrough
+
+		default:
+			progressingMessages = append(progressingMessages, fmt.Sprintf("Awaiting unit %q to be in active state %q.", managedRequiredUnit.FileName, activeStateActive))
+		}
+
+		klog.V(2).InfoS("Starting unit", "Name", managedRequiredUnit.FileName)
+		err = control.StartUnit(ctx, managedRequiredUnit.FileName)
```
tnozicka, line 284: are you intentionally starting units in active state?

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
rzetelskik: good point, fixed, thanks

rzetelskik: DeferCleanup called in an `It` registeres a node behaving like AfterEach. Given our framework calls cleanup methods and removes the client in an AfterEach node, and AfterEach nodes have not guaranteed order, this could result in the `cleanupFunc` being called after the namespace and client are cleared. Hence using DeferCleanup would require significant changes to how our framework and all the tests are set up. 
I gave it a spec context instead, which shouldn't be invalidated when the test times out (since the timeout is not configured on the node but inside it).

zimnx (approved):
lgtm 
/assign tnozicka

pkg/systemd/unit.go
tnozicka: ok

pkg/systemd/unit_test.go
```diff
@@ -119,35 +121,76 @@ managedUnits:
 	}
 }
 
-type NoopEnsureControl struct{}
+type NoopEnsureControl struct {
```
tnozicka, line 124: with state the `Noop` name is no longer fitting, it's more a "fake" now

tnozicka (commented):
/approve

tnozicka:
/lgtm
(it's just a nit)

pkg/systemd/unit_test.go
rzetelskik: I was considering it but I think noop is still a better fit since Enable/Start methods don't affect the state

rzetelskik: But that will be a good point when we extend it to handle active states differently and have to test it

PR #2137, opened 2024-10-01
Move resolution of developerMode and additionalScyllaDBArguments to Operator 

### LINKED ISSUE #1389: Changes to some of the config values do not trigger a rolling restart
**Describe the bug**
Some of the config values are currently computed in the sidecar by getting the ScyllaCluster object using scylla client and extracting them from spec, e.g. the developer mode: https://github.com/scylladb/scylla-operator/blob/75a39bf4b973d789833b761d5c284c36adf07703/pkg/sidecar/config/config.go#L220.
For this reason, changes to these fields in ScyllaCluster do not trigger a rolling restart.
Passing these config values to sidecar in a different way could trigger a rolling restart, while also allowing us to get rid of using the scyllaClient in sidecar.

**To Reproduce**
1. Deploy a ScyllaCluster and wait for it to roll out
2. Change one of the fields extracted from ScyllaCluster in sidecar, .e.g developer mode

**Expected behavior**
The cluster should undergo a rolling restart.

**Additional context**
This issue can be resolved as a part of a bigger effort https://github.com/scylladb/scylla-operator/issues/604

zimnx:
Moves the resolution of `developerMode` and `additionalScyllaDBArguments` from the sidecar to the Operator.
This change addresses an issue where modifications to these arguments did not trigger a rolling restart of the ScyllaDB cluster.
By moving the logic to the Operator, any changes to these fields now correctly trigger a rolling restart as intended.

Since the sidecar no longer needs to fetch the ScyllaCluster resource for these parameters, it no longer requires permissions to perform GET operations on ScyllaCluster objects.
However, to ensure a safe rolling upgrade of the Operator, these permissions are not removed yet.

Additionally, this commit removes support for the deprecated `cpuset` field. It is now treated as if it is always set to true regardless of its value.

Requires:

- [x] #2139 

Fixes #1389

### DIFF
--- a/helm/scylla-operator/templates/scyllacluster_member_clusterrole_def.yaml
+++ b/helm/scylla-operator/templates/scyllacluster_member_clusterrole_def.yaml
@@ -47,6 +47,9 @@ rules:
   - list
   - patch
   - watch
+# Sidecar no longer fetches ScyllaClusters. Necessary information is provided by the Operator to sidecar via ScyllaDB config or arguments.
+# However, to ensure smooth Operator updates, this cannot be removed earlier than in 1.16.
+# TODO: remove this permission in Operator >=1.16 (https://github.com/scylladb/scylla-operator/issues/2141)
 - apiGroups:
   - scylla.scylladb.com
   resources:
--- a/pkg/cmd/operator/sidecar.go
+++ b/pkg/cmd/operator/sidecar.go
@@ -64,6 +64,7 @@ func NewSidecarCmd(streams genericclioptions.IOStreams) *cobra.Command {
 		Use:   "sidecar",
 		Short: "Run the scylla sidecar.",
 		Long:  `Run the scylla sidecar.`,
+		Args:  cobra.ArbitraryArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			err := o.Validate()
 			if err != nil {
@@ -75,7 +76,7 @@ func NewSidecarCmd(streams genericclioptions.IOStreams) *cobra.Command {
 				return err
 			}
 
-			err = o.Run(streams, cmd)
+			err = o.Run(streams, cmd, args)
 			if err != nil {
 				return err
 			}
@@ -141,20 +142,18 @@ func (o *SidecarOptions) Complete() error {
 		return fmt.Errorf("can't build kubernetes clientset: %w", err)
 	}
 
-	o.scyllaClient, err = scyllaversionedclient.NewForConfig(o.RestConfig)
-	if err != nil {
-		return fmt.Errorf("can't build scylla clientset: %w", err)
-	}
-
 	o.clientsBroadcastAddressType = scyllav1alpha1.BroadcastAddressType(o.ClientsBroadcastAddressTypeString)
 	o.nodesBroadcastAddressType = scyllav1alpha1.BroadcastAddressType(o.NodesBroadcastAddressTypeString)
 
 	return nil
 }
 
-func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Command) error {
+func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Command, args []string) error {
 	klog.Infof("%s version %s", cmd.Name(), version.Get())
 	cliflag.PrintFlags(cmd.Flags())
+	for _, arg := range args {
+		klog.V(1).Infof("ARG: %q", arg)
+	}
 
 	stopCh := signals.StopChannel()
 	ctx, cancel := context.WithCancel(context.Background())
@@ -208,14 +207,14 @@ func (o *SidecarOptions) Run(streams genericclioptions.IOStreams, cmd *cobra.Com
 		return fmt.Errorf("can't get pod %q: %w", o.ServiceName, err)
 	}
 
-	member, err := identity.NewMember(service, pod, o.nodesBroadcastAddressType, o.clientsBroadcastAddressType)
+	member, err := identity.NewMember(service, pod, o.nodesBroadcastAddressType, o.clientsBroadcastAddressType, args)
 	if err != nil {
 		return fmt.Errorf("can't create new member from objects: %w", err)
 	}
 
 	klog.V(2).InfoS("Starting scylla")
 
-	cfg := config.NewScyllaConfig(member, o.kubeClient, o.scyllaClient, o.CPUCount, o.ExternalSeeds)
+	cfg := config.NewScyllaConfig(member, o.kubeClient, o.CPUCount, o.ExternalSeeds)
 	scyllaCmd, err := cfg.Setup(ctx)
 	if err != nil {
 		return fmt.Errorf("can't set up scylla: %w", err)
--- a/pkg/controller/scylladbdatacenter/resource.go
+++ b/pkg/controller/scylladbdatacenter/resource.go
@@ -613,6 +613,16 @@ func StatefulSetForRack(rack scyllav1alpha1.RackSpec, sdc *scyllav1alpha1.Scylla
 							Ports:           scyllaContainerPorts,
 							// TODO: unprivileged entrypoint
 							Command: func() []string {
+								var positionalArgs []string
+
+								if len(sdc.Spec.ScyllaDB.AdditionalScyllaDBArguments) > 0 {
+									positionalArgs = append(positionalArgs, sdc.Spec.ScyllaDB.AdditionalScyllaDBArguments...)
+								}
+
+								if sdc.Spec.ScyllaDB.EnableDeveloperMode != nil && *sdc.Spec.ScyllaDB.EnableDeveloperMode {
+									positionalArgs = append(positionalArgs, "--developer-mode=1")
+								}
+
 								cmd := []string{
 									"/usr/bin/bash",
 									"-euEo",
@@ -653,13 +663,28 @@ exec /mnt/shared/scylla-operator sidecar \
 --service-name=$(SERVICE_NAME) \
 --cpu-count=$(CPU_COUNT) \
 --loglevel=2 \
-` + func() string {
-										if len(sdc.Spec.ScyllaDB.ExternalSeeds) > 0 {
-											return fmt.Sprintf("--external-seeds=%s", strings.Join(sdc.Spec.ScyllaDB.ExternalSeeds, ","))
-										}
-										return ""
-									}(),
+` +
+										func() string {
+											var optionalArgs []string
+
+											if len(sdc.Spec.ScyllaDB.ExternalSeeds) > 0 {
+												optionalArgs = append(optionalArgs, fmt.Sprintf("--external-seeds=%s", strings.Join(sdc.Spec.ScyllaDB.ExternalSeeds, ",")))
+											}
+
+											return strings.Join(optionalArgs, ` \`)
+										}() +
+										func() string {
+											if len(positionalArgs) > 0 {
+												return ` -- "$@"`
+											}
+											return ""
+										}(),
 									),
+								}
+
+								if len(positionalArgs) > 0 {
+									cmd = append(cmd, "--")
+									cmd = append(cmd, positionalArgs...)
 								}
 
 								return cmd
--- a/pkg/controller/scylladbdatacenter/resource_test.go
+++ b/pkg/controller/scylladbdatacenter/resource_test.go
@@ -1316,6 +1316,49 @@ scylla-manager-agent \
 			expectedError: nil,
 		},
 		{
+			name: "new StatefulSet with non-empty additional scyllaDB arguments",
+			rack: newBasicRack(),
+			scyllaDBDatacenter: func() *scyllav1alpha1.ScyllaDBDatacenter {
+				sc := newBasicScyllaDBDatacenter()
+				sc.Spec.ScyllaDB.AdditionalScyllaDBArguments = []string{
+					"--batch-size-warn-threshold-in-kb=128",
+					"--batch-size-fail-threshold-in-kb",
+					"1024",
+					`--commitlog-sync="batch"`,
+				}
+				return sc
+			}(),
+			existingStatefulSet: nil,
+			expectedStatefulSet: func() *appsv1.StatefulSet {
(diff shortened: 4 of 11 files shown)


helm/scylla-operator/templates/scyllacluster_member_clusterrole_def.yaml
```diff
@@ -47,6 +47,8 @@ rules:
   - list
   - patch
   - watch
+# Sidecar no longer fetches ScyllaClusters, as necessary information is provided by the Operator to sidecar via arguments.
```
rzetelskik, line 50: nit: add a TODO to make this stand out a bit more (an issue as well?)

pkg/cmd/operator/sidecar.go
```diff
@@ -95,6 +97,8 @@ func NewSidecarCmd(streams genericclioptions.IOStreams) *cobra.Command {
 	cmd.Flags().StringSliceVar(&o.ExternalSeeds, "external-seeds", o.ExternalSeeds, "The external seeds to propagate to ScyllaDB binary on startup as \"seeds\" parameter of seed-provider.")
 	cmd.Flags().StringVarP(&o.NodesBroadcastAddressTypeString, "nodes-broadcast-address-type", "", o.NodesBroadcastAddressTypeString, "Address type that is broadcasted for communication with other nodes.")
 	cmd.Flags().StringVarP(&o.ClientsBroadcastAddressTypeString, "clients-broadcast-address-type", "", o.ClientsBroadcastAddressTypeString, "Address type that is broadcasted for communication with clients.")
+	cmd.Flags().StringSliceVar(&o.AdditionalScyllaDBArguments, "additional-scylladb-arguments", o.AdditionalScyllaDBArguments, "specify a list of arguments appended to the ScyllaDB binary during startup. When set, ScyllaDB may behave unexpectedly, and every such setup is considered unsupported. Instead, consider using customConfigMapRef for setting custom ScyllaDB configuration options.")
```
rzetelskik, line 100:
```suggestion
	cmd.Flags().StringSliceVar(&o.AdditionalScyllaDBArguments, "additional-scylladb-arguments", o.AdditionalScyllaDBArguments, "Specifies a list of arguments appended to the ScyllaDB binary during startup. When set, ScyllaDB may behave unexpectedly, and every such setup is considered unsupported. Instead, consider using customConfigMapRef for setting custom ScyllaDB configuration options.")
```
for consistency with the other descriptions (including the other one added in this commit)

pkg/sidecar/config/config.go
```diff
@@ -229,38 +221,22 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
-	// configured in the kubelet, otherwise scylla will crash.
-	if cluster.Spec.CpuSet {
-		cpusAllowed, err := getCPUsAllowedList("/proc/1/status")
-		if err != nil {
-			return nil, errors.WithStack(err)
-		}
-		if err := s.validateCpuSet(ctx, cpusAllowed, s.cpuCount); err != nil {
-			return nil, errors.WithStack(err)
-		}
-		args["cpuset"] = &cpusAllowed
```
rzetelskik, line 245: nit: since this is being removed, a PR/commit title could include a note so that it propagates to release notes

pkg/sidecar/config/config.go
```diff
@@ -229,38 +221,22 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
-		if !version.SupportFeatureUnsafe(semver.ScyllaVersionThatSupportsArgs) {
-			klog.InfoS("This scylla version does not support ScyllaArgs. ScyllaArgs is ignored", "version", cluster.Spec.Version)
-		} else {
-			appendScyllaArguments(ctx, s, cluster.Spec.ScyllaArgs, args)
-		}
+	if len(m.AdditionalScyllaDBArguments) > 0 {
+		appendScyllaArguments(ctx, s, strings.Join(m.AdditionalScyllaDBArguments, " "), args)
 	}
 
-	if _, err := os.Stat(scyllaIOPropertiesPath); err == nil && version.SupportFeatureSafe(semver.ScyllaVersionThatSupportsDisablingIOTuning) {
```
rzetelskik, line 260: var `semver.ScyllaVersionThatSupportsDisablingIOTuning` will no longer be used so it can be removed from semver package

pkg/sidecar/config/config.go
```diff
@@ -229,38 +221,22 @@ func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
...
+		return nil, errors.WithStack(err)
+	}
+	args["cpuset"] = &cpusAllowed
 
-	version := semver.NewScyllaVersion(cluster.Spec.Version)
-
-	klog.InfoS("Scylla version detected", "version", version)
-
-	if len(cluster.Spec.ScyllaArgs) > 0 {
-		if !version.SupportFeatureUnsafe(semver.ScyllaVersionThatSupportsArgs) {
```
rzetelskik, line 253: should this also be removed from validation then?

tnozicka:
> Additionally, this commit removes support for the deprecated cpuset field. It is now treated as if it is always set to true regardless of its value.

This should go in a dedicated PR, especially for visibility.

pkg/controller/scylladbdatacenter/resource_test.go
```diff
@@ -741,6 +746,8 @@ exec /mnt/shared/scylla-operator sidecar \
 										}() + ` \
 --nodes-broadcast-address-type=ServiceClusterIP \
 --clients-broadcast-address-type=ServiceClusterIP \
+--developer-mode=true \
+--additional-scylladb-arguments="--log-to-stdout=1 --log-to-stderr=1" \
```
tnozicka, line 750: this grows the flagset for something we plan to get rid of and is harder to escape - I think it will be better to use args that just propagate through, like

```
scylla-operator sidecar --loglevel=2 -- --developer-mode=1 --log-to-stdout=1 --log-to-stderr=1
```

and eventually became what we pass to scylladb directly

pkg/cmd/operator/sidecar.go
```diff
@@ -38,6 +38,8 @@ type SidecarOptions struct {
 	ExternalSeeds                     []string
 	NodesBroadcastAddressTypeString   string
 	ClientsBroadcastAddressTypeString string
+	AdditionalScyllaDBArguments       []string
+	DeveloperMode                     bool
```
tnozicka, line 42: I'd like to se this in our generated yaml config

helm/scylla-operator/templates/scyllacluster_member_clusterrole_def.yaml
zimnx: done

pkg/sidecar/config/config.go
zimnx: extracted to separate PR

pkg/sidecar/config/config.go
zimnx: done

pkg/sidecar/config/config.go
zimnx: done

zimnx:
> This should go in a dedicated PR, especially for visibility.

extracted to separate PR #2139

pkg/cmd/operator/sidecar.go
zimnx: Unfortunately developer-mode is not only parameter to ScyllaDB, but also it's image entrypoint. When it's specified in the config and not as a parameter, iotune and additional hardware checks are executed and prevent Scylla process from starting.

pkg/controller/scylladbdatacenter/resource_test.go
zimnx: Proposed format is ambigous to me, it's easy to confuse which parameters belongs to which binary. I prefer explicit name as it's clear where these should be used.

tnozicka: This is a standard way to pass the args to another binary - literarly from `kubectl run --help`
```
  # Start the nginx pod using the default command, but use custom arguments (arg1 .. argN) for that command
  kubectl run nginx --image=nginx -- <arg1> <arg2> ... <argN>
```

pkg/cmd/operator/sidecar.go
tnozicka: ok, we can get back to it when we start scylla directly

zimnx:
https://github.com/scylladb/scylla-operator/pull/2139 landed
/hold cancel

pkg/controller/scylladbdatacenter/resource_test.go
zimnx: with `run` command it's more obvious, with `sidecar` it's not. Anyways, changed to use positional args.

zimnx:
@rzetelskik @tnozicka ptal

pkg/cmd/operator/sidecar.go
```diff
@@ -95,6 +97,7 @@ func NewSidecarCmd(streams genericclioptions.IOStreams) *cobra.Command {
 	cmd.Flags().StringSliceVar(&o.ExternalSeeds, "external-seeds", o.ExternalSeeds, "The external seeds to propagate to ScyllaDB binary on startup as \"seeds\" parameter of seed-provider.")
 	cmd.Flags().StringVarP(&o.NodesBroadcastAddressTypeString, "nodes-broadcast-address-type", "", o.NodesBroadcastAddressTypeString, "Address type that is broadcasted for communication with other nodes.")
 	cmd.Flags().StringVarP(&o.ClientsBroadcastAddressTypeString, "clients-broadcast-address-type", "", o.ClientsBroadcastAddressTypeString, "Address type that is broadcasted for communication with clients.")
+	cmd.Flags().BoolVarP(&o.DeveloperMode, "developer-mode", "", o.DeveloperMode, "Specifies if ScyllaDB runs in developer mode.")
```
tnozicka, line 100: Can we prepend this to the arbitrary scylladb args in the operator and avoid plumbing in an extra sidecar flag that will eventually do the same inside the sidecar?

pkg/controller/scylladbdatacenter/resource_test.go
tnozicka: thanks, I think it will be even clearer when the sidecar (that's really a starter) becomes the starter

pkg/cmd/operator/sidecar.go
zimnx: good idea, moved to positional arg

rzetelskik (approved):
lgtm
/assign tnozicka

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -653,12 +653,32 @@ exec /mnt/shared/scylla-operator sidecar \
...
+											if len(sdc.Spec.ScyllaDB.AdditionalScyllaDBArguments) > 0 {
+												positionalArgs = append(positionalArgs, sdc.Spec.ScyllaDB.AdditionalScyllaDBArguments...)
+											}
+
+											if sdc.Spec.ScyllaDB.EnableDeveloperMode != nil && *sdc.Spec.ScyllaDB.EnableDeveloperMode {
+												positionalArgs = append(positionalArgs, "--developer-mode=1")
+											}
+
+											if len(positionalArgs) > 0 {
+												return fmt.Sprintf(" -- %s \n", strings.Join(positionalArgs, " "))
```
tnozicka, line 678: Shouldn't we escape each arg with `%q` so it propagates correctly?

Security wise, given that the args are supplied by the user we should prevent injecting bash snippets as well (subshell, reading env vars).

tnozicka: to avoid escaping, maybe supplying the args to the bash itself and then propagating them with `-- "$@"` might work
```bash
bash -c 'echo "$@"' -- foo bar
```

zimnx: done, tested with injected subshell printing secret and it doesn't work.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -653,15 +663,30 @@ exec /mnt/shared/scylla-operator sidecar \
...
+											if len(sdc.Spec.ScyllaDB.ExternalSeeds) > 0 {
+												optionalArgs = append(optionalArgs, fmt.Sprintf("--external-seeds=%s", strings.Join(sdc.Spec.ScyllaDB.ExternalSeeds, ",")))
+											}
+
+											return strings.Join(optionalArgs, ` \`)
+										}() +
+										func() string {
+											if len(positionalArgs) > 0 {
+												return ` -- "$@"`
+											}
```
tnozicka, line 679: nit: I think this can be always set even if the args are empty
(simplifies branching and testing matrix)

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -653,15 +663,30 @@ exec /mnt/shared/scylla-operator sidecar \
...
+											return ""
+										}(),
 									),
 								}
 
+								if len(positionalArgs) > 0 {
+									cmd = append(cmd, "--")
+									cmd = append(cmd, positionalArgs...)
+								}
```
tnozicka, line 688: nit: this should also be fine to set unconditionally

tnozicka (commented):
thanks for the updates
overall lgtm, but 2 nits would simplify branching and a testing matrix

pkg/controller/scylladbdatacenter/resource.go
zimnx: It would make testing optional args harder, as such test would have to inject the expected arg into string at the right place. Check unit test for external seeds for reference.

pkg/controller/scylladbdatacenter/resource.go
zimnx: I don't agree with previous comment, and I would leave this as is for consistency.

pkg/controller/scylladbdatacenter/resource.go
tnozicka: ok, I tend to prefer code coverage and simpler branching - I don't think unit test complexity should take over it but it's a nit

PR #2139, opened 2024-10-02
Remove support for deprecated `cpuset` field

### LINKED ISSUE #2140: Remove support for deprecated cpuset field

Remove support for cpuset field, assuming it's always set to true regardless of value set.

This field gives misleading feeling that it enables cpupinning which on Kubernetes is handled differently than on baremetals. Instead having a knob for it, Operator should set allowed cpus always.

_No response_

zimnx:
It is now treated as if it is always set to true regardless of its value.
Validation of ScyllaCluster resources is relaxed to not require full cores and resource requests == limits.
Conditions for getting Guaranteed QoS class are explained in documentation.

Fixes #2140

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -1217,7 +1217,7 @@ spec:
                     type: object
                   type: array
                 cpuset:
-                  description: 'cpuset determines if the cluster will use cpu-pinning. Deprecated: `cpuset` is deprecated and may be ignored in the future.'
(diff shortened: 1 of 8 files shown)


tnozicka (commented):
> It is now treated as if it is always set to true regardless of its value.

This should be reflected in the API comment as well, otherwise this lgtm

rzetelskik (approved):
lgtm
/assign tnozicka

zimnx:
> This should be reflected in the API comment as well, otherwise this lgtm

done

PR #2142, opened 2024-10-03
Set managed hash labels in manager tasks and only update tasks when managed hash changes

rzetelskik:
**Description of your changes:** Currently, Scylla Manager integration controller decides to update the tasks defined in ScyllaCluster's spec by checking for deep equality between the task definition in ScyllaCluster's spec and the task status obtained from the Manager's state.
Due to some discrepancies between how we and Scylla Manager keep tasks' properties, and the fact that some values would be converted, but not converted back when reading the status, this would often lead to task update hot loops, where a task update requests would be sent despite no changes in spec.

Such scenario can be observed by specifying a repair task with e.g. `smallTableThreshold` set:
```
  repairs:
  - failFast: true
    intensity: "1"
    name: repair
    numRetries: 1
    parallel: 1
    smallTableThreshold: 1GiB
```

The threshold would be converted to bytes, so that the corresponding status would be:
```
  repairs:
  - cron: '{"spec":"","start_date":"0001-01-01T00:00:00Z"}'
    failFast: true
    id: d0550c36-f799-4ac4-8b01-aab7aaf042bf
    intensity: "1"
    interval: ""
    name: repair
    numRetries: 0
    parallel: 1
    smallTableThreshold: "1073741824"
    startDate: "2024-10-02T15:39:22.907Z"
    timezone: ""
```

The value would not be converted back, causing the deep equality test to always return false, and so an update would be triggered on every comparison, causing a hot loop. 

There were also cases reported by users in which we suspected the constant updates were causing a high cpu load: https://github.com/scylladb/scylla-manager/issues/3920.

This PR addresses this issue by making use of `labels` recently added to the manager's task API. Similarily to how we handle managed k8s resources, the scylla-manager controller now computes a hash from the task's spec and sets a managed hash label. Therefore, an update request is only sent when a task spec changes in ScyllaCluster and the managed hash no longer matches the one set in scylla manager's state (and correspondingly, ScyllaCluster's status). 
Changes in the manager's state which are not reflected in our API (e.g. when a user manually updates a task with sctool) will not trigger an update, but they will be overwritten when we eventually reconcile.

This PR does not extend ScyllaCluster's API with the `labels` field for Scylla Manager's tasks, at this point we're only using them internally and don't expose them in spec to the users of our API. If there ever is a need to do add it, the managed hash would take precedence over colliding labels.

[...]

Issue #2096, opened 2024-08-27
Flake - should setup and maintain up to date TLS certificates


zimnx:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2066/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1828179211538403328

### Snippet of what failed.

```
 ScyllaCluster [It] should setup and maintain up to date TLS certificates
github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_tls.go:42
  Timeline >>
  STEP: Creating a new namespace @ 08/26/24 21:30:43.436
  Aug 26 21:30:43.466: INFO: Created namespace "e2e-test-scyllacluster-2ms98-0-85njt".
  STEP: Waiting for service account token Secret "e2e-user-token" in namespace "e2e-test-scyllacluster-2ms98-0-85njt". @ 08/26/24 21:30:43.522
  STEP: Waiting for default ServiceAccount in namespace "e2e-test-scyllacluster-2ms98-0-85njt". @ 08/26/24 21:30:44.178
  STEP: Waiting for kube-root-ca.crt in namespace "e2e-test-scyllacluster-2ms98-0-85njt". @ 08/26/24 21:30:44.492
  STEP: Creating an initial ScyllaCluster with a single node @ 08/26/24 21:30:44.502
  STEP: Waiting for the ScyllaCluster to roll out (RV=5639) @ 08/26/24 21:30:44.533
  Aug 26 21:35:13.491: INFO: ScyllaCluster e2e-test-scyllacluster-2ms98-0-85njt/basic-nztd8 (RV=11632) is rolled out
  STEP: Verifying the ScyllaCluster @ 08/26/24 21:35:13.491
  Aug 26 21:35:13.648: INFO: Found 1 pvc(s) in namespace "e2e-test-scyllacluster-2ms98-0-85njt"
  Aug 26 21:35:13.648: INFO: Found 1 pvc(s) for ScyllaCluster "e2e-test-scyllacluster-2ms98-0-85njt/basic-nztd8"
  STEP: Waiting for the ScyllaCluster(s) to reach consistency ALL @ 08/26/24 21:35:13.757
  Aug 26 21:35:14.163: INFO: ScyllaDB nodes have reached status consistency.
  STEP: Inserting data @ 08/26/24 21:35:14.259
[...]

PR #2156, opened 2024-10-15
Use Scylla Manager cluster labels for cluster reconciliation

### LINKED ISSUE #1902: Manager controller recreates clusters when manager cluster ID is missing from status
### What happened?

Currently, the manager cluster ID is saved in ScyllaCluster's status on cluster creation. If the controller fails to update ScyllaCluster's status, the ID is lost, or an older generation of the object is reconciled, the controller will delete the existing cluster from the manager state and create it again.

The issue and its root cause are similar https://github.com/scylladb/scylla-operator/issues/1752.

This not only adds a superfluous workload, but may introduce incorrectness, involving e.g. task retention.

/priority important-soon
/assign

Clusters in manager state should not be deleted once they've been created successfully.

n/a

### Scylla Operator version

master

n/a

n/a

Unfortunately, we now have no reliable way of telling whether a cluster existing in manager state corresponds to a K8S object if we don't have the ID.
This should be easy to fix with https://github.com/scylladb/scylla-manager/issues/3219, since we'll be able to save metadata in manager state, and so we'll be able to "reclaim" the cluster despite not having its ID.

rzetelskik:
**Description of your changes:** Currently, when the manager controller fails to save manager's cluster ID in ScyllaCluster's status after cluster's registration with manager, the cluster is deleted and recreated again. As update conflicts are not a rare occurrence, this often causes many unnecessary recreation attempts.
To make the reconciliation more robust, this PR changes this behaviour. Instead of using the ID from status, labels from manager state are used. A cluster is created with a label holding the owner's UID, which allows us to maintain and recognize cluster's identity without relying on the status of our API resources. In turn clusters are only deleted when the owner UID labels is not matching the UID of the current owner, in order to avoid name collisions.

The labels are also extended with a managed hash label to align the cluster update logic with changes recently introduced in https://github.com/scylladb/scylla-operator/pull/2142.

The logic related to creating "actions" is modified to produce one cluster-related action at once and requeue in order to only schedule any further actions on next iteration. The reasoning behind it is to try avoiding errors related to task actions in case of a required cluster action, e.g. when auth token needs to be updated first.

Additionally, the manager state computed in each reconciliation loop is reduced to only one cluster, since cluster names in manager are unique and propagating additional clusters to the state is redundant.

Unit tests are also extended to cover these scenarios and unified for consistency.

Resolves #1902 

/kind bug
/priority important-soon
/cc

### DIFF
--- a/pkg/controller/manager/status.go
+++ b/pkg/controller/manager/status.go
@@ -6,13 +6,35 @@ import (
 	"context"
 
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/pointer"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 )
 
-func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, managerState *state) *scyllav1.ScyllaClusterStatus {
+func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, state *managerClusterState) *scyllav1.ScyllaClusterStatus {
 	status := sc.Status.DeepCopy()
+
+	status.ManagerID = pointer.Ptr("")
+	status.Backups = []scyllav1.BackupTaskStatus{}
+	status.Repairs = []scyllav1.RepairTaskStatus{}
+
+	if state.Cluster == nil {
+		return status
+	}
+
+	ownerUIDLabelValue, hasOwnerUIDLabel := state.Cluster.Labels[naming.OwnerUIDLabel]
+	if !hasOwnerUIDLabel {
+		klog.Warningf("Cluster %q is missing an owner UID label.", state.Cluster.Name)
+	}
+
+	if ownerUIDLabelValue != string(sc.UID) {
+		// Cluster is not owned by ScyllaCluster, do not propagate its status.
+		return status
+	}
+
+	status.ManagerID = pointer.Ptr(state.Cluster.ID)
 
 	repairTaskClientErrorMap := map[string]string{}
 	for _, rts := range status.Repairs {
@@ -21,7 +43,6 @@ func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, managerState *s
 		}
 	}
 
-	status.Repairs = []scyllav1.RepairTaskStatus{}
 	for _, rt := range sc.Spec.Repairs {
 		repairTaskStatus := scyllav1.RepairTaskStatus{
 			TaskStatus: scyllav1.TaskStatus{
@@ -29,7 +50,7 @@ func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, managerState *s
 			},
 		}
 
-		managerTaskStatus, isInManagerState := managerState.RepairTasks[rt.Name]
+		managerTaskStatus, isInManagerState := state.RepairTasks[rt.Name]
 		if isInManagerState {
 			repairTaskStatus = managerTaskStatus
 		} else {
@@ -52,7 +73,6 @@ func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, managerState *s
 		}
 	}
 
-	status.Backups = []scyllav1.BackupTaskStatus{}
 	for _, bt := range sc.Spec.Backups {
 		backupTaskStatus := scyllav1.BackupTaskStatus{
 			TaskStatus: scyllav1.TaskStatus{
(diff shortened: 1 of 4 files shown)


pkg/controller/manager/sync.go
```diff
@@ -24,59 +25,78 @@ func (c *Controller) getAuthToken(sc *scyllav1.ScyllaCluster) (string, error) {
...
+		// Cluster names in manager state are unique, so it suffices to only find one with a matching name.
+		if mc.Name == clusterName {
+			managerCluster = mc
+			break
 		}
+	}
+
+	if managerCluster == nil {
+		return &managerClusterState{}, nil
+	}
```
zimnx, line 46: could be replace with slices.Find which would be much shorter

pkg/controller/manager/sync_test.go
```diff
@@ -4,675 +4,829 @@ package manager
...
 		},
 	}
 
-	for _, test := range tcs {
-		t.Run(test.Name, func(t *testing.T) {
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
 			ctx := context.Background()
```
zimnx, line 811: nit: add sane timeout

rzetelskik: nothing in this function consumes the context, it's just legacy, so I'd be leaning towards removing the param if anyhing - wdyt?

pkg/controller/manager/sync.go
rzetelskik: good point, fixed

pkg/controller/manager/status.go
```diff
@@ -6,30 +6,41 @@ import (
...
 
-func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, managerState *state) *scyllav1.ScyllaClusterStatus {
+func (c *Controller) calculateStatus(sc *scyllav1.ScyllaCluster, state *managerClusterState) *scyllav1.ScyllaClusterStatus {
 	status := sc.Status.DeepCopy()
 
+	status.ManagerID = pointer.Ptr("")
+	status.Backups = []scyllav1.BackupTaskStatus{}
+	status.Repairs = []scyllav1.RepairTaskStatus{}
+
+	if state.Cluster == nil || state.Cluster.Labels[naming.OwnerUIDLabel] != string(sc.UID) {
```
tnozicka, line 23: should we report an error or warning when the label is missing? (globally)

tnozicka (approved):
/approve

/assign @zimnx

pkg/controller/manager/sync_test.go
zimnx: I would be against removing context parameter, it's useful once its has its user and having it is recommended practice. I'm fine not touching the timeout too.

pkg/controller/manager/status.go
tnozicka: I mean it more to distinguish between the label being there (lookup in the map), before it's compared with the sc.UID, so you have the distinction in the logs between UID mismatch and `state.cluster.Labels` not having it.

rzetelskik: I see. Changed to warning on missing label and comment/debug log on mismatch.

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[cf03afe](https://github.com/scylladb/scylla-operator/commit/cf03afe3d9aefa9a13f710ac220a551fd3887ba2)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2156/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1846873675131785216)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2156). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

cluster provisioning failed
/retest

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[cf03afe](https://github.com/scylladb/scylla-operator/commit/cf03afe3d9aefa9a13f710ac220a551fd3887ba2)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2156/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1846888608951177216)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2156). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

tls test flake, can't possibly be related?
https://github.com/scylladb/scylla-operator/issues/2096#issuecomment-2425801644
/retest

zimnx (approved):
/lgtm

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[cf03afe](https://github.com/scylladb/scylla-operator/commit/cf03afe3d9aefa9a13f710ac220a551fd3887ba2)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2156/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1848259917215961088)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2156). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

/test images
/retest

PR #2162, opened 2024-10-22
Update Git Clone command

susiesolis:

Changing git clone from ssh to https

Resolves #
If you are not logged into to GH then git clone using ssh throws an access denied error which can lead to a poor user experience. Changing to https://github.com/scylladb/scylla-operator.git allows it to work without authentication.


tnozicka (approved):
/approve
/lgtm
thanks

PR #2164, opened 2024-10-28
Update ScyllaDB versions to OS 6.2.0 and Enterprise 2024.1.11

rzetelskik:
**Description of your changes:** This PR updates the default ScyllaDB version used in our tests, manifests, examples and docs to 6.2.0. It also bumps the version used for ScyllaDB Manager restore procedure test for versions 2024.1.Z or older to 2024.1.11. We should be testing against the most recent releases.

Resolves #

/kind machinery
/priority important-longterm


zimnx (approved):
/lgtm

PR #2165, opened 2024-10-28
Extend node tune controller with status conditions

### LINKED ISSUE #1920: Extend node tune controller with status conditions

Currently, the node tune controller only updates NodeConfig's NodeStatuses on reconciliation. It should be extended with conditions so that they can be propagated to NodeConfig.

/kind feature
/needs-priority

Being able to query NodeConfig conditions, instead of checking NodeStatuses.

_No response_

rzetelskik:
**Description of your changes:** Node tune controller does not currently set status conditions on NodeConfig's reconciliation. Instead it only set status' NodeStatuses. While this is enough for checking the status programmatically, setting and aggregating status conditions allows for a better UX, by giving the users an option to query NodeConfig conditions.
This PR extends the node tune controller with per-node status conditions and further aggregates them with the existing ones in NodeConfig controller.

It also adds handlers for NodeConfig events to node tune controller and aligns the existing e2e tests with the changes.

Resolves #1920 

/kind feature
/priority important-longterm
/cc

### DIFF
--- a/pkg/controller/nodeconfig/sync.go
+++ b/pkg/controller/nodeconfig/sync.go
@@ -107,7 +107,12 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 		return objectErr
 	}
 
-	status := ncc.calculateStatus(nc)
+	matchingNodes, err := ncc.getMatchingNodes(nc)
+	if err != nil {
+		return fmt.Errorf("can't get matching Nodes: %w", err)
+	}
+
+	status := ncc.calculateStatus(nc, matchingNodes)
 
 	if nc.DeletionTimestamp != nil {
 		return ncc.updateStatus(ctx, nc, status)
@@ -207,37 +212,41 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 		errs = append(errs, fmt.Errorf("can't sync DaemonSet(s): %w", err))
 	}
 
-	matchingNodes, err := ncc.getMatchingNodes(nc)
-	if err != nil {
-		errs = append(errs, fmt.Errorf("can't get matching Nodes: %w", err))
-		return utilerrors.NewAggregate(errs)
-	}
-
 	// Aggregate conditions.
 	var aggregationErrs []error
 
 	availableConditions := controllerhelpers.FindStatusConditionsWithSuffix(statusConditions, scyllav1alpha1.AvailableCondition)
 
-	// Use a default available node condition for nodes which haven't propagated their conditions to status yet.
-	nodeAvailableConditionTypeFunc := func(node *corev1.Node) string {
-		return fmt.Sprintf(internalapi.NodeAvailableConditionFormat, node.Name)
-	}
-	defaultNodeAvailableConditions := makeDefaultNodeConditions(
-		nc.Generation,
-		matchingNodes,
-		availableConditions,
-		nodeAvailableConditionTypeFunc,
-		func(node *corev1.Node) metav1.Condition {
-			return metav1.Condition{
-				Type:               nodeAvailableConditionTypeFunc(node),
-				Status:             metav1.ConditionFalse,
-				ObservedGeneration: nc.Generation,
-				Reason:             internalapi.AwaitingConditionReason,
-				Message:            fmt.Sprintf("Awaiting Available condition of node %q to be set.", naming.ObjRef(node)),
-			}
+	controllerNodeAvailableConditionTypeFuncs := []func(*corev1.Node) string{
+		func(n *corev1.Node) string {
+			return fmt.Sprintf(internalapi.NodeSetupAvailableConditionFormat, n.Name)
 		},
+		func(n *corev1.Node) string {
+			return fmt.Sprintf(internalapi.NodeTuneAvailableConditionFormat, n.Name)
+		},
+	}
+	newNodeAvailableConditionFunc := func(generation int64, node *corev1.Node) metav1.Condition {
+		return metav1.Condition{
+			Type:               fmt.Sprintf(internalapi.NodeAvailableConditionFormat, node.Name),
+			Status:             metav1.ConditionTrue,
+			Reason:             internalapi.AsExpectedReason,
+			Message:            "",
+			ObservedGeneration: generation,
+		}
+	}
+	nodeAvailableConditions, err := aggregateNodeStatusConditions(
+		nc.Generation,
+		availableConditions,
+		matchingNodes,
+		controllerNodeAvailableConditionTypeFuncs,
+		newNodeAvailableConditionFunc,
 	)
-	availableConditions = append(availableConditions, defaultNodeAvailableConditions...)
+	if err != nil {
+		aggregationErrs = append(aggregationErrs, fmt.Errorf("can't aggregate node available conditions: %w", err))
+	} else {
+		availableConditions = append(availableConditions, nodeAvailableConditions...)
+	}
+
 	availableCondition, err := controllerhelpers.AggregateStatusConditions(
 		availableConditions,
 		metav1.Condition{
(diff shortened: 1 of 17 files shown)


test/e2e/set/nodeconfig/nodeconfig_optimizations.go
```diff
@@ -356,12 +355,11 @@ var _ = g.Describe("NodeConfig Optimizations", framework.Serial, func() {
 			nc.Name,
 			controllerhelpers.WaitForStateOptions{TolerateDelete: false},
 			utils.IsNodeConfigRolledOut,
-			utils.IsNodeConfigDoneWithNodeTuningFunc(matchingNodes),
-			utils.IsNodeConfigDoneWithContainerTuningFunc(pod.Spec.NodeName, scyllaContainerID),
 		)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		verifyNodeConfig(ctx, f.KubeAdminClient(), nc)
+		o.Expect(controllerhelpers.IsNodeTunedForContainer(nc, pod.Spec.NodeName, scyllaContainerID))
```
zimnx, line 362: Assertion is missing

pkg/controller/nodetune/status.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+func (ncdc *Controller) calculateStatus(nc *v1alpha1.NodeConfig) *v1alpha1.NodeConfigStatus {
+	status := nc.Status.DeepCopy()
+	status.ObservedGeneration = nc.Generation
```
zimnx, line 17: This updates ObservedGeneration of entire NodeConfig, while this controller is only responsible for single Node. 
This field of NodeConfigStatus should be (and is already) reconciled by nodeconfig controller.

pkg/controller/nodesetup/conditions.go
```diff
@@ -3,6 +3,8 @@
 package nodesetup
 
 const (
+	// TODO(rzetelskik): change format to NodeSetup in next API version
```
zimnx, line 6: add issue ref where further discussion can be found and motivation behind the change can be explained

pkg/controller/nodetune/sync.go
```diff
@@ -124,18 +109,84 @@ func (ncdc *Controller) sync(ctx context.Context) error {
...
+		},
+	)
 	if err != nil {
 		errs = append(errs, fmt.Errorf("can't sync jobs: %w", err))
 	}
 
-	err = ncdc.updateNodeStatus(ctx, nodeStatus)
+	// Aggregate node conditions.
+	var aggregationErrs []error
+	nodeAvailableConditionType := fmt.Sprintf(internalapi.NodeTuneAvailableConditionFormat, ncdc.nodeName)
```
zimnx, line 132:
```suggestion
	nodeTuneAvailableConditionType := fmt.Sprintf(internalapi.NodeTuneAvailableConditionFormat, ncdc.nodeName)
```
Otherwise it's ambigous. The same with below ones.

zimnx (commented):
first batch

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
rzetelskik: thanks, fixed

pkg/controller/nodetune/status.go
rzetelskik: I see your point, but from the correctness perspective, is there anything inherently wrong about this?

If so it should also be fixed in the existing nodesetup controller.

pkg/internalapi/conditions.go
```diff
@@ -1,9 +1,14 @@
 package internalapi
 
 const (
-	NodeAvailableConditionFormat   = "Node%sAvailable"
-	NodeProgressingConditionFormat = "Node%sProgressing"
-	NodeDegradedConditionFormat    = "Node%sDegraded"
+	// TODO(rzetelskik): change format to NodeSetup in next API version
+	NodeSetupAvailableConditionFormat   = "Node%sAvailable"
+	NodeSetupProgressingConditionFormat = "Node%sProgressing"
+	NodeSetupDegradedConditionFormat    = "Node%sDegraded"
```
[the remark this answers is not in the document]
rzetelskik: > Why Node%s* became conditions for NodeSetup controller?

Because they are being set by it. They are then aggregated together with NodeTune's node conditions in the top-level NodeConfig conditions into the generic Available/Progressing/Degraded conditions.

I understand that what you're proposing is to add one more level of aggregation into `Node<NodeName><ConditionType>` in the NodeConfig controller, and have NodeSetup controller aggregate its conditions into `NodeSetup<NodeName><ConditionType>` instead. I can see the purpose behind it (although it doesn't change much functionally, it's just an additional layer), but naming-wise it would end up being confusing imo, since the existing NodeSetup controller conditions are already in the `<ControllerName>Node<NodeName><ConditionType>` form, and we can't change them?

pkg/controller/nodetune/sync.go
rzetelskik: I started with this and moved to `nodeAvailableConditionType` as I figured it should be clear from context, but in that case I don't mind going back to it.

pkg/controller/nodetune/status.go
tnozicka: ObservedGeneration shall only be updated by the main loop and relates to its fields that it protects. For fields updated from different loops you either need a separate ObservedGeneration (like for conditions) or to live without the protections it provides.

rzetelskik: done

pkg/controller/nodeconfig/sync.go
```diff
@@ -254,26 +283,55 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
 
 	progressingConditions := controllerhelpers.FindStatusConditionsWithSuffix(statusConditions, scyllav1alpha1.ProgressingCondition)
 
-	// Use a default progressing node condition for nodes which haven't propagated their conditions to status yet.
-	nodeProgressingConditionTypeFunc := func(node *corev1.Node) string {
-		return fmt.Sprintf(internalapi.NodeProgressingConditionFormat, node.Name)
+	// Use default controller node progressing conditions for nodes which haven't propagated their conditions to status yet.
```
tnozicka, line 286: The "default" condition is called `Unknown` and the nodeconfig controller should fill in such condition for every node that haven't reported it yet (not just the aggregate). `nodeconfig.calculateStatus` seems like a good place to do it. With the existing condition present in the status, the aggregation may get easier as well.

rzetelskik: Are you saying the NodeConfig controller should actually set the missing conditions in status? I wanted to avoid having any intersections between the conditions being set by NodeConfig and NodeTune controllers, the filled in defaults I'm only using for aggregation

tnozicka: I don't see the conflict. It's the same resource (RV), so no extra conflicts. Each condition has it's own observedGeneration.

From the user's point of view, there should be info about the condition even if in Unknown state. Say the pod can fail to start, or get blocked on resources and the field would be missing for a long time.

rzetelskik: done

pkg/controller/nodesetup/conditions.go
```diff
@@ -3,15 +3,15 @@
 package nodesetup
 
 const (
-	raidControllerNodeProgressingConditionFormat = "RaidControllerNode%sProgressing"
-	raidControllerNodeDegradedConditionFormat    = "RaidControllerNode%sDegraded"
+	raidControllerNodeSetupProgressingConditionFormat = "RaidControllerNodeSetup%sProgressing"
```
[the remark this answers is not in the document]
rzetelskik: done

rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[a0714b1](https://github.com/scylladb/scylla-operator/commit/a0714b1df0262ae362cd6c5662a8758927f1c24b)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2165/pull-scylla-operator-master-e2e-gke-parallel/1854175863751315456)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2165). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

New flake https://github.com/scylladb/scylla-operator/issues/2185, unrelated to these changes though.

pkg/controller/nodeconfig/sync.go
```diff
@@ -483,3 +531,37 @@ func getNodeConfigReconciledCondition(conditions []metav1.Condition, generation
...
+	var nodeAggregateConditions []metav1.Condition
+
+	var errs []error
+	for _, n := range nodes {
+		var controllerNodeConditions []metav1.Condition
+		for _, f := range controllerNodeConditionTypeFuncs {
+			controllerNodeConditionType := f(n)
+			controllerNodeCondition := apimeta.FindStatusCondition(conditions, controllerNodeConditionType)
+			if controllerNodeCondition == nil || controllerNodeCondition.ObservedGeneration < generation {
+				errs = append(errs, fmt.Errorf("controller node condition missing: %q", controllerNodeConditionType))
```
zimnx, line 545: error msg doesn't match generation condition

pkg/controller/nodeconfig/sync.go
```diff
@@ -288,8 +307,40 @@ func (ncc *Controller) sync(ctx context.Context, key string) error {
...
+
+	controllerNodeDegradedConditionTypeFns := []func(*corev1.Node) string{
+		func(n *corev1.Node) string {
+			return fmt.Sprintf(internalapi.NodeSetupDegradedConditionFormat, n.Name)
+		},
+		func(n *corev1.Node) string {
+			return fmt.Sprintf(internalapi.NodeTuneDegradedConditionFormat, n.Name)
+		},
+	}
+	nodeDegradedConditionFn := func(generation int64, node *corev1.Node) metav1.Condition {
```
zimnx, line 320:
```suggestion
	newNodeDegradedConditionFunc := func(generation int64, node *corev1.Node) metav1.Condition {
```

zimnx (commented):
copule of comments, rest lgtm

pkg/controller/nodeconfig/sync.go
rzetelskik: ok

pkg/controller/nodeconfig/sync.go
rzetelskik: done

zimnx (approved):
lgtm
/assign tnozicka

PR #2168, opened 2024-10-29
Fix prunning of tuning related Roles and RoleBindings

zimnx:
Prunning functions of Role and RoleBinding were using namespace from non-namespaced resource. It was changed to namespace used by our tuning resources.


tnozicka (approved):
/approve
/lgtm

zimnx:
The same flake
/retest
/test images

PR #2169, opened 2024-10-29
Remove unused fields from ScyllaDB and ScyllaDBManagerAgent datacenter wide properties

zimnx:
Datacenter wide properties of ScyllaDB and ScyllaDBManagerAgent should be specified in RackTemplate.
These fields were unused and put there by mistake.


tnozicka (approved):
/approve
(lgtm)

/assign @rzetelskik

rzetelskik (approved):
/lgtm

tnozicka:
Just for the record - this is fine to do because none of the API fields have been released yet. Otherwise the backward compatibility guarantees would apply.

PR #2176, opened 2024-10-31
Bump ScyllaDB utils image to latest stable

zimnx:
With scylladb/scylladb#17787 being fixed and new images already having the fix, we should bump the EOL image to most recent stable.

Tested locally, perftune passes without complaining about missing systemd in the image.


rzetelskik (approved):
lgtm
/assign tnozicka

tnozicka (approved):
/approve
/lgtm
thx

PR #2178, opened 2024-11-04
Add an E2E test for NodeConfig mount degraded condition propagation for mounts over corrupted xfs filesystems

### LINKED ISSUE #1558: Make sure xfs mount that won't mount because it needs repair is manifested on NodeConfig

When a NodeConfig utilizes mount which systemd unit fails, it needs to set a degraded condition.

xfs is notorious for breaking on unclean restarts and often needs manual action (running xfs_repair to recover). We should manifest that into tha API so it's observable. (And all the other mount failures)

```
● mnt-persistent\x2dvolumes.mount                                                                                                                        loaded failed failed    Managed mount by Scylla Operator
root@ubuntu-2204:~# systemctl status mnt-persistent\x2dvolumes.mount
Unit mnt-persistentx2dvolumes.mount could not be found.
root@ubuntu-2204:~# journal^C
root@ubuntu-2204:~# systemctl status 'mnt-persistent\x2dvolumes.mount'
× mnt-persistent\x2dvolumes.mount - Managed mount by Scylla Operator
     Loaded: loaded (/etc/systemd/system/mnt-persistent\x2dvolumes.mount; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Fri 2023-11-10 18:50:29 UTC; 29s ago
      Where: /mnt/persistent-volumes
       What: /dev/md127
        CPU: 5ms

Nov 10 18:50:29 ubuntu-2204 systemd[1]: Mounting Managed mount by Scylla Operator...
Nov 10 18:50:29 ubuntu-2204 systemd[1]: mnt-persistent\x2dvolumes.mount: Mount process exited, code=exited, status=32/n/a
Nov 10 18:50:30 ubuntu-2204 mount[3469]: mount: /mnt/persistent-volumes: mount(2) system call failed: Structure needs cleaning.
Nov 10 18:50:29 ubuntu-2204 systemd[1]: mnt-persistent\x2dvolumes.mount: Failed with result 'exit-code'.
Nov 10 18:50:29 ubuntu-2204 systemd[1]: Failed to mount Managed mount by Scylla Operator.
````

_No response_

### PR DESCRIPTION
**Pull request by rzetelskik**
**Description of your changes:** This PR adds an e2e test verifying that NodeConfig ends up with a degraded status condition when trying to mount over a corrupted xfs filesystem. The behaviour itself, as requested by #1158, was already there since merging #2134.

Resolves #1558 

/priority important-longterm

### DIFF
--- a/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
+++ b/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
@@ -242,8 +243,9 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
 	)
 
 	type degradedEntry struct {
-		nodeConfigFunc            func() *scyllav1alpha1.NodeConfig
-		preNodeConfigCreationFunc func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context)
+		nodeConfigFunc             func() *scyllav1alpha1.NodeConfig
+		preNodeConfigCreationFunc  func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context)
+		postNodeConfigCreationFunc func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context)
 	}
 
 	g.DescribeTable("should propagate mount controller's degraded condition", func(specCtx g.SpecContext, e *degradedEntry) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rzetelskik** [comment]
@zimnx @tnozicka ptal

File: test/e2e/set/nodeconfig/nodeconfig_disksetup.go [hunk 1]
```diff
@@ -456,6 +457,87 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
...
+						{
+							Device: "/dev/loops/disk",
+							Type:   scyllav1alpha1.XFSFilesystem,
+						},
+					},
+				}
+				return nc
+			},
+			preNodeConfigCreationFunc: nil,
+			postNodeConfigCreationFunc: func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context) {
```
tnozicka, line 483: putting a whole new test into an entry seems weird to me - I think this should be a new `It` block and if there is a common setup then `BeforeEach` seems to address that

rzetelskik: > BeforeEach seems to address that

iirc you can't nest BeforeEach in container nodes and it's broken otherwise since our framework uses it for initialisation, so you can't enforce any order between multiple BeforeEach blocks

> putting a whole new test into an entry seems weird to me - I think this should be a new It block

there's already a bunch of tests there that are testing propagation of degraded status conditions for a single mount controller, only for different scenarios - so why is that an issue now if it wasn't before? I agree it grows out of proportion a bit but it's either this or a lot of code duplication

tnozicka: you are right that it's preexisting

tnozicka (approved):
I think this could be done simpler by creating a broken XFS initially using a pod in the preNodeConfigCreationFunc and reusing more of the existing code (avoids the extra rollout wait and patch) but I don't feel strong

/approve
/assign @zimnx

rzetelskik:
> I think this could be done simpler by creating a broken XFS initially using a pod in the preNodeConfigCreationFunc and reusing more of the existing code (avoids the extra rollout wait and patch) but I don't feel strong

Depends on what you consider simpler - I started by trying to do it this way but at some point I realised the client pod started to turn into the nodeconfig controller because it had to perform the exact same steps to set it up. I think this way is clearer in terms of highlighting the broken setup. In terms of the test entry complexity it's admittedly worse - perhaps the test table could use a rehaul.

test/e2e/set/nodeconfig/nodeconfig_disksetup.go
```diff
@@ -456,6 +457,87 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
...
+				stdout, stderr, err := executeInPod(f.ClientConfig(), f.KubeClient().CoreV1(), clientPod, "xfs_repair", "-o", "force_geometry", "-f", "-n", hostDevicePath)
+				o.Expect(err).NotTo(o.HaveOccurred(), stdout, stderr)
+
+				framework.By("Corrupting XFS filesystem")
+				stdout, stderr, err = executeInPod(f.ClientConfig(), f.KubeClient().CoreV1(), clientPod, "xfs_db", "-x", "-c", "blockget", "-c", "blocktrash -s 12345678 -n 1000", hostDevicePath)
+				o.Expect(err).NotTo(o.HaveOccurred(), stdout, stderr)
+
+				framework.By("Verifying that XFS filesystem is corrupted")
+				stdout, stderr, err = executeInPod(f.ClientConfig(), f.KubeClient().CoreV1(), clientPod, "xfs_repair", "-o", "force_geometry", "-f", "-n", hostDevicePath)
+				o.Expect(err).To(o.HaveOccurred())
```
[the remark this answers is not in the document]
rzetelskik:
```
xfs_repair -n (no modify mode) will return a status of 1 if filesystem corruption was detected and 0 if no filesystem corruption was detected.
```
Doesn't seem like much we can do here, Bear in mind this is just a sanity check, if this assertion passes for an unrelated reason it doesn't make the test pass.

PR #2186, opened 2024-11-06
Add Conditions and ClusterDomain fields into ScyllaOperatorConfig 

zimnx:
Cluster domain can be used as a Kubernetes cluster indentifier.
Conditions may carry information regarding reconcilation of ClusterDomain automation.


pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -48,6 +55,9 @@ type ScyllaOperatorConfigStatus struct {
 	// prometheusVersion is the Prometheus version used by the operator to create a Prometheus instance.
 	// +optional
 	PrometheusVersion *string `json:"prometheusVersion"`
+
+	// clusterDomain is the Kubernetes cluster domain used by the operator as a cluster identifier.
+	ClusterDomain *string `json:"clusterDomain,omitempty"`
```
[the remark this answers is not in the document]
zimnx: added

zimnx:
@rzetelskik @tnozicka ptal, it's awaiting your feedback

pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -25,6 +25,10 @@ type ScyllaOperatorConfigSpec struct {
 	// Setting this field renders your cluster unsupported. Use at your own risk.
 	// +optional
 	UnsupportedPrometheusVersionOverride *string `json:"unsupportedPrometheusVersionOverride,omitempty"`
+
+	// clusterDomainOverride allows to adjust Kubernetes cluster domain used by the operator as a cluster identifier.
+	// +optional
+	ClusterDomainOverride *string `json:"clusterDomainOverride,omitempty"`
```
tnozicka, line 31:
```suggestion
	ClusterDomain *string `json:"clusterDomain,omitempty"`
```
this should be valid to set, and if set it should disable the autodetection, so it's not actually an override

pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -25,6 +25,10 @@ type ScyllaOperatorConfigSpec struct {
 	// Setting this field renders your cluster unsupported. Use at your own risk.
 	// +optional
 	UnsupportedPrometheusVersionOverride *string `json:"unsupportedPrometheusVersionOverride,omitempty"`
+
+	// clusterDomainOverride allows to adjust Kubernetes cluster domain used by the operator as a cluster identifier.
```
tnozicka, line 29: automatically detected, if not set

pkg/controller/scyllaoperatorconfig/controller.go
```diff
@@ -185,6 +193,28 @@ func (opc *Controller) Run(ctx context.Context, workers int) {
 		}()
 	}
 
+	opc.wg.Add(1)
+	go func() {
+		defer opc.wg.Done()
+		wait.UntilWithContext(ctx, func(ctx context.Context) {
+			klog.V(4).InfoS("Periodically enqueuing %q ScyllaOperatorConfig", naming.SingletonName)
+
+			soc, err := opc.scyllaOperatorConfigLister.Get(naming.SingletonName)
```
tnozicka, line 202: this may technically be not found but I don't think you need to list at all
````golang
	key, err := h.keyFunc(metav1.ObjectMeta{
		Namespace: "",
		Name:      naming.SingletonName,
	})
```

tnozicka (commented):
as this shares the queue with ScyllaOperatorConfig controller, it can do the DNS query more often then the 5 min interval but I can live with it for now (separate queue/object with the interval would be nicer)

pkg/api/scylla/v1alpha1/types_operatorconfig.go
zimnx: done

pkg/api/scylla/v1alpha1/types_operatorconfig.go
zimnx: done

pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -25,6 +25,11 @@ type ScyllaOperatorConfigSpec struct {
 	// Setting this field renders your cluster unsupported. Use at your own risk.
 	// +optional
 	UnsupportedPrometheusVersionOverride *string `json:"unsupportedPrometheusVersionOverride,omitempty"`
+
+	// clusterDomain allows to set Kubernetes cluster domain used by the operator as a cluster identifier.
+	// It's set automatically discovered if not provided.
+	// +optional
+	ClusterDomain *string `json:"clusterDomain,omitempty"`
```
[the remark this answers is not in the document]
tnozicka: I think having something that has a scheme (DNS) is useful. Knowing the cluster domain name may be helpful to communicate with service across cluster, if that becomes needed at some point (like shared manager or other shared services, multiDC, ...).

I can see that this may seem a bit misleading to some, although I'd not expect `ScyllaOperatorConfig.spec.clusterDomain` to configure the kubernetes DNS service itself.

a)
```suggestion
	// clusterDomain allows users to set the configured Kubernetes cluster domain explicitly, instead of letting the scylla operator to automatically discover it.
	// +optional
	ClusterDomain *string `json:"clusterDomain,omitempty"`
```

b)
```suggestion
	// configuredClusterDomain allows users to set the configured Kubernetes cluster domain explicitly, instead of letting the scylla operator to automatically discover it.
	// +optional
	ConfiguredClusterDomain *string `json:"configuredClusterDomain,omitempty"`
```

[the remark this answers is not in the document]
rzetelskik: > Knowing the cluster domain name may be helpful to communicate with services across clusters, if that becomes needed at some point

Why would the cluster domain be useful then? Doesn't the multicluster dns actually depend on the concept of a cluster identifier?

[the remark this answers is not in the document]
zimnx: The sole purpose of it is to identify from which Cluster resources are originating. Identifier doesn't need to be a domain, these usually has restrictions, but it could be anything human readable. We will use cluster domain as there's nothing better available, but users should be able to set their own string, not necessarily something that passes DNS name requirements.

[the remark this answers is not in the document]
tnozicka: > The sole purpose of it is to identify from which Cluster resources are originating

yes, we don't need the domain specified for today's task, but we already go through the hussle of discovering it. I don't think it's restricting to limit this filed to a DNS label (label value is similarly restricted already) and we get the benefit of having an API field with the cluster domain, if we need it in the future - otherwise we then have to introduce a new field which will effectively deprecate this one as it will be a superset of it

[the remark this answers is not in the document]
tnozicka: > Why would the cluster domain be useful then (or rather why assume it would be more useful than some other kind of identifier)? 

Say you want to talk to a manager / monitoring in a different cluster, or something in those lines - my main point is we already auto discover it, so let's just not degrade it by taking the scheme away.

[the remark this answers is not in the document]
zimnx: I agree, we will need cluster domain eventually, hence there's no point of having two fields for the same thing. Changed name and added validation.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
```diff
@@ -25,6 +25,10 @@ type ScyllaOperatorConfigSpec struct {
 	// Setting this field renders your cluster unsupported. Use at your own risk.
 	// +optional
 	UnsupportedPrometheusVersionOverride *string `json:"unsupportedPrometheusVersionOverride,omitempty"`
+
+	// configuredClusterDomain allows users to set the configured Kubernetes cluster domain explicitly, instead of letting the Scylla Operator to automatically discover it.
```
rzetelskik, line 29:
```suggestion
	// configuredClusterDomain allows users to set the configured Kubernetes cluster domain explicitly, instead of letting the Scylla Operator automatically discover it.
```

pkg/api/scylla/validation/scyllaoperatorconfig.go
```diff
@@ -63,6 +64,15 @@ func ValidateScyllaOperatorConfigSpec(spec *scyllav1alpha1.ScyllaOperatorConfigS
 		allErrs = append(allErrs, ValidateSemanticVersion(*spec.UnsupportedPrometheusVersionOverride, fldPath.Child("unsupportedPrometheusVersionOverride"))...)
 	}
 
+	if spec.ConfiguredClusterDomain != nil {
+		if len(*spec.ConfiguredClusterDomain) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("configuredClusterDomain"), "can't be empty"))
+		}
+		for _, msg := range apimachineryvalidationutils.IsValidLabelValue(*spec.ConfiguredClusterDomain) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("configuredClusterDomain"), *spec.ConfiguredClusterDomain, msg))
+		}
```
tnozicka, line 73: I think it should also check `IsFullyQualifiedDomainName`

pkg/controller/scyllaoperatorconfig/sync_clusterdomain.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func (socc *Controller) syncClusterDomain(
+	ctx context.Context,
+	soc *osscyllav1alpha1.ScyllaOperatorConfig,
+	status *osscyllav1alpha1.ScyllaOperatorConfigStatus,
+) ([]metav1.Condition, error) {
+	var progressingConditions []metav1.Condition
+
+	if soc.Spec.ConfiguredClusterDomain != nil && len(*soc.Spec.ConfiguredClusterDomain) != 0 {
```
tnozicka, line 19: technically (given it's forbidden in validation), if it's not nil we should use it and if it's empty value it should error out (instead of silently autodiscovering it when specified)

pkg/api/scylla/validation/scyllaoperatorconfig.go
zimnx: FullyQualifiedName requires 3 segments between dots which is not what we want. Qualified name regexp is part of label value regexp so it's validated already.

pkg/api/scylla/v1alpha1/types_operatorconfig.go
zimnx: done

pkg/api/scylla/validation/scyllaoperatorconfig.go
zimnx: done

pkg/api/scylla/v1alpha1/types_operatorconfig.go
rzetelskik: it still says `the Scylla Operator` - you shouldn't use articles before proper nouns, but I won't hold merging because of that

rzetelskik (commented):
lgtm
/assign tnozicka

pkg/api/scylla/v1alpha1/types_operatorconfig.go
zimnx: Sorry missed that, fixed.

tnozicka (approved):
thanks for the updates 
/approve
/lgtm

PR #2195, opened 2024-11-25
Unify scheme usage

zimnx:
Webhook was using outdated scheme defined in the same package. The rest of the project uses one defined in pkg/scheme which is maintained and up to date. This changes webhook to use the right scheme to deserialize object.


tnozicka (approved):
/approve
/lgtm

PR #2200, opened 2024-11-26
[v1.15] Pin image tags to v1.15

zimnx:
In preparation for v1.15, manifest image was pinned to 1.15.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ SHELL :=/bin/bash -euEo pipefail -O inherit_errexit
 
 comma :=,
 
-IMAGE_TAG ?= latest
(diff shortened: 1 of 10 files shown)


rzetelskik (approved):
/lgtm

tnozicka:
/hold 
the docs substitutions need to be also pinned
https://github.com/scylladb/scylla-operator/blob/v1.15/docs/source/conf.py#L58

tnozicka:
/hold cancel
(this will be harder with sphinx-multiversion)

tnozicka:
filed #2204 to track the manifest versioning

PR #2201, opened 2024-11-26
Build v1.15 documentation

zimnx:
In preparation for v1.15 release, build docs from v1.15 branch and treat them as unstable.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -81,9 +81,9 @@ redirects_file = "./redirections.yaml"
 TAGS = []
 smv_tag_whitelist = multiversion_regex_builder(TAGS)
 # Whitelist pattern for branches (set to None to ignore all branches)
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -81,9 +81,9 @@
 TAGS = []
 smv_tag_whitelist = multiversion_regex_builder(TAGS)
 # Whitelist pattern for branches (set to None to ignore all branches)
-BRANCHES = ['master', 'v1.12', 'v1.13', 'v1.14']
```
rzetelskik, line 84: should we drop 1.12?

rzetelskik (approved):

docs/source/conf.py
zimnx: not now, I'll drop it once v1.15 is stable.

rzetelskik: ok

Issue #2180, opened 2024-11-05
SCT tests for Operator are failing during K8s cluster creation while installing dynamic local volume provisioner (AWS)


grzywin:
### What happened?

Example argus run: https://argus.scylladb.com/tests/scylla-cluster-tests/9d9291b1-27b4-4cd4-9e3c-2ca69700b2ae
Failing step: 
```
Command: 'kubectl --kubeconfig=/home/ubuntu/sct-results/20241103-010504-970249/functional-master-9d9291b1/.kube/config --cache-dir=/home/ubuntu/sct-results/20241103-010504-970249/.kube/http-cache--functional-master-9d9291b1 --namespace=local-csi-driver rollout status daemonset.apps/local-csi-driver'
Stdout:
Waiting for daemon set spec update to be observed...
Waiting for daemon set "local-csi-driver" rollout to finish: 3 out of 4 new pods have been updated...
Waiting for daemon set "local-csi-driver" rollout to finish: 0 of 4 updated pods are available...
```
In must-gather logs we can see an error in NodeConfig

```
  - lastTransitionTime: "2024-11-03T01:27:27Z"
    message: 'MountControllerNodeip-10-0-0-28.eu-north-1.compute.internalDegraded:
      can''t create mounts: can''t ensure units: can''t get unit statuses: can''t
      list units by names "mnt-persistent\\x2dvolumes.mount": Unknown method ''ListUnitsByNames''
      or interface ''org.freedesktop.systemd1.Manager''.'
    observedGeneration: 1
    reason: MountControllerNodeip-10-0-0-28.eu-north-1.compute.internalDegraded_Error
    status: "True"
    type: Nodeip-10-0-0-28.eu-north-1.compute.internalDegraded
```

### What did you expect to happen?

Cluster created without error in NodeConfig

### How can we reproduce it (as minimally and precisely as possible)?

You can rerun argus job.

### Scylla Operator version

v1.15.0-alpha.0-93-g24fd817-latest

But I have seen that this error start occur since version **v1.15.0-alpha.0-56-g8bdba4f-latest**

### Kubernetes platform name and version

k8s: 1.27
platform: AWS/EKS
manager: 3.3
scylla: 2024.1.12

### Please attach the must-gather archive.

Must-gather archive can be downloaded from here (Logs tab):
https://argus.scylladb.com/tests/scylla-cluster-tests/9d9291b1-27b4-4cd4-9e3c-2ca69700b2ae

[...]

PR #2211, opened 2024-11-27
Use a portable method to get systemd unit statuses

rzetelskik:
**Description of your changes:** Use a more portable `ListUnits` method to get unit statuses in systemd package.
The currently used `'ListUnitsByNames'` is only supported by systemd v230 or higher, which breaks EKS NodeConfig deployments with even the most recent Amazon Linux 2.

Resolves https://github.com/scylladb/scylla-operator/issues/2180

/kind bug
/priority critical-urgent
/cc

### DIFF
--- a/pkg/systemd/daemon.go
+++ b/pkg/systemd/daemon.go
@@ -9,6 +9,8 @@ import (
 	"github.com/coreos/go-systemd/v22/dbus"
 	godbus "github.com/godbus/dbus/v5"
 	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 var ErrNotExist = errors.New("unit does not exist")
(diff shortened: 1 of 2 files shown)


rzetelskik:
@grzywin could you run the failing SCT suite against `quay.io/scylladb-dev/ci:scylla-operator-pr-f780249a25b998d200d95b5ea98406a@sha256:0893926a05729ebf743a6cdf1683f8b3a786865b615afe9b88258851f573c845`?

rzetelskik:
/hold
for SCT suite to pass with this fix

tnozicka (approved):
It's a bit unfortunate to always call the method twice just to find out if you can use the new or old method and it also introduces 2 different code paths that need to be tested and more space for diversion between the platforms.
I'd personally use only the old (compatible) method that can be used everywhere, until it gets deprecated/removed, at which point we can likely just move to the new one everywhere.

/approve
/assign @zimnx

rzetelskik:
@tnozicka My reasoning was that otherwise you end up always issuing the additinal calls for missing units, which can also result in an inconsistent state. But the obvious downside is we never test that path until we run tests in EKS. I can switch to using it unconditionally, I don't feel particularly strong about it.

grzywin:
@rzetelskik AFAIK currently in SCT there is no way to specify scylla-operator branch on which I would run the tests. Instead, we are using helm charts, so until it won't be published there I have no way to test it. Unless I am missing something and there actually is somewhere proper helm chart which I could use?

rzetelskik:
@grzywin you should be able to use the helm chart for current master head and just override the operator image

zimnx (approved):
Thanks!
/lgtm

grzywin:
> @grzywin you should be able to use the helm chart for current master head and just override the operator image

I am checking it.

tnozicka (approved):

rzetelskik:
> @grzywin you should be able to use the helm chart for current master head and just override the operator image

never mind, looks like it's not possible to run SCT with an image with a different name than `scylla-operator` :facepalm: 

https://github.com/scylladb/scylla-cluster-tests/blob/27ac27270270683635e5972da7141672eb11c50c/sdcm/cluster_k8s/__init__.py#L646-L654 

https://github.com/scylladb/scylla-operator/blob/master/helm/scylla-operator/templates/operator.deployment.yaml#L34

well, we'll test it on master

/hold cancel

PR #2221, opened 2024-12-02
Wait for nodetuning in NodeConfig e2e and trigger a manual DaemonSet requeue

### LINKED ISSUE #2216: Tuning takes 9 minutes for a single node on GKE
https://github.com/scylladb/scylla-operator/pull/2212#issuecomment-2507307166

now that we actually wait for ignition - the tuning took over 9 minutes which made the job timeout
https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2212/pull-scylla-operator-master-e2e-gke-serial/1862151638882455552/artifacts/e2e/cluster-0/namespaces/e2e-test-nodeconfig-z4wpb-0-j4m2b/pods/basic-8l7bn-us-east-1-us-east-1a-0/scylladb-ignition.current

tnozicka:
When we block node tuning in NodeConfig e2e, there is still a DaemonSet created that keeps failing because of the Pod ResourceQuota, with an exponential backoff rate limiter. Given the DaemonSet is not wired for ResourceQuota updates, when we unblock it, it won't react sooner then the rate limiter allows retrying the error - at this point the default max delay for DaemonSet controller backoff is `1000*time.Second`.

This PR triggers an artificial update for the NodeConfig's DaemonSet, so the DaemonSet controller gets out of the backoff right away and also ads explicit wait for tuning to be rolled out before waiting for an updated ScyllaCluster to better indicate which component failed.

Resolves #2216

### DIFF
--- a/pkg/controllerhelpers/wait.go
+++ b/pkg/controllerhelpers/wait.go
@@ -11,6 +11,7 @@ import (
 	scyllav1client "github.com/scylladb/scylla-operator/pkg/client/scylla/clientset/versioned/typed/scylla/v1"
 	scyllav1alpha1client "github.com/scylladb/scylla-operator/pkg/client/scylla/clientset/versioned/typed/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/helpers"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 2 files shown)


rzetelskik:
looks like a duplicate of https://github.com/scylladb/scylla-operator/pull/2245 to me

tnozicka:
yeah, forgot I've already extracted this

Issue #2118, opened 2024-09-10
[Flake] ScyllaCluster [It] should support scaling


tnozicka:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2117/pull-scylla-operator-master-e2e-gke-parallel/1833430612615106560#1:test-build-log.txt%3A2423

### Snippet of what failed.

```
   [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc0000ca200>: 
      waiting has been interupted ([false]): timed out waiting for the condition
      {
          msg: "waiting has been interupted ([false]): timed out waiting for the condition",
          err: <wait.errInterrupted>{
              cause: <*errors.errorString | 0xc00094cce0>{
                  s: "timed out waiting for the condition",
              },
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_scaling.go:39 @ 09/10/24 09:42:48.531
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.init.func13.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_scaling.go:39 +0x323 
```

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2200/pull-scylla-operator-v1.15-e2e-gke-parallel-clusterip/1861378310697652224 
/remove-lifecycle rotten

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2236/pull-scylla-operator-v1.15-e2e-gke-parallel/1864335744349245440

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2367.

PR #2225, opened 2024-12-03
Propagate operator loglevel

### LINKED ISSUE #2224: Inherit loglevel setting from the operator in the sidecars
Scylla Operator creates multiple instance of its own binary, like the sidecars where we need to also adjust the loglevel. Without going into fine tuning this, inheriting the loglevel from the operator is gonna let us easily adjust it without bloating the API surface.

tnozicka:
This PR propagates operator loglevel into the sidecars to allow adjusting it there as well.

Resolves #2224

### DIFF
--- a/pkg/cmdutil/helpers.go
+++ b/pkg/cmdutil/helpers.go
@@ -3,14 +3,17 @@
 package cmdutil
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
+	"github.com/scylladb/scylla-operator/pkg/helpers"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"k8s.io/apimachinery/pkg/util/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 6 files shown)


tnozicka:
https://github.com/scylladb/scylla-operator/issues/2118#issuecomment-2514578169
/retest

pkg/cmdutil/helpers.go
```diff
@@ -74,11 +77,25 @@ func InstallKlog(cmd *cobra.Command) {
...
-	if f != nil {
-		return f.Value.String()
+	if f == nil {
+		return 0, errors.New(`can't lookup klog "v" flag`)
 	}
 
-	return ""
+	s := f.Value.String()
+	if len(s) == 0 {
+		return 0, nil
```
rzetelskik, line 88: nit: shouldn't a default have *some* verbosity?

rzetelskik (approved):
one nit, lgtm
/assign zimnx

pkg/cmdutil/helpers.go
tnozicka: the default was always 0, this doesn't change it. we only usually set it to 2 explicitly in the deployments

zimnx (approved):
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/2061#issuecomment-2523127350
/retest

zimnx (approved):
/lgtm

PR #2238, opened 2024-12-04
Explicitly disable developerMode when it's not set in the API object

### LINKED ISSUE #2237: Developer mode is enabled on ScyllaClusters when it's not provided in the API
Our API contains a knob to enable developer mode, which by default, when not provided should be disabled. 
Since #2137 flag to Scylla container image entrypoint is no longer set explicitly, it's only added when developerMode should be explicitly enabled. 
Unfortunately, ScyllaDB container image entrypoint has opposite behavior, and it enables developer mode unless it's explicitly disabled, leading ScyllaClusters to always run in developerMode.

zimnx:
ScyllaDB contianer image enables developerMode if it's not explicitly disabled. It's the opposite logic to what we had in the past, which was changed in #2137 by mistake. This restores previous logic of explicitly disabling developerMode unless it's enabled in the API object.

Resolves #2237


tnozicka (approved):
/lgtm
/assign @rzetelskik

PR #2242, opened 2024-12-05
[v1.15] Explicitly disable developerMode when it's not set in the API object

zimnx:
Backport of https://github.com/scylladb/scylla-operator/pull/2238


rzetelskik (approved):
/lgtm

Issue #825, opened 2021-11-08
Document and test storage resizing on the same nodes


tnozicka:
Given [StatefulSet don't reconcile storage changes](https://github.com/kubernetes/kubernetes/issues/68737) we can't yet easily handle it in the operator seamlessly and we should at least document a manual procedure. Something in the lines of:
- orphan delete the scyllacluster
- orphan delete the statefulsets
- create the scyllacluster with higher storage request
- resize the existing PVC, if the storage class supports it, replace the nodes if it doesn't
- trigger a scyllacuster rollout

mflendrich:
Reopened - the scope is to carry the community PR https://github.com/scylladb/scylla-operator/pull/2243 through completion.

mflendrich:
Will be tracked in https://github.com/scylladb/scylla-operator/issues/2453.

PR #1053, opened 2022-10-03
Document manual procedure of storage resizing 

Choraden:
**Description of your changes:**
At this moment we can't easily handle storage resizing it the operator. This PR documents a manual procedure.

**Which issue is resolved by this Pull Request:**
Resolves #825

PR #2243, opened 2024-12-05
Document manual procedure of storage resizing (2nd take)

gdubicki:

Based on https://github.com/scylladb/scylla-operator/pull/1053 by @Choraden, expanded to include both storage increase and decrease cases.

I have tested the decreasing case on ScyllaDB v6.1.2 and Scylla Operator v1.14.0.

Resolves https://github.com/scylladb/scylla-operator/issues/825

### DIFF
--- a/docs/source/resources/scyllaclusters/storage-resizing.md
+++ b/docs/source/resources/scyllaclusters/storage-resizing.md
@@ -1,0 +1,65 @@
+# Resizing storage in ScyllaCluster
+
+Due to limitations of the underlying StatefulSets, ScyllaClusters don't allow storage changes. The following procedure describes how to adjust the storage manually.
+
+Let's use the following pods and statefulsets in the resize example below. 
+   ``` 
+   kubectl get pods                                                      
+                 
+   NAME                                    READY   STATUS    RESTARTS   AGE
+   simple-cluster-us-east-1-us-east-1a-0   2/2     Running   0          12m
+   simple-cluster-us-east-1-us-east-1a-1   2/2     Running   0          11m
+   simple-cluster-us-east-1-us-east-1a-2   2/2     Running   0          9m27s
+    
+   kubectl get statefulset
+   
+   NAME                                  READY   AGE
+   simple-cluster-us-east-1-us-east-1a   3/3     12m
+   ``` 
+
+1. Orphan delete Scylla Cluster
+    ```
+    kubectl delete scyllacluster/simple-cluster --cascade='orphan'
+   
+    scyllacluster.scylla.scylladb.com "simple-cluster" deleted
+    ```
+2. Orphan delete Statefulsets
+    ```
+    kubectl delete statefulset --selector scylla/cluster=simple-cluster --cascade='orphan'
+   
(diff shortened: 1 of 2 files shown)


gdubicki:
Sorry, it's not ready yet. I need to combine it with the new index.

docs/source/resources/scyllaclusters/storage-resizing.md
```diff
@@ -0,0 +1,99 @@
...
+   kubectl get statefulset
+   
+   NAME                                  READY   AGE
+   simple-cluster-us-east-1-us-east-1a   3/3     12m
+   ``` 
+
+
+## Increasing the storage size
+
+1. Orphan delete ScyllaCluster
```
zimnx, line 23: I think it's important to mention that existing scyllacluster spec needs to be saved somewhere and aligned to new storage requirements. Otherwise someone executing very first step of a procedure may loose the spec and find out later it's no longer available.

docs/source/resources/scyllaclusters/storage-resizing.md
```diff
@@ -0,0 +1,99 @@
+# Resizing storage in ScyllaCluster
+
+Due to limitations of the underlying StatefulSet, ScyllaClusters don't allow storage changes. The following procedure describes how to adjust the storage manually.
+
+Let's use the following pods and StatefulSet in the resize examples below.
```
zimnx, line 5: It should be generic enough to explain it needs to be done for every rack (StatefulSet) that requires changes.

docs/source/resources/scyllaclusters/storage-resizing.md
```diff
@@ -0,0 +1,99 @@
+# Resizing storage in ScyllaCluster
+
+Due to limitations of the underlying StatefulSet, ScyllaClusters don't allow storage changes. The following procedure describes how to adjust the storage manually.
```
zimnx, line 3: One of the requirements is that StorageClass used supports resising. It should be mentioned here and explained how to verify that.

docs/source/resources/scyllaclusters/storage-resizing.md
```diff
@@ -0,0 +1,99 @@
...
+   ```
+   kubectl apply -f clusterDefinition.yaml --server-side=true
+   ```
+   or, you use Helm to manage your ScyllaCluster, edit your values YAML with the cluster definition and its `racks[0].storage.capacity` field and apply the change with:
+   ```
+   helm upgrade simple-cluster scylla/scylla --values values.cluster.yaml
+   ```
+   The StatefulSet will be recreated by the Scylla Operator with the new storage size.
+
+5. (Only if PVC does not support volume expansion) You need to replace every Scylla node, one by one, following the procedure described in [replacing nodes](/resources/scyllaclusters/nodeoperations/replace-node). Start with the last node in the StatefulSet and move towards the first node. This will recreate each node and their PVCs with the new size.
```
zimnx, line 70: There should be separate procedure for those StorageClasses, not embedded in this one.

zimnx (commented):
Thanks for contirbution, I think it requires slightly more work to explain procedure better.

PR #2245, opened 2024-12-06
Speed up NodeConfig optimization test

tnozicka:
This PR generates an explicit event for the DaemonSet controller which is limited in the tuning test because a number of failures on the resourcequota we use to temporarily block it. This will speed up the response time and avoid waiting in the rate limited queue (up to 1000s with the defaults) which could lead into a startup probe failing and killing the container.


zimnx (approved):
/lgtm

PR #2246, opened 2024-12-06
Fix sidecar annotation controller rate limits for better response times

tnozicka:
The sidecar controller is syncing a single key for the ScyllaDB node where it's running. By design its retries will marginally differ from what you need for a shared queue, yet that's what we had. This means that in cases where it got failing, it could be rate limited up to 1000s at which point it is not responsive and may cause issues with probes as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,6 +49,7 @@ require (
 	go.uber.org/config v1.4.0
 	golang.org/x/oauth2 v0.24.0
 	golang.org/x/sys v0.27.0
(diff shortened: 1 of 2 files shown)


rzetelskik (approved):
lgtm
/assign zimnx

zimnx (approved):
/lgtm

tnozicka:
https://github.com/scylladb/scylla-operator/issues/2096#issuecomment-2527637153
/retest

PR #2267, opened 2024-12-13
[WIP] Fix topology constraints in orphaned PV replacement e2e test

rzetelskik:
**Description of your changes:** wip

Resolves #

/kind failing-test
/priority important-soon
/cc


test/e2e/set/scyllacluster/scyllacluster_pv.go
```diff
@@ -82,17 +83,18 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", func() {
...
-					{
-						LabelSelector: &metav1.LabelSelector{
-							MatchLabels: map[string]string{
-								cloneLabelKey: f.Namespace(),
-							},
+		if sc.Spec.Datacenter.Racks[0].Placement == nil {
+			sc.Spec.Datacenter.Racks[0].Placement = &scyllav1.PlacementSpec{}
+		}
+		sc.Spec.Datacenter.Racks[0].Placement.PodAffinity = &corev1.PodAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
```
tnozicka, line 90: wouldn't it be more plugable to append here?

test/e2e/set/scyllacluster/scyllacluster_pv.go
```diff
@@ -178,6 +180,10 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", func() {
 							},
 							TerminationGracePeriodSeconds: pointer.Ptr(int64(1)),
 							RestartPolicy:                 corev1.RestartPolicyNever,
+							Tolerations:                   slices.Clone(sc.Spec.Datacenter.Racks[0].Placement.Tolerations),
+							NodeSelector: map[string]string{
+								"scylla.scylladb.com/node-type": "scylla",
+							},
```
tnozicka, line 186: why wasn't this there before?

also if you want the same node, then set nodeName and avoid the scheduler

rzetelskik: > why wasn't this there before?

Because we were running everything in the same pool due to the misconfigured taints. Now this workload needs to run in scylla-dedicated pool, which is normally dictated by our workloads using `scylladb-local-xfs` StorageClass. This one does not have it, so it needs to be set explicitly. Otherwise it runs in the generic worker pool, and the PodAffinity on ScyllaCluster also causes it to run there, which in turn causes it to error due to the worker nodes not having `aio-max-nr` set up.

test/e2e/set/scyllacluster/scyllacluster_pv.go
rzetelskik: ok

test/e2e/set/scyllacluster/scyllacluster_pv.go
tnozicka: ok, I thought this was doing something slightly different

tnozicka: shouldn't this technically be on the scyllacluster and just copied here?

rzetelskik: we could set the nodeAffinity on scylla clusters, yes, since we already set the nodeSelector on NodeConfigs either way. But in that case, given we just made the effort to unify taints/tolerations, shouldn't we set it in all our examples/manifests/fixtures?

rzetelskik: Or do you mean just in this particular test?

rzetelskik: (fixed for this test's scope)

tnozicka: > But in that case, given we just made the effort to unify taints/tolerations, shouldn't we set it in all our examples/manifests/fixtures?

yep, everywhere (the fixture). These ScyllaClusters actually do target this node selector. (It "accidentally" works without saying it because of the storage affinity.)

rzetelskik: Ok, I'll send a PR setting node affinity on all ScyllaClusters and remove it from this test

rzetelskik: I made it a part of this PR since otherwise we wouldn't be able to merge either without overriding CI.

tnozicka (approved):
/approve
(lgtm)

/assign @zimnx

test/e2e/set/scyllacluster/scyllacluster_pv.go
tnozicka: maybe https://github.com/scylladb/scylla-operator/pull/2275 would unblock it but it's fine with me

tnozicka:
/lgtm
(this blocks the queue)

PR #2275, opened 2024-12-17
Adjust key buffer sizes in CI based on parallelism

### LINKED ISSUE #2274: CI is running parallel tests with undersized key buffers
Closing as I am stepping down from the project. Feel free to open the issue on your own, if you are interested.

tnozicka:
This PR adjust the key buffer sizes based on the parallelism, so the cache has a real chance to be effectively used. This should help with many flake while keeping the cpu allocation. We can adjust the cpu allocation separately - this just makes a better use of what we have.

Resolves #2274

### DIFF
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -35,10 +35,19 @@ for f in $( find "${DEPLOY_DIR}"/ -type f -name '*.yaml' ); do
 done
 
 yq e --inplace '.spec.template.spec.containers[0].args += ["--qps=200", "--burst=400"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
-yq e --inplace '.spec.template.spec.containers[0].args += ["--crypto-key-buffer-size-min=3", "--crypto-key-buffer-size-max=6", "--crypto-key-buffer-delay=2s"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
+
+CRYPTO_KEY_BUFFER_SIZE_MIN=6
+export CRYPTO_KEY_BUFFER_SIZE_MIN
+CRYPTO_KEY_BUFFER_SIZE_MAX=10
+export CRYPTO_KEY_BUFFER_SIZE_MAX
+if [[ -n "${SO_E2E_PARALLELISM-}" ]]; then
+  CRYPTO_KEY_BUFFER_SIZE_MIN=$(( "${CRYPTO_KEY_BUFFER_SIZE_MIN}" * "${SO_E2E_PARALLELISM}" ))
+  CRYPTO_KEY_BUFFER_SIZE_MAX=$(( "${CRYPTO_KEY_BUFFER_SIZE_MAX}" * "${SO_E2E_PARALLELISM}" ))
+fi
+yq e --inplace '.spec.template.spec.containers[0].args += ["--crypto-key-buffer-size-min="+env(CRYPTO_KEY_BUFFER_SIZE_MIN), "--crypto-key-buffer-size-max="+env(CRYPTO_KEY_BUFFER_SIZE_MAX), "--crypto-key-buffer-delay=2s"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
 
 if [[ -n ${SCYLLA_OPERATOR_FEATURE_GATES+x} ]]; then
-    yq e --inplace '.spec.template.spec.containers[0].args += "--feature-gates="+ strenv(SCYLLA_OPERATOR_FEATURE_GATES)' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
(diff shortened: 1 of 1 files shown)


hack/ci-deploy.sh
```diff
@@ -35,10 +35,19 @@ for f in $( find "${DEPLOY_DIR}"/ -type f -name '*.yaml' ); do
...
+export CRYPTO_KEY_BUFFER_SIZE_MAX
+if [[ -n "${SO_E2E_PARALLELISM-}" ]]; then
+  CRYPTO_KEY_BUFFER_SIZE_MIN=$(( "${CRYPTO_KEY_BUFFER_SIZE_MIN}" * "${SO_E2E_PARALLELISM}" ))
+  CRYPTO_KEY_BUFFER_SIZE_MAX=$(( "${CRYPTO_KEY_BUFFER_SIZE_MAX}" * "${SO_E2E_PARALLELISM}" ))
+fi
+yq e --inplace '.spec.template.spec.containers[0].args += ["--crypto-key-buffer-size-min="+env(CRYPTO_KEY_BUFFER_SIZE_MIN), "--crypto-key-buffer-size-max="+env(CRYPTO_KEY_BUFFER_SIZE_MAX), "--crypto-key-buffer-delay=2s"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
 
 if [[ -n ${SCYLLA_OPERATOR_FEATURE_GATES+x} ]]; then
-    yq e --inplace '.spec.template.spec.containers[0].args += "--feature-gates="+ strenv(SCYLLA_OPERATOR_FEATURE_GATES)' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
+    yq e --inplace '.spec.template.spec.containers[0].args += "--feature-gates="+ env(SCYLLA_OPERATOR_FEATURE_GATES)' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
```
rzetelskik, line 50: why the change here?

hack/ci-deploy.sh
```diff
@@ -35,10 +35,19 @@ for f in $( find "${DEPLOY_DIR}"/ -type f -name '*.yaml' ); do
...
-yq e --inplace '.spec.template.spec.containers[0].args += ["--crypto-key-buffer-size-min=3", "--crypto-key-buffer-size-max=6", "--crypto-key-buffer-delay=2s"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
+
+CRYPTO_KEY_BUFFER_SIZE_MIN=6
+export CRYPTO_KEY_BUFFER_SIZE_MIN
+CRYPTO_KEY_BUFFER_SIZE_MAX=10
+export CRYPTO_KEY_BUFFER_SIZE_MAX
+if [[ -n "${SO_E2E_PARALLELISM-}" ]]; then
+  CRYPTO_KEY_BUFFER_SIZE_MIN=$(( "${CRYPTO_KEY_BUFFER_SIZE_MIN}" * "${SO_E2E_PARALLELISM}" ))
+  CRYPTO_KEY_BUFFER_SIZE_MAX=$(( "${CRYPTO_KEY_BUFFER_SIZE_MAX}" * "${SO_E2E_PARALLELISM}" ))
+fi
```
rzetelskik, line 46: this also needs to be done in ci-deploy-release script, especially that we only saw the tests fail in periodics

rzetelskik (commented):
To make this effective, shouldn't we give the operator some warmup time? Otherwise it seems like it may actually make things worse?
So far we only saw the test with a short timeout fail in periodics which actually have bigger buffers (10/30), so empirically it looks like the smaller buffers could have actually helped get the controllers more cpu time not allocated to generating certs?

hack/ci-deploy.sh
tnozicka: strenv doesn't fail if the env var is missing

hack/ci-deploy.sh
tnozicka: > that we only saw the tests fail in periodics

this fails on presubmits as well, see the linked issue

tnozicka:
> To make this effective, shouldn't we give the operator some warmup time?

This is orthogonal, even without it there is still time where we wait for cluster rollout, generate 3 certs and wait to have to little of them later.

> Otherwise it seems like it may actually make things worse?

not for the certs to my knowledge, it just makes the load constant but I don't think there are other compute extensive task to compete with

> So far we only saw the test with a short timeout fail in periodics which actually have bigger buffers (10/30)

nope, see the "resolves issue" which is a presubmit https://github.com/scylladb/scylla-operator/issues/2274#issue-2744202350
failing with buffer size 3
(and there is a lot of filed flakes that have not been identified yet, and this may affect all of them)

hack/ci-deploy.sh
rzetelskik: ok, haven't seen that before (still needs to be wired in the other script)

tnozicka:
but this might make more mess without the limits / quaranteed qos, let's wait for that

PR #2279, opened 2024-12-17
Fix 'OS Metrics' dashboards in ScyllaDBMonitoring

### LINKED ISSUE #2278: `OS Metrics` dashboard from ScyllaDBMonitoring is empty
When ScyllaDB Monitoring is deployed through v1alpha1.ScyllaDBMonitoring, `OS Metrics` dashboard panels are empty.

Image

zimnx:
`OS Metrics` dashboard panels require `job="node_exporter.*"` label for metrics coming from node-exporter. Our endpoint for node-exporter had `job="scylla"` which didn't match and as a result dashboard was completely empty. Additional relabelling makes sure that job label has correct value.

After this change, `OS Metrics` dashboards starts showing the data.
Selection_103

Resolves #2278

### DIFF
--- a/assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml
+++ b/assets/monitoring/prometheus/v1/scylladb.servicemonitor.yaml
@@ -30,6 +30,11 @@ spec:
       regex:  '(.+)'
       targetLabel: dc
       replacement: '${1}'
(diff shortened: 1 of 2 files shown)


tnozicka (approved):
I think we generally shouldn't access k8s node data from namespaced / non-admin CRs as it break privilege separation and can likely be a CVE, but this one just plumbs in the already exposed data to a dashboard so the surface stays the same.

/approve
(lgtm)

/assign @rzetelskik

PR #2283, opened 2024-12-19
Always terminate leftover jobs in scripts

tnozicka:
**Description of your changes:**
When a script is terminated we should make sure there are no leftover jobs running on the background. This is especially important in CI where such processes will block container exit and the job will timeout, without logs.

**Which issue is resolved by this Pull Request:**
Resolves #2282

PR #2284, opened 2024-12-19
Extract CI setup into a dedicated script for testing release deployment 

tnozicka:
This extract the setup code to test the ci-deploy-release.sh script from our release repo. It will align it with the other scripts and make sure the changes are tested before hand. It will also wire the termination cleanup from our libs.

### Requires
- [x] #2283
- [x] https://github.com/scylladb/scylla-operator-release/pull/366

### DIFF
--- a/hack/.ci/deploy-release-gke.sh
+++ b/hack/.ci/deploy-release-gke.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2024 ScyllaDB
+#
+
+set -euExo pipefail
+shopt -s inherit_errexit
+shopt -s extglob
+
+if [ -z "${ARTIFACTS+x}" ]; then
+  echo "ARTIFACTS can't be empty" > /dev/stderr
+  exit 2
+fi
+
+source "$( dirname "${BASH_SOURCE[0]}" )/lib/e2e.sh"
+parent_dir="$( dirname "${BASH_SOURCE[0]}" )"
+
+trap gather-artifacts-on-exit EXIT
+
+SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=${parent_dir}/manifests/cluster/nodeconfig.yaml}"
+export SO_NODECONFIG_PATH
+
+# We need to strip the tag because skopeo doesn't support references with both a tag and a digest.
+SO_IMAGE="${SO_IMAGE/:*([^:\/])@/@}"
+echo "${SO_IMAGE}"
+ARTIFACTS_DEPLOY_DIR="${ARTIFACTS}/deploy" timeout --foreground -v 10m "${parent_dir}/../ci-deploy-release.sh" "${SO_IMAGE}"


tnozicka:
https://github.com/scylladb/scylla-operator/issues/2285#issuecomment-2554302276
/retest

tnozicka:
/override ci/prow/e2e-gke-release-script-latest/override
(borked until all the fixes land)

hack/.ci/deploy-release-gke.sh
```diff
@@ -0,0 +1,29 @@
...
+source "$( dirname "${BASH_SOURCE[0]}" )/lib/e2e.sh"
+parent_dir="$( dirname "${BASH_SOURCE[0]}" )"
+
+trap gather-artifacts-on-exit EXIT
+
+SO_NODECONFIG_PATH="${SO_NODECONFIG_PATH=${parent_dir}/manifests/cluster/nodeconfig.yaml}"
+export SO_NODECONFIG_PATH
+
+SCYLLA_OPERATOR_FEATURE_GATES="${SCYLLA_OPERATOR_FEATURE_GATES:-AllAlpha=true,AllBeta=true}"
+export SCYLLA_OPERATOR_FEATURE_GATES
```
rzetelskik, line 24: Contrary to `/hack/ci-deploy.sh`, it's not being used in `/hack/ci-deploy-release.sh` script. And it's also already being set in `/hack/.ci/run-e2e-gke-release.sh`. It also doesn't feel platform-dependent?

rzetelskik (commented):
I think this should be wired in `/hack/.ci/run-e2e-gke-release.sh`, which is using `/hack/ci-deploy-release.sh` directly now.
Also shouldn't we first switch the script in CI to make sure this actually fixes the issue?

hack/.ci/deploy-release-gke.sh
tnozicka: removed

tnozicka:
> Also shouldn't we first switch the script in CI to make sure this actually fixes the issue?

We should, thx. https://github.com/scylladb/scylla-operator-release/pull/366

hack/.ci/deploy-release-gke.sh
rzetelskik: When this script gets wired into `/hack/.ci/run-e2e-gke-release.sh`, wouldn't it make sense to move all of [these env vars](https://github.com/scylladb/scylla-operator/blob/master/hack/.ci/run-e2e-gke-release.sh#L20-L27)  into this script and actually wire `SCYLLA_OPERATOR_FEATURE_GATES` to be used in `/hack/ci-deploy-release.sh`? Doesn't have to be in this PR but it seems like it would unify the usage better

tnozicka:
/retest
(ci config was updated)

hack/.ci/deploy-release-gke.sh
tnozicka: not against it eventually, but I can also see how the machines/nodes, or disks can differ

rzetelskik: > not against it eventually, but I can also see how the machines/nodes, or disks can differ

both of those scripts target the same platform though? Moving it to this script would just cover the setup better with `e2e-gke-release-script-latest`, no?

rzetelskik:
> I think this should be wired in /hack/.ci/run-e2e-gke-release.sh, which is using /hack/ci-deploy-release.sh directly now.

This one still stands, rest lgtm

hack/.ci/deploy-release-gke.sh
tnozicka: say the regular GKE one will have big machines with many cores or fancy/multiple NVME - you don't need to spin such a beast to verify you can run a deploy-release script

rzetelskik: ok, fair

tnozicka:
>> I think this should be wired in /hack/.ci/run-e2e-gke-release.sh, which is using /hack/ci-deploy-release.sh directly now.

>This one still stands, rest lgtm

`/hack/.ci/run-e2e-gke-release.sh` and `hack/.ci/deploy-release-gke.sh` are independent entrypoints for the CI so I don't think they should call each other. The generic script to deploy a release is `ci-deploy-release.sh` and that's used by both.

rzetelskik:
> /hack/.ci/run-e2e-gke-release.sh and hack/.ci/deploy-release-gke.sh are independent entrypoints for the CI so I don't think they should call each other. The generic script to deploy a release is ci-deploy-release.sh and that's used by both.

I understand your point that they don't have to be set up the same way, but aren't you losing coverage that way? You could amend the env variables in `/hack/.ci/run-e2e-gke-release.sh` and `/hack/ci-deploy-release.sh` such that the release jobs start to fail but `e2e-gke-release-script-latest` passes (e.g. remove nodeconfig var from `/hack/.ci/run-e2e-gke-release.sh`)  - doesn't that break the whole point of it?

rzetelskik:
> > /hack/.ci/run-e2e-gke-release.sh and hack/.ci/deploy-release-gke.sh are independent entrypoints for the CI so I don't think they should call each other. The generic script to deploy a release is ci-deploy-release.sh and that's used by both.
> 
> I understand your point that they don't have to be set up the same way, but aren't you losing coverage that way? You could amend the env variables in `/hack/.ci/run-e2e-gke-release.sh` and `/hack/ci-deploy-release.sh` such that the release jobs start to fail but `e2e-gke-release-script-latest` passes (e.g. remove nodeconfig var from `/hack/.ci/run-e2e-gke-release.sh`) - doesn't that break the whole point of it?

discussed internally, the `/hack/.ci/` scripts are meant to be entrypoints for jobs 
/lgtm

tnozicka:
unexported variable, fixed in https://github.com/scylladb/scylla-operator-release/pull/367
/retest

PR #2288, opened 2024-12-19
Use strict decoding in e2e tests' fixture registry

### LINKED ISSUE #2289: Fixture registry used in e2e tests does not use strict mode deserialization
Deserialization of fixtures used in our e2e tests does not use strict mode, which has led to one of the fixtures, namely ScyllaDBDatacenter template, getting out of sync with the scheme.

We should enable strict mode deserialization there.

/kind bug
/assign

rzetelskik:
**Description of your changes:** This PR enables strict mode deserialization in codecs factory used in e2e tests' fixture registry. It also fixes a fixture that got out of sync with the scheme and would start causing the tests to fail otherwise.

Resolves #2289 

/cc

### DIFF
--- a/test/e2e/fixture/scylla/scylladbdatacenter.yaml.tmpl
+++ b/test/e2e/fixture/scylla/scylladbdatacenter.yaml.tmpl
@@ -11,25 +11,9 @@ spec:
   datacenterName: us-east-1
   scyllaDB:
     image: "docker.io/scylladb/scylla:{{ .scyllaDBVersion }}"
-    developerMode: true
-    resources:
-      requests:
-        cpu: 10m
-        memory: 100Mi
-      limits:
-        cpu: 1
-        memory: 1Gi
-    storage:
-      capacity: 1Gi
-      {{- if .storageClassName }}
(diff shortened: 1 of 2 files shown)


test/e2e/fixture/scylla/registry.go
```diff
@@ -7,7 +7,7 @@ import (
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/assets"
-	"github.com/scylladb/scylla-operator/test/e2e/scheme"
+	"github.com/scylladb/scylla-operator/pkg/scheme"
```
tnozicka, line 10: Every binary has it's own scheme - they may differ depending on which things it needs to decode. Some of the test schemes should not be in the production.

rzetelskik: ok, enabled strict decoding in the test scheme instead

rzetelskik:
ping @tnozicka

Issue #2296, opened 2025-01-03
E2E scripts cause unnecessary rolling restart of the operator by patching log level


rzetelskik:
The log level adjustment in the e2e func

https://github.com/scylladb/scylla-operator/blob/cc08f3ed8686f9ae1bb81d24a55241fecf21355d/hack/.ci/lib/e2e.sh#L215-L218

causes an unnecessary rolling restart of the operator. It prevents us from giving operator guaranteed QoS in CI with resources >= half of what's available because PDB blocks the restart and everything gets stuck.

We should move this to CI scripts to avoid the rolling restart.

~Perhaps together with https://github.com/scylladb/scylla-operator/issues/836.~ Edit: until we have managed SM it wouldn't simplify things too much, so this is unrelated.

/kind machinery

PR #2276, opened 2024-12-17
Make Scylla Operator resources configurable in CI scripts

rzetelskik:
**Description of your changes:** This PR makes Scylla Operator resources configurable in CI scripts. It also adds a shared env file, sourced by scripts running e2e tests, which by default configures the operator with a guaranteed QoS class. This is a step towards stabilising our CI.

**Which issue is resolved by this Pull Request:**
Resolves https://github.com/scylladb/scylla-operator-release/issues/363

/kind machinery
/priority important-soon
/cc

Issue #836, opened 2021-11-15
Allow changing loglevel


tnozicka:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
Loglvevl should be changable in the API.
We should expose loglevel in the scyllaoperatorconfig for:
- operator
- operator sidecar
- nodeconfigdaemon
- ...

rzetelskik:
I wanted to move the loglevel adjustment from the e2e func
https://github.com/scylladb/scylla-operator/blob/cc08f3ed8686f9ae1bb81d24a55241fecf21355d/hack/.ci/lib/e2e.sh#L215-L218
to CI scripts, as now the patch is causing an unnecessary restart of the operator, and I bumped into the above TODO comment.
@tnozicka with https://github.com/scylladb/scylla-operator/pull/2225, does this still apply?

PR #2297, opened 2025-01-03
Move overriding Scylla Operator's log level to CI scripts and make it configurable

rzetelskik:
**Description of your changes:** Currently, the Operator's and manager controller's loglevels are patched after they have already been deployed, which causes a rolling restart. It prevents us from giving operator guaranteed QoS in CI with resources >= half of what's available because PDB blocks the restart and everything gets stuck.

This PR moves the adjustment to CI scripts and makes the value configurable through e2e scripts.

Resolves https://github.com/scylladb/scylla-operator/issues/2296

/cc

### DIFF
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -37,6 +37,13 @@ for f in $( find "${DEPLOY_DIR}"/ -type f -name '*.yaml' ); do
     sed -i -E -e "s~docker\.io/scylladb/scylla-operator:[^ @]+$~${OPERATOR_IMAGE_REF}~" "${f}"
 done
 
+# TODO: Replace it with ScyllaOperatorConfig field when available.
+# Ref: https://github.com/scylladb/scylla-operator/issues/2314.
+SO_SCYLLA_OPERATOR_LOGLEVEL="${SO_SCYLLA_OPERATOR_LOGLEVEL:-4}"
+export SO_SCYLLA_OPERATOR_LOGLEVEL
+yq e --inplace '.spec.template.spec.containers[0].args += "--loglevel=" + env(SO_SCYLLA_OPERATOR_LOGLEVEL)' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
+yq e --inplace '.spec.template.spec.containers[0].args += "--loglevel=" + env(SO_SCYLLA_OPERATOR_LOGLEVEL)' "${DEPLOY_DIR}/manager/50_controller_deployment.yaml"
+
 yq e --inplace '.spec.template.spec.containers[0].args += ["--qps=200", "--burst=400"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
 yq e --inplace '.spec.template.spec.containers[0].args += ["--crypto-key-buffer-size-min=3", "--crypto-key-buffer-size-max=6", "--crypto-key-buffer-delay=2s"]' "${DEPLOY_DIR}/operator/50_operator.deployment.yaml"
 
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -211,15 +211,6 @@ function run-e2e {
   kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o=yaml | kubectl_create -f=-
   kubectl create -n=e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o=yaml | kubectl_create -f=-
 
-
-  # Raise loglevel in CI.
-  # TODO: Replace it with ScyllaOperatorConfig field when available.
-  kubectl -n=scylla-operator patch --field-manager="${FIELD_MANAGER}" deployment/scylla-operator --type=json -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--loglevel=4"}]'
-  kubectl -n=scylla-operator rollout status deployment/scylla-operator
-
-  kubectl -n=scylla-manager patch --field-manager="${FIELD_MANAGER}" deployment/scylla-manager-controller --type=json -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--loglevel=4"}]'
-  kubectl -n=scylla-manager rollout status deployment/scylla-manager-controller
-
   kubectl create -n=e2e secret generic kubeconfigs ${KUBECONFIGS[@]/#/--from-file=} --dry-run=client -o=yaml | kubectl_create -f=-
   kubeconfigs_in_container_path=$( IFS=','; basenames=( "${KUBECONFIGS[@]##*/}" ) && in_container_paths="${basenames[@]/#//var/run/secrets/kubeconfigs/}" && echo "${in_container_paths[*]}" )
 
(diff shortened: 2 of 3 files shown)


rzetelskik:
Proof that loglevel propagated in `ci-deploy.sh`:
- operator: https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2297/pull-scylla-operator-master-e2e-gke-parallel/1876986614622720000/artifacts/must-gather/0/namespaces/scylla-operator/pods/scylla-operator-58486446b9-6qtmv/scylla-operator.current
- manager controller: https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2297/pull-scylla-operator-master-e2e-gke-parallel/1876986614622720000/artifacts/must-gather/0/namespaces/scylla-manager/pods/scylla-manager-controller-55bc7bcccd-m22pc/scylla-manager-controller.current

hack/.ci/run-e2e-gke-release.sh
```diff
@@ -23,6 +23,9 @@ export SO_NODECONFIG_PATH
 SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME=scylladb-local-xfs}"
 export SO_SCYLLACLUSTER_STORAGECLASS_NAME
 
+SO_SCYLLA_OPERATOR_LOGLEVEL="${SO_SCYLLA_OPERATOR_LOGLEVEL:-4}"
+export SO_SCYLLA_OPERATOR_LOGLEVEL
+
```
mflendrich, line 28: What do you think about creating a `_run-e2e-common.inc.sh` file with some of the bash boilerplate? (to get rid of the "repeated constants" code smell)

that would then be included using something like 
```sh
source "$( dirname "${BASH_SOURCE[0]}" )/_run-e2e-common.inc.sh"
```

hack/ci-deploy.sh
```diff
@@ -37,6 +37,12 @@ for f in $( find "${DEPLOY_DIR}"/ -type f -name '*.yaml' ); do
     sed -i -E -e "s~docker\.io/scylladb/scylla-operator:[^ @]+$~${OPERATOR_IMAGE_REF}~" "${f}"
 done
 
+# TODO: Replace it with ScyllaOperatorConfig field when available.
```
mflendrich, line 40: Maybe we could make this a GitHub issue?

mflendrich (commented):
Assuming that the in place editing pattern for the operator & manager kustomizations is already present in these scripts (as opposed to creating a dependent kustomization), I agree that following the established way of doing things is the right decision (whilst a rewrite to a declarative approach is strongly preferable for maintainability reasons).

rzetelskik:
> whilst a rewrite to a declarative approach is strongly preferable for maintainability reasons

I have a couple of similar PRs in queue, so maybe it's worth considering. I'm not sure how would you see this done though?

mflendrich:
> I'm not sure how would you see this done though?

I have started looking and it was more complex than I managed to produce a complete recipe for in a timeboxed effort. But basically the idea is that:

- the kustomization file is either static or generated from a single "flat" template
- the patches are static (and - where appropriate - replaced with kustomize transformers like [images](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/images/), [replacements](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) etc.
- what varies between executions is: decisions whether to apply specific patches or not and what values to fill in - but keeping the structure of the kustomization fixed.

I am more than happy to try explore this tomorrow, or pair program. Kustomize has one important limitation that it doesn't like accepting external context that we'd need to work around somehow.

hack/ci-deploy.sh
rzetelskik: We already had https://github.com/scylladb/scylla-operator/issues/836, but it was closed on Tomas' departure.
I created https://github.com/scylladb/scylla-operator/issues/2314, I'll add a ref in a comment.

hack/.ci/run-e2e-gke-release.sh
rzetelskik: The idea behind separate scripts was to make the config platform-dependent and allow for configuring it in the CI jobs. The values here are meant as platform-dependent defaults, but as you can see, most of them repeat across different platforms. 
This particular one could be embedded in `ci-deploy.sh` and `ci-deploy-release.sh` scripts, but it won't hold for some other examples (e.g. resources, as in https://github.com/scylladb/scylla-operator/pull/2276, because it will prevent us from using it locally and/or on less powerful setups).
I think having the shared sourced script could clean things up a bit, but I'm wondering if it's not going to obfuscate where the values are coming from in the end (given they could then come from the CI job, the shared script, or the e2e script)?
I'll make it your call. :wink:

rzetelskik:
> > I'm not sure how would you see this done though?
> 
> I have started looking and it was more complex than I managed to produce a complete recipe for in a timeboxed effort. But basically the idea is that:
> 
> * the kustomization file is either static or generated from a single "flat" template
> * the patches are static (and - where appropriate - replaced with kustomize transformers like [images](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/images/), [replacements](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) etc.
> * what varies between executions is: decisions whether to apply specific patches or not and what values to fill in - but keeping the structure of the kustomization fixed.

I though about it before but I figured it wouldn't make that much of a difference since the patches would still have to be conditionally modified and concatenated to the kustomize file. It might make things more understandable, but I don't think it will change the flow in these scripts too much, so it seems mostly orthogonal to this PR.
How about we take this as a followup to these and other PRs in queue?

hack/.ci/run-e2e-gke-release.sh
mflendrich: The option of embedding the default log level in `ci-deploy.sh` and `ci-deploy-release.sh` directly (and inlining the loglevel patch in the kustomization yaml) looks most promising to me (because this config is unlikely to be environment-dependent and adding excessive logic here doesn't add clear benefit)

Let's do it (that is: drop the conditional, embed the loglevel patch directly in the kustomization) if you think the same way. Structuring the config in a streamlined way would require a fundamental restructuring of the deploy machinery - something we won't do in this PR.

rzetelskik: done

rzetelskik:
@mflendrich I believe I addressed all of your comments - let me know if we're on the same page now

mflendrich (approved):
looks good :+1:

/lgtm
/approve

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[e981aa9](https://github.com/scylladb/scylla-operator/commit/e981aa91a177fe041666a2b1d587ecf26524a70e)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2297/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1877766313292599296)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2297). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

https://github.com/scylladb/scylla-operator/issues/2304#issuecomment-2586512507
known manager flake
/test images
/retest

PR #2309, opened 2025-01-08
Fix disk setup e2e assertion

### LINKED ISSUE #2308: E2E failures - Node Setup should propagate mount controller's degraded condition
Two tests out of ` Node Setup should propagate mount controller's degraded condition` suite are failing on Openshift.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-latest-e2e-openshift-aws-serial/1876955592392708096

zimnx:
Assertion validates if all matching Nodes of test NodeConfig have Degraded condition set, but only on one of them test tries to mess around the disks. In case when there're more than 1 matching Nodes this always fails.

To fix that, test picks up one of the nodes and validates whether expected conditions are set for chosen one.

Resolves #2308

### DIFF
--- a/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
+++ b/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
@@ -248,8 +248,8 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
 
 	type degradedEntry struct {
 		nodeConfigFunc             func() *scyllav1alpha1.NodeConfig
-		preNodeConfigCreationFunc  func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context)
-		postNodeConfigCreationFunc func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context)
+		preNodeConfigCreationFunc  func(ctx context.Context, nc *scyllav1alpha1.NodeConfig, nodeUnderTest *corev1.Node) func(context.Context)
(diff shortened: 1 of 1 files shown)


test/e2e/set/nodeconfig/nodeconfig_disksetup.go
```diff
@@ -428,11 +427,12 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
 
 				return nc
 			},
-			preNodeConfigCreationFunc: func(ctx context.Context, nc *scyllav1alpha1.NodeConfig) func(context.Context) {
+			preNodeConfigCreationFunc: func(ctx context.Context, nc *scyllav1alpha1.NodeConfig, nodeUnderTest *corev1.Node) func(context.Context) {
```
mflendrich, line 430: Did you consider passing just a resource name instead of the whole node resource?

I am not familiar enough with the code to make a judgment call, but passing an entire resource to the deferred function where just a name is needed might lead to (unjustified) implied guarantee of the resource contents being up-to-date.

mflendrich:
LGTM - one nit attached

mflendrich (approved):
/lgtm

zimnx:
Let me validate if test indeed passed on non-required openshift serial before merge
/hold

zimnx:
One of the tests that was supposed to be fixed by this change failed, but on later stage and with different root cause. The other test will be fixed in other PR. Since required jobs are green, i'm merging this and will continue working on fixing entire openshift suite.

/hold cancel

PR #2355, opened 2025-01-29
Support configuring ScyllaDB and Scylla Manager versions for e2e tests with flags

### LINKED ISSUE #2320: Support configuring ScyllaDB and Scylla Manager versions for e2e tests with flags
We need to set up early testing against ScyllaDB and Manager pre-release so we need to have this configurable when running `scylla-operator-tests`. These flags should override the default version defined in the repository.

Ref https://github.com/scylladb/scylla-operator/issues/1937

grzywin:
This PR makes the flags for ScyllaDB and Scylla Manager versions configurable. Hardcoded values (versions) in config.yaml will be overwritten by the ones passed as arguments.

Resolves #2320

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -207,6 +207,13 @@ function run-e2e {
   SO_BUCKET_NAME="${SO_BUCKET_NAME:-}"
   SO_E2E_PARALLELISM="${SO_E2E_PARALLELISM:-0}"
 
+  config_file="$(realpath "$(dirname "${BASH_SOURCE[0]}")/../../../assets/config/config.yaml")"
+  SCYLLADB_VERSION="${SCYLLADB_VERSION:-$(yq '.operator.scyllaDBVersion' "$config_file")}"
+  SCYLLADB_MANAGER_VERSION="${SCYLLADB_MANAGER_VERSION:-$(yq '.operator.scyllaDBManagerVersion' "$config_file")}"
+  SCYLLADB_MANAGER_AGENT_VERSION="${SCYLLADB_MANAGER_AGENT_VERSION:-$(yq '.operator.scyllaDBManagerAgentVersion' "$config_file")}"
+  SCYLLADB_UPDATE_FROM_VERSION="${SCYLLADB_UPDATE_FROM_VERSION:-$(yq '.operatorTests.scyllaDBVersions.updateFrom' "$config_file")}"
+  SCYLLADB_UPGRADE_FROM_VERSION="${SCYLLADB_UPGRADE_FROM_VERSION:-$(yq '.operatorTests.scyllaDBVersions.upgradeFrom' "$config_file")}"
+
   kubectl create namespace e2e --dry-run=client -o=yaml | kubectl_create -f=-
   kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o=yaml | kubectl_create -f=-
   kubectl create -n=e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o=yaml | kubectl_create -f=-
@@ -274,6 +281,11 @@ spec:
     - "--object-storage-bucket=${SO_BUCKET_NAME}"
     - "--gcs-service-account-key-path=${gcs_sa_in_container_path}"
     - "--s3-credentials-file-path=${s3_credentials_in_container_path}"
+    - "--scylladb-version=${SCYLLADB_VERSION}"
+    - "--scylladb-manager-version=${SCYLLADB_MANAGER_VERSION}"
+    - "--scylladb-manager-agent-version=${SCYLLADB_MANAGER_AGENT_VERSION}"
+    - "--scylladb-update-from-version=${SCYLLADB_UPDATE_FROM_VERSION}"
+    - "--scylladb-upgrade-from-version=${SCYLLADB_UPGRADE_FROM_VERSION}"
     image: "${SO_IMAGE}"
     imagePullPolicy: Always
     volumeMounts:
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -113,6 +113,19 @@ if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-}" ]]; then
 elif [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]]; then
   yq e --inplace 'del(.spec.datacenter.racks[0].storage.storageClassName)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
 fi
+
+if [[ -n "${SCYLLADB_VERSION:-}" ]]; then
+  yq e --inplace '.spec.version = env(SCYLLADB_VERSION)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
+fi
+
+if [[ -n "${SCYLLA_MANAGER_VERSION:-}" ]]; then
+  yq e --inplace '.spec.template.spec.containers[0].image |= "docker.io/scylladb/scylla-manager:" + env(SCYLLA_MANAGER_VERSION)' "${DEPLOY_DIR}/manager/50_manager_deployment.yaml"
+fi
+
+if [[ -n "${SCYLLA_MANAGER_AGENT_VERSION:-}" ]]; then
+  yq e --inplace '.spec.agentVersion = env(SCYLLA_MANAGER_AGENT_VERSION)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
+fi
+
 kubectl_create -f "${DEPLOY_DIR}"/manager
 
 kubectl -n=scylla-manager wait --timeout=10m --for='condition=Progressing=False' scyllaclusters.scylla.scylladb.com/scylla-manager-cluster
--- a/pkg/cmd/tests/options.go
+++ b/pkg/cmd/tests/options.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/onsi/ginkgo/v2"
+	configassets "github.com/scylladb/scylla-operator/assets/config"
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	"github.com/scylladb/scylla-operator/pkg/genericclioptions"
 	"github.com/scylladb/scylla-operator/pkg/helpers/slices"
@@ -16,6 +18,16 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/rest"
 )
+
+const (
+	// https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L68
+	referenceTagRegexp = `[\w][\w.-]{0,127}`
+
+	// https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L81
+	digestRegexp = `[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}`
+)
+
+var tagWithOptionalDigestRegexp = regexp.MustCompile("^" + referenceTagRegexp + "(?:@" + digestRegexp + ")?$")
 
 type IngressControllerOptions struct {
 	Address           string
@@ -55,6 +67,11 @@ type TestFrameworkOptions struct {
 	objectStorageType           framework.ObjectStorageType
 	gcsServiceAccountKey        []byte
 	s3CredentialsFile           []byte
+	ScyllaDBVersion             string
+	ScyllaDBManagerVersion      string
+	ScyllaDBManagerAgentVersion string
+	ScyllaDBUpdateFrom          string
+	ScyllaDBUpgradeFrom         string
 }
 
 func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent string) *TestFrameworkOptions {
@@ -69,12 +86,17 @@ func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent stri
 			ClientsBroadcastAddressType: string(scyllav1.BroadcastAddressTypePodIP),
 			StorageClassName:            "",
 		},
-		ObjectStorageBucket:      "",
-		GCSServiceAccountKeyPath: "",
-		S3CredentialsFilePath:    "",
-		objectStorageType:        framework.ObjectStorageTypeNone,
-		gcsServiceAccountKey:     []byte{},
-		s3CredentialsFile:        []byte{},
+		ObjectStorageBucket:         "",
+		GCSServiceAccountKeyPath:    "",
+		S3CredentialsFilePath:       "",
+		objectStorageType:           framework.ObjectStorageTypeNone,
+		gcsServiceAccountKey:        []byte{},
+		s3CredentialsFile:           []byte{},
+		ScyllaDBVersion:             configassets.Project.Operator.ScyllaDBVersion,
+		ScyllaDBManagerVersion:      configassets.Project.Operator.ScyllaDBManagerVersion,
+		ScyllaDBManagerAgentVersion: configassets.Project.Operator.ScyllaDBManagerAgentVersion,
+		ScyllaDBUpdateFrom:          configassets.Project.OperatorTests.ScyllaDBVersions.UpdateFrom,
+		ScyllaDBUpgradeFrom:         configassets.Project.OperatorTests.ScyllaDBVersions.UpgradeFrom,
 	}
 }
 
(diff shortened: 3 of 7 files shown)


pkg/cmd/tests/options.go
```diff
@@ -164,6 +181,28 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
 		errors = append(errors, fmt.Errorf("gcs-service-account-key-path and s3-credentials-file-path can't be set simultanously"))
 	}
 
+	versionPattern := regexp.MustCompile(`^\d+\.\d+\.\d+(-[a-zA-Z0-9]+)?$`)
```
rzetelskik, line 184: This doesn't cover tag + digest entirely (e.g. latest wouldn't be caught by that), see the following for reference:
- https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/reference.go#L6C44-L6C70
- https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L39
- https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L9
I'd import `distribution/reference` to handle that and look there to see how to match and capture both correctly. It might differ between ScyllaDB and manager versions. From the top of my head I'm not sure we handle [tag + ":"] + digest for scylla versions - you'll have to try.

hack/.ci/lib/e2e.sh
```diff
@@ -274,6 +274,11 @@ spec:
     - "--object-storage-bucket=${SO_BUCKET_NAME}"
     - "--gcs-service-account-key-path=${gcs_sa_in_container_path}"
     - "--s3-credentials-file-path=${s3_credentials_in_container_path}"
+    - "--scylla-db-version=${SCYLLA_DB_VERSION}"
+    - "--scylla-manager-version=${SCYLLA_MANAGER_VERSION}"
+    - "--scylla-manager-agent-version=${SCYLLA_MANAGER_AGENT_VERSION}"
+    - "--scylla-update-from=${SCYLLA_UPDATE_FROM}"
+    - "--scylla-upgrade-from=${SCYLLA_UPGRADE_FROM}"
```
rzetelskik, line 281: You have to declare these env vars, otherwise the script will fail on undeclared

hack/ci-deploy.sh
```diff
@@ -113,6 +113,19 @@ if [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME:-}" ]]; then
 elif [[ -n "${SO_SCYLLACLUSTER_STORAGECLASS_NAME+x}" ]]; then
   yq e --inplace 'del(.spec.datacenter.racks[0].storage.storageClassName)' "${DEPLOY_DIR}/manager/50_scyllacluster.yaml"
 fi
+
+if [[ -n "${SCYLLA_DB_VERSION:-}" ]]; then
```
rzetelskik, line 117: SCYLLADB_VERSION

pkg/cmd/tests/options.go
```diff
@@ -118,6 +130,11 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(&o.ObjectStorageBucket, "object-storage-bucket", "", o.ObjectStorageBucket, "Name of the object storage bucket.")
 	cmd.PersistentFlags().StringVarP(&o.GCSServiceAccountKeyPath, "gcs-service-account-key-path", "", o.GCSServiceAccountKeyPath, "Path to a file containing a GCS service account key.")
 	cmd.PersistentFlags().StringVarP(&o.S3CredentialsFilePath, "s3-credentials-file-path", "", o.S3CredentialsFilePath, "Path to the AWS credentials file providing access to the S3 bucket.")
+	cmd.PersistentFlags().StringVarP(&o.ScyllaDBVersion, "scylla-db-version", "", o.ScyllaDBVersion, "Version of ScyllaDB to use in tests.")
```
rzetelskik, line 133: s/scylla-db-version/scylladb-version

pkg/cmd/tests/options.go
```diff
@@ -118,6 +130,11 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
...
+	cmd.PersistentFlags().StringVarP(&o.ScyllaManagerVersion, "scylla-manager-version", "", o.ScyllaManagerVersion, "Version of Scylla Manager to use in tests.")
+	cmd.PersistentFlags().StringVarP(&o.ScyllaManagerAgentVersion, "scylla-manager-agent-version", "", o.ScyllaManagerAgentVersion, "Version of Scylla Manager Agent to use in tests.")
+	cmd.PersistentFlags().StringVarP(&o.ScyllaDBUpdateFrom, "scylla-update-from", "", o.ScyllaDBUpdateFrom, "Version of Scylla Update to use in tests.")
```
rzetelskik, line 136: s/scylla-update-from/scylladb-update-from-version

pkg/cmd/tests/options.go
```diff
@@ -118,6 +130,11 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
...
+	cmd.PersistentFlags().StringVarP(&o.ScyllaDBUpgradeFrom, "scylla-upgrade-from", "", o.ScyllaDBUpgradeFrom, "Version of Scylla Upgrade to use in tests.")
```
rzetelskik, line 137: s/scylla-upgrade-from/scylladb-upgrade-from-version

pkg/cmd/tests/options.go
rzetelskik: also `Version of Scylla Update to use in tests.` doesn't make much sense to me (applies to upgrade as well), how about "Version of ScyllaDB to update from"?

pkg/cmd/tests/options.go
rzetelskik: nit: You don't need to say "in tests" either, it's pretty clear as it's a test runner command (applies to all flags)

pkg/cmd/tests/options.go
rzetelskik: still applies partially, s/scylla/scylladb

pkg/cmd/tests/options.go
```diff
@@ -164,6 +181,33 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
 		errors = append(errors, fmt.Errorf("gcs-service-account-key-path and s3-credentials-file-path can't be set simultanously"))
 	}
 
+	if o.ScyllaDBVersion != "" {
```
rzetelskik, line 184: why only do this conditionally? Empty string is an invalid "version", no?

pkg/cmd/tests/options.go
```diff
@@ -164,6 +181,33 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
...
+		if _, err := reference.ParseAnyReference(o.ScyllaDBVersion); err != nil {
```
rzetelskik, line 185: this is incorrect, ParseAnyReference is going to "accept" (as in not return an error) much more than just `tag ['@' digest]`. That way you'd accept a whole reference `docker.io/scylladb/scylladb:latest`, just a name `docker.io/scylladb/scylladb`, just a remote name `scylladb/scylladb`, just a path component `scylladb` etc. 

We want to validate this matches `tag ['@' digest]`.

Again, see https://github.com/scylladb/scylla-operator/pull/2355#discussion_r1935998932 with references.

pkg/cmd/tests/options.go
```diff
@@ -186,6 +230,22 @@ func (o *TestFrameworkOptions) Complete(args []string) error {
...
+	}
+	if o.ScyllaManagerAgentVersion == "" {
+		o.ScyllaManagerAgentVersion = configassets.Project.Operator.ScyllaDBManagerAgentVersion
+	}
+	if o.ScyllaDBUpdateFrom == "" {
+		o.ScyllaDBUpdateFrom = configassets.Project.OperatorTests.ScyllaDBVersions.UpdateFrom
+	}
+	if o.ScyllaDBUpgradeFrom == "" {
+		o.ScyllaDBUpgradeFrom = configassets.Project.OperatorTests.ScyllaDBVersions.UpgradeFrom
+	}
```
rzetelskik, line 247: you already assign defaults in the constructor, this is redundant

rzetelskik (commented):
@grzywin thanks for the updates.

Please don't resolve comments - it should be done by the reviewer to track what was actually done.

Also the commit name "Update generated" is misleading - you're updating dependencies, not generated files.

hack/.ci/lib/e2e.sh
```diff
@@ -234,6 +234,12 @@ function run-e2e {
   ingress_custom_annotations='haproxy.org/ssl-passthrough=true,route.openshift.io/termination=passthrough'
   ingress_controller_address="$( kubectl -n=haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' ):9142"
 
+  SCYLLADB_VERSION="${SCYLLADB_VERSION:-}"
+  SCYLLA_MANAGER_VERSION="${SCYLLA_MANAGER_VERSION:-}"
+  SCYLLA_MANAGER_AGENT_VERSION="${SCYLLA_MANAGER_AGENT_VERSION:-}"
+  SCYLLA_UPDATE_FROM_VERSION="${SCYLLA_UPDATE_FROM_VERSION:-}"
```
rzetelskik, line 240: s/SCYLLA/SCYLLADB

hack/.ci/lib/e2e.sh
```diff
@@ -234,6 +234,12 @@ function run-e2e {
...
+  SCYLLA_UPGRADE_FROM_VERSION="${SCYLLA_UPGRADE_FROM_VERSION:-}"
```
rzetelskik, line 241: s/SCYLLA/SCYLLADB

pkg/cmd/tests/options.go
grzywin: @rzetelskik Tests on CI are failing without this part of code. I also ran some tests locally and without it, when we pass empty value to parameter e.g. `scylladb-version=` (like it is in CI) the tests are failing.

pkg/cmd/tests/options.go
grzywin: On CI the test are triggered in following form:
`I0131 22:48:32.144278       1 tests/tests_run.go:358] "Started process" ID=1 Command="/usr/bin/scylla-operator-tests run scylla-operator/conformance/multi-datacenter-parallel --kubeconfig=/var/run/secrets/kubeconfigs/europe-west4.kubeconfig,/var/run/secrets/kubeconfigs/europe-west3.kubeconfig,/var/run/secrets/kubeconfigs/europe-west1.kubeconfig --loglevel=2 --color=false --artifacts-dir=/tmp/artifacts --parallelism=0 --feature-gates=AllAlpha=true,AllBeta=true --ingress-controller-address=10.107.206.204:9142 --ingress-controller-ingress-class-name=haproxy --ingress-controller-custom-annotations=haproxy.org/ssl-passthrough=true,route.openshift.io/termination=passthrough --scyllacluster-node-service-type=Headless --scyllacluster-nodes-broadcast-address-type=PodIP --scyllacluster-clients-broadcast-address-type=PodIP --scyllacluster-storageclass-name=scylladb-local-xfs --object-storage-bucket= --gcs-service-account-key-path= --s3-credentials-file-path= --scylladb-version= --scylla-manager-version= --scylla-manager-agent-version= --scylla-update-from-version= --scylla-upgrade-from-version= --parallel-server-address=127.0.0.1:40981 --loglevel=0 --random-seed=1738363712 --parallel-shard=1"`

The parameters have an empty value "--scylladb-version= " so without this "if not empty" condition assertion will fail and we will get `invalid scylladb-version format`

pkg/cmd/tests/options.go
grzywin: I will take a closer look into this.

rzetelskik: I see `distribution/reference` doesn't export the actual expressions for tag and digest but only the compiled structs, so you can either import the compiled vars and get the expression with `.String()`
```
regexp.MustCompile("^" + reference.TagRegexp.String() + "(?:" + reference.DigestRegexp.String() + ")?$")
```

or, simpler, just provide references in comments:

```
const (
  // https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L68.
  referenceTagRegexp = `[\w][\w.-]{0,127}`
  
  ...
)

var tagWithOptionalDigestRegexp = regexp.MustCompile("^" + referenceTagRegexp + "(?:" + referenceDigestRegexp + ")?$")
```

Either works, second one avoids the back and forth so it's maybe a bit better.
Verify the regex string though, I wrote it from the top of my head and I'm not sure it even compiles.

pkg/cmd/tests/options.go
rzetelskik: > On CI the test are triggered in following form:

That's how you set it up in this PR, it didn't exist before :wink: 
Empty string **is** an invalid version - we expect `tag ['@' digest]`. We provide a default for an unset flag, but if it's explicitly set to an empty value, then we shouldn't override it but just fail.

So what we need to do instead is to only add the flags conditionally in the e2e runner or set the env var to the defaults coming from the config file. Either way this should happen in the scripts, not in the test runner.

pkg/cmd/tests/options.go
rzetelskik: See https://github.com/scylladb/scylla-operator/pull/2355#discussion_r1939313680

hack/.ci/lib/e2e.sh
grzywin: fixed

hack/.ci/lib/e2e.sh
grzywin: fixed

pkg/cmd/tests/options.go
grzywin: removed

pkg/cmd/tests/options.go
grzywin: I removed https://github.com/distribution/reference dependency and used regex directly in our code as in option 2 which you showed. I checked multiple formats:
image

pkg/cmd/tests/options.go
grzywin: Now, when you are writing about it sound obvious ;) Fixed (I edited e2e.sh)

pkg/cmd/tests/options.go
```diff
@@ -164,6 +181,31 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
 		errors = append(errors, fmt.Errorf("gcs-service-account-key-path and s3-credentials-file-path can't be set simultanously"))
 	}
 
+	const referenceTagRegexp = `[\w][\w.-]{0,127}`
+	const digestRegexp = `[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}`
+
```
rzetelskik, line 186: this needs a reference to the source (as a comment, I put it in the code snippet earlier for a reason)

rzetelskik: nit: could be moved to a single const block at the top of the file

hack/.ci/lib/e2e.sh
```diff
@@ -234,6 +234,14 @@ function run-e2e {
...
   ingress_controller_address="$( kubectl -n=haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' ):9142"
 
+  CONFIG_FILE="$(realpath "$(dirname "${BASH_SOURCE[0]}")/../../../assets/config/config.yaml")"
+
+  SCYLLADB_VERSION="${SCYLLADB_VERSION:-$(yq '.operator.scyllaDBVersion' "$CONFIG_FILE")}"
+  SCYLLADB_MANAGER_VERSION="${SCYLLADB_MANAGER_VERSION:-$(yq '.operator.scyllaDBManagerVersion' "$CONFIG_FILE")}"
+  SCYLLADB_MANAGER_AGENT_VERSION="${SCYLLADB_MANAGER_AGENT_VERSION:-$(yq '.operator.scyllaDBManagerAgentVersion' "$CONFIG_FILE")}"
+  SCYLLADB_UPDATE_FROM_VERSION="${SCYLLADB_UPDATE_FROM_VERSION:-$(yq '.operatorTests.scyllaDBVersions.updateFrom' "$CONFIG_FILE")}"
+  SCYLLADB_UPGRADE_FROM_VERSION="${SCYLLADB_UPGRADE_FROM_VERSION:-$(yq '.operatorTests.scyllaDBVersions.upgradeFrom' "$CONFIG_FILE")}"
+
```
rzetelskik, line 244: nit: move to top of the file, where `SO_BUCKET_NAME` and other default assignments are

hack/.ci/lib/e2e.sh
```diff
@@ -234,6 +234,14 @@ function run-e2e {
   ingress_custom_annotations='haproxy.org/ssl-passthrough=true,route.openshift.io/termination=passthrough'
   ingress_controller_address="$( kubectl -n=haproxy-ingress get svc haproxy-ingress --template='{{ .spec.clusterIP }}' ):9142"
 
+  CONFIG_FILE="$(realpath "$(dirname "${BASH_SOURCE[0]}")/../../../assets/config/config.yaml")"
```
rzetelskik, line 237: nit: it's a local variable so please use a lowercase name for consistency

rzetelskik (commented):
thanks for the updates, a few more comments, mostly nits

hack/.ci/lib/e2e.sh
grzywin: done

pkg/cmd/tests/options.go
```diff
@@ -164,6 +181,31 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
...
+	var tagWithOptionalDigestRegexp = regexp.MustCompile("^" + referenceTagRegexp + "(?:@" + digestRegexp + ")?$")
```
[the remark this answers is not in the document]
grzywin: done

hack/.ci/lib/e2e.sh
grzywin: done

pkg/cmd/tests/options.go
grzywin: done

pkg/cmd/tests/options.go
```diff
@@ -164,6 +191,41 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
 		errors = append(errors, fmt.Errorf("gcs-service-account-key-path and s3-credentials-file-path can't be set simultanously"))
 	}
 
+	if !tagWithOptionalDigestRegexp.MatchString(o.ScyllaDBVersion) {
+		errors = append(errors, fmt.Errorf(
+			"invalid scylladb-version format: %q. Expected format: <tag>[@<digest>], e.g., '6.2.0' or '6.2.0@sha256:abcdef...'",
```
rzetelskik, line 196: nit: imo `<tag>[@<digest>]` is sufficient, but if you prefer to leave the example it's ok

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
```diff
@@ -5,10 +5,8 @@ package scyllacluster
 import (
 	"context"
 	"fmt"
-
```
rzetelskik, line 8: shouldn't be deleted

pkg/cmd/tests/options.go
```diff
@@ -55,8 +57,23 @@ type TestFrameworkOptions struct {
...
 
+const (
+	// https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L68
+	referenceTagRegexp = `[\w][\w.-]{0,127}`
+
+	// https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go#L81
+	digestRegexp = `[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}`
+)
+
+var tagWithOptionalDigestRegexp = regexp.MustCompile("^" + referenceTagRegexp + "(?:@" + digestRegexp + ")?$")
```
rzetelskik, line 75: nit: conventionally at the top of the file

pkg/cmd/tests/options.go
```diff
@@ -2,8 +2,10 @@ package tests
 
 import (
 	"fmt"
+	configassets "github.com/scylladb/scylla-operator/assets/config"
```
rzetelskik, line 5: shouldn't be grouped with std imports

PR #2484, opened 2025-02-14
Extend Rack spec with rack specific ExposeOptions 

### LINKED ISSUE #2479: Allow Operator in the ScyllaCluster CDR to allow rack-specific annotations
We need a way to define rack-specific annotations in the ScyllaCluster CRD. Currently, we can only apply annotations globally to all services, but we need the flexibility to:

1. Add external DNS annotations specific to each Rack service.
2. Specify NLB subnet mappings per service, instead of mapping all Availability Zones (AZs) to each NLB service.

Ref: https://github.com/scylladb/scylla-enterprise/issues/5215

zimnx:

ScyllaDB cluster deployments using LoadBalancer exposure type can take advantage of specifying Service annotations specific to given rack which helps LoadBalancer controllers to provide topology constraints for underlying cloud primitives.

Resolves #2479

/cc mflendrich


pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -185,6 +185,22 @@ func MemberService(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName, name string
...
 	}
 
+	rackSpec, _, ok := slices.Find(sdc.Spec.Racks, func(rs scyllav1alpha1.RackSpec) bool {
+		return rs.Name == rackName
+	})
+	if !ok {
+		return nil, fmt.Errorf("can't find rack spec having %q name", rackName)
+	}
+
+	if sdc.Spec.RackTemplate != nil {
```
mflendrich, line 195: This `if` looks redundant - the nil check is repeated as the 1st instruction within `applyRackTemplateOnRackSpec`.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -2039,6 +2055,34 @@ func applyRackTemplateOnRackSpec(rackTemplate *scyllav1alpha1.RackTemplate, rack
 					},
 				}
 			}(),
+			ExposeOptions: func() *scyllav1alpha1.RackExposeOptions {
+				if rackTemplate.ExposeOptions == nil && rack.ExposeOptions == nil {
+					return nil
+				}
+
+				rackExposeOptions := &scyllav1alpha1.RackExposeOptions{}
+				for _, reo := range []*scyllav1alpha1.RackExposeOptions{rackTemplate.ExposeOptions, rack.ExposeOptions} {
```
mflendrich, line 2064: The `reo` and `rackExposeOptions` variable names are synonymous, which poses a risk of mixing them up, and makes this harder to read than it needs to be because the reader needs to reinvent those semantics.

I suggest `src` and `dest` or something similar to clearly denote the semantic difference.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -2039,6 +2055,34 @@ func applyRackTemplateOnRackSpec(rackTemplate *scyllav1alpha1.RackTemplate, rack
...
+					if reo == nil {
+						continue
+					}
```
mflendrich, line 2067: This `if` and the below `if` express very similar things but with the logic inverted.

How about:
```go
if reo == nil || reo.NodeService == nil {
    continue
}
```

mflendrich (commented):
Some comments follow :+1:

pkg/controller/scylladbdatacenter/resource.go
zimnx: It's intended, as global ExposeOptions currently have more options than NodeService. Extending the code in case more rack-level expose options would be available, wouldn't require changing `if` you're proposing, and is more explicit about the expected flow. Leaving as is.

pkg/controller/scylladbdatacenter/resource.go
zimnx: done

pkg/controller/scylladbdatacenter/resource.go
zimnx: yep, removed

pkg/controller/scyllacluster/migrate.go
```diff
@@ -241,6 +241,25 @@ func MigrateV1ScyllaClusterSpecToV1Alpha1ScyllaDBDatacenterSpec(scName string, s
...
+							}
+
+							return &scyllav1alpha1.RackExposeOptions{
+								NodeService: func() *scyllav1alpha1.RackNodeServiceTemplate {
+									if rack.ExposeOptions.NodeService == nil {
+										return nil
+									}
+									return &scyllav1alpha1.RackNodeServiceTemplate{
+										ObjectTemplateMetadata: scyllav1alpha1.ObjectTemplateMetadata{
+											Labels:      rack.ExposeOptions.NodeService.Labels,
```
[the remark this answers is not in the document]
zimnx: better safe than sorry, added cloining

mflendrich (approved):
Looking good.

/lgtm

zimnx:
multi-datacenter-parallel CI job requires https://github.com/scylladb/scylla-operator-release/pull/397

PR #2491, opened 2025-02-20
Add support of named and digest scylladb image reference with google cr

### LINKED ISSUE #2455: Support sha references to scylla images
_Issue originally authored by **tnozicka** as https://github.com/scylladb/scylla-operator/issues/821_

Currently scylla images can be specified only with a tag that has to be a semver. This dosn't work for named references like `latest`, rolling update references like `4.4` (sort of) and for sha references. As images can be overwritten sha reference is the only reliable way to always reference the same image.

As we need to know the information in the operator we can't just download and run the image to find out but when buliding the image we could set an image label that the operator would retrieve. The label is stored in the image manifest that can be downloaded separately from a container registry at low cost and the result can be cached.

Needs cooperation from scylla.

(Look at libs like `github.com/containers/image/v5 `)

grzywin:
Add support of named and digest scylladb image reference with google cr. Supported syntax: \<tag\>\<digest\>, \<digest\>, \<tag\> (also tags like "latest")

Resolves #2455


assets/config/config.yaml
```diff
@@ -13,6 +13,6 @@ operator:
   prometheusVersion: "v2.54.1"
 operatorTests:
   scyllaDBVersions:
-    updateFrom: "6.2.0-rc2"
-    upgradeFrom: "6.1.2"
+    updateFrom: "sha256:cf00b8bfcce4d27e8527cad913e83db89af842e6cfb64ff4736bce83594b33d1"
```
zimnx, line 16: It's unclear what it's being tested, use tag+sha format

assets/config/config.yaml
```diff
@@ -1,5 +1,5 @@
 operator:
-  scyllaDBVersion: "6.2.0"
+  scyllaDBVersion: "sha256:5b53a7c60d9f9555bb87791ff29b2e633c6f472aec00de7afaf4db1addc6d594"
```
zimnx, line 2: It's unclear what version it is. Use tag+sha format

Makefile
```diff
@@ -463,7 +468,7 @@ endef
 # $1 - values.yaml
 define update-scylla-helm-versions
 	$(YQ) eval-all -i -P '\
-	select(fi==0).scyllaImage.tag = ( select(fi==1) | .operator.scyllaDBVersion ) | \
+	select(fi==0).scyllaImage.tag = ( select(fi==1) | "$(SCYLLADB_SEM_VER)" ) | \
```
zimnx, line 471: Would you mind elaborating why this is changed?

pkg/semver/version.go
```diff
@@ -47,3 +66,91 @@ func (sv ScyllaVersion) SupportFeatureUnsafe(featureVersion semver.Version) bool
 func (sv ScyllaVersion) SupportFeatureSafe(featureVersion semver.Version) bool {
 	return !sv.unknown && sv.version.GTE(featureVersion)
 }
+
+func GetImageVersionAndDigest(imageName, version string) (string, string, error) {
```
zimnx, line 70: Based on the name it's unclear what is returned from this function.

zimnx: Please also add unit tests for it.

pkg/semver/version.go
```diff
@@ -47,3 +66,91 @@ func (sv ScyllaVersion) SupportFeatureUnsafe(featureVersion semver.Version) bool
...
+	// If the version is already in <tag>@<digest> format, extract the semantic version
+	if strings.Contains(version, "@") {
+		semVersion, err := computeSemVersion(version)
+		if err != nil {
+			return "", "", err
+		}
+		return semVersion, version, nil
+	}
+
+	imageReference := fmt.Sprintf("scylladb/%s", imageName)
```
zimnx, line 80: Lets not hardcode anything. ScyllaCluster spec accepts also repository which might not contain scylladb substring at all.

pkg/semver/version.go
```diff
@@ -47,3 +66,91 @@ func (sv ScyllaVersion) SupportFeatureUnsafe(featureVersion semver.Version) bool
...
+		return "", fmt.Errorf("could not extract semantic version from: %s", fullVersion)
+	}
+
+	reSuffix := regexp.MustCompile(`[~-]([a-zA-Z]+\d*)`)
+	suffixMatch := reSuffix.FindStringSubmatch(fullVersion)
+	if len(suffixMatch) > 1 {
+		semVersion = fmt.Sprintf("%s-%s", semVersion, suffixMatch[1])
+	}
+	return semVersion, nil
+}
```
zimnx, line 156: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
Why not using semver dependancy we already have?

test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -172,7 +171,7 @@ var _ = g.Describe("ScyllaCluster", func() {
 	},
 		g.Entry(describeEntry, &entry{
 			scyllaImageRepository: scyllaOSImageRepository,
-			scyllaVersion:         configassests.Project.Operator.ScyllaDBVersion,
+			scyllaVersion:         framework.TestContext.ScyllaDBVersion,
```
zimnx, line 174: Why this is changed?

zimnx (commented):
first batch

Makefile
grzywin: When we execute `make update` command it is updating the manifests. Without this change the ScyllaDB version is taken directly from config.yaml file which at this moment I changed to digest. So in manifest we will have which won't work. 
spec:
  version: sha256:5b53a7c60d9f9555bb87791ff29b2e633c6f472aec00de7afaf4db1addc6d594

pkg/semver/version.go
grzywin: I will take another look on that but semver dependency was not working for me here as for RCs ScyllaDB version label has a bit unusual format:
`6.2.0~rc2-0.20241002.93700ff5d1ce@sha256:92dfd26d64c6d076c5a641b3526c63fd827c777bc8c0fbce3bfddc6c9913e9b6`
As far as I remember tilde sign is quite unusual/unsupported.

test/e2e/set/scyllacluster/scyllacluster_replace.go
grzywin: With my previous commit https://github.com/scylladb/scylla-operator/commit/b2d7fb172313421375acdde597c48a70f5b5b804 I added flags support.
This line is actually an oversight which should be changed/added in the commit above.

Makefile
zimnx: > Note: Just to clarify. My solution is not changing almost any logic directly in operator. I just added logic to handle digest and convert it to 'supported' version before it reaches the operator/manifests.

Lets do other way around. Having manipulations in tests and scripts doesn't bring any value to the user. If Operator would support all valid formats of image references, then any transformation in test/scripts wouldn't be needed, and users would benefit from it.

zimnx: I think as part of this feature, we could also restructure config.yaml to accept full Scylla image ref instead just version. Places where split between repository and other part would be handled by the scripts around assets.

grzywin: Ok, I thought the goal of this change was to allow us to use the digest/tag/'latest', so that we have more control over the exact version we're testing, rather than to please the users.

PR #2493, opened 2025-02-20
Use deterministic names for remote Namespace and RemoteOwner

### LINKED ISSUE #2492: Remote datacenter is recreated due to Operator stale cache
Stale cache is expected at the boot, because Remote informers HasSynced functions retruns if all registered clusters are synced. Since informers are dynamically updated when controller reconciling RemoteKubernetesCluster notice objects, it may be misleading about the expected state.

The current implementation uses non-deterministic names for remote Namespace and RemoteOwner resources, which leads to significant issues when the Informer cache becomes stale. Specifically, the Operator may incorrectly determine that a resource does not exist due to the outdated cache. As a result, it creates a duplicate resource with a different name using the GenerateName field.

Once the Informer cache is eventually updated, the Operator detects the presence of two resources with similar purposes. To resolve this, it purges one of the duplicates. However, because these resources are often cluster-wide or serve as root objects owning dependent resources, deleting one triggers a cascading deletion of all associated objects.

zimnx:

Non-deterministic naming leads to problems when the Informer cache becomes outdated. The Operator might assume resources are missing, resulting in the creation of duplicate resources with different names via GenerateName. Once the Informer cache refreshes, the Operator detects the duplicates, removes one, and inadvertently triggers a cascading deletion of dependent objects. Since resources with GenerateName are typically cluster-wide or root objects controlling others, this could potentially wipe out entire datacenters. Switching to deterministic naming avoids duplicate resources by allowing the api-server to reject creation attempts if a resource already exists. Once the Informer cache updates, the next reconciliation cycle will either update the existing resource or skip creation altogether.

Fixes #2492

/cc mflendrich

### DIFF
--- a/pkg/controller/scylladbcluster/resource.go
+++ b/pkg/controller/scylladbcluster/resource.go
@@ -15,13 +15,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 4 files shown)


mflendrich (approved):
/lgtm

mflendrich:
assuming that the `MultiDC cluster [It] should form when external seeds are provided to ScyllaClusters [MultiDatacenter]` test failure is unrelated?

/hold

zimnx:
It's definitely unrelated. PR changes are only within multi region ScyllaDBCluster controller, while the failing test doesn't use this resource at all.

PR #2501, opened 2025-02-25
Build v1.16 documentation

mflendrich:
In preparation for the v1.16 release, build docs from v1.16 branch and treat them as unstable.

Part of #2495


zimnx (approved):
/lgtm

PR #2507, opened 2025-02-26
Bump `go.mod` deps

mflendrich:
Addresses https://pkg.go.dev/vuln/GO-2025-3487.

/cc zimnx
/kind cleanup

PR #2509, opened 2025-02-26
[v1.15] Bump `go.mod` deps

mflendrich:
See #2507


mflendrich:
We'll only upgrade `golang.org/x/crypto`

PR #2510, opened 2025-02-26
[1.14] Bump `go.mod` dependencies

mflendrich:
See #2507

/cc zimnx
/kind cleanup


zimnx:
/lgtm
But Go version needs to be bumped in CI as well

PR #2512, opened 2025-02-26
[1.13] Bump `go.mod` dependencies

mflendrich:
See #2507 

/cc zimnx
/kind cleanup


zimnx:
1.13 is EOL, we don't even have CI for it.
https://operator.docs.scylladb.com/master/support/releases.html#supported-releases

mflendrich:
Fair, I somehow incorrectly assumed that a check for 1.13 was present as well.

PR #2523, opened 2025-02-28
Refactor RemoteKubernetesCluster setup in E2E tests

zimnx:
Elimnates huge block setting up RemoteKubernetesCluster's used by the e2e tests by moving it to util function.

/cc mflendrich

PR #2524, opened 2025-02-28
Mirror ConfigMaps and Secrets referenced in ScyllaDBCluster into remote datacenters

### LINKED ISSUE #2255: [Multi-DC] Configuration Synchronization Across Multiple Datacenters
In order to reduce manual intervention and the risk of misconfigurations we need to improve user experience by ensuring `ConfigMap` is synchronized across multiple datacenters.

zimnx:
The ScyllaDBCluster controller has been extended with a new capability of mirroring the ConfigMaps and Secrets referenced by the ScyllaDBCluster, into remote datacenters.
This improvement provides users with the convenience of managing the configuration of an entire cluster from a single centralized location.

Note: any modifications made to the ConfigMaps or Secrets do not automatically initiate a rollout to apply those changes.
A manual trigger is required to implement the updates - for example using `forceRedeploymentReason`.

Requires:
- [x] https://github.com/scylladb/scylla-operator/pull/2523

Fixes #2255

/cc

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -443,6 +443,8 @@ rules:
   - endpoints
   - namespaces
   - services
+  - secrets
+  - configmaps
   verbs:
   - create
   - delete
--- a/deploy/operator/00_operator_remote.clusterrole_def.yaml
+++ b/deploy/operator/00_operator_remote.clusterrole_def.yaml
@@ -47,6 +47,8 @@ rules:
   - endpoints
   - namespaces
   - services
+  - secrets
+  - configmaps
   verbs:
   - create
   - delete
--- a/pkg/controller/scylladbcluster/controller.go
+++ b/pkg/controller/scylladbcluster/controller.go
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	corev1informers "k8s.io/client-go/informers/core/v1"
+
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	scyllaclient "github.com/scylladb/scylla-operator/pkg/client/scylla/clientset/versioned"
 	scyllav1alpha1informers "github.com/scylladb/scylla-operator/pkg/client/scylla/informers/externalversions/scylla/v1alpha1"
@@ -54,6 +56,8 @@ type Controller struct {
 
 	scyllaDBClusterLister      scyllav1alpha1listers.ScyllaDBClusterLister
 	scyllaOperatorConfigLister scyllav1alpha1listers.ScyllaOperatorConfigLister
+	configMapLister            corev1listers.ConfigMapLister
+	secretLister               corev1listers.SecretLister
 
 	remoteRemoteOwnerLister        remotelister.GenericClusterLister[scyllav1alpha1listers.RemoteOwnerLister]
 	remoteScyllaDBDatacenterLister remotelister.GenericClusterLister[scyllav1alpha1listers.ScyllaDBDatacenterLister]
@@ -62,6 +66,8 @@ type Controller struct {
 	remoteEndpointSliceLister      remotelister.GenericClusterLister[discoveryv1listers.EndpointSliceLister]
 	remoteEndpointsLister          remotelister.GenericClusterLister[corev1listers.EndpointsLister]
 	remotePodLister                remotelister.GenericClusterLister[corev1listers.PodLister]
+	remoteConfigMapLister          remotelister.GenericClusterLister[corev1listers.ConfigMapLister]
+	remoteSecretLister             remotelister.GenericClusterLister[corev1listers.SecretLister]
 
 	cachesToSync []cache.InformerSynced
 
@@ -78,6 +84,8 @@ func NewController(
 	scyllaRemoteClient remoteclient.ClusterClientInterface[scyllaclient.Interface],
 	scyllaDBClusterInformer scyllav1alpha1informers.ScyllaDBClusterInformer,
 	scyllaOperatorConfigInformer scyllav1alpha1informers.ScyllaOperatorConfigInformer,
+	configMapInformer corev1informers.ConfigMapInformer,
+	secretInformer corev1informers.SecretInformer,
 	remoteRemoteOwnerInformer remoteinformers.GenericClusterInformer,
 	remoteScyllaDBDatacenterInformer remoteinformers.GenericClusterInformer,
 	remoteNamespaceInformer remoteinformers.GenericClusterInformer,
@@ -85,6 +93,8 @@ func NewController(
 	remoteEndpointSliceInformer remoteinformers.GenericClusterInformer,
 	remoteEndpointsInformer remoteinformers.GenericClusterInformer,
 	remotePodInformer remoteinformers.GenericClusterInformer,
+	remoteConfigMapInformer remoteinformers.GenericClusterInformer,
+	remoteSecretInformer remoteinformers.GenericClusterInformer,
 ) (*Controller, error) {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartStructuredLogging(0)
@@ -98,6 +108,8 @@ func NewController(
 
 		scyllaDBClusterLister:      scyllaDBClusterInformer.Lister(),
 		scyllaOperatorConfigLister: scyllaOperatorConfigInformer.Lister(),
+		configMapLister:            configMapInformer.Lister(),
+		secretLister:               secretInformer.Lister(),
 
 		remoteRemoteOwnerLister:        remotelister.NewClusterLister(scyllav1alpha1listers.NewRemoteOwnerLister, remoteRemoteOwnerInformer.Indexer().Cluster),
 		remoteScyllaDBDatacenterLister: remotelister.NewClusterLister(scyllav1alpha1listers.NewScyllaDBDatacenterLister, remoteScyllaDBDatacenterInformer.Indexer().Cluster),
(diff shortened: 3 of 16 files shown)


zimnx:
ready to review
/cc rzetelskik
/cc mflendrich

test/e2e/set/scylladbcluster/multidatacenter/multidatacenter_scylladbcluster_config.go
```diff
@@ -0,0 +1,212 @@
...
+
+				for k, v := range secret.Data {
+					decodedValue, err := io.ReadAll(base64.NewDecoder(base64.StdEncoding, bytes.NewReader(mirroredSecret.Data[k])))
+					o.Expect(err).NotTo(o.HaveOccurred())
+					o.Expect(mirroredSecret.Data).To(o.HaveKey(k))
+					o.Expect(decodedValue).To(o.Equal(v))
+				}
+			}
+
+			framework.By("Verifying if configuration of ScyllaDB Manager Agent is being used by %q datacenter", dc.Name)
```
rzetelskik, line 191: nit: this is a bit implicit with the auth token, maybe the comment could specify auth token propagation?

rzetelskik (commented):
A few comments, rest lgtm

pkg/controller/scylladbcluster/controller.go
```diff
@@ -98,6 +107,8 @@ func NewController(
 
 		scyllaDBClusterLister:      scyllaDBClusterInformer.Lister(),
 		scyllaOperatorConfigLister: scyllaOperatorConfigInformer.Lister(),
+		configMapLister:            configMapInformer.Lister(),
+		secretLister:               secretInformer.Lister(),
```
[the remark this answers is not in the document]
zimnx: added comment

pkg/controller/scylladbcluster/resource.go
```diff
@@ -911,3 +913,192 @@ func getRemoteNamespaceAndController(controllerProgressingType string, sc *scyll
...
+	}
+
+	for k, v := range mirroredSecrets {
+		requiredRemoteSecrets[k] = append(requiredRemoteSecrets[k], v...)
+	}
+
+	return progressingConditions, requiredRemoteSecrets, nil
+}
+
+func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, remoteNamespaces map[string]*corev1.Namespace, remoteControllers map[string]metav1.Object, localSecretLister corev1listers.SecretLister, managingClusterDomain string) ([]metav1.Condition, map[string][]*corev1.Secret, error) {
```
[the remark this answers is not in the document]
zimnx: extended issue description about missing mirror of it

rzetelskik (approved):
lgtm
/assign mflendrich

deploy/operator/00_operator_remote.clusterrole_def.yaml
```diff
@@ -47,6 +47,8 @@ rules:
   - endpoints
   - namespaces
   - services
+  - secrets
```
mflendrich, line 50: This is all verbs across all namespaces for secrets. This may have been discussed before, but I don't think that creating such a wide clusterrole is a good idea from the security design perspective, and I'd avoid it even at the cost of UX.

Some alternatives off the top of my head (with various degrees of disruption):
- switch from secrets to CRDs for the specific bits we need to propagate
- only grant on the per-namespace basis (might require manual grants by the user)

deploy/operator.yaml
```diff
@@ -443,6 +443,8 @@ rules:
   - endpoints
   - namespaces
   - services
+  - secrets
+  - configmaps
```
mflendrich, line 447: This looks wider than necessary - see my other comment below.

Suggestion: on the helm template level, grant access to secrets on the per-resource level (ideally with `resourceNames`).

mflendrich (commented):
I think that we're granting too wide RBAC access and should design for the necessary minimum, not for catch-all.

deploy/operator/00_operator_remote.clusterrole_def.yaml
zimnx: We can't predict in which namespaces users will create their ScyllaDBCluster/ScyllaDBDatacenter nor predict names of their Secrets and ConfigMaps used within our CRs. Restricting those would be a terrible UX choice and I'm fully against it.

zimnx:
clean rebase to pick up e2e timeout fix

pkg/controller/scylladbcluster/sync_configmap.go
```diff
@@ -0,0 +1,88 @@
+package scylladbcluster
```
mflendrich, line 1: The two files `sync_configmap.go` and `sync_secret.go` are copies of each other with (a practical equivalent of) find-replace "configmap" into "secret".

I think that DRYing out this logic far outweighs the benefit of maintaining both implementations that should require only minimal polymorphism, if any. Is there a good reason not to do that?

pkg/naming/labels.go
```diff
@@ -91,9 +91,10 @@ func mergeLabels(l1, l2 map[string]string) map[string]string {
 	return res
 }
 
-func ManagedResourcesLabels(managingClusterDomain string) map[string]string {
+func RemoteManagedResourcesLabels(managingClusterDomain string) map[string]string {
 	return map[string]string{
-		ManagedByClusterLabel: managingClusterDomain,
+		"app.kubernetes.io/managed-by": RemoteOperatorAppNameWithDomain,
```
mflendrich, line 96: Can we add `"app.kubernetes.io/managed-by"` to `pkg/naming/constants.go`?

pkg/controller/scylladbcluster/resource.go
```diff
@@ -911,3 +913,192 @@ func getRemoteNamespaceAndController(controllerProgressingType string, sc *scyll
...
+			configMapsToMirror = append(configMapsToMirror, *dc.RackTemplate.ScyllaDB.CustomConfigMapRef)
+		}
+
+		for _, rack := range dc.Racks {
+			if rack.ScyllaDB != nil && rack.ScyllaDB.CustomConfigMapRef != nil {
+				configMapsToMirror = append(configMapsToMirror, *rack.ScyllaDB.CustomConfigMapRef)
+			}
+		}
+
+		for _, cmName := range configMapsToMirror {
```
mflendrich, line 974: If we're mirroring the secrets/CMs verbatim, we open ourselves us to the possibility of the user effectively injecting secrets/CMs of arbitrary names. This means in practice that write access to the ScyllaCluster resource effectively gives the possibility to (unintentionally or maliciously) overwrite other applications' configuration in a remote cluster that can have unbounded consequences. (less severe if using a unique namespace but exacerbated by the possibility to set a namespace name).

While this individual problem can be resolved by creating deterministic names for the remote CMs/secrets referenced by the scylla config, I am not immediately confident if that is enough of a remedy.

mflendrich (commented):
We decided offline that, in order to allow the remote SA access only to a predetermined (by the user) list of namespaces (as opposed to access to CMs/secrets across the entire cluster), we can introduce a possibility for the user to specify the set of namespaces that the remote operator will have access to.

But over the course of this review I came to a realization that this is not enough, and is making another problem more prominent -- the possibility to hijack other cms/secrets - as described in one of the comments attached.

Also attached two less important maintainability comments, but the remote resource overwrite issue seems critical to me.

deploy/operator/00_operator_remote.clusterrole_def.yaml
mflendrich: Discussed offline with a conclusion that we'd pave a way towards a possibility to restrict the set of namespaces that operator remotely writes to to a predefined list. But the issue mentioned in https://github.com/scylladb/scylla-operator/pull/2524#discussion_r2017409079 has uncovered yet another concern, so the resolution here is TBD based on our higher-order approach.

deploy/operator.yaml
mflendrich: duplicate of https://github.com/scylladb/scylla-operator/pull/2524#discussion_r2014115891

pkg/controller/scylladbcluster/resource.go
zimnx: > If we're mirroring the secrets/CMs verbatim, we open ourselves us to the possibility of the user effectively injecting secrets/CMs of arbitrary names. This means in practice that write access to the ScyllaCluster resource effectively gives the possibility to (unintentionally or maliciously) overwrite other applications' configuration in a remote cluster that can have unbounded consequences

It's not possible because Operator only work on resources matching our selector and having our controller ref. Such victim resource would have to be in Namespace managed by Operator, have controllerRef pointing to RemoteOwner and have labels matching ScyllaDBCluster selector. 
If someone would have access to edit the resource to add those, they surely does have access to modify resource content.

pkg/controller/scylladbcluster/sync_configmap.go
zimnx: It's how our most of sync_* files look like. In the past I already tried several times and failed due to generics limitations.

pkg/naming/labels.go
zimnx: done

mflendrich (approved):
As discussed offline (at length), this implementation comes with an unconditional requirement that remote kubernetes cluster access must have [very broad permissions](https://github.com/scylladb/scylla-operator/pull/2524/files#r2014115891) effectively permitting reading of secrets, including SA tokens, _across all namespaces_, which paves a theoretically exploitable way towards privilege escalation (by reading 3rd party access tokens) or other attacks (by overwriting 3rd party configuration stored in secrets/CMs). While mechanisms in the operator exist that will likely (in the typical case) make it very hard to overwrite or leak 3rd party secrets, direct access to the remote kubernetes cluster kubeconfig opens the floodgates.

My first and foremost recommendation is to adjust the design so that
- remote operator access is restricted to a user-defined set of namespaces from day 1
- RBAC-controlled access to all secrets/configmaps is not necessary (e.g. by using deterministically named secrets/CMs in the remote cluster)

but the approach predates my engagement with the project, so I'm not directly questioning it, assuming that these limitations are well-known.

To unblock possible future improvement here, over the course of review we've made adjustments give the user control over the remote namespace name (created by operator in the remote cluster), making it possible to pave a future way (see #2577) to incrementally implement the two traits expressed in the bullet points above.

Therefore I'm conditionally approving this PR with an expectation that it's decided whether those limitations (materialized by #2577) should be a prerequisite for release or not.

Also created #2576 for a code health issue (not directly introduced by this PR but somewhat exacerbated) to be addressed at a later time.

/lgtm
(with the above caveat)

/hold for acknowledgement

mflendrich:
/lgtm
(it was not picked up from https://github.com/scylladb/scylla-operator/pull/2524#pullrequestreview-2747320423)

zimnx:
ack
/hold cancel

PR #2525, opened 2025-02-28
Refactor e2e package structure

zimnx:
Multi datacenter tests were moved to subpackage and their file name prefix is removed. Verifiers for v1alpha1.ScyllaDBCluster were moved to separate package to be consistent with v1.ScyllaCluster.

/cc mflendrich


mflendrich (approved):
Thanks!

/lgtm

PR #2528, opened 2025-03-05
Clear data keyspace snapshots for first node in rack during ScyllaDB version upgrade

### LINKED ISSUE #2527: Data keyspace snapshots are not removed for 0'th node post ScyllaDB version upgrade
During ScyllaDB minor/major version upgrade Operator takes data keyspaces snapshots before upgrading version for particular node. Once node is upgraded and Ready it's removed.
Due to a bug, snapshots are not removed for 0'th node in each rack. 

Steps to reproduce:
1. Deploy ScyllaDB cluster having 3 racks having 3 nodes.
2. Create user keyspace.
2. Upgrade ScyllaDB version minor/major version.
3. Post upgrade list snapshots on every node.

On 0'th node in each rack, there're user keyspace snapshots having "so_" prefix, indicating those are created by Operator. Other nodes doesn't have them.

zimnx:

Fixed a bug where user keyspace snapshots persisted after a ScyllaDB upgrade.
The issue stemmed from an early return that skipped the postUpgradeHook call after a StatefulSet having partition 0 rolled out, leaving snapshots uncleared on the 0'th node.
Extended E2E upgrade validation to check for lingering snapshots, which now passes after this fix.

Resolves #2527

/cc mflendrich

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_upgrades.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_upgrades.go
@@ -109,6 +109,17 @@ var _ = g.Describe("ScyllaCluster upgrades", func() {
 				o.Expect(err).NotTo(o.HaveOccurred())
 			}
 			scyllaclusterverification.VerifyCQLData(ctx, di)
+
+			framework.By("Validating if all snapshots created by upgrade hooks are cleared for every node")
+			scyllaClient, hosts, err := utils.GetScyllaClient(ctx, f.KubeClient().CoreV1(), sc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			for _, host := range hosts {
+				snapshots, err := scyllaClient.ListSnapshots(ctx, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				o.Expect(snapshots).To(o.BeEmpty())
+			}
 		},
 		// Test 1 and 3 member rack to cover e.g. handling PDBs correctly.
 		g.Entry(describeEntry, &entry{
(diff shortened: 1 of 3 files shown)


test/e2e/set/scyllacluster/scyllacluster_upgrades.go
```diff
@@ -5,7 +5,6 @@ package scyllacluster
 import (
 	"context"
 	"fmt"
-
```
mflendrich, line 8: Why remove this whitespace? Do we have a goimports-like standard for the import block structure?

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
```diff
@@ -109,6 +108,16 @@ var _ = g.Describe("ScyllaCluster upgrades", func() {
 				o.Expect(err).NotTo(o.HaveOccurred())
 			}
 			scyllaclusterverification.VerifyCQLData(ctx, di)
+
+			scyllaClient, hosts, err := utils.GetScyllaClient(ctx, f.KubeClient().CoreV1(), sc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			for _, host := range hosts {
+				snapshots, err := scyllaClient.ListSnapshots(ctx, host)
```
mflendrich, line 116: For comprehensibility 1 year from now, I suggest adding a comment explaining the intention - snapshots AFAICT aren't synonymous with post upgrade hooks, especially given the possibility of giving up snapshots during upgrade in the future altogether.

For example:

```
// Ensure that the post upgrade hooks have been run on every single node
```

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
zimnx: fixed, my IDE was missing import reformat action on save.

mflendrich (approved):
Thanks :+1: 
/lgtm

PR #2529, opened 2025-03-05
Add manual X.Y tagging to the release procedure

mflendrich:
Now that we're disabling automated rolling tag update, it has to become a manual step in the procedure.

This PR adds the new requirement to the checklist.

/kind cleanup
/priority important-soon
/cc zimnx


mflendrich:
Decided offline with @zimnx that we will keep the rolling tags as they are, and reconsider when we have a more complete redesign of the release process allowing for automated patch releases.

Issue #2496, opened 2025-02-24
[Test Flake] artifact collection


mflendrich:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-latest-e2e-gke-arm64-parallel/1893903885731368960

### Snippet of what failed.

```
 + kubectl -n=gather-artifacts cp --retries=42 -c=wait-for-artifacts must-gather:/tmp/artifacts /logs/artifacts/must-gather/0
tar: Removing leading `/' from member names
E0224 07:06:29.843524     988 v2.go:167] "Unhandled Error" err="read message: %!w(<nil>)"
E0224 07:06:29.843536     988 v2.go:150] "Unhandled Error" err="read message: %!w(<nil>)"
error: error reading from error stream: read message: %!w(<nil>) 
```

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2506/pull-scylla-operator-v1.16-e2e-gke-multi-datacenter-parallel/1894689216319721472

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2525/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1896878165884669952

mflendrich:
Fixed by #2540 (hopefully).

PR #2540, opened 2025-03-07
Change retries of e2e artifact collection

zimnx:
Instead using retries flag which doesn't fully restart the stream, lets try retrying entire command.

/cc mflendrich

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -73,7 +73,10 @@ EOF
 
   exit_code="$( wait-for-container-exit-with-logs gather-artifacts must-gather must-gather )"
 
-  kubectl -n=gather-artifacts cp --retries=42 -c=wait-for-artifacts must-gather:/tmp/artifacts "${1}"
+  for i in {1..60}; do
+    kubectl -n=gather-artifacts cp -c=wait-for-artifacts must-gather:/tmp/artifacts "${1}" && break || echo "Attempt $i to collect artifacts failed, retrying"
+  done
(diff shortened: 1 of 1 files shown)


mflendrich (commented):
I have looked into kubectl sources and I'm inclined to think that this will not work.

Note that the `--retries` flag controls the behavior of `TarPipe.Read`
https://github.com/kubernetes/kubectl/blob/deef152435a8265c9cc91ff12e43c87132dbb0a8/pkg/cmd/cp/cp.go#L406-L421

while the error (likely) comes from the websocket implementation here
https://github.com/kubernetes/kubernetes/blob/ea49618a749b84c3cb5bd781380891eb51289950/staging/src/k8s.io/client-go/tools/remotecommand/websocket.go#L315
which, in case of the error we're experiencing, returns in a way that makes subsequent repeated retries get the same "pipe closed" situation where retries change nothing.

I think that the two options below have a better chance of being successful:
- move the retries one level up - retry the `kubectl cp` command from the level of this script,
- change protocol from websocket to SPDY by feature flag (a shot in the dark but might or might not change something). I read [this comment](https://github.com/kubernetes/kubernetes/issues/60140#issuecomment-2565537460) as somewhat promising that it might address the issue.

Therefore my suggestion is to refrain from merging this, and instead do either SPDY or retries above the kubectl level (or both?)

mflendrich (approved):
/lgtm

and TIL about the move from SPDY to ws

mflendrich:
xref #2496

PR #2541, opened 2025-03-07
Graceful e2e shutdown on timeout

zimnx:
With scylladb/scylla-operator-release#406 e2e script receives SIGINT when timeout happens. It's not propagated to actual e2e binary because it's being run as Pod in spawned e2e Kubernetes cluster. This registers INT signal handler which will gracefully delete e2e Pod giving it a chance to print out where it's being stuck and collect e2e artifacts.

/cc mflendrich

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -73,9 +73,7 @@ EOF
 
   exit_code="$( wait-for-container-exit-with-logs gather-artifacts must-gather must-gather )"
 
-  for i in {1..60}; do
-    kubectl -n=gather-artifacts cp -c=wait-for-artifacts must-gather:/tmp/artifacts "${1}" && break || echo "Attempt $i to collect artifacts failed, retrying"
-  done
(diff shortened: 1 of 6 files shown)


hack/.ci/lib/e2e.sh
```diff
@@ -99,6 +99,10 @@ function gather-artifacts-on-exit {
   cleanup-bg-jobs "${ec}"
 }
 
+function gracefully-shutdown-e2es-on-interrupt {
```
mflendrich, line 102: SRP: the interrupt is semantically part of the `trap` expression, not this function definition. Suggested rename

```suggestion
function gracefully-shutdown-e2es {
```

+ update to the `trap` call

mflendrich (approved):
/lgtm

zimnx:
Signal wasn't set due to wrong timeout on e2e run script, fixed by https://github.com/scylladb/scylla-operator-release/pull/407
/retest

zimnx:
Didn't work, pod got terminated and removed and because of it logs are missing in artifacts :facepalm: I'll fix that

mflendrich (approved):
/lgtm

PR #2555, opened 2025-03-18
[1.16] Propagate racks.exposeOptions in Helm

mflendrich:
Follow-up to https://github.com/scylladb/scylla-operator/pull/2484/ (#2479)

Part of #2554

Ensures that the Helm template for ScyllaCluster propagates the rack-specific `exposeOptions`.

PR #2556, opened 2025-03-18
Render `.spec.datacenter.racks` directly in Helm template

### LINKED ISSUE #2554: Scylla Helm Chart Missing Support for exposeOption in spec.datacenter.racks
Description:
A customer who has upgraded to Scylla Operator 1.16 noticed that the current Helm chart does not support the exposeOption field inside spec.datacenter.racks. The missing field prevents configuring certain exposure options at the rack level.

Expected Behavior:
The exposeOption field should be available in the Helm chart under spec.datacenter.racks 

Helm template reference: https://github.com/scylladb/scylla-operator/blob/master/helm/scylla/templates/scyllacluster.yaml#L70-L121

mflendrich:
Replacing field-wise rendering of `.spec.datacenter.racks` with a direct render to make maintenance simpler (not need to copy fields over to the helm template when adding them in the CRD).

Part of #2554 
Follow-up to https://github.com/scylladb/scylla-operator/pull/2484 (https://github.com/scylladb/scylla-operator/issues/2479)

An equivalent but more conservative change is landing in 1.16: https://github.com/scylladb/scylla-operator/pull/2555.
This PR is a solution to #2554 for `master`. #2555 (and not a backport of this PR) fixes #2554 for 1.16.

This should produce an exactly identical render as before, with the below differences:
- the fields in an object will be in a different order
- the `.racks[].exposeOptions` field that didn't exist in the preexisting template is now rendered correctly
- semantically equivalent indentation changes (at least in `.racks[].placement.tolerations`, maybe elsewhere as well)

/cc zimnx
/kind bug
/priority important-soon

### DIFF
--- a/deploy/manager-dev.yaml
+++ b/deploy/manager-dev.yaml
@@ -310,18 +310,8 @@ spec:
   datacenter:
     name: manager-dc
     racks:
-      - name: manager-rack
-        members: 1
-        storage:
-          capacity: 5Gi
-          storageClassName: scylladb-local-xfs
-        resources:
-          limits:
-            cpu: 200m
-            memory: 200Mi
-          requests:
-            cpu: 10m
-            memory: 100Mi
+      - members: 1
+        name: manager-rack
         placement:
           nodeAffinity:
             requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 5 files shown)


zimnx (approved):
/lgtm

zimnx:
Helm is not used in e2e so changes cannot break it
/override ci/prow/e2e-gke-parallel-clusterip

PR #2564, opened 2025-03-26
Move e2e tests RemoteKubernetesCluster Secret to dedicated namespace

zimnx:
Creates a dedicated namespace for RemoteKubernetesCluster Secrets in the multidatacenter test suite, storing the kubeconfig Secret there instead of alongside the ScyllaDBCluster's. Previously, deleting the test namespace also removed the Secret, breaking the ScyllaDBCluster finalizer’s ability to validate remote cluster objects, stalling cleanup and causing timeouts. By isolating the Secret in its own namespace, we ensure it persists for the finalizer, fixing the timeouts.

### DIFF
--- a/test/e2e/utils/clusters.go
+++ b/test/e2e/utils/clusters.go
@@ -3,10 +3,11 @@ package utils
 import (
 	"context"
 	"fmt"
+	"time"
+
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/gather/collect"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"time"
 
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/naming"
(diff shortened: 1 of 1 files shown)


rzetelskik (commented):
This looks like it should fix the issue but I'm wondering - isn't this something we might be bumping into in the future? Wouldn't it be more error proof to just change the implementation of `func (c *Cluster) Cleanup`? If we switched from FIFO to LIFO, cleaners would be invoked in the reverse order of their registration? We could rename the other functions to make it clear that it behaves like a stack `PushCleaners` instead of `AddCleaners` e.g.

zimnx:
> This looks like it should fix the issue but I'm wondering - isn't this something we might be bumping into in the future? Wouldn't it be more error proof to just change the implementation of func (c *Cluster) Cleanup? If we switched from FIFO to LIFO, cleaners would be invoked in the reverse order of their registration? We could rename the other functions to make it clear that it behaves like a stack, e.g. PushCleaners instead of AddCleaners

As discussed offline, LIFO would work if we would delete resources explicitly, but we don't as ScyllaClusters and other resources are removed when test namespace is removed.

rzetelskik:
/lgtm
Thanks

PR #2582, opened 2025-04-10
Align naming of Multidatacenter Config Sync E2E

zimnx:
"multidatacenter" prefix was removed as file is in "multidatacenter" directory.


rzetelskik (approved):
/lgtm

PR #2585, opened 2025-04-10
Update docs theme to 1.8.6

dgarcia360:
Updates docs theme to the latest version. 

Details: https://sphinx-theme.scylladb.com/stable/upgrade/CHANGELOG.html

### DIFF
--- a/docs/poetry.lock
+++ b/docs/poetry.lock
@@ -13,13 +13,13 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.8.0"
+version = "4.9.0"
 description = "High level compatibility layer for multiple asynchronous event loop implementations"
 optional = false
 python-versions = ">=3.9"
 files = [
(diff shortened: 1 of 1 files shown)


docs/poetry.lock
```diff
@@ -1,25 +1,27 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
```
rzetelskik, line 1: We still use poetry 1.8.3, trying to update (see https://github.com/scylladb/scylla-operator/blob/f8edb3bee935e3c778a749e9d7abcb8acde44620/docs/README-dev.md#update-dependencies) with these changes results in this warning:
```
The lock file might not be compatible with the current version of Poetry.
Upgrade Poetry to ensure the lock file is read properly or, alternatively, regenerate the lock file with the `poetry lock` command.
```

Not really sure if this is an issue but I'd stick to using the same version.

dgarcia360: Done!

PR #2592, opened 2025-04-15
Introduce Dependabot to bump go.mod & vendored packages

mflendrich:
Now that Dependabot supports gomod vendoring and PR grouping, all technical blockers preventing us from productive automated dependency management seem to be off the table.

Therefore, this PR adds a Dependabot config that will result in weekly grouped PRs with version bumps.
There are 2 PRs with weekly cadence expected: one bumping Kubernetes dependencies and one for the rest.
Vendoring is enabled and I'd expect it to run fine with our CI - we shall see with the first run.

This only manages the dependencies on the main branch; we'd still need to backport manually.

Also includes a weekly auto-bumper for GH Actions.

/cc zimnx rzetelskik

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,29 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: weekly
+    day: wednesday
+    time: '08:00' # UTC
+  labels:
+  - priority/important-longterm
+  - kind/dependency-bump
+  groups:
+    k8s.io:
+        applies-to: version-updates
+        patterns:
+          - "k8s.io/*"
+          - "sigs.k8s.io/*"
+    gomod:
+        applies-to: version-updates
+        patterns:
+          - "*"
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
+  labels:
+  - priority/important-longterm
+  - kind/dependency-bump
+vendor: true


mflendrich:
If it occurs to us that a certain class of dependencies requires additional regeneration, we'll be able to put them in a separate group and do one of 2 things
- exclude them altogether, or
- manually edit the PR to include additional codegen

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
...
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: weekly
+    day: wednesday
+    time: '08:00' # UTC
+  labels:
+  - priority/important-longterm
+  - area/dependency
+  - kind/cleanup
```
zimnx, line 12: Bumping dependency may introduce feature, it won't clean anything in our codebase. 
```suggestion
  - kind/feature
```

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
...
+        patterns:
+          - "*"
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
+  labels:
+  - priority/important-longterm
+  - area/dependency
+  - kind/cleanup
```
zimnx, line 30:
```suggestion
  - kind/feature
```

zimnx (commented):
Lets try it out, two comments

.github/dependabot.yml
rzetelskik: How's a dependency bump a feature from the operator's user perspective? Remember this also propagates to the release notes. I think cleanup fits fine, as per the label's comment, it "Categorizes issue or PR as related to cleaning up code, process, or technical debt."

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
+version: 2
+updates:
```
rzetelskik, line 2: any chance we can align the PR titles and commit subject lines to our contributing guide?

rzetelskik (approved):
> There are 2 PRs with weekly cadence expected: one bumping Kubernetes dependencies and one for the rest.

What's the reasoning behind the split?

> If it occurs to us that a certain class of dependencies requires additional steps (like extra code generation), we'll be able to put them in a separate group and do one of 2 things
> 
> - exclude them altogether, or
> - manually edit the PR to include additional codegen

I'm not very optimistic about this given it will also require independently updating builder images, but let's see I guess.

@mflendrich two other questions:
1. did you leave out documentation dependencies on purpose?
2. did you by any chance look into if this could somehow be used to bump
a) the base image of scylla-operator image?
b) the asset images?

.github/dependabot.yml
zimnx: We use dependencies for bunch of stuff, If for example there's performance improvement in the dependent library, then Operator can gain this performance improvement as well if it's in its hot branch. Performance improvements are features. 
Or if Scylla Manager starts supporting new bucket location, then bumping manager client may relax validation unblocking new feature in Operator. 

We use dependencies for features they provide, so even a bug fix there is a feature for us. 

> I think cleanup fits fine, as per the label's comment, it "Categorizes issue or PR as related to cleaning up code, process, or technical debt."

I don't agree, dependency bump doesn't fall into any of these categories.

mflendrich: I feel about this in two different ways at once:

1. In the current system, I think that dependency bumps deserve their own category (say `kind/dependency-bump`) 
2. Usefulness of dep bumps in the release notes commit-per-commit is minimal, - what matters is:
    - Scylla ecosystem default version changes
    - Scylla ecosystem supported version range changes
    - non-Scylla-ecosystem co-installed software default versions (prom, grafana, etc)
    - and maybe a diff of go.mod, and (even less likely) base image updates.

My suggestion is to, in this order:

In scope of this PR:
1. introduce `kind/dependency-bump`
2. if we don't want to introduce `kind/dependency-bump`, keep `kind/cleanup` because version bumps IMO are noise to a user skimming through a list of functional changes

Outside of the scope of this PR: get the dep bumps excluded from the changelog and include a report of version changes

mflendrich: Replying directly to @zimnx's comment as we were responding simultaneously :). This is a paraphrase of the above, addressing the comment directly.

I agree that version bumps can bring useful value to the user. I however don't believe that (in a typical case) they're functional changes and therefore don't belong under `kind/feature`.

Therefore I suggest that we find a way to keep `kind/feature` reserved for functional changes and to present dependencies in a useful way.

(I think we may be hitting the limits of our changelog tooling)

.github/dependabot.yml
mflendrich: I understand based on [docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#commit-message--) and empirical evidence (dependabot PRs in other repos) that:

- the PR title/commit desc is already very close to our contributing guide; the sentence structure is fine, the only difference is capitalization of the 1st letter
- the config expressive power doesn't allow to adjust this detail

If it hurts our eyes, my take would be to sanitize the PR title in the changelog generation tool. But as discussed in https://github.com/scylladb/scylla-operator/pull/2592#discussion_r2045051784, I think that representation of deps bumps in the changelog has other solutions on the table that might render this a moot point.

TL;DR: sorry but no :sweat_smile:

rzetelskik:
thanks for the updates, lgtm
/assign zimnx

.github/dependabot.yml
zimnx: I would be ok with new kind

mflendrich: Done - PTAL @zimnx 

Also adding the label to labelsync: https://github.com/scylladb/scylla-operator-release/pull/413

zimnx (approved):
/lgtm

PR #2593, opened 2025-04-16
Fix `vendor` syntax in the dependabot config

mflendrich:
Fixes the below issue.

> Your .github/dependabot.yml contained invalid details
> 
> Dependabot encountered the following error when parsing your .github/dependabot.yml:
> 
>     The property '#/' contains additional properties ["vendor"] outside of the schema when none are allowed

Also bumps the cron trigger a bit so that we can get dependabot to kick in today with the corrected configuration.

follow-up to #2592

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,10 +2,11 @@ version: 2
 updates:
 - package-ecosystem: gomod
   directory: /
+  vendor: true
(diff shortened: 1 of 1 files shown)


grzywin (approved):

zimnx (approved):
/lgtm

PR #2599, opened 2025-04-16
Fix Available Condition of RemoteKubernetesCluster after recovery

zimnx:

Sets ClientHealthcheckControllerAvailable condition status when RemoteKubernetesCluster healthcheck probes succeeds.

Previously aggregated Available condition was set to True when there was no ClientHealthcheckControllerAvailable having False status. When access to Kubernetes is broken and then restored, nothing removed this failed condition, hence RemoteKubernetesCluster was constantly not Available. To fix that, healthcheck controller sets this condition according to state of given healthcheck iteration.


mflendrich (approved):
/lgtm

PR #2608, opened 2025-04-25
Update docs dependencies

dgarcia360:
This is a mandatory dependency update to resolve a critical Dependabot alert. For more details, see the [Dependabot alerts](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/viewing-and-updating-dependabot-alerts).


zimnx (approved):
/approve
/lgtm

PR #2616, opened 2025-04-29
Stop building docs for 1.13 and 1.14

rzetelskik:
**Description of your changes:** 1.13 and 1.14 are EOL now, we should stop building docs for corresponding branches to reduce the maintenance cost (e.g. backporting https://github.com/scylladb/scylla-operator/pull/2613).

Resolves #

/kind documentation
/priority important-longterm


zimnx (approved):
/lgtm

PR #2641, opened 2025-05-14
Support backup and repair task scheduling with global ScyllaDB Manager for v1alpha1.ScyllaDBDatacenter

### LINKED ISSUE #2571: Support backup and repair task scheduling for v1alpha1.ScyllaDBDatacenter
## Acceptance criteria
- [ ] Introduce ScyllaDBManagerTask API supporting backups and repairs
- [ ] Add validation of ScyllaDBManagerTasks
  - Includes references to ScyllaDBDatacenters
- [ ] Add a controller scheduling tasks defined with ScyllaDBManagerTask for ScyllaDBDatacenters
  - Includes implementation of controllers and finalizers
- [ ] Add an E2E test for scheduling backup and repair tasks with ScyllaDBDatacenter

rzetelskik:
**Description of your changes:** This PR introduces support for scheduling backup and repair tasks for v1alpha1.ScyllaDBDatacenters with global ScyllaDBManager instance. A new v1alpha1.ScyllaDBManagerTask custom resource is introduced for that purpose. 

As this is an intermediate step towards migrating ScyllaCluster's manager-related spec to the new CR, the validation and controllers support various overrides through annotations to allow for backwards compatibility with the legacy API.

A note for reviewers: please verify the implementation against [the enhancement proposal](https://github.com/scylladb/scylla-operator/tree/613a6adc5429279b9e62b8272d49fb4d4dd4c650/enhancements/proposals/2549-scylladbcluster-backup-repair) and the existing implementation, including the manager-controller and ScyllaCluster validation.

Requires:
- [x] https://github.com/scylladb/scylla-operator/pull/2590
- [x] https://github.com/scylladb/scylla-operator/pull/2676

Resolves #2571

/kind feature
/priority important-soon
/cc


pkg/api/scylla/v1alpha1/scylla.scylladb.com_scylladbmanagertasks.yaml
```diff
@@ -0,0 +1,275 @@
...
+                        type: string
+                      type: array
+                    keyspace:
+                      description: |-
+                        keyspace specifies a list of `glob` patterns used to include or exclude tables from repair.
+                        The patterns match keyspaces and tables. Keyspace names are separated from table names with a dot e.g. `!keyspace.table_prefix_*`.
+                      items:
+                        type: string
+                      type: array
+                    location:
```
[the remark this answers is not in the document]
rzetelskik: According to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required, the lack of `+optional` comment tag and `optional` struct tag is enough to consider it required. Leaving as is.

pkg/api/scylla/validation/scylladbmanagertask_validation.go
```diff
@@ -0,0 +1,271 @@
...
+
+	return allErrs
+}
+
+func ValidateScyllaDBManagerRepairTaskOptions(repairOptions *scyllav1alpha1.ScyllaDBManagerRepairTaskOptions, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, ValidateScyllaDBManagerTaskSchedule(&repairOptions.ScyllaDBManagerTaskSchedule, fldPath)...)
+
+	if repairOptions.Intensity != nil && *repairOptions.Intensity < 0 {
```
czeslavo, line 207: This could be covered on the API server side if we annotated the field with `// +kubebuilder:validation:Minimum=0`.

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
```diff
@@ -0,0 +1,179 @@
...
+	// +optional
+	Intensity *int64 `json:"intensity,omitempty"`
+
+	// parallel specifies the maximum number of ScyllaDB repair jobs that can run at the same time (on different token ranges and replicas).
+	// Each node can take part in at most one repair at any given moment. By default, the maximum possible parallelism is used.
+	// The maximal effective parallelism depends on keyspace replication strategy and cluster topology.
+	// When set to a value greater than the maximum supported by the node, parallel is capped at the maximum supported value.
+	// Refer to repair documentation for details.
+	// +optional
+	Parallel *int64 `json:"parallel,omitempty"`
```
czeslavo, line 113: Shouldn't we verify this is not negative?

czeslavo (commented):
I reviewed the API surface and validation logic. Left some suggestions/questions.

BTW, wouldn't it make sense to split this PR into smaller ones, e.g.:
- CRD API
- validation logic + tests
- controller implementation + unit tests
- E2E test

This way, we could iterate more frequently and focus on a narrower context each time. As someone new to the codebase, I find it difficult to hold the full feature context at once, and reviewing such a large volume of code thoroughly in one go takes too much time.

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
```diff
@@ -0,0 +1,179 @@
...
+	// observedGeneration is the most recent generation observed for this ScyllaDBManagerTask. It corresponds to the
+	// ScyllaDBManagerTask's generation, which is updated on mutation by the API Server.
+	// +optional
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+
+	// conditions hold conditions describing ScyllaDBManagerTask state.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// taskID reflects the internal identification number of the task in ScyllaDB Manager state.
```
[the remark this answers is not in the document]
rzetelskik: > Is this ID useful to a user?

yes, it's a reflection of the state in manager

> Could we point them to some docs or document it inline on how to make use of the ID when something goes wrong with their task?

I don't think the API is a good place for it. The documentation should refer to the API, not the other way around imo.

pkg/api/scylla/validation/scylladbmanagertask_validation.go
rzetelskik: we moved away from mixing the two - we use the webhook either way. Is there a benefit to doing bits of validation on the apiserver side anyway?

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
```diff
@@ -0,0 +1,179 @@
...
+	// `<dc>:` is optional and allows for specifying different limits in selected datacenters. If `<dc>:` is not set, the limit is global.
+	// For instance, `[]string{"dc1:2", "5"}` corresponds to two parallel nodes in `dc1` datacenter and five parallel nodes in the other datacenters.
+	// +optional
+	SnapshotParallel []string `json:"snapshotParallel,omitempty"`
+
+	// uploadParallel specifies a list of upload parallelism limits in the following format: `[<dc>:]<limit>`.
+	// `<dc>:` is optional and allows for specifying different limits in selected datacenters. If `<dc>:` is not set, the limit is global.
+	// For instance, `[]string{"dc1:2", "5"}` corresponds to two parallel nodes in `dc1` datacenter and five parallel nodes in the other datacenters.
+	// +optional
+	UploadParallel []string `json:"uploadParallel,omitempty"`
```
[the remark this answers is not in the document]
czeslavo: :+1: Given we can capture an error on a task creation when calling the manager, I think it makes sense to not duplicate validation logic on the operator side.

> manager is extremely bad in terms of API guarantees

Is this something that we experienced in the past? I'm worried that we rely on the manager API, but it doesn't keep guarantees. :grimacing:

pkg/api/scylla/validation/scylladbmanagertask_validation.go
czeslavo: The advantage is that if a user doesn't install the webhook server, they still get the validation (with fewer network hops, but it's not that important). I understand the approach where we implement everything in one place, and it makes sense if we assume that the webhook server installation is required.

Anyway, if validations are simple enough, they can be implemented with https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules and https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/ on the server side, and effectively reduce the complexity of the deployment (dropping webhook server deployment and webhook configuration). But this requires going all-in to get the benefits.

pkg/api/scylla/validation/scylladbmanagertask_validation_test.go
```diff
@@ -0,0 +1,1215 @@
+// Copyright (C) 2025 ScyllaDB
+
+package validation
```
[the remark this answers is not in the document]
rzetelskik: You can achieve the same thing by just testing the exported functions, I don't think it's much more than a convention thing to have an explicit separation

pkg/api/scylla/validation/scylladbmanagertask_validation.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+type ValidateScyllaDBManagerTaskObjectMetaOptions struct {
+	ValidateScyllaDBManagerTaskObjectMetaAnnotationsOptions ValidateScyllaDBManagerTaskObjectMetaAnnotationsOptions
+}
+
+type ValidateScyllaDBManagerTaskObjectMetaAnnotationsOptions struct {
+	IsScyllaDBManagerTaskScheduleCronNil              bool
+	IsScyllaDBManagerTaskRepairIntensityNil           bool
+	IsScyllaDBManagerTaskRepairSmallTableThresholdNil bool
```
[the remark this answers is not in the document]
rzetelskik: About the naming - fair point, I started this trying to implement actual functions following the options pattern, but figured the interactions would be very convoluted, as the actual validation logic would be created in different places. So I ended up just adding boolean flags.
I'm opposed to passing these as parameters though as having this struct allows for clear separation between validation of annotations and options coming from particular parts of spec. It also allows for having a constructor for these modifies and propagating only the relevant parts.
Perhaps different naming would help? How about `ValidateScyllaDBManagerBackupTaskOptionsFlags`, `ValidateScyllaDBManagerBackupTaskOptionsModifiers`?

pkg/api/scylla/validation/scylladbmanagertask_validation_test.go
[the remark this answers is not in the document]
czeslavo: Testing only public interfaces in most cases is a good approach (allows focusing on the perspective of a user). Putting tests inside a `{pkg}_test`  package enforces that tests won't compile if you were to use a non-exported symbol. So I would say it's not just a convention - it's a way to express intent and keep tests clean.

pkg/api/scylla/validation/scylladbmanagertask_validation.go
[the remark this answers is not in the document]
czeslavo: `ValidateScyllaDBManagerBackupTaskOptionsFlags` :+1:

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
rzetelskik: I double checked and neither the API server, nor sctool prevent this from being negative. We could on our side, but I'm not exactly sure it makes sense. @Michal-Leszczynski ?

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
```diff
@@ -0,0 +1,179 @@
...
+	Location []string `json:"location"`
+
+	// rateLimit specifies the limit for the upload rate, expressed in megabytes (MiB) per second, at which the snapshot files can be uploaded from a ScyllaDB node to its backup destination, in the following format: `[<dc>:]<limit>`.
+	// `<dc>:` is optional and allows for specifying different upload limits in selected datacenters.
+	// +optional
+	RateLimit []string `json:"rateLimit,omitempty"`
+
+	// retention specifies the number of backups to store.
+	// +optional
+	Retention *int64 `json:"retention,omitempty"`
```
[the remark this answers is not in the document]
rzetelskik: Yeah, manager server doesn't accept a negative value, so it's better we do that. This will require another override annotation though.

pkg/api/scylla/validation/scylladbmanagertask_validation.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+var (
+	supportedScyllaDBManagerTaskTypes = []scyllav1alpha1.ScyllaDBManagerTaskType{
+		scyllav1alpha1.ScyllaDBManagerTaskTypeBackup,
+		scyllav1alpha1.ScyllaDBManagerTaskTypeRepair,
+	}
+)
+
+type ValidateScyllaDBManagerTaskObjectMetaOptions struct {
```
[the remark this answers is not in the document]
rzetelskik: good point, fixed

pkg/api/scylla/validation/scylladbmanagertask_validation.go
rzetelskik: We tried moving to CEL entirely at some point but it proved not expressive enough

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: Done. Did you see any other concerning fields?

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: I think the furthest we could go here is saying something along the lines of `It can be used to identify the task in manual interactions with ScyllaDB Manager`, but isn't that mostly redundant with the existing comment?

pkg/api/scylla/validation/scylladbmanagertask_validation.go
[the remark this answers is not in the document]
rzetelskik: done

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: I'm not exactly convinced by the struct idea (migration would be really inconvenient, as we'd have to encode the slices, pass them through overriding annotations and decode them in the controller), but I think we should consider strengthening the validation overall - it's now or never. My worry is that not doing this on our side leads to spamming the manager with invalid requests until the controller hits rate limiting. Manager client doesn't allow us to recognise an invalid request either.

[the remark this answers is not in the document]
rzetelskik: @zimnx @mflendrich wdyt? About strenghtening the validation in general and @czeslavo's proposed changes to the API?

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[5bf5142](https://github.com/scylladb/scylla-operator/commit/5bf514213848e3bf54813c9f5f8a2f5adabc83f7)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2641/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1929610065006825472)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2641). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

known manager flake with colliding snapshot IDs
/test images
/retest

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
czeslavo: There are other fields, such as `UploadParallel`, where we accept integers as strings, so we won't be able to validate those without changing the schema. We're good.

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
czeslavo: Maybe it is _a bit_, but I like it anyway as it emphasizes why this ID could be useful and how. :+1:

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
Michal-Leszczynski: Well, SM does not check for negative parallel. Moreover, for negative parallel, it behaves as advertised - the repair gets stuck because no repair jobs can run... We should validate it on SM side in order to protect user from themselves.

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: ok, added

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
rzetelskik: added validation on our side

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
Michal-Leszczynski: >  I think the DC idea makes sense, but manager is extremely bad in terms of API guarantees.

That's true :c

But in terms of per DC options, I wouldn't go in this direction.
It might happen (even if it's unlikely) that we would like to extend this syntax in order to support `hostID:X` mapping or `shardCnt:X` mapping (e.g. set given rate limiting for nodes with given amount of shards). It could make sense since we are going in the direction of making heterogeneous clusters work better, and the per DC options might not be enough then.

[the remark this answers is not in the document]
rzetelskik: > It might happen (even if it's unlikely) that we would like to extend this syntax in order to support hostID:X mapping or shardCnt:X mapping

Wouldn't this be an orthogonal change anyway? As in it would be an extension of global options, orthogonal to per-DC options?

pkg/api/scylla/validation/scylladbmanagertask_validation_test.go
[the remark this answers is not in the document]
rzetelskik: Ok, I don't mind this, but I'd rather the entire package follows this, so I'd separate it from this PR. Is a followup PR ok?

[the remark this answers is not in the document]
czeslavo: Sure!

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
mflendrich: Three answers to three problems being considered here.

------------

> [wdyt? About] @czeslavo's proposed changes to the API

Based on the assumption that this PR intends to pass `UploadParallel` as an opaque string to Manager - Operator will not be building or interpreting this string. Please keep me honest here.

- This opaque string is a part of Manager's public API. Is the cost of inventing a separate API with translation, testing, reacting to changes justified?
- Do we expect users to be familiar with Manager's language to use our backup task definition language?

My gut call for the first bullet point is "no" and for the second one is "yes". Which leads me to a recommendation "this API is good enough if it's a Manager thing, let Manager be responsible for its style".

------------

> strenghtening the validation in general

I see it as a balancing act 

- paying the cost of support tickets / writing documentation 
- user sadness over misconfiguration (failure mode, prevalence)
- failure mode that @rzetelskik pointed out (flooding leading to rate throttling) that looks bad enough

vs

- the failure mode in case the string format deviates in the future from our validation
- the maintenance effort to keep the validator complete

My favorite thing in the world would be if Manager rejected invalid strings and we could propagate this back to the user.
If we can't have that - I kinda feel that having this validation is a cost worth paying because the pain will be real otherwise, but take this with a grain of salt because of my shallow context.

------------

> The API was already agreed on in the enhancement proposal so I'm reluctant to making any changes here if I'm being honest

This is more of a general comment about this sentiment, not applicable to this PR: I believe that API format is among the costliest things to make a mistake in, even costlier than pushing ugly code. Unless the API is released at some promise of stability, if we have a compelling reason to make the API less saddening to use, the cost of doing that (and rewriting some small portion of the code) feels worth paying, over dealing with an API we were wrong about.

Note that earlier in this comment I am tipping the balance against changing the originally designed API. What I'm trying to do here is demote _"if something is part of an approved design doc, it's set in stone"_ and promote _"let's iterate on the design as we identify problems along the way"_.

If our process makes it harder than necessary to iterate over design, let's fix the process.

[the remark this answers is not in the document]
rzetelskik: > Which leads me to a recommendation "this API is good enough if it's a Manager thing, let Manager be responsible for its style".

It is a manager thing, but bear in mind manager's API is dictated mostly by their CLI, so I assume this is what shaped the server side. @czeslavo's proposal feels much more k8s-like and a strongly typed struct does make the API more self-explanatory, provided the users are not already expecting the CLI-like formatting.

> failure mode that @rzetelskik pointed out (flooding leading to rate throttling) that looks bad enough

This is something we can limit, to an extent, by making errors nonretriable in case of the manager client returning 400s. 

> My favorite thing in the world would be if Manager rejected invalid strings and we could propagate this back to the user.

We can to an extent - we propagate the failures on the manager server side to the object's status (through degraded conditions). What we're considering here is adding best-effort syntax validation to the webhook. Receiving an error from kube's API server on object creation is a much better experience than checking the status of an already created object for errors propagated from manager client.
We can't validate whether the DCs/keyspaces etc make sense for the registered cluster without going into cross-validation, but with a reasonably low cost we can validate the syntax.

> If we can't have that - I kinda feel that having this validation is a cost worth paying because the pain will be real otherwise

Agreed, given we won't get a chance to strengthen the validation later, I feel like it's better to pay the price now and do this preemptively. So it seems like we all agree on strengthening the validation, what's left to be decided is the API.

rzetelskik:
- Manager client errors are now stripped from non-idempotent info to avoid status update hot loops.
- `BadRequest` error responses from manager client are now non-retriable to avoid requeuing on validation errors.

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: @zimnx ping

rzetelskik:
-  Rebased

rzetelskik:
- Rebased

rzetelskik:
- Fixed a nil access bug in recently added logs
- Added progressing conditions on task creation/update

rzetelskik:
- Made e2e validation of task propagation to ScyllaDB Manager state stricter

pkg/controller/scylladbmanagertask/conditions.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2025 ScyllaDB
+
+package scylladbmanagertask
+
+const (
+	managerControllerProgressingCondition            = "ManagerControllerProgressing"
+	managerControllerDegradedCondition               = "ManagerControllerDegraded"
+	scyllaDBManagerTaskFinalizerProgressingCondition = "ScyllaDBManagerTaskFinalizerProgressing"
+	scyllaDBManagerTaskFinalizerDegradedCondition    = "ScyllaDBManagerTaskFinalizerDegraded"
```
czeslavo, line 9: Wouldn't it make sense to include these constants in the public API, so that if someone relies on our conditions, they don't have to hardcode those? Or do we treat these conditions as "private" and expect users to only rely on the top-level ones (`Progressing`/`Available`/`Degraded`)?

pkg/controller/scylladbmanagertask/sync.go
```diff
@@ -0,0 +1,146 @@
...
+			return nil
+		}
+
+		return fmt.Errorf("can't get ScyllaDBManagerTask %q: %w", naming.ManualRef(namespace, name), err)
+	}
+
+	status := smtc.calculateStatus(smt)
+
+	if smt.DeletionTimestamp != nil {
+		err = controllerhelpers.RunSync(
```
czeslavo, line 50: Shouldn't we check if the error is not nil?

pkg/controller/scylladbmanagertask/controller.go
```diff
@@ -0,0 +1,253 @@
...
+		scyllaDBManagerClusterRegistrationLister: scylladbManagerClusterRegistrationInformer.Lister(),
+
+		cachesToSync: []cache.InformerSynced{
+			scyllaDBManagerTaskInformer.Informer().HasSynced,
+			scylladbManagerClusterRegistrationInformer.Informer().HasSynced,
+		},
+
+		eventRecorder: eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "scylladbmanagertask-controller"}),
+
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "scylladbmanagertask"),
```
czeslavo, line 90: Nit: please note `NewNamedRateLimitingQueue` is deprecated in favor of `NewTypedRateLimitingQueueWithConfig`. I know there's a lot of use of it across the codebase, so it may be worth tackling them in one separate PR.

pkg/controller/scylladbmanagertask/controller.go
```diff
@@ -0,0 +1,253 @@
...
+	scylladbManagerClusterRegistrationInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    smtc.addScyllaDBManagerClusterRegistration,
+		UpdateFunc: smtc.updateScyllaDBManagerClusterRegistration,
+		DeleteFunc: smtc.deleteScyllaDBManagerClusterRegistration,
+	})
+
+	return smtc, nil
+}
+
+func (smtc *Controller) processNextItem(ctx context.Context) bool {
```
czeslavo, line 127: I believe we could benefit from extracting the controllers' boilerplate (`processNextItem`, `Run`) to a reusable generic struct. By copying it for each new controller (and potentially slightly modifying it), we make testing this logic harder because it's widely spread in different packages. Having this covered only in E2E is not enough IMHO, as it's hard to trigger every single branch that we handle here.

I'm not sure if this requires action in this PR (most likely not), but I wanted to surface my concerns.

pkg/controller/scylladbmanagertask/controller.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+const (
+	ControllerName = "ScyllaDBManagerTaskController"
+
+	// maxSyncDuration enforces preemption. Do not raise the value! Controllers shouldn't actively wait,
+	// but rather use the queue.
+	// Unfortunately, Scylla Manager calls are synchronous, internally retried and can take ages.
+	// Contrary to what it should be, this needs to be quite high.
+	maxSyncDuration = 2 * time.Minute
```
czeslavo, line 42: What's the _expected_ response time for the manager calls that we make in this controller? How often may it happen that they will take close to 2 minutes? Does it happen only in error conditions?

If ~2min call times are expected in success scenarios, it might be worth considering extracting this to a background worker that would notify the controller via queue on finish to avoid blocking the controller (to not reduce the controller's throughput). This way we could also surface that the call is happening in the background via a condition.

pkg/controller/scylladbmanagertask/conditions.go
rzetelskik: > expect users to only rely on the top-level ones (Progressing/Available/Degraded)

yes across the board, excluding some legacy ones that we maintain

pkg/controller/scylladbmanagertask/sync.go
rzetelskik: definitely, good catch

pkg/controller/scylladbmanagertask/controller.go
rzetelskik: > so it may be worth tackling them in one separate PR

agreed, do you think that's a prerequisite or should we handle this separately?

pkg/controller/scylladbmanagertask/controller.go
rzetelskik: I agree with the sentiment, but that's quite a bit of an effort and I wouldn't tie this to this PR

pkg/controller/scylladbmanagertask/controller.go
rzetelskik: The value is taken from the existing, standalone manager-controller implementation, where it's covering multiple calls in the same iteration - so all operations on a ScyllaCluster would fall into the single timeout. Now this is only supposed to cover a single task, which comes down to a single call in an iteration. 
My idea for this is was start with this value to reduce this PR to the minimum changes necessary and establish a correctness baseline. Once we do this, we should definitely start lowering the value. 2 minutes does seem totally over the top though, so maybe there is no point in starting that high. Wdyt?

pkg/controller/scylladbmanagertask/sync.go
rzetelskik: fixed

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
zimnx: Agreed - sticking to the Manager style API makes a lot of sense, especially since it's documented in their documentation and closely aligned with how Manager tasks work. 

> But in terms of per DC options, I wouldn't go in this direction. It might happen (even if it's unlikely) that we would like to extend this syntax in order to support `hostID:X` mapping or `shardCnt:X` mapping (e.g. set given rate limiting for nodes with given amount of shards). It could make sense since we are going in the direction of making heterogeneous clusters work better, and the per DC options might not be enough then.

This is great example why we should just pass fields directly to Manager and let it be responsible for their semantic. They may add more syntax/semantic from version to version, so we can't forbid tasks not having particular prefix. So let's leave format and semantic to Manager.

On validation, I’m in favor of proactively rejecting clearly invalid objects. Catching issues early avoids unnecessary noise and failed attempts at sync. Early validation immediately informs the user about the problem. If a task is accepted but later fails due to invalid data, the issue only becomes visible after the object is created - and only if the user checks the status, which might not happen right away. Reacting early gives users a much faster and clearer feedback loop.

Although I think we don’t need to go crazy - checking for things like an unknown DC (which we can’t infer directly from the task spec) doesn’t seem practical. But if something is definitely invalid per the Manager spec, it’s better to block it upfront.

pkg/controller/scylladbmanagertask/controller.go
czeslavo: I think it will be better to tackle that separately. :+1:

pkg/controller/scylladbmanagertask/controller.go
czeslavo: I think the best would be to lower this based on some data that we may gather once it's shipped. Given we don't know the right limit, it might be better to keep it higher for now to not disturb users. Perhaps we could create an issue to address this once we gain more insight?

pkg/controller/scylladbmanagertask/controller.go
czeslavo: I've created an issue to track this: https://github.com/scylladb/scylla-operator/issues/2685

rzetelskik: thanks, I'll add a comment with the issue ref

pkg/controller/scylladbmanagertask/controller.go
rzetelskik: I've created a tracker (https://github.com/scylladb/scylla-operator/issues/2686) and added it in a comment. I believe this is a right way to approach this, especially that it's already made its way into ScyllaDBManagerClusterRegistration controller.

pkg/controller/scylladbmanagertask/controller.go
rzetelskik: done

pkg/controller/scylladbmanagertask/controller.go
czeslavo: I agree, it's a separate problem. I've created an issue to track it: https://github.com/scylladb/scylla-operator/issues/2687

pkg/controller/scylladbmanagertask/sync_finalizer.go
```diff
@@ -0,0 +1,122 @@
...
+		progressingConditions = append(progressingConditions, metav1.Condition{
+			Type:               scyllaDBManagerTaskFinalizerProgressingCondition,
+			Status:             metav1.ConditionTrue,
+			ObservedGeneration: smt.Generation,
+			Reason:             "DeletedScyllaDBManagerClientTask",
+			Message:            "Deleted the task from ScyllaDB Manager state.",
+		})
+
+		return progressingConditions, nil
+	}
```
czeslavo: Nit: if we flipped the if condition here, we'd have the indented branch shorter, which is preferred, leaving the "happy path" unindented.

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+
+		status.TaskID = pointer.Ptr(managerTaskID.String())
+		progressingConditions = append(progressingConditions, metav1.Condition{
+			Type:               managerControllerProgressingCondition,
+			Status:             metav1.ConditionTrue,
+			ObservedGeneration: smt.Generation,
+			Reason:             "CreatedScyllaDBManagerTask",
+			Message:            fmt.Sprintf("Created a ScyllaDB Manager task: %s (%s).", requiredManagerTask.Name, managerTaskID),
+		})
+		return progressingConditions, nil
```
czeslavo: For readability's sake (it's indentation-heavy and we return at the end of the branch), I think it could make sense to extract this to a helper method, something like `handleManagerClientTaskNotFound`.

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+		}
+
+		progressingConditions = append(progressingConditions, metav1.Condition{
+			Type:               managerControllerProgressingCondition,
+			Status:             metav1.ConditionTrue,
+			ObservedGeneration: smcr.Generation,
+			Reason:             "DeletedCollidingScyllaDBManagerTask",
+			Message:            fmt.Sprintf("Deleted a colliding ScyllaDB Manager task with no OwnerUID label: %s (%s).", managerTask.Name, managerTaskID),
+		})
+		return progressingConditions, nil
```
czeslavo: Ditto - this could be extracted to a helper method, e.g., `handleManagerTaskMissingOwnerUID`.

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+
+	idx := slices.IndexFunc(tasks.TaskListItemSlice, func(item *managerclient.TaskListItem) bool {
+		return item.Name == taskName
+	})
+
+	if idx >= 0 {
+		return tasks.TaskListItemSlice[idx], true, nil
+	}
+
+	return nil, false, nil
```
czeslavo: I think that flipping the condition here would be more aligned with "The happy path is left-aligned" approach.

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+}
+
+type ScyllaDBManagerClientTaskOverrideOption func(*scyllav1alpha1.ScyllaDBManagerTask, *managerclient.Task)
```
czeslavo: Both this and `WithScheduleStartDateNowSyntaxRetention` are public but are only used in this package, should we make them private?

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+
+func makeScyllaDBManagerClientTask(smt *scyllav1alpha1.ScyllaDBManagerTask, clusterID string, overrideOptions ...ScyllaDBManagerClientTaskOverrideOption) (*managerclient.Task, error) {
+	var err error
+	var managerClientTaskType string
+
+	managerClientTaskName := scyllaDBManagerClientTaskName(smt)
+	managerClientTaskSchedule := &managerclient.Schedule{}
+	managerClientTaskProperties := map[string]any{}
+
+	var scheduleOverrideOptions []ScyllaDBManagerClientScheduleOverrideOption
```
czeslavo: Aggregating these declarations in a single block might make sense, WDYT?
```suggestion
	var (
		err error
		managerClientTaskType string
		
		managerClientTaskName = scyllaDBManagerClientTaskName(smt)
		managerClientTaskSchedule = &managerclient.Schedule{}
		managerClientTaskProperties = map[string]any{}
		
		scheduleOverrideOptions []ScyllaDBManagerClientScheduleOverrideOption
	)
```

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+	managedHash, err := hashutil.HashObjects(requiredManagerTask)
+	if err != nil {
+		return nil, fmt.Errorf("can't calculate managed hash: %w", err)
+	}
+	requiredManagerTask.Labels[naming.ManagedHash] = managedHash
+
+	return requiredManagerTask, nil
+}
+
+type ScyllaDBManagerClientScheduleOverrideOption func(*managerclient.Schedule) error
```
czeslavo: Ditto - make it private?

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+}
+
+// parseStartDate parses the start date string into a strfmt.DateTime.
+// The function handles two formats:
+// 1. "now[+duration]", e.g. "now+1h", "now+2d". For "now" or "now+duration", where duration is equivalent to 0, it returns an empty DateTime deliberately to indicate the current time should be resolved by the server.
+// 2. RFC3339 formatted timestamp (e.g., "2023-05-08T17:24:00Z").
+// The nowUTCFunc parameter is a function that returns the current time in UTC.
+func parseStartDate(s string, nowUTCFunc func() time.Time) (strfmt.DateTime, error) {
+	if strings.HasPrefix(s, "now") {
+		now := nowUTCFunc()
```
czeslavo:
```suggestion
func parseStartDate(s string, nowFn func() time.Time) (strfmt.DateTime, error) {
	if strings.HasPrefix(s, "now") {
		now := nowFn().UTC()
```
If we changed it this way, we'd make this function immune to passing functions that do not return UTC. I think it makes sense as the TZ info is embedded in `time.Time` anyways.

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+
+		if s == "now" {
+			return strfmt.DateTime{}, nil
```
czeslavo, line 412: Is the empty struct an expected value to make the task executed "now"?

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+
+	if options.UploadParallel != nil {
+		managerClientTaskProperties["upload_parallel"] = options.UploadParallel
+	}
+
+	return managerClientTaskProperties, nil
+}
+
+type ScyllaDBManagerClientPropertiesOverrideOption func(map[string]any) error
```
czeslavo: Ditto - make private?

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -0,0 +1,608 @@
...
+
+		if managerTask.Schedule == nil {
+			managerTask.Schedule = &managerclient.Schedule{}
+		}
+
+		managerTask.Schedule.StartDate = pointer.Ptr(existingStartDate)
+	}
+}
+
+func makeScyllaDBManagerClientTask(smt *scyllav1alpha1.ScyllaDBManagerTask, clusterID string, overrideOptions ...ScyllaDBManagerClientTaskOverrideOption) (*managerclient.Task, error) {
```
czeslavo: Thought: `makeScyllaDBManagerClientTask` and all its dependencies would be a good candidate for a separate PR. We could implement it in a separate, helper package, test it there, just not plug it into the controller yet. (no action required)

pkg/controllerhelpers/scylladbmanager.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/scylladb/scylla-manager/v3/pkg/managerclient"
+	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func GetScyllaDBManagerClient(_ context.Context, _ *scyllav1alpha1.ScyllaDBManagerClusterRegistration) (*managerclient.Client, error) {
+	url := fmt.Sprintf("http://%s.%s.svc/api/v1", naming.ScyllaManagerServiceName, naming.ScyllaManagerNamespace)
+	managerClient, err := managerclient.NewClient(url, func(httpClient *http.Client) {
+		httpClient.Transport = http.DefaultTransport
```
czeslavo, line 21: Using global `http.DefaultTransport` might be error-prone. I suggest cloning it. If we want to share the transport between all manager clients (i.e., to cache connections), we might create our own global in this package and use it solely for that purpose.
```suggestion
		httpClient.Transport = http.DefaultTransport.(*http.Transport).Clone()
```

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: seems purely style related

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: yes, this is almost an exact copy of the existing implementation https://github.com/scylladb/scylla-operator/blob/cbb1cbe08fd018d2dda28d972668a8f0cbdfb415/pkg/controller/manager/types_old.go#L309-L334

pkg/controllerhelpers/scylladbmanager.go
rzetelskik: This code is only moved, not added in this PR. It comes directly from the existing implementation (standalone manager-controller). I suggest we handle it separately, similar to max sync duration etc. We could also reevaluate why we're not using the default manager client config.

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: good idea

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: I don't think any of our guidelines mentions it as a preferred way to declare variables either, but if so, please correct me

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: Looks like we have a quorum - the API is staying as is. I'll add syntax validation, with obligatory exclusion annotations for backward compatibility, where it's possible without cross-resource checks.

pkg/controller/scylladbmanagertask/sync_manager.go
czeslavo: Yes, it's purely style-related. If you don't agree, go ahead and leave it as is. :)

pkg/api/scylla/v1alpha1/types_scylladbmanagertask.go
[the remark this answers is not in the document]
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_finalizer.go
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: leaving as is then

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controllerhelpers/scylladbmanager.go
rzetelskik: created https://github.com/scylladb/scylla-operator/issues/2693 and added a ref

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

rzetelskik:
/hold
looks like some of the annotation names are too long, I'll have to edit that

czeslavo:
/lgtm
:rocket:

PR #2646, opened 2025-05-19
Document ScyllaCluster volume expansion procedure

### LINKED ISSUE #2453: Document and test storage resizing on the same nodes
_Issue originally authored by **tnozicka** as https://github.com/scylladb/scylla-operator/issues/825_

Given [StatefulSet don't reconcile storage changes](https://github.com/kubernetes/kubernetes/issues/68737) we can't yet easily handle it in the operator seamlessly and we should at least document a manual procedure. Something in the lines of:
- orphan delete the scyllacluster
- orphan delete the statefulsets
- create the scyllacluster with higher storage request
- resize the existing PVC, if the storage class supports it, replace the nodes if it doesn't
- trigger a scyllacuster rollout

zimnx:

Due to limitations of the underlying StatefulSet, ScyllaClusters don't allow storage changes. The following procedure describes how to adjust the storage **manually**.

Resolves #2453


docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
```diff
@@ -0,0 +1,69 @@
...
+Let's use the following ScyllaCluster as an example.
+```bash
+$ kubectl get scyllacluster
+NAME     READY   MEMBERS   RACKS   AVAILABLE   PROGRESSING   DEGRADED   AGE
+scylla   3       3         1       True        False         False      62m
+```
+
+## Increasing the storage size
+
+1. Since we are going to delete the ScyllaCluster object, make sure to have a copy of the existing one. You will need it later.
```
czeslavo, line 24: How about adding an explicit command that can be used to dump the current state of the ScyllaCluster?

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
```diff
@@ -0,0 +1,69 @@
...
+
+2. Orphan delete ScyllaCluster
```
czeslavo, line 26: I think we could be more explicit and tell why we orphan-delete instead of cascading  (I believe it's because we'd like to keep manual control over what objects are deleted, as we want to preserve some of the ScyllaCluster's children). I suspect it may not be obvious to the reader.

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
```diff
@@ -0,0 +1,69 @@
...
+    
+    persistentvolumeclaim/data-scylla-us-east-1-a-0 patched
+    ```
+   
+7. Update saved ScyllaCluster definition with the new storage size `spec.datacenter.racks[*].storage.capacity` field, and apply it:
+   ```bash
+   kubectl apply --server-side -f scyllaClusterDefinition.yaml 
+   ```
+   
+   The StatefulSets will be recreated by the Scylla Operator with the new storage size.
```
czeslavo, line 66:
```suggestion
   The ScyllaDBDatacenter and StatefulSets will be recreated by the Scylla Operator with the new storage size.
```

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
```diff
@@ -0,0 +1,69 @@
...
+8. Depending on your storage provisioner, affected Pods may or may not require a restart.
+   Verify if storage was resized by checking volume size from within the affected Pod.
```
czeslavo, line 68: How about adding an explicit command here? E.g. `kubectl exec -it <pod-name> -- df -h` .

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
zimnx: done

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
zimnx: done

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
czeslavo: But people reading it will have to do it anyway, and it might not be obvious to them how to achieve that, no? Maybe something like below could cover this?

```bash
kubectl get scyllacluster scylla -o json | jq 'del(
    .metadata.creationTimestamp,
    .metadata.generation,
    .metadata.uid,
    .metadata.resourceVersion,
    .status
)' | yq -P > scyllaClusterDefinition.yaml
```

zimnx: It's obvious if you're using gitops, where you keep either helm values or scylla cluster spec in repository. Then it's enough to trigger resync to recreate the object.

However not everybody does, so maybe it's better to include this step. 
Added.

zimnx:
I'll split commands and outputs into separate blocks to allow for easier copying. 
/hold

docs/source/resources/scyllaclusters/nodeoperations/volume-expansion.md
czeslavo: @rzetelskik :+1: You're right, the following works as well and doesn't require jq.
```shell
kubectl get scyllacluster scylla -o yaml | yq 'del(
    .metadata.creationTimestamp,
    .metadata.generation,
    .metadata.uid,
    .metadata.resourceVersion,
    .status
)' > scyllaClusterDefinition.yaml
```

zimnx: good idea, done

czeslavo (approved):
/lgtm

PR #2660, opened 2025-05-28
Merge the Dependabot schedules

mflendrich:
Because k8s.io version bumps are less painful than originally expected, it no longer makes sense to bump them separately from the rest of the go modules.

This PR therefore merges the two package groups into one.

Also changes the schedule so that a PR should be ready when maintainers in Europe show up to work on Wednesday.

/cc rzetelskik
/kind machinery
/priority important-longterm


rzetelskik:
/lgtm
Thanks!

PR #2668, opened 2025-06-02
Extend ScyllaDBCluster Webhook E2E with update being declined

zimnx:
E2E is extended to validate if update of existing ScyllaDBCluster changing immutable field is denied.

### DIFF
--- a/test/e2e/set/scylladbcluster/scylladbcluster_webhook.go
+++ b/test/e2e/set/scylladbcluster/scylladbcluster_webhook.go
@@ -6,19 +6,21 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/go-cmp/cmp"
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	"github.com/scylladb/scylla-operator/test/e2e/framework"
 	"github.com/scylladb/scylla-operator/test/e2e/utils"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apiserver/pkg/storage/names"
 )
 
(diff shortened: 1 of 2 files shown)


test/e2e/set/scylladbcluster/scylladbcluster_webhook.go
```diff
@@ -50,10 +53,43 @@ var _ = g.Describe("ScyllaDBCluster webhook", func() {
...
-		}}))
+		}}
+
+		o.Expect(err).To(o.Equal(expectedValidationError), cmp.Diff(err, expectedValidationError))
 
 		framework.By("Accepting a creation of valid ScyllaCluster")
 		validSC, err = f.ScyllaClient().ScyllaV1alpha1().ScyllaDBClusters(f.Namespace()).Create(ctx, validSC, metav1.CreateOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Rejecting an update of ScyllaDBCluster's clusterName")
```
czeslavo, line 64: Question to ensure I understand correctly the intention: this is just an arbitrarily picked validation rule that we trigger to ensure that `Update` validation is properly wired up and works from the API perspective, right? We do not intend to cover all possible validation rules in E2E, as that's done on the unit test level.

If yes, could we add a comment explaining this intention here?

BTW I can't find this case covered in [unit tests](https://github.com/scylladb/scylla-operator/blob/d89e58bf157925537e8bcd9275fce234b3a316ab/pkg/api/scylla/validation/scylladbcluster_validation_test.go#L2533), shall we add it?

zimnx: yes, that's the intention. I changed test name to better reflect it, and added unit test as it's indeed missing.

czeslavo (approved):
/lgtm

zimnx:
This contains only changes in unit tests and non-multi-datacenter suite. multi-datacenter suite is currently broken, i'll override it here and will try to fix it in separate PR. 

PR #2669, opened 2025-06-03
Enable importas linter

czeslavo:

Enable `importas` linter to enforce some packages to be always imported with a predefined alias.

This is good for keeping the aliases consistent across the codebase as it reduces cognitive load when an alias has to be defined, and makes reading code easier in the long run.

Note to reviewers: I picked some common packages, but if you think there's more that we could enforce, I'm open to adding such.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -4,11 +4,31 @@ linters:
   enable:
     - unused
     - govet
+    - importas
     # More linters can be added, see supported ones: https://golangci-lint.run/usage/linters/
   exclusions:
     presets:
       - common-false-positives
       - std-error-handling
+  settings:
+    importas:
+      no-unaliased: true
+      alias:
+        # # e.g. k8s.io/api/core/v1 -> corev1
+        - pkg: k8s.io/api/(\w+)/(v[\w\d]+)
+          alias: ${1}${2}
+        # e.g. github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1 -> scyllav1alpha1
+        - pkg: github.com/scylladb/scylla-operator/pkg/api/(\w+)/(v\d+[\w\d]*)
+          alias: ${1}${2}
+        # e.g. k8s.io/apimachinery/pkg/util/errors -> apimachineryutilerrors
+        - pkg: k8s.io/apimachinery/pkg/util/(\w+)
+          alias: apimachineryutil${1}
+        - pkg: k8s.io/apimachinery/pkg/apis/meta/v1
+          alias: metav1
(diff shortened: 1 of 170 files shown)


.golangci.yaml
```diff
@@ -4,11 +4,29 @@ linters:
...
+      alias:
+        # # e.g. k8s.io/api/core/v1 -> corev1
+        - pkg: k8s.io/api/(\w+)/(v[\w\d]+)
+          alias: ${1}${2}
+        # e.g. github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1 -> scyllav1alpha1
+        - pkg: github.com/scylladb/scylla-operator/pkg/api/(\w+)/(v\d+[\w\d]*)
+          alias: ${1}${2}
+        # e.g. k8s.io/apimachinery/pkg/util/errors -> k8serrors
+        - pkg: k8s.io/apimachinery/pkg/util/(\w+)
+          alias: k8s${1}
```
rzetelskik: I don't like this one, there are many different ways of aliasing these in our repo, but none similar to this, and it doesn't seem specific enough. The dominant ones are:
- `apimachineryutilvalidation "k8s.io/apimachinery/pkg/util/validation"`
- `apierrors "k8s.io/apimachinery/pkg/util/errors"`
- `utilruntime "k8s.io/apimachinery/pkg/util/runtime"` or `utilerrors "k8s.io/apimachinery/pkg/util/errors"` 

I'd personally stick with `apimachineryutil${1}` then. @zimnx what's your opinion on this?

.golangci.yaml
```diff
@@ -4,11 +4,29 @@ linters:
...
+        - pkg: k8s.io/apimachinery/pkg/apis/meta/v1
```
rzetelskik, line 26: nit: `apimachinerymetav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"` (or metav1validation) might also be worth adding to set up a pattern

.golangci.yaml
zimnx: IMO k8s is too generic. I would use `apimachineryutil${1}` for these.

czeslavo: Agreed, changed to `apimachineryutil${1}`. :+1:

.golangci.yaml
czeslavo: :+1: Added with the `metav1validation` alias.

zimnx:
multidc suite is broken, these changes doesn't affect it, so overriding.

czeslavo:
Rebased onto master after resolving conflicts.

PR #2672, opened 2025-06-04
Fix NonRetriableError matching

rzetelskik:
**Description of your changes:** This PR fixes the function matching non-retriable errors. `errors.Is` is meant for comparing with sentinel errors as it checks for equality. The NonRetriableError, being a wrapper, fails the equality test.

The PR also implements `Unwrap()` method for NonRetriableError to allow errors.As/errors.Is to be (correctly) used for wrapped errors.

Resolves #

/kind bug
/priority important-soon

### DIFF
--- a/pkg/controller/scylladbmanagerclusterregistration/controller.go
+++ b/pkg/controller/scylladbmanagerclusterregistration/controller.go
@@ -168,13 +168,12 @@ func (smcrc *Controller) processNextItem(ctx context.Context) bool {
 	case apierrors.IsAlreadyExists(err):
 		klog.V(2).InfoS("Hit already exists, will retry in a bit", "Key", key, "Error", err)
 
-	default:
-		if controllertools.IsNonRetriable(err) {
-			klog.InfoS("Hit non-retriable error. Dropping the item from the queue.", "Error", err)
-			smcrc.queue.Forget(key)
-			return true
-		}
+	case controllertools.IsNonRetriable(err):
+		klog.InfoS("Hit non-retriable error. Dropping the item from the queue.", "Error", err)
+		smcrc.queue.Forget(key)
+		return true
 
+	default:
 		utilruntime.HandleError(fmt.Errorf("syncing key '%v' failed: %v", key, err))
 
 	}
--- a/pkg/controllertools/errors.go
+++ b/pkg/controllertools/errors.go
@@ -8,6 +8,10 @@ type NonRetriableError struct {
 	error
 }
 
+func (e NonRetriableError) Unwrap() error {
+	return e.error
+}
+
 func NonRetriable(err error) error {
 	return NonRetriableError{err}
 }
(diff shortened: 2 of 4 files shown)


zimnx (approved):
/lgtm

rzetelskik:
/hold
I'm questioning now what the expected behaviour should actually be.
E.g. if a non-retriable error is a part of an error aggregate (`errors.Join(nonRetriableErr, regularErr)`), what should `IsNonRetriable` on the aggregate return?

czeslavo:
> /hold I'm questioning now what the expected behaviour should actually be. E.g. if a non-retriable error is a part of an error aggregate (`errors.Join(nonRetriableErr, regularErr)`), what should `IsNonRetriable` on the aggregate return?

I'd say one non-retryable error makes the whole batch non-retryable (why: retryable errors might resolve eventually, but the non-retryable one will remain).

rzetelskik:
> I'd say one non-retryable error makes the whole batch non-retryable (why: retryable errors might resolve eventually, but the non-retryable one will remain).

I'm a bit worried about potential misuse with different scopes though. Say you return a non-retriable error in one reconciliation loop, while the other reconciliation loop could potentially fix it. Although this is clearly an invalid use of the non-retriable error, wouldn't this be somewhat hard to catch?

czeslavo:
> I'm a bit worried about potential misuse with different scopes though. Say you return a non-retriable error in one reconciliation loop, while the other reconciliation loop could potentially fix it. Although this is clearly an invalid use of the non-retriable error, wouldn't this be somewhat hard to catch?

In such situations where another controller can fix a non-retryable condition in another one, we should ensure it gets notified to rerun the reconciliation that was halted due to the non-retryable error. And I believe in most cases that should be covered by watching resources related to the controller's resource, no?

rzetelskik:
> In such situations where another controller can fix a non-retryable condition in another one, we should ensure it gets notified to rerun the reconciliation that was halted due to the non-retryable error. And I believe in most cases that should be covered by watching resources related to the controller's resource, no?

My bad - I meant different functions inside a single reconciliation loop. But in this case we're going into defence against human errors.

czeslavo:
lgtm :+1:

rzetelskik:
> 
> ci/prow/e2e-gke-serial	[9f14394](https://github.com/scylladb/scylla-operator/commit/9f1439421f4ef40dcaad1716db3e45be7527d10a)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2672/pull-scylla-operator-master-e2e-gke-serial/1930264339353178112)	true	`/test e2e-gke-serial`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2672). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

known flake
/test images
/retest

zimnx (approved):
/lgtm

Issue #2674, opened 2025-06-04
doc: add redirections for pages that are moved or deleted


annastuchlik:
**Problem**
When you change a URL of a page in the Operator documentation by changing the documentation structure, moving, or removing pages, the use gets a 404 when trying to access the previous URL. 
See this example in the Manager docs: https://github.com/scylladb/scylla-manager/issues/4389

**Fix**
Every time you move or remove a page, you should add a redirection to this file:
https://github.com/scylladb/scylla-operator/blob/master/docs/_utils/redirects.yaml

See the [redirections in the scylladb repo ](https://github.com/scylladb/scylladb/blob/master/docs/_utils/redirects.yaml)as reference.

**Todo**
- If possible, add the missing redirections (if you can identify them). At a minimum, add a redirection specified in https://github.com/scylladb/scylla-manager/issues/4389 (it was reported by a customer).
- If impossible to identify all missing redirections, add as many as you can and remember to add them in the future. I'll be reporting individual cases once I spot them in other documentation sets.

rzetelskik:
> Problem
> When you change a URL of a page in the Operator documentation by changing the documentation structure, moving, or removing pages, the use gets a 404 when trying to access the previous URL.
> See this example in the Manager docs: https://github.com/scylladb/scylla-manager/issues/4389

The real problem here seems to be the fact that this doesn't cause any errors when building the docs. @annastuchlik don't we have any tools that could check for dead links?

annastuchlik:
> > Problem
> > When you change a URL of a page in the Operator documentation by changing the documentation structure, moving, or removing pages, the use gets a 404 when trying to access the previous URL.
> > See this example in the Manager docs: [scylladb/scylla-manager#4389](https://github.com/scylladb/scylla-manager/issues/4389)
> 
> The real problem here seems to be the fact that this doesn't cause any errors when building the docs. [@annastuchlik](https://github.com/annastuchlik) don't we have any tools that could check for dead links?

@rzetelskik The real problem is missing redirections. The old links may be used not only in the docs, but also on our website or other resources, such as ScyllaDB blog or forum. In addition, they may be used by existing users who have them saved, e.g., as bookmarks.
We don't have control over all possible resources, so redirections are critical.

Link checking is another issue. We do have tools, but they are not fully automated - I'll look into how it can be improved.
Still, the only way to detect a dead link (to Operator in other tool) in the Manager docs is to run link check on the Manager docs.

rzetelskik:
[...]

PR #2675, opened 2025-06-04
Add redirections for pages that are moved or deleted

ylebi:
Add redirections for old restore and manager URLs. 
We probably have more than those two, so feel free to open new issue when you catch it. 

Resolves https://github.com/scylladb/scylla-operator/issues/2674

### DIFF
--- a/docs/_utils/redirects.yaml
+++ b/docs/_utils/redirects.yaml
@@ -2,8 +2,8 @@
 #old path: new path
 #
 
-/stable/resources/scyllaclusters/nodeoperations/restore.html: https://operator.docs.scylladb.com/stable/resources/scyllaclusters/nodeoperations/restore.html
-/stable/manager.html: https://operator.docs.scylladb.com/stable/architecture/manager.html
+/stable/resources/scyllaclusters/nodeoperations/restore.html: /stable/resources/scyllaclusters/nodeoperations/restore.html
+/stable/manager.html: /stable/architecture/manager.html
 
 # removing redirection html script files
 # test: /


docs/_utils/redirects.yaml
```diff
@@ -2,5 +2,8 @@
 #old path: new path
 #
 
+/stable/resources/scyllaclusters/nodeoperations/restore.html: https://operator.docs.scylladb.com/stable/resources/scyllaclusters/nodeoperations/restore.html
+/stable/manager.html: https://operator.docs.scylladb.com/stable/architecture/manager.html
```
czeslavo: I believe we could use relative links as targets.
```suggestion
/stable/resources/scyllaclusters/nodeoperations/restore.html: /stable/resources/scyllaclusters/nodeoperations/restore.html
/stable/manager.html: /stable/architecture/manager.html
```

ylebi: We could use both, see https://github.com/scylladb/scylladb/blob/master/docs/_utils/redirects.yaml. But I agree with you that it looks cleaner with relative links as targets. Committed.

annastuchlik:
@ylebi @czeslavo The redirection added in this PR is incorrect.  
This update redirects the page it itself, creating an endless loop:
```
/stable/resources/scyllaclusters/nodeoperations/restore.html: /stable/resources/scyllaclusters/nodeoperations/restore.html
```

As a result the page stopped, working:
https://operator.docs.scylladb.com/stable/resources/scyllaclusters/nodeoperations/restore.html

The correct redirection is:
```
/stable/nodeoperations/restore.html: /stable/resources/scyllaclusters/nodeoperations/restore.html
```

PR #2676, opened 2025-06-04
Strip ScyllaDB Manager client errors propagated to objects' statuses from non-idemponent information and unify log levels

rzetelskik:
**Description of your changes:** ScyllaDB Manager client can contain unique TraceIDs, which, when propagated to status conditions, may lead to status update hot loops, as the trace ID changes on every reconciliation, leading to subsequent updates. To prevent this, in this PR I add a method that allows for stripping the error of non-idempotent information. To retain the trace IDs, the errors are now also logged with debug level. The existing messages' levels are unified.

Resolves #

/kind bug
/priority important-soon


rzetelskik:
/hold
I'll add some unit tests

Issue #2677, opened 2025-06-04
doc: fix the redirection


annastuchlik:
Please fix the redirection added in https://github.com/scylladb/scylla-operator/pull/2675

> > @ylebi @czeslavo The redirection added in this PR is incorrect. This update redirects the page it itself, creating an endless loop:
> > 
> > ```
> > /stable/resources/scyllaclusters/nodeoperations/restore.html: /stable/resources/scyllaclusters/nodeoperations/restore.html
> > ```
> > 
> > As a result the page stopped, working: https://operator.docs.scylladb.com/stable/resources/scyllaclusters/nodeoperations/restore.html
> > 
> > The correct redirection is:
> > 
> > ```
> > /stable/nodeoperations/restore.html: /stable/resources/scyllaclusters/nodeoperations/restore.html
> > ```
>

PR #2678, opened 2025-06-04
Fix redirect loop in docs

czeslavo:
**Description of your changes:** Fixes the recently added redirect loop.

Resolves https://github.com/scylladb/scylla-operator/issues/2677.

### DIFF
--- a/docs/_utils/redirects.yaml
+++ b/docs/_utils/redirects.yaml
@@ -1,8 +1,7 @@
 ### a dictionary of redirections
 #old path: new path
 #
-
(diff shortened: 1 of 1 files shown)


czeslavo:
/hold
 ~I'm not sure if the second redirect shouldn't be `/stable/architecture/` instead of `/stable/architecture.html` as https://operator.docs.scylladb.com/stable/architecture/ is fine, while https://operator.docs.scylladb.com/stable/architecture.html returns 404.~ @annastuchlik 

Nevermind, it's https://operator.docs.scylladb.com/stable/architecture/manager.html, which is working fine.

PR #2679, opened 2025-06-05
Ensure runtime namespaces are collected regardless of ScyllaDBCluster rollout success

zimnx:
Previously, runtime namespaces created during ScyllaDBCluster rollout were only registered for collection after a successful rollout, making it hard to debug failed deployments.
This change introduces a NotFound-tolerant namespace collector and moves namespace registration earlier in the flow, ensuring they are always collected if created.

### DIFF
--- a/pkg/controller/scylladbcluster/resource.go
+++ b/pkg/controller/scylladbcluster/resource.go
@@ -44,15 +44,15 @@ func MakeRemoteRemoteOwners(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alph
 }
 
 func MakeRemoteNamespaces(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alpha1.ScyllaDBClusterDatacenter, managingClusterDomain string) ([]*corev1.Namespace, error) {
-	suffix, err := naming.GenerateNameHash(sc.Namespace, dc.Name)
+	name, err := naming.RemoteNamespaceName(sc, dc)
 	if err != nil {
-		return nil, fmt.Errorf("can't generate namespace name suffix: %w", err)
+		return nil, fmt.Errorf("can't get namespace name: %w", err)
 	}
 
 	return []*corev1.Namespace{
 		{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:   fmt.Sprintf("%s-%s", sc.Namespace, suffix),
+				Name:   name,
 				Labels: naming.ScyllaDBClusterDatacenterLabels(sc, dc, managingClusterDomain),
 			},
 		},
--- a/pkg/naming/names.go
+++ b/pkg/naming/names.go
@@ -305,3 +305,12 @@ func scyllaDBManagerClusterRegistrationName(kind, name string) (string, error) {
 	fullNameWithSuffix := fmt.Sprintf("%s-%s", fullName[:min(len(fullName), apimachineryutilvalidation.DNS1123SubdomainMaxLength-len(nameSuffix)-1)], nameSuffix)
 	return fullNameWithSuffix, nil
 }
+
+func RemoteNamespaceName(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alpha1.ScyllaDBClusterDatacenter) (string, error) {
+	suffix, err := GenerateNameHash(sc.Namespace, dc.Name)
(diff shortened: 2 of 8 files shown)


pkg/naming/names.go
```diff
@@ -305,3 +305,12 @@ func scyllaDBManagerClusterRegistrationName(kind, name string) (string, error) {
...
 }
+
+func RemoteNamespaceName(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alpha1.ScyllaDBClusterDatacenter) (string, error) {
+	suffix, err := GenerateNameHash(sc.Namespace, dc.Name)
+	if err != nil {
+		return "", fmt.Errorf("can't generate namespace name suffix: %w", err)
+	}
+
+	return fmt.Sprintf("%s-%s", sc.Namespace, suffix), nil
+}
```
rzetelskik, line 316: Imo it should share the code with existing `scyllaDBManagerClusterRegistrationName`. It should also be trimmed to max allowed length (not sure what it is in case of a namespace, probably the same), especially given https://github.com/scylladb/scylla-operator/issues/2722. A helper with variadic arguments + max length perhaps?

test/e2e/framework/cleanup.go
```diff
@@ -106,6 +106,31 @@ func NewNamespaceCleanerCollector(restConfig *rest.Config, client kubernetes.Int
...
+		},
+	}
+}
+
+func (nc *NotFoundTolerantNamespaceCollector) Collect(ctx context.Context, artifactsDir string, _ string) {
+	By("Collecting dumps from namespace %q.", nc.NamespaceName)
+
+	err := DumpNamespace(ctx, nc.RestConfig, cacheddiscovery.NewMemCacheClient(nc.Client.Discovery()), nc.DynamicClient, nc.Client.CoreV1(), artifactsDir, nc.NamespaceName)
+	if apierrors.IsNotFound(err) {
+		By("Namespace %q doesn't exists, it won't be collected.", nc.NamespaceName)
```
rzetelskik: g.By should idiomatically be used for describing specs, not logging

pkg/controller/scylladbcluster/resource.go
```diff
@@ -42,15 +42,15 @@ func MakeRemoteRemoteOwners(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alph
 }
 
 func MakeRemoteNamespaces(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alpha1.ScyllaDBClusterDatacenter, managingClusterDomain string) ([]*corev1.Namespace, error) {
-	suffix, err := naming.GenerateNameHash(sc.Namespace, dc.Name)
+	name, err := naming.RemoteNamespaceName(sc, dc)
```
rzetelskik, line 47: please adjust the error message

pkg/naming/names.go
zimnx: This doesn't aim to fix any issue other than remote namespace collection. I simply moved existing code to be shared across two usages. Leaving as is.

pkg/controller/scylladbcluster/resource.go
zimnx: done

test/e2e/framework/cleanup.go
zimnx: done

zimnx:
There's a panic in JustAfterEach, need to debug it.

/hold

zimnx:
Fixed variable initialization 
/hold cancel

pkg/naming/names.go
rzetelskik: Avoiding bugs while you're already creating the helper doesn't seem like much of a stretch imo, the existing code is not set in stone. But ok, let's handle this in #2722.

rzetelskik (approved):
/lgtm

pkg/naming/names.go
zimnx: It's clearly an existing limitation. If I would fix it here, limitation lift wouldn't be visible in release notes. Lets handle items iteratively, not block changes due to existing known bugs.

zimnx:
Known flakes, I'll override it because this change doesn't affect it, although it makes it easier to debug. I'll keep working on the fix in other PR.

PR #2681, opened 2025-06-05
[WIP] Migrate v1.ScyllaCluster to in-operator integration with global ScyllaDB Manager instance

rzetelskik:

Resolves #

/kind feature
/priority important-soon


pkg/cmd/operator/operator.go
```diff
@@ -301,6 +303,12 @@ func (o *OperatorOptions) Execute(ctx context.Context, streams genericclioptions
 }
 
 func (o *OperatorOptions) run(ctx context.Context, streams genericclioptions.IOStreams) error {
+	if ok, err := isStandaloneScyllaDBManagerControllerDeployed(ctx, o.kubeClient.AppsV1()); err != nil {
+		return fmt.Errorf("can't check for standalone ScyllaDB Manager controller presence: %w", err)
+	} else if ok {
+		return fmt.Errorf("Standalone ScyllaDB Manager controller Deployment %q should not be running alongside Scylla Operator. Delete it before starting Scylla Operator.", naming.ManualRef(naming.ScyllaManagerNamespace, naming.StandaloneScyllaDBManagerControllerName))
+	}
```
czeslavo, line 310: Should we create an issue to track dropping this check at some point, e.g., after 2 minor releases or so?

pkg/controller/scyllacluster/migrate_test.go
```diff
@@ -202,9 +206,7 @@ func TestMigrateV1ScyllaClusterToV1Alpha1ScyllaDBDatacenter(t *testing.T) {
 			}(),
 			expectedScyllaDBDatacenter: func() *scyllav1alpha1.ScyllaDBDatacenter {
 				sd := newBasicScyllaDBDatacenterWithNoStatus()
-				sd.Annotations = map[string]string{
-					"internal.scylla-operator.scylladb.com/host-networking": "true",
-				}
+				metav1.SetMetaDataAnnotation(&sd.ObjectMeta, "internal.scylla-operator.scylladb.com/host-networking", "true")
```
czeslavo: As we're modifying these lines, we could opportunistically turn hardcoded strings into constants references (applies to all cases), WDYT?
```suggestion
				metav1.SetMetaDataAnnotation(&sd.ObjectMeta, naming.TransformScyllaClusterToScyllaDBDatacenterHostNetworkingAnnotation, "true")
```

rzetelskik: So far we've been hardcoding these values in unit tests to protect against changes that could affect the behaviour, so this is a "feature". I recall that @mflendrich recently wanted to move away from these change detectors, but I'm not sure it applied to unit tests. Anyway, at this point it seems to be more of a culture/habit change than a fix, so I definitely wouldn't want to touch that in this PR. But I'm happy to have this discussion otherwise.

mflendrich: 1. string constants vs literals

My personal opinion is that repeating string literals in a codebase is a code smell, and it includes tests (on all levels for which this is practical - I am not convinced that a repeated string tests anything of value). This code smell is prevalent across the entire codebase, and we've had bugs caused by typos between repetitions. ( :cry: )

2. assuming that there's no strong pushback, is it a good idea to make opportunistic code smell fixes locally

I view it as a spectrum / trade-off.

On one extreme - our culture of keeping "consistency" even in tiniest details is an absolute killer of progress in terms of quality. If we don't do anything about it, we'll keep producing thousands of lines of code a week that largely repeats existing patterns (exacerbating problems we already have). Low-impact tasks like "let's refactor 300 repetitions of this small code pattern across the codebase" will rarely get created and even rarer get enough traction to be completed.

On the other extreme, I totally get that in a codebase (especially a debt-ridden one), consistency is a survival strategy allowing some degree of predictability of changes.

-----------

I'd be curious to hear where on this spectrum y'all would like to see our coding standards going forward.

pkg/controller/scyllacluster/migrate.go
```diff
@@ -464,3 +472,223 @@ func migrateV1Alpha1ScyllaDBDatacenterStatusToV1ScyllaClusterStatus(sdc *scyllav
...
+		ID:    smt.Status.TaskID,
+		Error: taskStatusErr,
+	})
+
+	scyllaV1TaskStatusAnnotation, ok := smt.Annotations[naming.ScyllaDBManagerTaskStatusAnnotation]
+	if ok {
+		err := json.NewDecoder(strings.NewReader(scyllaV1TaskStatusAnnotation)).Decode(&taskStatus)
+		if err != nil {
+			return *new(T), false, fmt.Errorf("can't decode ScyllaDBManagerTask status annotation: %w", err)
+		}
```
czeslavo, line 691: How about making the possible cases more explicit by not reusing `taskStatus` and returning early? Or is the intention to override `taskStatus` partially in `Decode`?
```suggestion
		statusFromAnnotation := new(T)
		if err := json.NewDecoder(strings.NewReader(scyllaV1TaskStatusAnnotation)).Decode(&statusFromAnnotation); err != nil {
			return *new(T), false, fmt.Errorf("can't decode ScyllaDBManagerTask status annotation: %w", err)
		}
		return statusFromAnnotation
```

pkg/controller/scyllacluster/sync.go
```diff
@@ -235,12 +250,18 @@ func (scmc *Controller) sync(ctx context.Context, key string) error {
 	configMaps = oslices.Filter(configMaps, isOwnedByAnyFunc[*corev1.ConfigMap](allowedOwnerUIDs))
 	services = oslices.Filter(services, isOwnedByAnyFunc[*corev1.Service](allowedOwnerUIDs))
 
+	// ScyllaDBManagerClusterRegistrations are not owned by ScyllaCluster or ScyllaDBDatacenter, so we list all.
+	scyllaDBManagerClusterRegistrations, err := scmc.scyllaDBManagerClusterRegistrationLister.ScyllaDBManagerClusterRegistrations(sc.Namespace).List(labels.Everything())
+	if err != nil {
+		objectErrs = append(objectErrs, fmt.Errorf("can't list ScyllaDBManagerClusterRegistrations: %w", err))
+	}
```
czeslavo, line 257: It could make sense to have a helper function that would filter out `ScyllaDBManagerClusterRegistration`s not related to the `ScyllaCluster` at this stage (as that's what happens for other objects). This way, we maintain the imperative of passing down only the objects related to the cluster we're processing, WDYT?

pkg/controller/scylladbmanagertask/sync_manager.go
```diff
@@ -640,3 +651,110 @@ func scyllaDBManagerClientTaskName(smt *scyllav1alpha1.ScyllaDBManagerTask) stri
...
+		repairTaskStatus, err := newScyllaV1RepairTaskStatus(t)
+		if err != nil {
+			return "", fmt.Errorf("can't make scyllav1.RepairTaskStatus: %w", err)
+		}
+
+		err = json.NewEncoder(&buf).Encode(repairTaskStatus)
+		if err != nil {
+			return "", fmt.Errorf("can't encode scyllav1.RepairTaskStatus: %w", err)
+		}
+
```
czeslavo, line 708: Better safe than sorry.
```suggestion
	default:
		return "", fmt.Errorf("unsupported task type: %q", t.Type)
```

pkg/controllerhelpers/annotations_test.go
```diff
@@ -0,0 +1,126 @@
...
+			t.Parallel()
+
+			patch, err := PrepareSetAnnotationPatch(tc.obj, tc.annotationKey, tc.annotationValue)
+			if !reflect.DeepEqual(err, tc.expectedErr) {
+				t.Fatalf("expected error %v, got %v", tc.expectedErr, err)
+			}
+
+			if !equality.Semantic.DeepEqual(patch, tc.expected) {
+				t.Errorf("expected and got differ: %s", cmp.Diff(tc.expected, patch))
+			}
```
czeslavo, line 71: I can see it's a pattern in the codebase, but it feels wrong to check equality using one function, and displaying diff using a different one, as they might detect other differences (e.g., `cmp.Diff` will output the ones that `equality.Semantic.DeepEqual` ignores).

Not sure if we want to do something with that, just wanted to surface it.

test/e2e/set/scyllacluster/scyllamanager.go
```diff
@@ -207,8 +210,53 @@ var _ = g.Describe("Scylla Manager integration", func() {
...
+		)
 		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(tasks.TaskListItemSlice).To(o.BeEmpty())
+
+		framework.By("Verifying that the cluster was removed from the global ScyllaDB Manager state")
+		o.Eventually(func(eo o.Gomega, ctx context.Context) {
+			_, err := managerClient.GetCluster(ctx, managerClusterID)
+			eo.Expect(err).To(o.HaveOccurred())
+			eo.Expect(err).To(o.Satisfy(managerclienterrors.IsNotFound))
+		}).
```
czeslavo: I can see we do not have a complementary check for verifying whether the cluster got registered in the manager in the first place, ergo we do not get 404  (we have a check relying on `IsScyllaClusterRegisteredWithManager` helper that reads `ScyllaCluster`'s status, but doesn't verify state on the manager side). Should we add one?

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -779,13 +805,49 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
...
+		framework.By("Verifying that the cluster was removed from the global ScyllaDB Manager state")
+		o.Eventually(func(eo o.Gomega, ctx context.Context) {
+			_, err := managerClient.GetCluster(ctx, managerClusterID)
+			eo.Expect(err).To(o.HaveOccurred())
+			eo.Expect(err).To(o.Satisfy(managerclienterrors.IsNotFound))
+		}).
+			WithContext(ctx).
+			WithTimeout(utils.SyncTimeout).
+			WithPolling(5 * time.Second).
+			Should(o.Succeed())
```
czeslavo: Couldn't we extract a helper deleting a `ScyllaCluster` and verifying its deletion? We've got this duplicated in `Scylla Manager integration` test.

pkg/controller/scyllacluster/migrate_test.go
czeslavo: I can see some value in the approach that lets us use string literals in unit tests to have some kind of a sanity check to protect us from unintentional public API changes. We should not use literals in the production code, that's for sure.

Another problem I can see is that we do not make annotations/labels' names part of our API. If we did so, we could use https://pkg.go.dev/golang.org/x/exp/apidiff to detect possible breaking changes automatically in our `pkg/api/...` packages.

pkg/controller/scyllacluster/migrate.go
rzetelskik: > Or is the intention to override taskStatus partially in Decode?

It is, both the status annotation and degraded condition (e.g. on an invalid update) can be present, and for backward compatibility both should be propagated.
I know it's not the nicest code but it avoids duplication of the entire logic for the two task types.

pkg/controller/scyllacluster/sync.go
rzetelskik: The issue is there is no direct relation between SC and SMCR. 
It can't be done using a selector as SC is not the owner of SMCR. We could set selector labels on SMCR and just not try to adopt it but it feels incorrect?
To get the related SMCR, SDC for SC needs to be found first and then the SMCR can be found using the naming convention. Coincidentally the names of SC and SDC are equal, so this could technically be possible, but that again feels like breaking the actual relations.

test/e2e/set/scyllacluster/scyllamanager.go
rzetelskik: Manager's ClusterID can only be propagated to ScyllaCluster's status if the cluster was registered with manager, so I'm not sure there's much benefit to it. Or would you do it just for confidence/correctness (fine with me)?

pkg/controller/scyllacluster/migrate.go
czeslavo: Can we leave a comment to explain this? The intention might not be obvious to the reader.

pkg/controller/scyllacluster/migrate_test.go
rzetelskik: > I'd be curious to hear where on this spectrum y'all would like to see our coding standards going forward.

I think the consistency thing sets the bar in two ways.

One is for new contributions in which case I think the contributor should either follow the existing pattern or be prepared to explain why not following it is a better option. The danger here is burdening them with modifying the already existing, unrelated code, which in my opinion can lead to even more resistance.

The other is when a reviewer expects moving away from the existing pattern in a specific case. In that case, provided the proposed changed is beneficial (and the reviewer can clearly explain why), by no means should a contributor use consistency as an argument to not yield. I suppose this is where the issue is now - too much conservatism just for the sake of it.

> I am not convinced that a repeated string tests anything of value

I though Greg's comment was referring to a string in an expected result, but I now see it refers to the test's input, in which case I see no benefit to copying string literals either.

I can change the tests to use references then.

> Another problem I can see is that we do not make annotations/labels' names part of our API.

Bear in mind most of the labels/annotations I'm using and testing here are internal.

pkg/controllerhelpers/annotations_test.go
rzetelskik: I don't think I ever saw it cause any issues. We use cmp.Diff just to print "nicer" comparison, the worst case scenario is `equality` finds a difference which `cmp` doesn't surface. I'd still rather stick with equality than cmp.

pkg/controller/scyllacluster/sync.go
czeslavo: We extract SDC a few lines above, don't we? 
```go
	if sdc, ok := scyllaDBDatacenterMap[sc.Name]; ok {
		allowedOwnerUIDs = append(allowedOwnerUIDs, sdc.UID)
	}
```
We could use it to call `getScyllaDBManagerClusterRegistration` right after `scmc.scyllaDBManagerClusterRegistrationLister.ScyllaDBManagerClusterRegistrations` (or possibly tie them together in a helper method, e.g. `getScyllaDBManagerClusterRegistrationForScyllaDataCenter`)?

test/e2e/set/scyllacluster/scyllamanager.go
czeslavo: I think it could protect us from potential issues with the implementation of the manager integration. As this is an E2E test, we should verify the side effects and not trust the internal implementation (i.e. assume ID gets correctly propagated to the status), because it may change (e.g., we switch to generating cluster ID on the operator side and propagate it to the status before registration, or similar change).

pkg/controllerhelpers/annotations_test.go
czeslavo: I agree the test logic is correct, we may just hit a misleading message if there's a difference between `cmp` and `equality` logic. If you believe it's not a problem, let it be as-is.

pkg/controller/scyllacluster/sync.go
rzetelskik: added filtering

pkg/controller/scylladbmanagertask/sync_manager.go
rzetelskik: done

pkg/controller/scyllacluster/migrate_test.go
rzetelskik: > I can change the tests to use references then.

done

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
rzetelskik: I propose to deduplicate the test instead. At the moment every test is verifying that the cluster was registered with ScyllaDB Manager before it proceeds to schedule tasks. Instead, I added a dedicated test here testing ScyllaCluster registration and deregistration from ScyllaDB Manager state.
I suggest that the existing tests (for backup and repair tasks) no longer explicitly verify that the cluster was registered and instead treat the registration as an intermediate step to task scheduling. That part would be changed in a separate PR to ensure conformance with the existing tests.

test/e2e/set/scyllacluster/scyllamanager.go
rzetelskik: done

pkg/cmd/operator/operator.go
rzetelskik: Good point, I'll create an issue. This should probably cover more than just the gate, I'll try to cover it in the same issue.

pkg/controller/scyllacluster/migrate.go
rzetelskik: done

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
rzetelskik: Added a separate test for registration/deregistration.

czeslavo: Makes sense. :+1: I like the proposed approach as it promotes a "single reason to fail" rule of thumb for tests.

rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[49d2d39](https://github.com/scylladb/scylla-operator/commit/49d2d39adafe94e655074c199a36ab46deabd8c8)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2681/pull-scylla-operator-master-e2e-gke-parallel/1937060619316367360)	unknown	`/test e2e-gke-parallel`
> ci/prow/e2e-gke-parallel-clusterip	[49d2d39](https://github.com/scylladb/scylla-operator/commit/49d2d39adafe94e655074c199a36ab46deabd8c8)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2681/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1937060619635134464)	unknown	`/test e2e-gke-parallel-clusterip`
> ci/prow/e2e-gke-multi-datacenter-parallel	[49d2d39](https://github.com/scylladb/scylla-operator/commit/49d2d39adafe94e655074c199a36ab46deabd8c8)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2681/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1937060620377526272)	unknown	`/test e2e-gke-multi-datacenter-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2681). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

all known flakes + timeouts
relevant, manager-related ones all failed on snapshot tag collisions
/retest

rzetelskik:
/hold
for https://github.com/scylladb/scylla-operator/pull/2702

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
rzetelskik: https://github.com/scylladb/scylla-operator/issues/2707

rzetelskik:
rebased
/hold cancel
/cc czeslavo

rzetelskik:
> 
> ci/prow/e2e-gke-multi-datacenter-parallel	[81c12d5](https://github.com/scylladb/scylla-operator/commit/81c12d52b85e3925e1d3051c9d69616486529616)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2681/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1937485022164422656)	true	`/test e2e-gke-multi-datacenter-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2681). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

/retest

PR #2682, opened 2025-06-09
Update OWNERS to the new structure

mflendrich:
Implements the agreed changes to the OWNERS file.

/auto-cc


OWNERS_ALIASES
```diff
@@ -1,13 +1,22 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 aliases:
-  operator-reviewers:
-  - zimnx
-  - rzetelskik
-  - mflendrich
+  operator-team:
```
rzetelskik, line 4: Do we need this for anything?

OWNERS_ALIASES
```diff
@@ -1,13 +1,22 @@
...
+  # All members of the team.
   - czeslavo
-  operator-approvers:
+  - mflendrich
+  - rzetelskik
   - zimnx
+  operator-review-autoassignees:
+  # Subset of operator-team that will be automatically assigned to PRs.
```
czeslavo: I think this way it will be unambiguous - "being assigned" is not equal to "being requested for review" in the GH PRs world. 
```suggestion
  # Subset of operator-team that will be automatically requested for reviews in PRs.
```

OWNERS_ALIASES
```diff
@@ -1,13 +1,22 @@
...
+  - czeslavo
   - rzetelskik
+  - zimnx
+  operator-approvers:
+  # Subset of operator-team that has "approve" rights.
+  # Rule for adding new members:
+  # Demonstrated meaningful contributions to the majority of the codebase
+  # over a period of at least 6 months.
```
czeslavo: Maybe this way, to avoid ambiguity?
```suggestion
  # over a period of at least 6 months, OR a team lead.
```

OWNERS_ALIASES
```diff
@@ -1,13 +1,22 @@
...
-  operator-approvers:
+  - mflendrich
+  - rzetelskik
   - zimnx
+  operator-review-autoassignees:
+  # Subset of operator-team that will be automatically requested for reviews in PRs.
+  - czeslavo
   - rzetelskik
+  - zimnx
+  operator-approvers:
```
rzetelskik, line 15: bear in mind this is also used for `required_reviewers` for dependency changes, if your goal was to not be automatically assigned to reviews, maybe this would make more sense?

```
$ cat OWNERS_ALIASES
# See the OWNERS docs at https://go.k8s.io/owners

aliases:
  operator-review-autoassignees:
  - czeslavo
  - rzetelskik
  - zimnx
  
  operator-leaders:
  - mflendrich
  
  operator-maintainers: 
  # Rule for adding new members:
  # Demonstrated meaningful contributions to the majority of the codebase
  # over a period of at least 6 months.
  - rzetelskik
  - zimnx
```

```
$ cat OWNERS
# See the OWNERS docs at https://go.k8s.io/owners

filters:
  ".*":
    approvers:
    - operator-leaders
    - operator-maintainers
    reviewers:
    - operator-review-autoassignees
  "go\\.(mod|sum)$":
    required_reviewers:
    - operator-maintainers
    labels:
    - area/dependency
```

OWNERS_ALIASES
mflendrich: Technically? No.

But in other projects (like internal documentation, etc) this team may be useful and having this file consistent across repos feels like the right thing to do. And I do want to have a group for the entire team for vanity reasons.

Therefore I'd like to keep it as is.

OWNERS_ALIASES
mflendrich: I'd err on the side of simplicity and not treat the "not assigned to reviews by default" dogmatically, but practically. Also I'm happy to take on some of the dependency review burden.

I'm fine as it is and would rather keep it this way for the reasons described above.

OWNERS_ALIASES
mflendrich: Done.

OWNERS_ALIASES
mflendrich: Done.

OWNERS_ALIASES
rzetelskik: ok

OWNERS_ALIASES
rzetelskik: this split is being resolved now using a comment (the `OR a team lead` part), so imo separate groups for a leader and maintainers doesn't add complexity, but I don't mind as long as the functionality is as expected

rzetelskik:
lgtm, but let's give the whole team a chance to review this.
Please squash the commits though.

mflendrich:
> Please squash the commits though.

done

> lgtm, but let's give the whole team a chance to review this. 

@zimnx @czeslavo PTAL :eyes:

czeslavo (approved):

PR #2683, opened 2025-06-09
fix: changed name of namespaces into chart templates: scylla-operator, scylla-manager

kyourselfer:


kyourselfer:
checked by deploy different Namespaces

zimnx (changes requested):
Those names are reserved, and we use them in several places in code. Changes only in helm are not sufficient to have fully functional and supported setup.

kyourselfer:
Thanks for quick response!)
I'll close for this!

kyourselfer:
names of Namespaces scylla-operator, scylla-manager are reserved, and we use them in several places in code.

PR #2684, opened 2025-06-09
Fix ID propagation to ScyllaDBClusterRegistration's status

rzetelskik:
**Description of your changes:** Currently, if a colliding cluster exists in ScyllaDB Manager state, its ID will be incorrectly propagated to ScyllaDBManagerClusterRegistration status just before the cluster is deleted. This PR fixes it. It also adds a missing error check and missing progressing conditions.

Resolves #

/kind bug
/priority important-soon


zimnx (approved):
/lgtm

PR #2689, opened 2025-06-10
Migrates NewNamedRateLimitingQueue to NewTypedRateLimitingQueueWithConfig

### LINKED ISSUE #2685: Migrate deprecated NewNamedRateLimitingQueue calls to NewTypedRateLimitingQueueWithConfig
## Problem description

`k8s.io/client-go/util/workqueue.NewNamedRateLimitingQueue` is deprecated, and we should stop using it as it may get removed. We should rely on `NewTypedRateLimitingQueueWithConfig` instead.

## Acceptance criteria
- [ ] There are no `k8s.io/client-go/util/workqueue.NewNamedRateLimitingQueue`  calls in the codebase
- [ ] Instead of loosely typed `TypedRateLimitingQueueConfig[any]`, `TypedRateLimitingQueueConfig[types.NamespacedName]` in all controllers

RishabKrish1650:
Hello! 

First time contributing to an open source project. Please let me know if there are any issues.

This PR migrates NewNamedRateLimitingQueue to NewTypedRateLimitingQueueWithConfig as the former is deprecated now. In order to keep the behaviour consistent with the behaviour of the old code, the new typed queue uses the "any" type.

Resolves #2685


rzetelskik:
Thanks @RishabKrish1650 

> In order to keep the behaviour consistent with the behaviour of the old code, the new typed queue uses the "any" type.

Is there anything that prevents us from actually taking advantage of the strong typing (actual controller's GVK, not any or strings that we use as keys now)?

rishabkrish:
Hello @rzetelskik,

Not that I am aware of.

For transparency's sake, since I am not familiar with the codebase, I'm afraid that making this change might introduce some side effect elsewhere, but I am willing to give it a shot. Hopefully, if I make any mistake the review/tests will catch it. 

If its OK with you I will proceed with incorporating the types.NamespacedName into the queues as mentioned in the new acceptance criteria?

rishabkrish:
Hello @rzetelskik,

Sorry to bother. Had a quick question regarding adding the types.NamespacedNamed type as the key type to the queue. As you know the previous implementation relied on string keys (I believe the current code expects them in the format namespace/name). I tried to introduce the type for one controller and the changes cascade into multiple files since I need to change the corresponding handlers and KeyFunc.

My question is this: is the acceptance criteria just to introduce strong typing in the queue in the respective controller, and pass the key in the required format wherever necessary(i.e. something like key.Namespace + "/" + key.Name)? Basically, I want to know how extensive of a change is required. I am assuming that we don't want a massive change pushed in relation to this issue.

Please let me know how to proceed here.

rishabkrish:
Hello @rzetelskik 

I've incorporate strong typing using the types.NamespacedName type as per the acceptance criteria. I had to define a new deletion key function (DeletionHandlingTypedKeyFunc) and a new struct for if the deletion event was missed(DeletedStronglyTypedFinalStateUnknown) for a strongly typed key. 

Please let me know if there are any changes required.

Thanks

vendor/k8s.io/client-go/tools/cache/controller.go
```diff
@@ -354,6 +356,19 @@ func DeletionHandlingMetaNamespaceKeyFunc(obj interface{}) (string, error) {
 	return MetaNamespaceKeyFunc(obj)
 }
 
+// DeletionHandlingTypedKeyFunc checks for
+// DeletedFinalStateUnknownNamespacedName objects before calling
+// MetaNamespaceKeyFuncWithStrongType. It returns a
+// types.NamespacedName, which is a struct with Namespace and Name
+// fields, instead of a string key. This function is basically the same as
+// DeletionHandlingMetaNamespaceKeyFunc but for keys of type types.NamespacedName
+func DeletionHandlingTypedKeyFunc(obj interface{}) (types.NamespacedName, error) {
```
rzetelskik: You can't just edit the vendored dependencies.
Instead, you could add a similar func to `controllehelpers/handlers.go`

vendor/k8s.io/client-go/tools/cache/delta_fifo.go
```diff
@@ -727,3 +728,11 @@ type DeletedFinalStateUnknown struct {
 	Key string
 	Obj interface{}
 }
+
+// DeletedStronglyTypedFinalStateUnknown is like DeletedFinalStateUnknown, but
+// uses a types.NamespacedName as the key. This is useful for objects that are
+// namespaced, as it allows for more precise identification of the object.
+type DeletedStronglyTypedFinalStateUnknown struct {
```
rzetelskik: Ditto re: vendored dependencies.

vendor/k8s.io/client-go/tools/cache/store.go
```diff
@@ -116,6 +117,19 @@ func MetaNamespaceKeyFunc(obj interface{}) (string, error) {
 	return objName.String(), nil
 }
 
+// MetaNamepaceKeyFuncWithStrongType is a default KeyFunc similar to
+// MetaNamespaceKeyFunc but for keys of type types.NamespacedName.
+func MetaNamespaceKeyFuncWithStrongType(obj interface{}) (types.NamespacedName, error) {
+	if key, ok := obj.(ExplicitKey); ok {
+		return types.NamespacedName{Namespace: "", Name: string(key)}, nil
+	}
```
rzetelskik: I don't think that'd be possible (or correct) with a queue of `types.NamespacedName`

rzetelskik (changes requested):
> If its OK with you I will proceed with incorporating the types.NamespacedName into the queues as mentioned in the new acceptance criteria?

Sure. Imo there are two main goals here:
1. stop using the the deprecated methods
2. get rid of key typecasting (from any to string)

For the above, using workqueue.TypedRateLimitingInterface[string] should be fine.

Optionally we could also reduce the amount of key splitting and joining (by using a struct type, e.g. `types.NamespacedName` like you did) instead of string. This is more convoluted though, as the k8s store/cache implementations still depend on the key being a string.

> Please let me know how to proceed here.

Changing the handlers implementation is fine, no worries.

> I've incorporate strong typing using the types.NamespacedName type as per the acceptance criteria. I had to define a new deletion key function (DeletionHandlingTypedKeyFunc) and a new struct for if the deletion event was missed(DeletedStronglyTypedFinalStateUnknown) for a strongly typed key.

Overall this seems to be going in the right direction, but see my comments on editing the vendored dependencies. especially with regards to `cache.DeletedFinalStateUnknown`.

I looked into it briefly myself and because of the k8s store/cache implementation, and the `cache.DeletedFinalStateUnknown` being used, it seems like we can't really move away from struct-string conversions either way. So feel free to give completing this a shot, or, if you find it too clumsy, backtrack to using `workqueue.TypedRateLimitingInterface[string]`. Either is fine with me.

Also, as per our contributing guide, please try to keep the PR in a reviewable state every time you request a review; in this case that means squashing the commits and adjusting the PR title and commit messages to adhere to the guide. Thanks!

vendor/k8s.io/client-go/tools/cache/controller.go
rishabkrish: Moved this to the handlers file. I've reverted all changes in the vendor dependencies.

vendor/k8s.io/client-go/tools/cache/delta_fifo.go
rishabkrish: Moved this to the handlers file. I've reverted all changes in the vendor dependencies.

vendor/k8s.io/client-go/tools/cache/store.go
rishabkrish: You are correct, removed this.

rishabkrish:
Thanks for the review @rzetelskik. I've addressed the changes.

> Feel free to give completing this a shot, or, if you find it too clumsy, backtrack to using `workqueue.TypedRateLimitingInterface[string]`

I've gone ahead with using the `types.NamespacedName` type for now.

> Also, as per our contributing guide, please try to keep the PR in a reviewable state every time you request a review; in this case that means squashing the commits and adjusting the PR title and commit messages to adhere to the guide.

I've squashed the commits. Do you think the PR title needs to change?

If any further changes are required, please let me know.

rishabkrish:
Hello @rzetelskik 

I've made the changes you've requested. Additionally, I've added some comments to explain the functionality of the new methods and struct in the helpers file.

As for the tests, not sure why they are failing. As far as I can see the failure is not related to any change in this PR. I've seen that there are some open issues that indicate that these test failures are a known issue. 

Please let me know if there are any further changes needed on this PR.

Thanks

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
 func (h *Handlers[T]) HandleDelete(obj any, enqueueFunc EnqueueFuncType) {
 	h.HandleDeleteWithDepth(2, obj, enqueueFunc)
 }
+
+// DeletedStronglyTypedFinalStateUnknown is like DeletedFinalStateUnknown, but
+// uses a types.NamespacedName as the key. This is useful for objects that are
+// namespaced, as it allows for more precise identification of the object.
```
rzetelskik:
```suggestion
```
comments shouldn't explain the code but add additional context if necessary, this is very clear from the code alone

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
...
+	Key types.NamespacedName
+	Obj interface{}
+}
+
+// DeletionHandlingTypedKeyFunc checks for
+// DeletedStronglyTypedFinalStateUnknown objects before calling
+// MetaNamespaceKeyFuncWithStrongType. This function is essentially the
+// same as the function DeletionHandlingMetaNamespaceKeyFunc
+// but it handles keys of the types.NamespacedName type.
+// This is required for queues that use strongly typed keys of the same type
```
rzetelskik:
```suggestion
```
ditto about the comment

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
...
+	if d, ok := obj.(DeletedStronglyTypedFinalStateUnknown); ok {
+		return d.Key, nil
+	}
+	return MetaNamespaceKeyFuncWithStrongType(obj)
+}
+
+// MetaNamespaceKeyFuncWithStrongType is a KeyFunc that returns
+// types.NamespacedName. This function is similar to
+// MetaNamespaceKeyFunc, but it returns a structured key instead of a string.
+// It is useful for queues that use strongly typed keys of the same type.
```
rzetelskik:
```suggestion
```
ditto about the comment

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
...
+func MetaNamespaceKeyFuncWithStrongType(obj interface{}) (types.NamespacedName, error) {
```
rzetelskik: I don't think the name reflects what this function does, it's not any more "strongly typed" than a function returning a string is

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
...
 	h.HandleDeleteWithDepth(2, obj, enqueueFunc)
 }
+
+// DeletedStronglyTypedFinalStateUnknown is like DeletedFinalStateUnknown, but
+// uses a types.NamespacedName as the key. This is useful for objects that are
+// namespaced, as it allows for more precise identification of the object.
+type DeletedStronglyTypedFinalStateUnknown struct {
+	Key types.NamespacedName
+	Obj interface{}
+}
```
rzetelskik:
```suggestion
type DeletedFinalStateUnknown struct {
	Key types.NamespacedName
	Obj interface{}
}
```
it's not any more "strongly typed" than the regular `cache.DeletedFinalStateUnknown`, so I'd suggest just redefining it here, the name collision is fine as we'll just use our implementation

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
...
+// types.NamespacedName. This function is similar to
+// MetaNamespaceKeyFunc, but it returns a structured key instead of a string.
+// It is useful for queues that use strongly typed keys of the same type.
+func MetaNamespaceKeyFuncWithStrongType(obj interface{}) (types.NamespacedName, error) {
+	objName, err := cache.ObjectToName(obj)
+	if err != nil {
+		return types.NamespacedName{}, err
+	}
+	return types.NamespacedName{Namespace: objName.Namespace, Name: objName.Name}, nil
+}
```
rzetelskik: seems a bit redundant to use `cache.ObjectToName`
```suggestion
func ObjectToNamespacedName(obj interface{}) (types.NamespacedName, error) {
	metaAccessor, err := meta.Accessor(obj)
	if err != nil {
		return types.NamespacedName{}, fmt.Errorf("object has no meta: %w", err)
	}

	return types.NamespacedName{
		Namespace: metaAccessor.GetNamespace(),
		Name:      metaAccessor.GetName(),
	}, nil
}
```

pkg/controllerhelpers/handlers.go
```diff
@@ -202,3 +203,36 @@ func (h *Handlers[T]) HandleDeleteWithDepth(depth int, obj any, enqueueFunc Enqu
...
+// MetaNamespaceKeyFuncWithStrongType. This function is essentially the
+// same as the function DeletionHandlingMetaNamespaceKeyFunc
+// but it handles keys of the types.NamespacedName type.
+// This is required for queues that use strongly typed keys of the same type
+func DeletionHandlingTypedKeyFunc(obj interface{}) (types.NamespacedName, error) {
+	if d, ok := obj.(DeletedStronglyTypedFinalStateUnknown); ok {
+		return d.Key, nil
+	}
+	return MetaNamespaceKeyFuncWithStrongType(obj)
+}
```
rzetelskik:
```suggestion
func DeletionHandlingObjectToNamespacedName(obj interface{}) (types.NamespacedName, error) {
	if d, ok := obj.(DeletedFinalStateUnknown); ok {
		return d.Key, nil
	}

	namespacedName, err := ObjectToNamespacedName(obj)
	if err != nil {
		return types.NamespacedName{}, fmt.Errorf("can't convert object to namespaced name: %w", err)
	}

	return namespacedName, nil
}
```
consider this with the alternative to 'MetaNamespaceKeyFuncWithStrongType' I proposed

rzetelskik (changes requested):
Thanks for the updates @RishabKrish1650, good effort overall.

> I've squashed the commits. Do you think the PR title needs to change?

Ideally it should say `Migrate rate limiting queue to typed version`, as it should use the imperative.

> As for the tests, not sure why they are failing. As far as I can see the failure is not related to any change in this PR. I've seen that there are some open issues that indicate that these test failures are a known issue.

Yes, the multi-datacenter parallel suite is absolutely broken right now and fails very consistently. The regular parallel suite less so, but still flaky. Don't worry about it, I'll make sure this PR doesn't introduce any new flakes and take care of merging this once we get it to that stage.

> Please let me know if there are any further changes needed on this PR.

Left a couple of comments

pkg/controllerhelpers/handlers.go
rishabkrish: I am fine with this change, just a question though (forgive me if it's silly).

The reason why I used the ObjectToName function to extract the relevant info was to ensure that the methodology of extracting the namespace and name would remain consistent regardless of whether it was in the strongly typed queues or somewhere else. 

My question is would it not be advantageous to reuse this methodology to ensure that the names are extracted the same way everywhere? Especially since this is a vendor dependency method. This way if the underlying implementation of ObjectToName changes, we would not have to change anything in the code. 

Of course, the assumption here is that the implementation changes but the same data is returned. If the returned value of the function changes then we will definitely have to make changes.

Let me know if the question is unclear.

pkg/controllerhelpers/handlers.go
rishabkrish: Will make this change based on response to question asked in the MetaNamespaceKeyFuncWithStrongType comment.

pkg/controllerhelpers/handlers.go
rishabkrish: Removed the comment

pkg/controllerhelpers/handlers.go
rishabkrish: Removed the comment

pkg/controllerhelpers/handlers.go
rishabkrish: Removed the comment

pkg/controllerhelpers/handlers.go
rishabkrish: Renamed

pkg/controllerhelpers/handlers.go
rishabkrish: Renamed this to ObectToNamespacedName. I agree with the name change, had a question regarding the implementation of this method which I have asked below.

pkg/controllerhelpers/handlers.go
rzetelskik: >  This way if the underlying implementation of ObjectToName changes, we would not have to change anything in the code.

I'm not sure I follow. We're in control of the keys in the queue and how we prepare a key from an object implementing metav1.Object. I don't see how a change in the implementation of `cache.ObjectToName` could require a change on our side, can you give me a pointer?

czeslavo: I believe what @rzetelskik proposed is the right path to follow - it skips the unnecessary interim `ObjectName` instantiation.

rishabkrish: This is just over complication on my part but the thought process is this:

So if I am not wrong the proposed changes are basically a 1:1 of the `cache.ObjectToName` function but now it returns a `types.NamespacedName` object. So if there's some change in the `cache.ObjectToName` function(for example if the behaviour of `meta.Accessor` changes), since we've essentially lifted that logic and tweaked it a little to suit our need, wouldn't we also need to make a similar change to our custom implementation? Basically, if what we have copied has changed, we need to change what we have copied.

Again, just an over complication on my end. I'll make the proposed changes if ok with you.

rishabkrish: > I believe what @rzetelskik proposed is the right path to follow - it skips the unnecessary interim `ObjectName` instantiation.

Got it, making the change accordingly.

rishabkrish: Changed the method as per proposal.

pkg/controllerhelpers/handlers.go
rishabkrish: Incorporated the proposed changes.

pkg/controllertools/observer.go
```diff
@@ -36,9 +36,9 @@ func NewObserver(name string, eventsClient corev1client.EventInterface, syncFunc
 
 	return &Observer{
 		name: name,
-		queue: workqueue.NewRateLimitingQueueWithConfig(
-			workqueue.DefaultControllerRateLimiter(),
-			workqueue.RateLimitingQueueConfig{
+		queue: workqueue.NewTypedRateLimitingQueueWithConfig(
+			workqueue.DefaultTypedControllerRateLimiter[any](),
```
rzetelskik: it can take a string as that's what's being used as a key

rishabkrish: done

rzetelskik:
/lgtm
/approve
Thanks @RishabKrish1650!

PR #2694, opened 2025-06-17
Fix ScyllaDB Manager task adoption in case of missing owner UID label

rzetelskik:
**Description of your changes:** Standalone ScyllaDB Manager controller does not label tasks created in ScyllaDB Manager with an OwnerUID label. Therefore the existing controller logic is broken and the tasks migrated from v1.ScyllaCluster would get recreated. This PR introduces an internal annotation that allows forcing adoption. It also changes the logic in case when the ownerUID labels exists but doesn't match the object to return an error instead of adopting.

Resolves #

/kind bug
/priority important-soon

### DIFF
--- a/pkg/controller/scylladbmanagertask/sync_manager.go
+++ b/pkg/controller/scylladbmanagertask/sync_manager.go
@@ -105,7 +105,12 @@ func (smtc *Controller) syncManager(
 	}
 
 	ownerUIDLabelValue, hasOwnerUIDLabel := managerTask.Labels[naming.OwnerUIDLabel]
-	if !hasOwnerUIDLabel {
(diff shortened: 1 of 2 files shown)


czeslavo (commented):
Do I understand correctly that we will set `internal.scylla-operator.scylladb.com/scylladb-manager-task-missing-owner-uid-force-adopt` only for tasks migrated from `v1.ScyllaCluster`?

rzetelskik:
> Do I understand correctly that we will set `internal.scylla-operator.scylladb.com/scylladb-manager-task-missing-owner-uid-force-adopt` only for tasks migrated from `v1.ScyllaCluster`?

Yes

Issue #2687, opened 2025-06-10
Extract controllers boilerplate logic to a reusable struct


czeslavo:
## Problem statement

Currently, all of our controllers share a very similar boilerplate that gets copied over every time a new controller is created. This code contains logic that is not unit tested. We may introduce changes/fixes to only a subset of the controllers, overlooking the rest, even though the logic should be unified. 

The boilerplate contains the public `Run` function and its private dependencies: `processNextItem` and `runWorker`.  

## Suggested solution

Extract a 

## Acceptance criteria
- [ ] The common controller's logic responsible for integration with the queue and processing items is extracted to a reusable struct
- [ ] The new struct logic is unit tested
- [ ] All controllers sharing this common logic, rely on the new struct

rzetelskik:
In terms of unification, it is also worth mentioning that we already have a limited version of it, capable of handling singleton keys only, extracted to a lib: https://github.com/scylladb/scylla-operator/blob/cbb1cbe08fd018d2dda28d972668a8f0cbdfb415/pkg/controllertools/observer.go.

PR #2702, opened 2025-06-23
Revert "Migrate rate limiting queue to typed version"

rzetelskik:
Reverts scylladb/scylla-operator#2689.

`cache.DeletedFinalStateUnknown` events can also be generated by the informers. With this change we wouldn't handle those.

/kind bug
/priority important-soon
/cc zimnx

### DIFF
--- a/pkg/controller/manager/controller.go
+++ b/pkg/controller/manager/controller.go
@@ -17,7 +17,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	apimachineryutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	apimachineryutilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	apimachineryutilwait "k8s.io/apimachinery/pkg/util/wait"
@@ -43,7 +42,7 @@ const (
 )
 
 var (
-	keyFunc                    = controllerhelpers.DeletionHandlingObjectToNamespacedName
+	keyFunc                    = cache.DeletionHandlingMetaNamespaceKeyFunc
 	scyllaClusterControllerGVK = scyllav1.GroupVersion.WithKind("ScyllaCluster")
 )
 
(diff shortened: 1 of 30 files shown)


czeslavo:
> cache.DeletedFinalStateUnknown events can also be generated by the informers. With this change, we wouldn't handle those.

Could you elaborate on why having a strongly typed queue (`{namespace, name}` in `types.NamespacedName` instead of an implicit `{namespace, name}` as a `string`) might be a problem in handling `cache.DeletedFinalStateUnknown`? I'd like to understand.

rzetelskik:
> Could you elaborate on why having a strongly typed queue ({namespace, name} in types.NamespacedName instead of an implicit {namespace, name} as a string) might be a problem in handling cache.DeletedFinalStateUnknown? I'd like to understand.

It's not preventing us from handling `cache.DeletedFinalStateUnknown`, but in this PR we introduced a proprietary type and stopped handling `cache.DeletedFinalStateUnknown`, which can still be produced by the DeltaFIFO used by informers (and how the key in this struct is produced is out of our control).
I'm reverting this PR to fix this in #2703, it'll just be clearer what's being changed and why that way.

See also https://github.com/scylladb/scylla-operator/pull/2703 - the issue is we still have to handle `cache.DeletedFinalStateUnknown`, which requires switching from queue's key type to string. We have a bunch of places where this conversion would have to be carried out and leaving it up to the callee seems fragile to me until we have https://github.com/scylladb/scylla-operator/issues/2687 implemented and can control the type conversion in a single place.
If you disagree feel free to comment on #2703.

czeslavo:
> > Could you elaborate on why having a strongly typed queue ({namespace, name} in types.NamespacedName instead of an implicit {namespace, name} as a string) might be a problem in handling cache.DeletedFinalStateUnknown? I'd like to understand.
> 
> It's not preventing us from handling `cache.DeletedFinalStateUnknown`, but in this PR we introduced a proprietary type and stopped handling `cache.DeletedFinalStateUnknown`, which can still be produced by the DeltaFIFO used by informers (and how the key in this struct is produced is out of our control). I'm reverting this PR to fix this in #2703, it'll just be clearer what's being changed and why that way.
> 
> See also #2703 - the issue is we still have to handle `cache.DeletedFinalStateUnknown`, which requires switching from queue's key type to string. We have a bunch of places where this conversion would have to be carried out and leaving it up to the callee seems fragile to me until we have #2687 implemented and can control the type conversion in a single place. If you disagree feel free to comment on #2703.

Thanks, it's clear now.

PR #2703, opened 2025-06-23
Migrate depracted RateLimitingInterface to TypedRateLimitingInterface

rzetelskik:
**Description of your changes:** This PR migrates the depracted RateLimitingInterface to TypedRateLimitingInterface.
We initially wanted to change the queue key type from implicit string to `types.NamespacedName`. 
However, with us not being in control of how the keys in `cache.DeletedFinalStateUnknown` struct are created by the DeltaFIFO, it seems a bit fragile to do this, as it would be up to the callee to parse/stringify the key in the expected manner.
Before we handle https://github.com/scylladb/scylla-operator/issues/2687, it seems like less of a hassle to stick with string as a key type.

Resolves https://github.com/scylladb/scylla-operator/issues/2685

/kind cleanup
/priority important-longterm

/hold for https://github.com/scylladb/scylla-operator/pull/2702


czeslavo:
lgtm 
/assign zimnx

PR #2708, opened 2025-06-24
Omit empty value for switch parameters

### LINKED ISSUE #2706: `scyllaArgs` not accepting flags without values
`scyllaArgs` expects every `--flag` to have a value and mangles flags without values by adding an empty value to them.

This does not work with `--poll-mode` for which ScyllaDB expects no value, but Operator appends an empty string `""` value, causing ScyllaDB to fail to start up.

https://github.com/scylladb/scylla-operator/blob/master/pkg/sidecar/config/config.go#L191-L199

## Acceptance Criteria

- [ ] `--poll-mode` works if passed in `scyllaArgs`

zimnx:

When scyllaArgs contains switch parameter having no value, flag should only contain it's name without empty string as a value. 
This fixes issue with setting for example `--poll-mode` parameter in ScyllaDB.

Before change:
```
I0623 19:54:01.943747       1 config/config.go:265] "Scylla entrypoint" Command="/docker-entrypoint.py --listen-address=0.0.0.0 --seeds=10.107.87.173 --smp=1 --broadcast-address=10.107.87.173 --cpuset=0-3 --developer-mode=1 --poll-mode= --overprovisioned=1 --prometheus-address=0.0.0.0 --broadcast-rpc-address=10.107.87.173"
[...]
terminate called after throwing an instance of 'boost::wrapexcept<boost::program_options::invalid_command_line_syntax>'
what():  the argument for option '--poll-mode' should follow immediately after the equal sign
2025-06-23 19:54:13,113 WARN exited: scylla (terminated by SIGABRT (core dumped); not expected)
```
After change:
```
I0623 20:37:02.952290       1 config/config.go:269] "Scylla entrypoint" Command="/docker-entrypoint.py --smp=1 --prometheus-address=0.0.0.0 --cpuset=0-3 --broadcast-rpc-address=10.107.87.173 --poll-mode --developer-mode=1 --listen-address=0.0.0.0 --seeds=10.107.87.173 --overprovisioned=1 --broadcast-address=10.107.87.173"
```
And ScyllaDB starts just fine.

Resolves #2706

### DIFF
--- a/pkg/sidecar/config/config.go
+++ b/pkg/sidecar/config/config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"maps"
 	"os"
 	"os/exec"
 	"path"
@@ -174,28 +175,30 @@ func mergeSnitchConfigProperties(userConfig, operatorConfig *properties.Properti
 
 var scyllaArgumentsRegexp = regexp.MustCompile(`--([^= ]+)(="[^"]+"|=\S+|\s+"[^"]+"|\s+[^\s-]+|\s+-?\d*\.?\d+[^\s-]+|)`)
 
-func convertScyllaArguments(scyllaArguments string) map[string]string {
-	output := make(map[string]string)
+func parseScyllaArguments(scyllaArguments string) map[string]*string {
+	output := make(map[string]*string)
 	for _, value := range scyllaArgumentsRegexp.FindAllStringSubmatch(scyllaArguments, -1) {
 		if value[2] == "" {
-			output[value[1]] = ""
-		} else if value[2][0] == '=' {
-			output[value[1]] = strings.TrimSpace(value[2][1:])
+			output[value[1]] = nil
 		} else {
-			output[value[1]] = strings.TrimSpace(value[2])
+			output[value[1]] = pointer.Ptr(strings.TrimSpace(strings.TrimPrefix(value[2], "=")))
 		}
 	}
 	return output
 }
 
-func appendScyllaArguments(ctx context.Context, s *ScyllaConfig, scyllaArgs string, scyllaFinalArgs map[string]*string) {
-	for argName, argValue := range convertScyllaArguments(scyllaArgs) {
-		if existing := scyllaFinalArgs[argName]; existing == nil {
-			scyllaFinalArgs[argName] = pointer.Ptr(strings.TrimSpace(argValue))
+// mergeArguments merges two sets of arguments.
+// Arguments managed by the operator take precedence and cannot be overridden by user-provided arguments.
+func mergeArguments(operatorManagedArguments, userProvidedArguments map[string]*string) map[string]*string {
+	finalArguments := maps.Clone(operatorManagedArguments)
+	for argName, argValue := range userProvidedArguments {
+		if _, ok := operatorManagedArguments[argName]; !ok {
+			finalArguments[argName] = argValue
 		} else {
-			klog.Infof("ScyllaArgs: argument '%s' is ignored, it is already in the list", argName)
+			klog.Infof("ScyllaArgs: argument '%s' is ignored, it's managed by Operator", argName)
 		}
 	}
+	return finalArguments
 }
 
 func (s *ScyllaConfig) setupEntrypoint(ctx context.Context) (*exec.Cmd, error) {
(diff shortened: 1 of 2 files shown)


pkg/sidecar/config/config.go
```diff
@@ -191,7 +191,7 @@ func convertScyllaArguments(scyllaArguments string) map[string]string {
 func appendScyllaArguments(ctx context.Context, s *ScyllaConfig, scyllaArgs string, scyllaFinalArgs map[string]*string) {
 	for argName, argValue := range convertScyllaArguments(scyllaArgs) {
 		if existing := scyllaFinalArgs[argName]; existing == nil {
-			scyllaFinalArgs[argName] = pointer.Ptr(strings.TrimSpace(argValue))
```
czeslavo, line 194: AFAICS, we do not have a unit test covering `appendScyllaArguments`, and we drop the trimming here. Could we cover this function, e.g., by converting the existing `TestScyllaArguments` to call `appendScyllaArguments` instead of `convertScyllaArguments`? This way, we could also cover skipping user-provided arguments when they're already set.

rzetelskik: nit: this would also be easier to test if we changed `appendScyllaArguments` to `appendScyllaArgument` taking a single extra parameter, which would be entirely possible with `additionalScyllaDBArguments` (which we're joining with a separator now)

zimnx: > we drop the trimming here.

It's not really dropped, it was already a noop, as `argValue` comes from `convertScyllaArguments` which only returns trimmed values. 

I renamed functions to better reflect their semantics and added unit tests for changed functions

> nit: this would also be easier to test if we changed appendScyllaArguments to appendScyllaArgument taking a single extra parameter, which would be entirely possible with additionalScyllaDBArguments (which we're joining with a separator now)

not really easier. In case of `--param value` and function accepting single string we would have to keep context from previous execution. 
I also wouldn't want to change that much, because this logic is already fragile.

rzetelskik: >  In case of --param value and function accepting single string we would have to keep context from previous execution.

~Do we allow `additionalScyllaDBArguments` to be defined that way? Would this be valid `[]string{"--param", "value"}`?~

Ok, I see we don't validate this at all. So yeah, we'd have to join and match substrings either way.

czeslavo: > It's not really dropped, it was already a noop, as argValue comes from convertScyllaArguments which only returns trimmed values.

Yes, I just meant that it might be worrying that we're changing something that isn't covered by a unit test, and we can only deduce that by analyzing the code. 

> I renamed functions to better reflect their semantics and added unit tests for changed functions

:+1:

czeslavo:
lgtm
/assign @rzetelskik

pkg/sidecar/config/config.go
```diff
@@ -174,28 +175,32 @@ func mergeSnitchConfigProperties(userConfig, operatorConfig *properties.Properti
...
+// mergeArguments merges two sets of arguments.
+// Arguments managed by the operator take precedence and cannot be overridden by user-provided arguments.
+func mergeArguments(operatorManagedArguments, userProvidedArguments map[string]*string) map[string]*string {
+	finalArguments := maps.Clone(operatorManagedArguments)
+	for argName, argValue := range userProvidedArguments {
+		if _, ok := operatorManagedArguments[argName]; !ok {
+			finalArguments[argName] = argValue
 		} else {
-			klog.Infof("ScyllaArgs: argument '%s' is ignored, it is already in the list", argName)
+			klog.Infof("ScyllaArgs: argument '%s' is ignored, it's managed by Operator", argName)
```
rzetelskik, line 198: 1. Do you think we should log this if values are equal? Special case would be flags without values. I see the value in always logging this for consistency, but I'd like to make sure this is deliberate.
2. nit: Regardless of the answer to 1, maybe we should log the operator-managed value vs user-provided value?

pkg/sidecar/config/config.go
```diff
@@ -174,28 +175,32 @@ func mergeSnitchConfigProperties(userConfig, operatorConfig *properties.Properti
...
 		if value[2] == "" {
-			output[value[1]] = ""
+			output[value[1]] = nil
 		} else if value[2][0] == '=' {
-			output[value[1]] = strings.TrimSpace(value[2][1:])
+			output[value[1]] = pointer.Ptr(strings.TrimSpace(value[2][1:]))
 		} else {
-			output[value[1]] = strings.TrimSpace(value[2])
+			output[value[1]] = pointer.Ptr(strings.TrimSpace(value[2]))
 		}
```
rzetelskik: nit: could be simplified by just trimming the prefix

pkg/sidecar/config/config.go
zimnx: 1. Imo it's not necessary, there's a ScyllaArgs prefix meaninig something that user gave is ignored, if it happens to match what Operator provides, user value is still ignored.
2. Final result is logged just before Scylla is executed, so I think it's enough.

pkg/sidecar/config/config.go
zimnx: good point, done

PR #2717, opened 2025-06-25
Use advertised Scylla addresses on cleanup Jobs

### LINKED ISSUE #2712: Cleanup Jobs are using different connection method than what Scylla broadcasts to clients
Currently, the Cleanup Jobs connect to Scylla using the DNS of the Scylla node Service. This approach assumes that the ClusterIP of the Service is used for communication. However, in configurations where Scylla nodes are exposed via PodIPs or through a LoadBalancer, this method becomes inconsistent with how Scylla broadcasts its own addresses to clients.

As a result, when internal firewalls or network policies are in place, the Cleanup Jobs may experience connectivity issues, since they are not using the advertised IPs that Scylla expects clients to use.

**Expected Behavior:**

Cleanup Jobs should connect to Scylla using the same method (e.g., PodIP, LoadBalancer IP) that Scylla broadcasts to its clients. This ensures consistent behavior and avoids connectivity issues caused by mismatched network paths.

zimnx:

Cleanup Jobs now connect to Scylla using the same advertised IPs or DNS that Scylla broadcasts to clients (e.g., PodIP or LoadBalancer), rather than relying on the ClusterIP of the node Service. This prevents connectivity issues in environments with internal firewalls or custom network policies by ensuring the same network path is used as other client applications use.

Resolves #2712

### DIFF
--- a/pkg/controllerhelpers/scylla.go
+++ b/pkg/controllerhelpers/scylla.go
@@ -21,12 +21,21 @@ import (
 
 func GetScyllaHost(sdc *scyllav1alpha1.ScyllaDBDatacenter, svc *corev1.Service, pod *corev1.Pod) (string, error) {
 	// Assume API's default.
-	nodeBroadcastAddressType := scyllav1alpha1.BroadcastAddressTypeServiceClusterIP
+	nodeBroadcastAddressType := scyllav1alpha1.ScyllaDBDatacenterDefaultNodesBroadcastAddressType
 	if sdc.Spec.ExposeOptions != nil && sdc.Spec.ExposeOptions.BroadcastOptions != nil {
 		nodeBroadcastAddressType = sdc.Spec.ExposeOptions.BroadcastOptions.Nodes.Type
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controllerhelpers/scylla.go
```diff
@@ -29,6 +29,15 @@ func GetScyllaHost(sdc *scyllav1alpha1.ScyllaDBDatacenter, svc *corev1.Service,
 	return GetScyllaBroadcastAddress(nodeBroadcastAddressType, svc, pod)
 }
 
+func GetScyllaClientBroadcastHost(sdc *scyllav1alpha1.ScyllaDBDatacenter, svc *corev1.Service, pod *corev1.Pod) (string, error) {
+	clientsBroadcastAddressType := scyllav1alpha1.BroadcastAddressTypeServiceClusterIP
```
rzetelskik: can we move the default to a const? We have an ever growing list of places where this is defaulted

zimnx:
@rzetelskik turns out the default broadcast address is still ClusterIP. It's PodIP on API level when you specify exposeOptions, but when you don't, we keep using ClusterIP.
PTAL

Issue #2662, opened 2025-05-30
Multi datacenter ScyllaDBCluster [It] should reconcile healthy datacenters when any of DCs are down [MultiDatacenter]


rzetelskik:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928247846067245056#1:test-build-log.txt%3A2946

### Snippet of what failed.

```

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000a320e0>: 
      waiting has been interupted ([false]): timed out waiting for the condition
      {
          msg: "waiting has been interupted ([false]): timed out waiting for the condition",
          err: <wait.errInterrupted>{
              cause: <*errors.errorString | 0xc000537110>{
                  s: "timed out waiting for the condition",
              },
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_fault_tolerance.go:76 @ 05/30/25 01:46:08.043
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter.init.func3.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_fault_tolerance.go:76 +0x9b3
```

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-1.17-e2e-gke-multi-datacenter-parallel/1929417954424786944#1:test-build-log.txt%3A2231

zimnx:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2668/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1929832787989237760

czeslavo:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2669/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1930544406654357504

czeslavo:
[...]

PR #2725, opened 2025-06-27
Distinguish the control plane cluster in machinery and E2E framework for multi-datacenter scenarios

rzetelskik:
**Description of your changes:** wip

Resolves #

/kind machinery
/priority important-soon


hack/.ci/lib/e2e.sh
```diff
@@ -9,15 +9,61 @@ shopt -s inherit_errexit
...
+    # For multi-datacenter suites, designate the first kubeconfig as the "control plane" cluster.
+    KUBECONFIG="${KUBECONFIG:-${f}}"
+    WORKER_KUBECONFIGS["$( basename "${f}" '.kubeconfig' )"]="${f}"
   done
+
+  export KUBECONFIG
 fi
 
+# $1 - deployment script path
+function deploy-ci {
```
czeslavo: I feel we've got here lots of vague names already - `deploy-ci` could mean almost anything in the context of CI scripts - let's try to be more descriptive with the function name. How about something like `run-deploy-script-in-all-clusters`? 

Also, having a nice comment explaining what this function's dependencies are (`WORKER_KUBECONFIGS`, `KUBECONFIG`) could help with readability.

hack/.ci/lib/e2e.sh
```diff
@@ -9,15 +9,61 @@ shopt -s inherit_errexit
...
-  KUBECONFIGS=()
+declare -A WORKER_KUBECONFIGS
+
+# TODO: remove once CI job is updated to define KUBECONFIG and WORKER_KUBECONFIGS instead of KUBECONFIG_DIR
+if [[ -n "${KUBECONFIG_DIR+x}" ]]; then
+  unset KUBECONFIG
   for f in $( find "$( realpath "${KUBECONFIG_DIR}" )" -maxdepth 1 -type f -name '*.kubeconfig' ); do
-    KUBECONFIGS+=("${f}")
+    # For multi-datacenter suites, designate the first kubeconfig as the "control plane" cluster.
+    KUBECONFIG="${KUBECONFIG:-${f}}"
```
czeslavo, line 21: I think that for readability's sake, it could make sense to also prefix the "control plane" kubeconfig variable, e.g., `CONTROL_PLANE_KUBECONFIG`.

hack/.ci/lib/e2e.sh
```diff
@@ -9,15 +9,61 @@ shopt -s inherit_errexit
...
+
+  if [ -z "${SO_IMAGE+x}" ]; then
+    echo "SO_IMAGE can't be empty" > /dev/stderr
+    exit 2
+  fi
+
+  if [ -z "${ARTIFACTS+x}" ]; then
+    echo "ARTIFACTS can't be empty" > /dev/stderr
+    exit 2
+  fi
```
czeslavo, line 53: Shouldn't we verify if `KUBECONFIG` is set?

hack/.ci/lib/e2e.sh
```diff
@@ -9,15 +9,61 @@ shopt -s inherit_errexit
 source "$( dirname "${BASH_SOURCE[0]}" )/../../lib/bash.sh"
 source "$( dirname "${BASH_SOURCE[0]}" )/../../lib/kube.sh"
 
-if [ -z "${KUBECONFIG_DIR+x}" ]; then
-  KUBECONFIGS=("${KUBECONFIG}")
-else
-  KUBECONFIGS=()
+declare -A WORKER_KUBECONFIGS
```
czeslavo, line 14: Worth documenting here what the worker exactly means in our context.

hack/.ci/lib/e2e.sh
```diff
@@ -9,15 +9,61 @@ shopt -s inherit_errexit
...
+    ci_deploy_bg_pids+=( $! )
+  done
+
+  for pid in "${ci_deploy_bg_pids[@]}"; do
+    wait "${pid}"
+  done
+
+  if [[ -n "${worker_as_control_plane+x}" ]]; then
+    mkdir -p "${ARTIFACTS}/deploy/workers"
+    ln -s "${ARTIFACTS}/deploy/cluster" "${ARTIFACTS}/deploy/workers/${worker_as_control_plane}"
```
czeslavo: Please consider forcing link creation to handle cases where the link already exists.
```suggestion
    ln -sf "${ARTIFACTS}/deploy/cluster" "${ARTIFACTS}/deploy/workers/${worker_as_control_plane}"
```

hack/.ci/lib/e2e.sh
```diff
@@ -88,15 +134,29 @@ EOF
...
+    gather_artifacts_bg_pids+=( $! )
   done
 
   for pid in "${gather_artifacts_bg_pids[@]}"; do
     wait "${pid}"
   done
 
+  if [[ -n "${worker_as_control_plane+x}" ]]; then
+    mkdir -p "${ARTIFACTS}/must-gather/workers"
+    ln -s "${ARTIFACTS}/must-gather/cluster" "${ARTIFACTS}/must-gather/workers/${worker_as_control_plane}"
```
czeslavo: Ditto.

hack/.ci/lib/e2e.sh
```diff
@@ -9,15 +9,61 @@ shopt -s inherit_errexit
...
+  ci_deploy_bg_pids=( $! )
+
+  for name in "${!WORKER_KUBECONFIGS[@]}"; do
+    if [[ "${WORKER_KUBECONFIGS[$name]}" == "${KUBECONFIG}" ]]; then
+      # Skip if the control plane cluster is also among the worker clusters.
+      worker_as_control_plane="${name}"
+      continue
+    fi
+
+    KUBECONFIG="${WORKER_KUBECONFIGS[$name]}" ARTIFACTS_DEPLOY_DIR="${ARTIFACTS}/deploy/workers/${name}" SO_DISABLE_SCYLLADB_MANAGER_DEPLOYMENT=true timeout --foreground -v 10m "${1}" "${SO_IMAGE}" &
```
czeslavo: Please consider following formatting for readability's sake:
```suggestion
    KUBECONFIG="${WORKER_KUBECONFIGS[$name]}" \
    ARTIFACTS_DEPLOY_DIR="${ARTIFACTS}/deploy/workers/${name}" \
    SO_DISABLE_SCYLLADB_MANAGER_DEPLOYMENT=true \
    timeout --foreground -v 10m "${1}" "${SO_IMAGE}" &
```

hack/.ci/lib/e2e.sh
```diff
@@ -226,8 +286,19 @@ function run-e2e {
...
+  kubectl create -n=e2e secret generic worker-kubeconfigs ${WORKER_KUBECONFIGS[@]/#/--from-file=} --dry-run=client -o=yaml | kubectl_create -f=-
+  worker_kubeconfigs_in_container_paths=$(
+    res=()
+    for key in "${!WORKER_KUBECONFIGS[@]}"; do
+      basename="${WORKER_KUBECONFIGS[$key]##*/}"
+      in_container_path="${basename/#//var/run/secrets/worker-kubeconfigs/}"
+      res+=( "${key}=${in_container_path}" )
+    done
+    IFS=','
+    echo "${res[*]}"
```
czeslavo, line 319: Just to limit the scope?
```suggestion
    IFS=',' echo "${res[*]}"
```

hack/.ci/lib/e2e.sh
```diff
@@ -226,8 +286,19 @@ function run-e2e {
   kubectl create clusterrolebinding e2e --clusterrole=cluster-admin --serviceaccount=e2e:default --dry-run=client -o=yaml | kubectl_create -f=-
   kubectl create -n=e2e pdb my-pdb --selector='app=e2e' --min-available=1 --dry-run=client -o=yaml | kubectl_create -f=-
 
-  kubectl create -n=e2e secret generic kubeconfigs ${KUBECONFIGS[@]/#/--from-file=} --dry-run=client -o=yaml | kubectl_create -f=-
-  kubeconfigs_in_container_path=$( IFS=','; basenames=( "${KUBECONFIGS[@]##*/}" ) && in_container_paths="${basenames[@]/#//var/run/secrets/kubeconfigs/}" && echo "${in_container_paths[*]}" )
+  kubectl create -n=e2e secret generic kubeconfig --from-file=kubeconfig="${KUBECONFIG}" --dry-run=client -o=yaml | kubectl_create -f=-
+
+  kubectl create -n=e2e secret generic worker-kubeconfigs ${WORKER_KUBECONFIGS[@]/#/--from-file=} --dry-run=client -o=yaml | kubectl_create -f=-
+  worker_kubeconfigs_in_container_paths=$(
```
czeslavo: With bash, I find it nice to have elaborate comments that allow skipping parsing some cryptic syntax or googling for its meaning. :sweat_smile: 
```suggestion
  # Create a Secret including _all_ workers' kubeconfigs (including a CP if present in WORKER_KUBECONFIGS).
  kubectl create -n=e2e secret generic worker-kubeconfigs ${WORKER_KUBECONFIGS[@]/#/--from-file=} --dry-run=client -o=yaml | kubectl_create -f=-
  # Build a comma-separated string following a `<worker_name>=<kubeconfig_path_in_container>` format expected by `--worker-kubeconfigs`.
  worker_kubeconfigs_in_container_paths=$(
```

hack/.ci/lib/e2e.sh
```diff
@@ -296,6 +367,7 @@ spec:
     - "--scylladb-manager-agent-version=${SCYLLADB_MANAGER_AGENT_VERSION}"
     - "--scylladb-update-from-version=${SCYLLADB_UPDATE_FROM_VERSION}"
     - "--scylladb-upgrade-from-version=${SCYLLADB_UPGRADE_FROM_VERSION}"
+$(if [[ -n "${worker_kubeconfigs_in_container_paths}" ]]; then echo '    - "--worker-kubeconfigs='"${worker_kubeconfigs_in_container_paths}"'"'; fi)
```
czeslavo: I don't like this echo with spaces hardcoded. Maybe we could make the command args a bash array and build them upfront, and later use them to render YAML somehow?

czeslavo (commented):
A couple of comments on the bash scripts.

hack/.ci/lib/e2e.sh
rzetelskik: Not sure I understand how that would make it better, you'd still be facing the same issues with spacing etc?

hack/.ci/lib/e2e.sh
rzetelskik: If it already exists, then something went wrong, most likely a name collision. I'd definitely prefer to learn about it sooner than later. Are there any benefits to this that I'm not seeing?

hack/.ci/lib/e2e.sh
rzetelskik: KUBECONFIG is used by many tools, so the point is not to have to set it explicitly everywhere (notice KUBECONFIG=... being deleted in multiple places). Unless you mean having both, but then I don't really see where it could help?

hack/.ci/lib/e2e.sh
rzetelskik: done + added comments, please see if it's clear now

hack/.ci/lib/e2e.sh
rzetelskik: technically it could be empty and it would still work, no?

hack/.ci/lib/e2e.sh
rzetelskik: done

hack/.ci/lib/e2e.sh
rzetelskik: done

hack/.ci/lib/e2e.sh
rzetelskik: good point, thanks, done

test/e2e/set/remotekubernetescluster/remotekubernetescluster_healthcheck.go
```diff
@@ -20,43 +20,47 @@ var _ = g.Describe("RemoteKubernetesCluster", func() {
...
+		// Use the meta cluster to simulate the remote clusters.
+		availableClusters := map[string]framework.ClusterInterface{metaCluster.Name(): metaCluster}
+		rkcs = make([]*scyllav1alpha1.RemoteKubernetesCluster, 0, len(availableClusters))
+		rkcClusterMap = make(map[string]framework.ClusterInterface, len(availableClusters))
 
-		rkcs = make([]*scyllav1alpha1.RemoteKubernetesCluster, 0, availableClusters)
-		for idx := range availableClusters {
-			cluster := f.Cluster(idx)
-			userNs, _ := f.Cluster(idx).CreateUserNamespace(ctx)
+		for _, cluster := range availableClusters {
```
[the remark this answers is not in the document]
rzetelskik: leaving as is to avoid unnecessary changes to existing code

hack/.ci/lib/e2e.sh
czeslavo: I imagine running it locally multiple times could be frustrating if you have to manually clean up your local artifacts directory on every run. But I agree there's a good side to it that you mentioned. Leaving it up to you to decide.

hack/.ci/lib/e2e.sh
czeslavo: Yeah, I can see its practicality. IDK, maybe we could improve this lack of explicit meaning by commenting just beside a `WORKER_KUBECONFIGS` declaration that `KUBECONFIG` in this script's context is a  "meta"/"control plane" cluster, also explaining what it exactly means.

czeslavo: Ah, you've already done this. So we're good. :+1:

hack/.ci/lib/e2e.sh
czeslavo: Would it? You mean that without `KUBECONFIG` explicitly set, `kubectl` would use the default location, or do you mean something else?

hack/.ci/lib/e2e.sh
rzetelskik: made it conditional on REENTRANT env var (already exists in other places)

hack/.ci/lib/e2e.sh
czeslavo: I was thinking about something like below. Yeah, it still has to explicitly set spaces (I'd be happy if there was a nice way to not do it), but in the future, when we decide to add a new optional parameter, we do not have to do that separately for every argument. 

```bash

  # Build the array of command arguments
  e2e_command_args=(
    "--skip=${SO_SKIPPED_TESTS}"
    "--kubeconfig=/var/run/secrets/kubeconfig"
    "--loglevel=2"
    "--color=false"
    ...
  )

  # Add optional worker kubeconfigs flag if applicable
  if [[ -n "${worker_kubeconfigs_in_container_paths}" ]]; then
    e2e_command_args+=("--worker-kubeconfigs=${worker_kubeconfigs_in_container_paths}")
  fi

  kubectl_create -n=e2e -f=- <<EOF
apiVersion: v1
kind: Pod
...
spec:
  - name: e2e
    command:
    - "scylla-operator-tests"
    - "run"
    - "${SO_SUITE}"
$(for arg in "${e2e_command_args[@]}"; do echo "    - \"$arg\""; done)
    image: "${SO_IMAGE}"
...
EOF

```

hack/.ci/lib/e2e.sh
rzetelskik: > You mean that without KUBECONFIG explicitly set, kubectl would use the default location, or do you mean something else?

Yes, I mean it would use the default location. I've made it required by the entire lib file, let's see if it breaks anything

hack/.ci/lib/e2e.sh
rzetelskik: made it conditional on REENTRANT env var (already exists in other places)

hack/.ci/lib/e2e.sh
rzetelskik: ok, I don't mind, done

test/e2e/framework/framework.go
```diff
@@ -48,69 +48,91 @@ var _ ClusterInterface = &Framework{}
...
+func newCluster(name, artifactsDir string, restConfig *restclient.Config, namespaceLabels map[string]string) *Cluster {
+	return NewCluster(
+		name,
+		artifactsDir,
+		restConfig,
+		func(ctx context.Context, adminClient kubernetes.Interface, adminClientConfig *restclient.Config) (*corev1.Namespace, Client) {
+			return CreateUserNamespace(ctx, name, namespaceLabels, adminClient, adminClientConfig)
+		},
+	)
+}
```
czeslavo: Why not modify `NewCluster`'s signature to accept the same set of arguments? It's not obvious to me why two separate constructors are needed, where a private one calls a public one.

test/e2e/framework/framework.go
```diff
@@ -48,69 +48,91 @@ var _ ClusterInterface = &Framework{}
...
+	e2eArtifactsDir := ""
 	if len(TestContext.ArtifactsDir) != 0 {
-		f.e2eArtifactsDir = path.Join(TestContext.ArtifactsDir, "e2e")
-		err = os.Mkdir(f.e2eArtifactsDir, 0777)
+		e2eArtifactsDir = path.Join(TestContext.ArtifactsDir, "e2e")
+		err = os.Mkdir(e2eArtifactsDir, 0777)
 		if !os.IsExist(err) {
 			o.Expect(err).NotTo(o.HaveOccurred())
 		}
 	}
```
czeslavo: Maybe we could extract code ensuring the artifact directories are set up to a helper function returning an error, and call it in `NewFramework`, asserting an error didn't occur only once?

czeslavo (commented):
Two more nits in the go code.

test/e2e/framework/framework.go
rzetelskik: I agree that after removing artifacts dir creation it doesn't really make sense, but I'd rather call the constructor with some duplication than do more refactoring in this PR. I'll remove the private method here.

test/e2e/framework/framework.go
rzetelskik: Imo this should be done in a BeforeSuite, doing it in this constructor doesn't make sense, regardless of what we do with the errors. I tried setting it up but failed due to BeforeSuite being misused (as are all ginkgo directives) in our tests. Let's make this a followup.

test/e2e/framework/framework.go
rzetelskik: done

hack/.ci/lib/e2e.sh
rzetelskik: it actually broke it, reverted

czeslavo:
Failed on known flake: https://github.com/scylladb/scylla-operator/issues/2662
/retest

rzetelskik:
> 
> ci/prow/e2e-gke-parallel-clusterip	[9a49430](https://github.com/scylladb/scylla-operator/commit/9a494307b7b6438eff05ce381f7da7a5a0caaa17)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2725/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1940786176306712576)	unknown	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2725). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

known flake
/retest

PR #2726, opened 2025-06-27
[1.17] Make the Git references in docs point to the release branch instead of master

mflendrich:
Before this PR, the gitops guide and the quickstart point to the files on the `master` branch in the operator repo.

After this PR, these documents point to these files on the `v1.17` branch instead.

A separate PR will include this discovered need (to set the `revision` constant upon new minor version release) in the release instructions.


mflendrich:
It doesn't work this way - conf.py is taken from the `master` branch :cry:. I intend to retry with`html_context`.

PR #2727, opened 2025-06-27
[v1.17] Make the Git references in docs point at the release branch instead of `master`

mflendrich:
This PR is a second attempt at what #2726 did not achieve.

Before this PR, the branch part in the GitHub URLs was not correctly rendered, because it always pointed to "master".

The reason is that for all versions, `sphinx-multiversion` evaluates the `conf.py` originating from `master` and ignores the `conf.py` on the branch being rendered. This causes the `revision` value to always come from `master`, resulting in all links pointing to `master`. Which means that any change to the manifests can break all earlier minor versions, which is unacceptable.

This PR achieves 2 things
- replaces the incorrect `revision` field with `env.config.smv_current_version` which has the right branch name :tada: 
- works around the fact that (in our setup) MyST's `substitution` plugin (for reasons beyond my comprehension) only allows "elaborate" Jinja templates (=such that are anything more than trivial resolution of values coming directly from `myst_substitutions`) outside of `code-block`s. We still want to render the code block, so we make the templates render a `code-block` without it being directly interpretable as a `code-block`, which would prevent proper evaluation of the Jinja expression.

I am **very** unproud of the style of this solution, but it achieves:
- correct links (finally!!)
- rich rendering of the code block

at the expense of hackiness that makes me feel bad about it. I think that a reasonable next step is to reach out to the maintainers of our docs stack for a better solution to replace this abomination.

-----------

Notes to reviewers:
- I am more than happy to accept suggestions for a less hacky solution - please try them locally first. Many good ideas have been tried before I arrived at this.
- We need the effect of this PR for 1.18 (because of substantial breaking changes to the manifests in this version). We should reach out to the docs stack experts (cc @dgarcia360 @annastuchlik) to find a more elegant solution going forward, but not at the expense of delaying (or otherwise distracting us from) a timely and correct 1.18 release.
- Once this PR merges, I intend to port it to `master` and to `v1.16` as well.

### DIFF
--- a/docs/source/installation/gitops.md
+++ b/docs/source/installation/gitops.md
@@ -20,6 +20,7 @@ In case you already have a supported version of each of these dependencies insta
 
 #### Cert Manager
 
+% The form of this code block is a workaround to allow resolution of smv_current_version - https://github.com/scylladb/scylla-operator/issues/2752
 {{"""
 BEGIN_CODE_BLOCK
 kubectl apply --server-side -f=https://raw.githubusercontent.com/REPO/BRANCH/examples/third-party/cert-manager.yaml
@@ -44,7 +45,7 @@ done
 
 #### Prometheus Operator
 
-
+% The form of this code block is a workaround to allow resolution of smv_current_version - https://github.com/scylladb/scylla-operator/issues/2752
 {{"""
 BEGIN_CODE_BLOCK
 kubectl -n=prometheus-operator apply --server-side -f=https://raw.githubusercontent.com/REPO/BRANCH/examples/third-party/prometheus-operator.yaml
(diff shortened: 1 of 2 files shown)


docs/source/installation/gitops.md
```diff
@@ -20,13 +20,14 @@ In case you already have a supported version of each of these dependencies insta
 
 #### Cert Manager
 
-:::{code-block} shell
-:substitutions:
-
-kubectl apply --server-side -f=https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/third-party/cert-manager.yaml
-:::
+{{"""
```
zimnx, line 24: Lets leave a comment about this hack and reason why it's there. In case we would need another instance, it would be easier to understand why it's being hacked this way.

mflendrich: I like the idea. But it's not clear to me how to make a comment - this file uses a flurry of markup languages. Should this be a plain HTML `<!-- ... -->` comment or something different?

zimnx: This should work: https://mystmd.org/guide/blocks#comments

zimnx:
Lets first target master, verify it's working after deployment and only then backport.

docs/source/installation/gitops.md
mflendrich: Done (intentionally in a separate commit for visibility) - PTAL

mflendrich:
> Lets first target master, verify it's working after deployment and only then backport.

PTAL https://github.com/scylladb/scylla-operator/pull/2754

zimnx:
It works :rocket:  https://operator.docs.scylladb.com/stable/installation/gitops.html

Lets backport to v1.16 and v1.15 (or drop it)

PR #2728, opened 2025-06-30
Add runtime information collection to must-gather

zimnx:
Adds support for collecting runtime information via commands run in filtered Pods. Initial data includes `nodetool status`, `nodetool gossipinfo`, `df -h`, Scylla process limits, iotune results (io_properites.yaml), and kubelet `cpuManagerPolicy`.

/cc

### DIFF
--- a/pkg/gather/collect/collect.go
+++ b/pkg/gather/collect/collect.go
@@ -1,19 +1,14 @@
 package collect
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"os"
-	"path/filepath"
 
 	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/helpers"
-	oslices "github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	"github.com/scylladb/scylla-operator/pkg/kubeinterfaces"
 	"github.com/scylladb/scylla-operator/pkg/naming"
-	"github.com/scylladb/scylla-operator/pkg/pointer"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -25,6 +20,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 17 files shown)


zimnx:
multidc suite is broken, this will allow for further debugging why it's broken. 

pkg/gather/collect/collect.go
```diff
@@ -332,6 +389,84 @@ func (c *Collector) collectPod(ctx context.Context, u *unstructured.Unstructured
...
+	if err != nil {
+		return stdout, stderr, fmt.Errorf("can't execute command %q on Pod %q: %w", command, naming.ObjRef(pod), err)
+	}
+
+	return stdout, stderr, nil
+}
+
+func (c *Collector) collectContainerCommandOutputFunc(containerName string, filename string, command []string) func(context.Context, *corev1.Pod, *ResourceInfo) error {
+	return func(ctx context.Context, pod *corev1.Pod, resourceInfo *ResourceInfo) error {
+		klog.V(4).InfoS("Collecting ScyllaDB runtime information", "Namespace", pod.Namespace, "Pod", pod.Name, "Command", command)
```
czeslavo:
```suggestion
		klog.V(4).InfoS("Collecting container command output", "Namespace", pod.Namespace, "Pod", pod.Name, "Container", containerName, "Command", command)
```

pkg/gather/collect/collect.go
```diff
@@ -106,6 +120,30 @@ func NewCollector(
...
+		},
+		{
+			Filter:  controllerhelpers.IsScyllaContainerRunning,
+			Collect: c.collectContainerCommandOutputFunc(naming.ScyllaContainerName, "df.log", []string{"df", "-h"}),
+		},
+		{
+			Filter:  controllerhelpers.IsNodeConfigPod,
+			Collect: c.collectContainerCommandOutputFunc(naming.NodeConfigAppName, "kubelet-cpu_manager_state.log", []string{"cat", "/host/var/lib/kubelet/cpu_manager_state"}),
+		},
+	}
```
czeslavo: Just food for thought: maybe we should start extracting collectors for specific resources (e.g., `PodCollector`) to dedicated structs & files so we have better separation of concerns? `Collector` implementation grows, and it's becoming hard to read from my perspective as it touches various resources where some of them have special handling (like Pods).

I think that for readability's and segregation's sake, we could define these runtime collectors inline in `collectPodRuntimeInformation` for now (to avoid refactoring), so they do not leak to the top-level logic (which the `Collector` constructor is from my perspective).

pkg/gather/collect/collect.go
zimnx: done

pkg/gather/collect/collect.go
zimnx: extracted ResourceCollector interface and added PodCollector as first implementation.

pkg/gather/collect/podcollector.go
```diff
@@ -0,0 +1,323 @@
...
+		{
+			Filter:  controllerhelpers.IsNodeConfigPod,
+			Collect: pc.collectContainerCommandOutputFunc(naming.NodeConfigAppName, "kubelet-cpu_manager_state.log", []string{"cat", "/host/var/lib/kubelet/cpu_manager_state"}),
+		},
+	}
+
+	return pc
+}
+
+func retrieveContainerLogs(ctx context.Context, podClient corev1client.PodInterface, destinationPath string, podName string, logOptions *corev1.PodLogOptions) error {
```
czeslavo: I think it will better reflect what this function does and also make it easier to distinguish from very similarly sounding `collectContainerLogs`.
```suggestion
func writeContainerLogsToFile(ctx context.Context, podClient corev1client.PodInterface, destinationPath string, podName string, logOptions *corev1.PodLogOptions) error {
```

pkg/gather/collect/podcollector.go
```diff
@@ -0,0 +1,323 @@
...
+		err = retrieveContainerLogs(ctx, c.corev1Client.Pods(podMeta.Namespace), filepath.Join(logsDir, containerName+".previous"), podMeta.Name, logOptions)
+		if err != nil {
+			return fmt.Errorf("can't retrieve previous pod logs for container %q in pod %q: %w", containerName, naming.ObjRef(podMeta), err)
+		}
+	}
+
+	return nil
+}
+
+func (c *PodCollector) Collect(ctx context.Context, u *unstructured.Unstructured, resourceInfo *ResourceInfo) error {
```
czeslavo: How about reordering the functions in the file so that the public ones come right after the constructor? That way, someone reading it for the first time can "read the code like prose", starting from the top-level functions and jumping to their callees as needed.

czeslavo (commented):
Thanks for extracting the `PodCollector` :tada: Added two more minor nits.

pkg/gather/collect/podcollector.go
zimnx: done

pkg/gather/collect/podcollector.go
zimnx: done

PR #2729, opened 2025-06-30
Use ignition container to validate sysctls in e2e

zimnx:
During e2e, test exec'd into a Pod to validate whether sysctls were applied. It used `scylla` container which no longer has `sysctl` binary. Switched to `scylladb-ignition` which uses `scylla-operator` image having it.

/cc czeslavo


czeslavo (approved):
/lgtm

PR #2731, opened 2025-06-30
Use ephemeralcontainer to validate prlimits in e2e

zimnx:
`prlimit` binary we used in the e2e test to validate if process limit was changed is no longer available in Scylla container image. Instead relying on Scylla container image, we switch to `node-setup` image we control.

/cc czeslavo


czeslavo (approved):
/lgtm

PR #2741, opened 2025-07-03
Add an identity service for ScyllaDBCluster

rzetelskik:
**Description of your changes:** This PR adds an identity/discovery service for clients in the control plane cluster to connect to ScyllaDBCluster nodes without reaching to either of the DCs directly. This is required for the global ScyllaDB Manager instance.

**Requires**:
- [x] https://github.com/scylladb/scylla-operator/issues/2746

Resolves https://github.com/scylladb/scylla-operator/issues/2742

/kind feature
/priority important-soon

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -76,18 +76,11 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
+  - endpoints
+  - namespaces
   - secrets
-  verbs:
-  - get
-  - list
-  - watch
-  - create
-  - delete
-  - update
-  - patch
-- apiGroups:
-  - ""
-  resources:
+  - serviceaccounts
   - services
   verbs:
   - create
@@ -152,18 +145,6 @@ rules:
   - update
   - watch
 - apiGroups:
-  - ""
-  resources:
-  - configmaps
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
   - coordination.k8s.io
   resources:
   - leases
@@ -191,18 +172,6 @@ rules:
   - scylla.scylladb.com
   resources:
   - nodeconfigs
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
-  - serviceaccounts
   verbs:
   - create
   - delete
@@ -242,18 +211,6 @@ rules:
   - batch
   resources:
   - jobs
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
-  - namespaces
   verbs:
   - create
   - delete
@@ -357,6 +314,18 @@ rules:
   - poddisruptionbudgets/finalizers
   verbs:
   - update
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/deploy/operator/00_operator.clusterrole_def.yaml
+++ b/deploy/operator/00_operator.clusterrole_def.yaml
@@ -60,18 +60,11 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
+  - endpoints
+  - namespaces
   - secrets
-  verbs:
-  - get
-  - list
-  - watch
-  - create
-  - delete
-  - update
-  - patch
-- apiGroups:
-  - ""
-  resources:
+  - serviceaccounts
   - services
   verbs:
   - create
@@ -136,18 +129,6 @@ rules:
   - update
   - watch
 - apiGroups:
-  - ""
-  resources:
-  - configmaps
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
   - coordination.k8s.io
   resources:
   - leases
@@ -175,18 +156,6 @@ rules:
   - scylla.scylladb.com
   resources:
   - nodeconfigs
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
-  - serviceaccounts
   verbs:
   - create
   - delete
@@ -226,18 +195,6 @@ rules:
   - batch
   resources:
   - jobs
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
-  - namespaces
   verbs:
   - create
   - delete
@@ -341,3 +298,15 @@ rules:
   - poddisruptionbudgets/finalizers
   verbs:
   - update
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch


rzetelskik:
> 
> ci/prow/e2e-gke-serial	[195c88e](https://github.com/scylladb/scylla-operator/commit/195c88e7828c3bc29f8424c1570efc081f8218ef)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2741/pull-scylla-operator-master-e2e-gke-serial/1941464828417675264)	true	`/test e2e-gke-serial`
> ci/prow/e2e-gke-parallel-clusterip	[195c88e](https://github.com/scylladb/scylla-operator/commit/195c88e7828c3bc29f8424c1570efc081f8218ef)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2741/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1941464828379926528)	true	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2741). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

parallel-clusterip is failing due to https://github.com/scylladb/scylla-operator/issues/2746 and the `RemoteKubernetesCluster finalizer should block deletion when there are ScyllaDBClusters referencing the object` test setting up a ScyllaDBCluster with `ServiceClusterIP` node broadcast address type

rzetelskik:
@zimnx this will keep failing the clusterip suite an unit tests until https://github.com/scylladb/scylla-operator/issues/2746 is merged, but the PR itself is ready for review.
Please pay special attention to the labels/selectors for "local" resources that are introduced (namely `ScyllaDBClusterSelectorLabels` which previously referred to the remote resources). We may want to reconsider whether the labels that are a part of `ScyllaLabels()` now should also be a part of ScyllaDBCluster's selector.

rzetelskik:
rebased, both parallel-clusterip suite and unit tests should be passing now

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+
+// makeLocalServices creates a slice of control-plane Services for the ScyllaDBCluster.
+func makeLocalServices(sc *scyllav1alpha1.ScyllaDBCluster) ([]*corev1.Service, error) {
+	identityService, err := makeIdentityService(sc)
+	if err != nil {
+		return nil, fmt.Errorf("can't make identity Service for ScyllaDBCluster %q: %w", naming.ObjRef(sc), err)
+	}
+
+	return []*corev1.Service{identityService}, nil
+}
```
zimnx: nit: imo this version is more extensible:
```suggestion
func makeLocalServices(sc *scyllav1alpha1.ScyllaDBCluster) ([]*corev1.Service, error) {
	var localServices []*corev1.Service
	
	identityService, err := makeIdentityService(sc)
	if err != nil {
		return nil, fmt.Errorf("can't make identity Service for ScyllaDBCluster %q: %w", naming.ObjRef(sc), err)
	}
	localServices = append(localServices, identityService)

	return localServices, nil
```

pkg/controller/scylladbcluster/resource.go
```diff
@@ -398,6 +398,24 @@ var scyllaDBInterNodeCommunicationPorts = []portSpec{
...
+	{
+		name:     "cql-ssl",
+		protocol: corev1.ProtocolTCP,
+		port:     scylla.DefaultNativeTransportPortSSL,
+	},
+	{
+		name:     "agent-api",
+		protocol: corev1.ProtocolTCP,
+		port:     scylla.DefaultScyllaManagerAgentPort,
+	},
```
zimnx: Agent isn't client communication port. Only CQL and Alternator, which is missing here.

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+
+func makeEndpointSliceForIdentityService(sc *scyllav1alpha1.ScyllaDBCluster, remoteNamespaces map[string]*corev1.Namespace, remoteServiceLister remotelister.GenericClusterLister[corev1listers.ServiceLister], remotePodLister remotelister.GenericClusterLister[corev1listers.PodLister]) ([]metav1.Condition, *discoveryv1.EndpointSlice, error) {
+	var progressingConditions []metav1.Condition
+
+	identityServiceName, err := naming.IdentityServiceName(sc)
+	if err != nil {
+		return progressingConditions, nil, fmt.Errorf("can't get identity Service name for ScyllaDBCluster %q: %w", naming.ObjRef(sc), err)
+	}
+
+	nodeBroadcastType := scyllav1alpha1.BroadcastAddressTypePodIP
```
zimnx: Lets use simillar consts: https://github.com/scylladb/scylla-operator/blob/040eb55d67e224d3c18d95f7bebbcb292b3bb51d/pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go#L310-L311

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+	if sc.Spec.ExposeOptions != nil && sc.Spec.ExposeOptions.BroadcastOptions != nil {
+		nodeBroadcastType = sc.Spec.ExposeOptions.BroadcastOptions.Nodes.Type
```
zimnx: Reconciled endpoints are for client connections. We should use BroadcastOptions.Clients.Type here.

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+func makeLocalEndpointSlices(sc *scyllav1alpha1.ScyllaDBCluster, remoteNamespaces map[string]*corev1.Namespace, remoteServiceLister remotelister.GenericClusterLister[corev1listers.ServiceLister], remotePodLister remotelister.GenericClusterLister[corev1listers.PodLister]) ([]metav1.Condition, []*discoveryv1.EndpointSlice, error) {
+	progressingConditions, identityEndpointSlice, err := makeEndpointSliceForIdentityService(sc, remoteNamespaces, remoteServiceLister, remotePodLister)
+	if err != nil {
+		return progressingConditions, nil, fmt.Errorf("can't make local identity EndpointSlice for ScyllaDBCluster %q: %w", naming.ObjRef(sc), err)
+	}
+
+	return progressingConditions, []*discoveryv1.EndpointSlice{identityEndpointSlice}, nil
+}
+
+func makeEndpointSliceForIdentityService(sc *scyllav1alpha1.ScyllaDBCluster, remoteNamespaces map[string]*corev1.Namespace, remoteServiceLister remotelister.GenericClusterLister[corev1listers.ServiceLister], remotePodLister remotelister.GenericClusterLister[corev1listers.PodLister]) ([]metav1.Condition, *discoveryv1.EndpointSlice, error) {
```
zimnx: makeEndpointSliceForLocalIdentityService

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+	identityService, err := makeIdentityService(sc)
+	if err != nil {
+		return nil, fmt.Errorf("can't make identity Service for ScyllaDBCluster %q: %w", naming.ObjRef(sc), err)
+	}
+
+	return []*corev1.Service{identityService}, nil
+}
+
+// makeIdentityService creates a Service for the ScyllaDBCluster that allows clients to connect to it.
+func makeIdentityService(sc *scyllav1alpha1.ScyllaDBCluster) (*corev1.Service, error) {
```
zimnx: makeLocalIdentityService

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+	name, err := naming.IdentityServiceName(sc)
```
zimnx: Given name consists of hash, this Service is not easy to use and it's intended to be used by user applications. We should either drop hash in case of short name making name predictable, or add special label to be able to discover this Service using constant label selector. I would opt for the latter.

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+		endpoints = append(endpoints, dcEndpoints...)
+	}
+
+	err = apimachineryutilerrors.NewAggregate(errs)
+	if err != nil {
+		return progressingConditions, nil, err
+	}
+
+	return progressingConditions,
+		&discoveryv1.EndpointSlice{
```
zimnx: nit: extract to separate variable

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1080,3 +1098,161 @@ func makeMirroredRemoteSecrets(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1a
...
+}
+
+// makeLocalEndpointSlices creates a slice of control-plane EndpointSlices for the ScyllaDBCluster.
+func makeLocalEndpointSlices(sc *scyllav1alpha1.ScyllaDBCluster, remoteNamespaces map[string]*corev1.Namespace, remoteServiceLister remotelister.GenericClusterLister[corev1listers.ServiceLister], remotePodLister remotelister.GenericClusterLister[corev1listers.PodLister]) ([]metav1.Condition, []*discoveryv1.EndpointSlice, error) {
+	progressingConditions, identityEndpointSlice, err := makeEndpointSliceForIdentityService(sc, remoteNamespaces, remoteServiceLister, remotePodLister)
+	if err != nil {
+		return progressingConditions, nil, fmt.Errorf("can't make local identity EndpointSlice for ScyllaDBCluster %q: %w", naming.ObjRef(sc), err)
+	}
+
+	return progressingConditions, []*discoveryv1.EndpointSlice{identityEndpointSlice}, nil
```
zimnx: nit: ditto

pkg/controller/scylladbcluster/resource_test.go
```diff
@@ -2398,3 +2329,1677 @@ func TestMakeScyllaDBDatacenters(t *testing.T) {
...
+				{
+					Name:  "dc1",
+					Nodes: pointer.Ptr[int32](3),
+				},
+			},
+		},
+	}
+}
+
+func newFakeClusterIndexer(t testing.TB, clusterObjectMap map[string][]apimachineryruntime.Object) func(name string) cache.Indexer {
```
zimnx: s/TB/T

pkg/controller/scylladbcluster/sync.go
```diff
@@ -394,6 +453,45 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
 		}
 	}
 
+	err = controllerhelpers.RunSync(
```
zimnx: This sync function uses loop over struct keeping kind, conditions and syncFn, consider using the same approach for local syncs.

pkg/controller/scylladbcluster/sync.go
```diff
@@ -394,6 +453,45 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
...
+		&status.Conditions,
+		serviceControllerProgressingCondition,
+		serviceControllerDegradedCondition,
+		sc.Generation,
+		func() ([]metav1.Condition, error) {
+			return scc.syncLocalServices(ctx, sc, localServiceMap)
+		},
+	)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("can't sync Services: %w", err))
```
zimnx: local Services

pkg/controller/scylladbcluster/sync.go
```diff
@@ -394,6 +453,45 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
...
+		&status.Conditions,
+		endpointSliceControllerProgressingCondition,
+		endpointSliceControllerDegradedCondition,
+		sc.Generation,
+		func() ([]metav1.Condition, error) {
+			return scc.syncLocalEndpointSlices(ctx, sc, localEndpointSlicesMap, remoteNamespaces)
+		},
+	)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("can't sync EndpointSlices: %w", err))
```
zimnx: ditto

pkg/controller/scylladbcluster/sync.go
```diff
@@ -394,6 +453,45 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
...
+		&status.Conditions,
+		endpointsControllerProgressingCondition,
+		endpointsControllerDegradedCondition,
+		sc.Generation,
+		func() ([]metav1.Condition, error) {
+			return scc.syncLocalEndpoints(ctx, sc, localEndpointsMap, remoteNamespaces)
+		},
+	)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("can't sync Endpoints: %w", err))
```
zimnx: ditto

pkg/naming/labels.go
```diff
@@ -129,12 +129,26 @@ func ScyllaDBClusterDatacenterAnnotations(sc *scyllav1alpha1.ScyllaDBCluster, dc
...
 		maps.Copy(dcAnnotations, dc.Metadata.Annotations)
 	}
+
 	return dcAnnotations
 }
 
 func ScyllaDBClusterSelectorLabels(sc *scyllav1alpha1.ScyllaDBCluster) map[string]string {
 	clusterLabels := make(map[string]string)
 
+	maps.Copy(clusterLabels, ScyllaLabels())
```
zimnx: I'm not a fan of ScyllaLabels, as their keys and values are short and doesn't carry our domain, which seems to be a good practice. I would consider introducing ScyllaExtendedLabels or something along those lines having the domains.

pkg/naming/labels.go
```diff
@@ -129,12 +129,26 @@ func ScyllaDBClusterDatacenterAnnotations(sc *scyllav1alpha1.ScyllaDBCluster, dc
 	if dc.Metadata != nil {
 		maps.Copy(dcAnnotations, dc.Metadata.Annotations)
 	}
+
 	return dcAnnotations
 }
 
 func ScyllaDBClusterSelectorLabels(sc *scyllav1alpha1.ScyllaDBCluster) map[string]string {
```
zimnx: Lets use Local and Remote names to distinguish those and not rely on local being defaulted.

pkg/controller/scylladbcluster/resource_test.go
rzetelskik: I don't mind, but is there a reason? TB is an interface implemented by T

pkg/naming/labels.go
rzetelskik: What are you thinking, something like the following?
```
map[string]string{
	"app.kubernetes.io/name":       "scylla-operator.scylladb.com/scylladb",
	"app.kubernetes.io/managed-by": "scylla-operator.scylladb.com",
}
```

pkg/controller/scylladbcluster/resource.go
rzetelskik: Do you mean reusing `"scylla-operator.scylladb.com/scylla-service-type": "identity"`? Or introducing a dedicated one?

pkg/controller/scylladbcluster/resource.go
rzetelskik: isn't manager server technically a client in that case? Also, how would you call this set then?

pkg/controller/scylladbcluster/resource.go
rzetelskik: ok, done

pkg/controller/scylladbcluster/resource.go
rzetelskik: Good point, I overlooked it. It makes me think, should we also implement a separate helper? `calculateEndpointsForRemoteDCPods` takes nodeBroadcastType as parameter. It also seems like part of the implementation should differ, e.g. should we create the endpoints using member services, or identity services of specific DCs (doesn't apply to BroadcastAddressTypePodIP, of course)

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

pkg/controller/scylladbcluster/sync.go
rzetelskik: done

pkg/controller/scylladbcluster/sync.go
rzetelskik: done

pkg/controller/scylladbcluster/sync.go
rzetelskik: done

pkg/controller/scylladbcluster/sync.go
rzetelskik: done

pkg/controller/scylladbcluster/resource.go
zimnx: I think we need a dedicated selector, although it may consist of this label - we need the selector to return 1 Service in case when the same Kubernetes cluster is shared between control plane and worker.

pkg/controller/scylladbcluster/resource_test.go
zimnx: treat it as nit, it's the first instance of TB in our codebase (excluding few in vendor). 

btw. add a call to t.Helper() inside function

pkg/naming/labels.go
zimnx: yes

pkg/controller/scylladbcluster/resource.go
zimnx: it's kind of client, but it's maintenance traffic, not intended for users. You could have a slice for Manager ports and merge them whenever it's needed. 
Treat it as a nit for Agent, but make sure to add Alternator ports.

pkg/controller/scylladbcluster/resource.go
rzetelskik: used the correct broadcast address type + renamed the parameter in `calculateEndpointsForRemoteDCPods`. Using member services seems correct to me but please take the above comment into consideration @zimnx

pkg/naming/labels.go
rzetelskik: done, changed app name to match group

pkg/naming/labels.go
rzetelskik: done

pkg/controller/scylladbcluster/resource_test.go
rzetelskik: call is already there, changed type

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

pkg/controller/scylladbcluster/resource.go
rzetelskik: done

rzetelskik:
@zimnx I've addressed all your comments. I only noticed a couple of them after I've rebased the branch, so you'll have to look at multiple comparisons - sorry for that.
- https://github.com/scylladb/scylla-operator/compare/e4ea5eb321af48540de53dbdc2d5189b3fd21b43..3ed7056c13b0cf69fd7e2df5b18783434e3a9d61
- https://github.com/scylladb/scylla-operator/compare/e23feb5e998418127e25ad9eec35fdecda746d3e..950025b83c51a5a59ec0cc1c0e33bc1e476f0c7a

pkg/api/scylla/v1alpha1/types_cluster.go
```diff
@@ -7,6 +7,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	ScyllaDBClusterDefaultClientsBroadcastAddressType = BroadcastAddressTypePodIP
+	ScyllaDBClusterDefaultNodesBroadcastAddressType   = BroadcastAddressTypePodIP
```
zimnx, line 13: Lets also change the default values used in remote ScyllaDBDatacenter's' to use these consts.

rzetelskik: done

rzetelskik:
> 
> Test name | Commit | Details | Required | Rerun command
> --- | --- | --- | --- | ---
> ci/prow/e2e-gke-multi-datacenter-parallel | d599895f9af65f4e57db3c5e9325db2458f388f0 | [link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2741/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1942957195507273728) | true | `/test e2e-gke-multi-datacenter-parallel`
> 
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2741). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).
> 
> <details>
> 
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes-sigs/prow](https://github.com/kubernetes-sigs/prow/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).
> </details>
> 

/test images
/retest

Issue #2317, opened 2025-01-13
E2E Failure [Openshift] - Node Setup should propagate mount controller's degraded condition for mount unit configured with a corrupted filesystem [Serial]


zimnx:
```
 Node Setup should propagate mount controller's degraded condition for mount unit configured with a corrupted filesystem [Serial]
github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_disksetup.go:469
  STEP: Creating a new namespace @ 01/13/25 07:00:03.43
  Jan 13 07:00:03.441: INFO: Created namespace "e2e-test-nodesetup-sttcv-0-d4sjb".
  STEP: Waiting for service account token Secret "e2e-user-token" in namespace "e2e-test-nodesetup-sttcv-0-d4sjb". @ 01/13/25 07:00:03.476
I0113 07:00:03.483226       1 cache/reflector.go:351] Caches populated for *v1.Secret from k8s.io/client-go@v0.29.11/tools/cache/reflector.go:229
  STEP: Waiting for default ServiceAccount in namespace "e2e-test-nodesetup-sttcv-0-d4sjb". @ 01/13/25 07:00:03.523
I0113 07:00:03.527954       1 cache/reflector.go:351] Caches populated for *v1.ServiceAccount from k8s.io/client-go@v0.29.11/tools/cache/reflector.go:229
  STEP: Waiting for kube-root-ca.crt in namespace "e2e-test-nodesetup-sttcv-0-d4sjb". @ 01/13/25 07:00:03.529
I0113 07:00:03.533779       1 cache/reflector.go:351] Caches populated for *v1.ConfigMap from k8s.io/client-go@v0.29.11/tools/cache/reflector.go:229
  STEP: Verifying there is at least one scylla node @ 01/13/25 07:00:03.535
  Jan 13 07:00:03.542: INFO: There are 2 scylla nodes
  STEP: Snapshotting object scylla.scylladb.com/v1alpha1, Resource=nodeconfigs "cluster" @ 01/13/25 07:00:03.542
I0113 07:00:03.544927       1 framework/cleanup.go:102] "No existing object found" GVR="scylla.scylladb.com/v1alpha1, Resource=nodeconfigs" Instance="cluster"
  STEP: Deleting object scylla.scylladb.com/v1alpha1, Resource=nodeconfigs "cluster". @ 01/13/25 07:00:03.544
  STEP: Creating NodeConfig @ 01/13/25 07:00:03.546
[...]

Issue #2550, opened 2025-03-13
Flake - Node Setup should successfully start a failed mount unit with a corrupted filesystem when it's overwritten with a clean one [Serial]


zimnx:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2546/pull-scylla-operator-master-e2e-gke-serial/1899784532085706752

### Snippet of what failed.

```
 Node Setup should successfully start a failed mount unit with a corrupted filesystem when it's overwritten with a clean one [Serial]
github.com/scylladb/scylla-operator/test/e2e/set/nodeconfig/nodeconfig_disksetup.go:527
  STEP: Creating a new namespace @ 03/12/25 12:05:14.236
  Mar 12 12:05:14.254: INFO: Created namespace "e2e-test-nodesetup-gflq6-0-f9gm5".
  STEP: Waiting for service account token Secret "e2e-user-token" in namespace "e2e-test-nodesetup-gflq6-0-f9gm5". @ 03/12/25 12:05:14.561
I0312 12:05:14.571804       1 cache/reflector.go:376] Caches populated for *v1.Secret from k8s.io/client-go@v0.32.2/tools/cache/reflector.go:251
  STEP: Waiting for default ServiceAccount in namespace "e2e-test-nodesetup-gflq6-0-f9gm5". @ 03/12/25 12:05:14.594
I0312 12:05:14.621606       1 cache/reflector.go:376] Caches populated for *v1.ServiceAccount from k8s.io/client-go@v0.32.2/tools/cache/reflector.go:251
  STEP: Waiting for kube-root-ca.crt in namespace "e2e-test-nodesetup-gflq6-0-f9gm5". @ 03/12/25 12:05:14.628
I0312 12:05:14.647839       1 cache/reflector.go:376] Caches populated for *v1.ConfigMap from k8s.io/client-go@v0.32.2/tools/cache/reflector.go:251
  STEP: Verifying there is at least one scylla node @ 03/12/25 12:05:14.653
  Mar 12 12:05:14.664: INFO: There are 1 scylla nodes
  STEP: Snapshotting object scylla.scylladb.com/v1alpha1, Resource=nodeconfigs "cluster" @ 03/12/25 12:05:14.665
  STEP: Deleting object scylla.scylladb.com/v1alpha1, Resource=nodeconfigs "cluster". @ 03/12/25 12:05:14.671
[...]

Issue #2370, opened 2025-02-04
[Flake] Scylla Manager integration should register cluster, sync backup tasks and support manual restore procedure [It] using default ScyllaDB version [RequiresObjectStorage]


mflendrich:
_Issue originally authored by **tnozicka** as https://github.com/scylladb/scylla-operator/issues/2061_

### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2055/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1820487766287323136#1:test-build-log.txt%3A1002

### Snippet of what failed.
```
   [FAILED] %!(EXTRA string=Error: create restore target, units and views: init target: look for schema file: get schema file from gcs:so-e490a80d-2b84-4d9d-a784-d0d9d7c27ee3/backup/schema: many schema files with sm_20240805161410UTC snapshot tag: [cluster/6f019377-acd2-4f23-9aad-d6cb21a05008/task_a7b7283e-c7eb-4830-a975-4fd5d16f19af_tag_sm_20240805161410UTC_schema_with_internals.json.gz cluster/82d8dd25-42fc-4c83-87a9-1e76a0e6a15e/task_d24f0a86-f287-4e04-87f8-46220bddda35_tag_sm_20240805161410UTC_schema_with_internals.json.gz]
  Trace ID: EmDCY15VQlG_ct5tWy6W_A (grep in scylla-manager logs)
  )
  Unexpected error:
      <exec.CodeExitError>: 
      command terminated with exit code 1
      {
          Err: <*errors.errorString | 0xc000c7c180>{
              s: "command terminated with exit code 1",
          },
          Code: 1,
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager_object_storage.go:361 @ 08/05/24 16:15:57.416
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster.init.func21.1({{0x0, 0x0}, {0x0, 0x0}, 0x0, 0x0})
    	github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager_object_storage.go:361 +0x4b47
    reflect.Value.call({0x25ea820?, 0xc000b821e0?, 0x13?}, {0x2d32c3a, 0x4}, {0xc000b2be18, 0x1, 0x1?})
[...]

PR #2743, opened 2025-07-04
Fix loop devices manipulations in Node Setup E2Es

czeslavo:

When creating a loop device, it gets assigned `/dev/loop0`, `/dev/loop1` paths, and so on. NodeConfig controller creates a symlink `/dev/loops/{name} -> /dev/loop0` for easier access. Client container, when reading the symlink under `/host/dev/loops/{name}` may get `/dev/loop0` that accidentally always exists in the container, and because of that, tests can falsely succeed. That is not the device on the host, but in the container. 

That explains why we were getting `cannot open /host/dev/loops/disk: No such file or directory` sometimes. It happens when there are more than 1 loop device created on the host, e.g., when our symlink resolves to `/dev/loop1`. It will be the case because such a device doesn’t exist in the container.  Such a condition can occur when `should successfully start a failed mount unit with a corrupted filesystem when it's overwritten with a clean one` test case fails (it's a known flake: https://github.com/scylladb/scylla-operator/issues/2550).

We have to manually resolve symlinks in the container and prepend them with `/host` when we want to manipulate their filesystem from the client Pod.

It was discovered when verifying our serial E2E test suite passes on an Oracle's OKE cluster (https://github.com/scylladb/scylla-operator-team/issues/6).

Should fix https://github.com/scylladb/scylla-operator/issues/2317. 
Related to https://github.com/scylladb/scylla-operator/issues/2550.

/kind failing-test
/priority important-longterm


czeslavo:
/cc @rzetelskik 
as you bumped into this in https://github.com/scylladb/scylla-operator/issues/2317

zimnx (approved):
Great catch! 
lgtm

rzetelskik:
Thanks @czeslavo!
/lgtm
But let's hold for the OpenShift serial suite to see if it actually fixes https://github.com/scylladb/scylla-operator/issues/2317.
/hold

czeslavo:
/test e2e-gke-serial
known flake that I mentioned in the PR description: https://github.com/scylladb/scylla-operator/issues/2550

czeslavo:
Looks like OpenShift cluster got into a degraded state in CI and won't recover. Trying one more time.
/test e2e-openshift-aws-serial

czeslavo:
/test e2e-gke-serial e2e-gke-parallel
Failed on known flakes: 
- https://github.com/scylladb/scylla-operator/issues/2550
- https://github.com/scylladb/scylla-operator/issues/2370

czeslavo:
Rebased after merging https://github.com/scylladb/scylla-operator/pull/2750 to master.

Issue #2490, opened 2025-02-20
ScyllaDBMonitoring: support an external Prometheus instance


mflendrich:
Extend ScyllaDBMonitoring to support an external Prometheus instance by:

- Adding a toggle to disable automatic Prometheus object creation.
- Allowing users to specify an external Prometheus endpoint and credentials.
- Configuring the managed Grafana instance to use the external Prometheus as a data source.

Design considerations: https://github.com/scylladb/scylla-operator/issues/2319#issuecomment-2643287819

## Acceptance Criteria
- [ ] Works on OpenShift
- [ ] Works on regular Kubernetes as an alternative to the current way of working (where ScyllaDBMonitoring spins up its own Prometheus)
- [ ] Tested on CI

buckleyGI:
I got it working thanks to the [ScyllaDBMonitoring ](https://operator.docs.scylladb.com/stable/resources/scylladbmonitorings.html)CRD. It's actually an elegant solution.


This CRD is great to get a Prometheus and grafana up and running quickly.
But it isn't made to connect to an existing prometheus that is already running on your k8s cluster which is the majority of uses cases I believe.

But with a small modification to the Prometheus operator, the metrics and alerting+recording rules, are picked up for your existing prometheus.

Just set these keys for the Prometheus CR which makes it pick up any servicemonitors (metrics) and alerting+recording rules

```
ruleSelectorNilUsesHelmValues: false
serviceMonitorSelectorNilUsesHelmValues: false
```

That's it. The only thing that needs happen is to disable the prometheus and grafana made by the ScyllaDBMonitoring CRD.
This is currently not possible so I still have to find a solution for that. To the Scylla team, is it possible to and "enabled: false" keys to these objects?

So only the following objects of the DBMonitoring CRD should be generated in this case (the elegant part is that this is only a subset of what ScyllaDBMonitoring does today, there is nothing new to develop. Thanks to the discoverability and compostability of k8s CRD's)

```
prometheusrule.monitoring.coreos.com/scylla-sed-alerts
prometheusrule.monitoring.coreos.com/scylla-sed-latency
prometheusrule.monitoring.coreos.com/scylla-sed-table

servicemonitor.monitoring.coreos.com/scylla-sed-node-exporter
servicemonitor.monitoring.coreos.com/scylla-sed-scylladb
servicemonitor.monitoring.coreos.com/scylla-sed-service-monitor
```

This leaves the dozen or so grafana dashboards:

```
[...]

Issue #2663, opened 2025-05-30
ScyllaDBCluster finalizer [It] should delete remote Namespaces when ScyllaDBCluster is deleted [MultiDatacenter]


rzetelskik:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928147965210267648#1:test-build-log.txt%3A2628

### Snippet of what failed.

```
  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000b660e0>: 
      waiting has been interupted ([false]): timed out waiting for the condition
      {
          msg: "waiting has been interupted ([false]): timed out waiting for the condition",
          err: <wait.errInterrupted>{
              cause: <*errors.errorString | 0xc0000587e0>{
                  s: "timed out waiting for the condition",
              },
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_finalizer.go:39 @ 05/29/25 19:09:26.569
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter.init.func4.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_finalizer.go:39 +0x3b2
```

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928122130860347392

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928100594178854912

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928077092025536512

rzetelskik:
[...]

Issue #2661, opened 2025-05-30
[Flake] Multi datacenter ScyllaDBCluster [It] should boostrap all datacenters and form single cluster [MultiDatacenter]


rzetelskik:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928247846067245056#1:test-build-log.txt%3A2839

### Snippet of what failed.

```
  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000d9a0e0>: 
      waiting has been interupted ([false]): timed out waiting for the condition
      {
          msg: "waiting has been interupted ([false]): timed out waiting for the condition",
          err: <wait.errInterrupted>{
              cause: <*errors.errorString | 0xc0004c1ee0>{
                  s: "timed out waiting for the condition",
              },
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_boostrap.go:35 @ 05/30/25 01:46:08.703
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter.init.func1.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_boostrap.go:35 +0x3af
```

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928147965210267648#1:test-build-log.txt%3A2521

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928147965210267648

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928122130860347392

rzetelskik:
[...]

PR #2750, opened 2025-07-07
Do not install prometheus-operator in OpenShift

czeslavo:

It's a fix for broken CI logic that was previously working fine because we had `REENTRANT` set (https://github.com/scylladb/scylla-operator/commit/e6e4e3422b093ee3e73f28247fb960e39ae953ef removed that) for OpenShift jobs (so instead of `kubectl create` we were running `kubectl apply`).

Prometheus Operator installation and verification are completely skipped for OpenShift clusters - it's fine as long as we do not support external Prometheus instances in `ScyllaDBMonitoring` (https://github.com/scylladb/scylla-operator/issues/2490) and do not test that on OpenShift. 

Extracted from https://github.com/scylladb/scylla-operator/pull/2743.

### DIFF
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -34,8 +34,13 @@ if [[ "${SO_DISABLE_SCYLLADB_MANAGER_DEPLOYMENT}" != "true" ]]; then
   cp ./deploy/manager/dev/*.yaml "${ARTIFACTS_DEPLOY_DIR}/manager"
 fi
 
+if [[ -n "${SO_DISABLE_PROMETHEUS_OPERATOR:-}" ]]; then
+  echo "Skipping copying prometheus-operator manifests to ${ARTIFACTS_DEPLOY_DIR}"
+else
+  cp ./examples/third-party/prometheus-operator/*.yaml "${ARTIFACTS_DEPLOY_DIR}/prometheus-operator"
+fi
+
 cp ./deploy/operator/*.yaml "${ARTIFACTS_DEPLOY_DIR}/operator"
-cp ./examples/third-party/prometheus-operator/*.yaml "${ARTIFACTS_DEPLOY_DIR}/prometheus-operator"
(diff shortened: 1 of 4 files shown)


zimnx (approved):
/lgtm

czeslavo:
/test e2e-gke-multi-datacenter-parallel
Flake: https://github.com/scylladb/scylla-operator/issues/2661

hack/ci-deploy.sh
```diff
@@ -83,7 +83,13 @@ if [[ -n ${SCYLLA_OPERATOR_FEATURE_GATES+x} ]]; then
     yq e --inplace '.spec.template.spec.containers[0].args += "--feature-gates="+ env(SCYLLA_OPERATOR_FEATURE_GATES)' "${ARTIFACTS_DEPLOY_DIR}/operator/50_operator.deployment.yaml"
 fi
 
-kubectl_create -n prometheus-operator -f "${ARTIFACTS_DEPLOY_DIR}/prometheus-operator"
+# Do not install prometheus-operator if the platform already has it (e.g., OpenShift).
+if [[ -n "${SO_DISABLE_PROMETHEUS_OPERATOR:-}" ]]; then
+  echo "Skipping prometheus-operator deployment"
```
rzetelskik, line 93: I think we should also skip copying the manifests to the deploy dir

czeslavo: Assumed it doesn't change much if we do not install them, but agree it's better to be fully consistent and not have non-installed manifests in the CI artifacts. Fixed.

czeslavo:
/test e2e-gke-parallel-clusterip
Flake: https://github.com/scylladb/scylla-operator/issues/2096

czeslavo:
/test e2e-gke-multi-datacenter-parallel
Flake: https://github.com/scylladb/scylla-operator/issues/2663

rzetelskik:
@zimnx shall we override the multi-datacenter job? It shouldn't be affected

Issue #2563, opened 2025-03-25
add flexibility to define a custom/alternate namespace for the operator vs scylla-operator with the helm operator


tluck:
### What should the feature do?

allow one to easily use an alternate namespace for the operator deployment.
e.g.: 

helm install scylla-operator scylla-operator --create-namespace --namespace myspecificname -f scylla-operator.yaml

### What is the use case behind this feature?

some organizations have specific naming requirements for their k8s environments and will not allow the default name ("hard-coded")

mflendrich:
Reopening because we're seeing a pattern in user requests for control over the namespace name.

Issue #2347, opened 2025-01-20
OpenShift M4: Published into `certified-operators`


mflendrich:
## Acceptance Criteria
- [ ] The release processes for dependencies/operands ensure that every version (including patch) is uploaded as a certified image.
- [ ] The upstream applications necessary as dependencies/operands are repackaged and certified as necessary. A procedure exists for repackaging future versions similarly.
- [ ] The release process includes the steps that publish Operator into the certified-operators registry.
- [ ] A successful release of the above has completed, leading to a certified operator being available.

```[tasklist]
```

mflendrich:
First attempt - using `quay.io/scylladb-dev/olm-bundle-operator:v1.17.2`

OKD 4.19.0-okd-scos.7
Preflight 1.12.1

Invocation

```console
env PFLT_INDEXIMAGE=quay.io/scylladb-dev/olm-bundle-operator:catalog-latest KUBECONFIG=$HOME/.kube/config-openshift-4.19 preflight check operator quay.io/scylladb-dev/olm-bundle-operator:v1.17.2 > result.json
```

Preflight result

```json
{
    "image": "quay.io/scylladb-dev/olm-bundle-operator:v1.17.2",
    "passed": false,
    "certification_hash": "d41d8cd98f00b204e9800998ecf8427e",
    "test_library": {
        "name": "github.com/redhat-openshift-ecosystem/openshift-preflight",
        "version": "1.12.1",
        "commit": "1adc5822eb992c19c97ce16b83b0158406c4ada0"
    },
    "results": {
        "passed": [
            {
                "name": "ScorecardBasicSpecCheck",
                "elapsed_time": 6231,
                "description": "Check to make sure that all CRs have a spec block."
            }
        ],
        "failed": [
            {
                "name": "ValidateOperatorBundle",
                "elapsed_time": 791,
                "description": "Validating Bundle image that checks if it can validate the content and format of the operator bundle",
                "help": "Check ValidateOperatorBundle encountered an error. Please review the preflight.log file for more information.",
                "suggestion": "Valid bundles are defined by bundle spec, so make sure that this bundle conforms to that spec. More Information: https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md",
                "knowledgebase_url": "https://sdk.operatorframework.io/docs/olm-integration/tutorial-bundle/",
                "check_url": "https://sdk.operatorframework.io/docs/olm-integration/tutorial-bundle/"
            },
            {
[...]

Issue #1663, opened 2023-12-28
Limit scylla operator to namespace scoped


sourabhsarkar:
### What should the feature do?

Hi, I have a requirement where in a single kubernetes cluster multiple teams can have separate scylla operator deployments and their own management. I am trying to make the operator namespace scoped but not finding any such namespaces to watch switch in the crd. Any inputs related to the same would be helpful.

### What is the use case behind this feature?

In multi-tenant clusters limiting scylla operator to particular namespaces to watch can benefit tenants manage their own scylla operator and cluster deployments.

tnozicka:
As I've already mentioned operator is a cluster level extension that needs it's CustomResourceDefinitions updated with it. You can't run multiple operators using the same CRD as the version would collide on the global resource.
https://forum.scylladb.com/t/can-the-scylla-operator-be-installed-namespace-scoped/1137/2?u=tnozicka

> In multi-tenant clusters limiting scylla operator to particular namespaces to watch can benefit tenants manage their own scylla operator and cluster deployments.

Can you explain why you can't install the operator as cluster admin (you need it to install and update the CRD itself anyways) and grant user permissions to create ScyllaCluster CRs?

steve-gray:
@sourabhsarkar there have been multiple threads like this, including one where I tried at length to get @tnozicka to see any other point of view on this, but unfortunately they did not subscribe to any of the points raised, and ultimately declined to make the tiny changes involved to support this. I've tried to find it, but I assume it got purged.

Specifically:

- Having the operator be cluster-wide is not a best practice, and its perfectly possible to have one thats configured to run at restricted scopes. Many other operators permit this for this exact reason (in fact, I'm hard pressed to find one that *doesn't*)

- The forced issue here prevents people having multiple instances, with different configurations/scopes/permissions, and takes the choice away from cluster operators to segregate access, even where they do control the whole cluster.

- The argument about "CRD versions" is a straw-man, and it's easily something that can just be the cluster administrators concern regardless of the operator versions in use (i.e. if i want to update the CRDs, I need to update all my operators, but thats really no big deal). 

- Best practice is actually to divorce the helm installation for CRDs from the main software installation, and be diligent about compatibility considerations. This issue is actually applicable even in a single install, as updating CRDs in a way thats not backward-compatible can impact even a single global install.

- The software has various hard-coded forced assumptions about namespace names, precluding any scenario where the person deploying the software is doing so in a cluster where they have to follow externally set naming conventions.

[...]

Issue #1579, opened 2023-11-18
Allow setting custom namespaces for charts


deefdragon:
### What should the feature do?

The operator and manager (at minimum) are currently unable to be changed the namespace that they are deployed in. This breaks my naming convention that I am using to manage how my operators are deployed. 

Users should be able to apply a namespace value, or the included helm flag should be used for the namespace.

### What is the use case behind this feature?

This is best used for keeping a consistent naming convention for namespaces that operators are deployed to (including putting them all in the same namespace). Outside of cleaner namespaces, this also makes it easier to apply policies on the resources.

### Anything else we need to know?

This is a surprising overload given most helm charts take in and honor the built-in --namespace flag automatically.

tnozicka:
dupe of #987

deefdragon:
As has been articulated in #667 and #1305, and is reflected in the fact that this, #758, #987 and #1305 exists, this is a poor decision and goes against the adopted standard.

I am not going to dictate to scylla how to run this project, but I _am_ going to say that breaking with the standards that every other operator follows, _even after receiving reasonable pushback_, does not reflect well for scylla. 

I installed the scylla operator to mess around with scylla, its tooling, and the environment around it. I wished to learn about and examine it for potential use cases for my work. Given the situation here, I will no longer be doing so at this time.

tnozicka:
Definitely your choice, but beyond just pointing to "I want to configure everything" and given this is opensource you don't pay for, it might be worth to change it into how I can help to solve this differently. At this point making the namespaces configurable breaks other features like:
- I think there is still some internal DNS that uses cross namespace domains that depend on the namespace
- must-gather collection that needs to know which system namespaces to collect (would need discovery API otherwise)
- operator creates other namespaces, like `scylla-operator-node-tuning` that are also fixed (had to be an API filed somewhere)
- cluster level objects like cluster roles also have prefixes
- probably more

So, the price to do all of this is high compared to, "I just want to have this configurable" and the value it brings is orders of magnitude lower. Operators are platform level extensions and I do not agree all of them make this configurable, probably the simple ones do because they don't face these challenges. There are fixed namespaces like `default`, `kube-system` and others. To pick an enterprise example about a few dozens of operators in OCP, all of them start with `openshift-` prefix that you can't change.

I am not saying this can't be done, and you are welcome to address it as it's opensource, but I'd rather invest my time into something that helps with actual issues then just naming when I decide how I spend my time.

[...]

PR #2755, opened 2025-07-07
Update helm templates for operator and manager to support custom namespace

debMan:
# Update helm templates for operator and manager to support custom namespace

This PR updates the helm templates for the `scylla-operator` and `scylla-manager` to be able to be installed in any custom namespace defined in the `helm` cli. 

Part of #2563 

**PS:**
We have tested this customized approach in an openshift cluster and seems working well.

### DIFF
--- a/helm/scylla-manager/templates/manager_configmap.yaml
+++ b/helm/scylla-manager/templates/manager_configmap.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: scylla-manager-config
-  namespace: scylla-manager
+  namespace: {{ .Release.Namespace }}
 data:
   scylla-manager.yaml: |-
     http: :5080
--- a/helm/scylla-manager/templates/manager_deployment.yaml
+++ b/helm/scylla-manager/templates/manager_deployment.yaml
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: scylla-manager
-  namespace: scylla-manager
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "scylla-manager.labels" . | nindent 4 }}
 spec:
--- a/helm/scylla-manager/templates/manager_networkpolicy.yaml
+++ b/helm/scylla-manager/templates/manager_networkpolicy.yaml
@@ -1,7 +1,7 @@
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  namespace: scylla-manager
+  namespace: {{ .Release.Namespace }}
   name: scylla-manager-to-scylla-pod
 spec:
   policyTypes:
@@ -14,4 +14,4 @@ spec:
   - from:
     - namespaceSelector:
         matchLabels:
-          kubernetes.io/metadata.name: scylla-manager
+          kubernetes.io/metadata.name: {{ .Release.Namespace }}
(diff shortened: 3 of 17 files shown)


rzetelskik:
/ok-to-test
But for the record there's no way you'll retain all functionality without wiring the customizable namespaces in the controllers.
Just to provide an example, the global ScyllaDB Manager instance is expected to be in the reserved namespace - https://github.com/scylladb/scylla-operator/blob/4b853a05c0d63b225e22ccc095ab23a8ac2e7874/pkg/controllerhelpers/scylladbmanager.go#L18.
Here are some pointers for more context:
- https://github.com/scylladb/scylla-operator/issues/1579
- https://github.com/scylladb/scylla-operator/issues/1663

You'll probably find at least a few more related issues/PRs.

debMan:
Oh, get it. Thanks for the clarification.
We are running a multi-tenant OpenShift cluster and serving a broad range of users. All available operators for end-users are deployed in namespaces with a custom prefix in their names. This is required for applying cluster-wide policies for infra namespaces.

I've read  #667, but it can the reasons mentioned there can be implemented another way, like specifying the operator namespace in the manager components with a flag or env variable, or in the `must-gather` command.

Is there any plan for this?

rzetelskik:
> I've read https://github.com/scylladb/scylla-operator/pull/667, but it can the reasons mentioned there can be implemented another way, like specifying the operator namespace in the manager components with a flag or env variable, or in the must-gather command.
> 
> Is there any plan for this?

cc @mflendrich

mflendrich:
This is not the first time that this request has bubbled up - I empathize with the "multi-tenant / structured environment" requirement; if we have a consistent & correct approach to the obstacle that operator relies on the hardcoded namespace names, I see no reason why we would force this restriction on users.

Also, freedom to choose the namespace for Operator and Manager may be necessary for smooth cooperation with OLM. #2347

@rzetelskik  @zimnx - if you see other obstacles (than the need to plumb through the namespace name, as @rzetelskik pointed out above), let's bring them to light.

Once those obstacles are ironed out,, I don't see why we wouldn't want a contribution for this issue. There is no current timeline for the team to work on this, but I would be glad to see it solved by a community PR.

rzetelskik:
@mflendrich let's keep one of the issues open as a tracker, all previously created ones were being closed religiously and it seems like it's making it harder to get context that way. I suppose we could aggregate the related discussions? 

As for the implementation, for a change of this size I'd expect to see an [enhancement proposal](https://github.com/scylladb/scylla-operator/tree/91fa083dca872e891016110725f96d6d54837778/enhancements) first to be reasonably sure everything is covered.

mflendrich:
> @mflendrich let's keep one of the issues open as a tracker, all previously created ones were being closed religiously and it seems like it's making it harder to get context that way. I suppose we could aggregate the related discussions?

I left one open: https://github.com/scylladb/scylla-operator/issues/2563

> As for the implementation, for a change of this size I'd expect to see an [enhancement proposal](https://github.com/scylladb/scylla-operator/tree/91fa083dca872e891016110725f96d6d54837778/enhancements) first to be reasonably sure everything is covered.

I don't disagree that we want an understanding that might or might not command an enhancement proposal, but let's not start the discussion with throwing red tape around and instead try to brainstorm about where might be dragons.

rzetelskik:
> I don't disagree that we want an understanding that might or might not command an enhancement proposal, but let's not start the discussion with throwing red tape around and instead try to brainstorm about where might be dragons.

[This comment](https://github.com/scylladb/scylla-operator/issues/1579#issuecomment-1818788460) names a few.

PR #2757, opened 2025-07-08
test sm 3.6.0

Michal-Leszczynski:
This PR is created just for triggering operator CI against SM 3.6.0 candidate.

/cc
/test all


Michal-Leszczynski:
Failure is not related to SM as `ci/prow/e2e-gke-serial` does not test SM:
```
 Node Setup [It] should successfully start a failed mount unit with a corrupted filesystem when it's overwritten with a clean one [Serial] 
```

Issue #2664, opened 2025-05-30
Multi datacenter ScyllaDBCluster [It] should mirror ConfigMaps and Secrets referenced by ScyllaDBCluster into remote datacenters [MultiDatacenter]


rzetelskik:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928147965210267648#1:test-build-log.txt%3A2735

### Snippet of what failed.

```

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000c000e0>: 
      waiting has been interupted ([false]): timed out waiting for the condition
      {
          msg: "waiting has been interupted ([false]): timed out waiting for the condition",
          err: <wait.errInterrupted>{
              cause: <*errors.errorString | 0xc000449c40>{
                  s: "timed out waiting for the condition",
              },
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_config.go:156 @ 05/29/25 19:09:23.816
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter.init.func2.1()
    	github.com/scylladb/scylla-operator/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_config.go:156 +0x1bda
```

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928147965210267648

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928077092025536512

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2658/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1928051242148302848

rzetelskik:
[...]

PR #2758, opened 2025-07-08
Add per-worker object storage support to E2E framework 

czeslavo:

Adds support for per-worker object storage settings to the E2E framework by adding the following flags to the test runner:

- `worker-gcs-service-account-key-paths`
- `worker-s3-credentials-file-paths`
- `worker-object-storage-buckets`

Resolves #


czeslavo:
/retest 
Flakes: https://github.com/scylladb/scylla-operator/issues/2664, https://github.com/scylladb/scylla-operator/issues/2370

pkg/cmd/tests/object_storage_options.go
```diff
@@ -0,0 +1,142 @@
...
+		workerS3CredentialsFiles:    map[string][]byte{},
+	}
+}
+
+func (oso *ObjectStorageOptions) AddFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVarP(&oso.ObjectStorageBucket, "object-storage-bucket", "", oso.ObjectStorageBucket, "Name of the object storage bucket.")
+	cmd.PersistentFlags().StringVarP(&oso.GCSServiceAccountKeyPath, "gcs-service-account-key-path", "", oso.GCSServiceAccountKeyPath, "Path to a file containing a GCS service account key.")
+	cmd.PersistentFlags().StringVarP(&oso.S3CredentialsFilePath, "s3-credentials-file-path", "", oso.S3CredentialsFilePath, "Path to the AWS credentials file providing access to the S3 bucket.")
+
+	cmd.PersistentFlags().StringToStringVarP(&oso.WorkerGCSServiceAccountKeyPaths, "worker-gcs-service-account-key-paths", "", oso.WorkerGCSServiceAccountKeyPaths, "Map of worker name to GCS service account key paths.")
```
rzetelskik: I wonder, does it call for some additional comment explaining that the bucket should be available from the worker cluster (or the reason for having buckets per cluster and how do they relate to the "global" one)? The boundary between infra, implementation and manager-specific stuff is somewhat blurred, so I'm hesitant. It's clear to me as I know the context well, but I'm wondering if it will also be clear for others.

pkg/cmd/tests/object_storage_options.go
```diff
@@ -0,0 +1,142 @@
...
+	if len(oso.ObjectStorageBucket) > 0 && len(oso.GCSServiceAccountKeyPath) == 0 && len(oso.S3CredentialsFilePath) == 0 {
+		errors = append(errors, fmt.Errorf("either gcs-service-account-key-path or s3-credentials-file-path must be set when object-storage-bucket is provided"))
+	}
+	if len(oso.GCSServiceAccountKeyPath) > 0 && len(oso.S3CredentialsFilePath) > 0 {
+		errors = append(errors, fmt.Errorf("gcs-service-account-key-path and s3-credentials-file-path can't be set simultanously"))
+	}
+
+	if len(oso.WorkerGCSServiceAccountKeyPaths) > 0 || len(oso.WorkerS3CredentialsFilePaths) > 0 || len(oso.WorkerObjectStorageBuckets) > 0 {
+		if len(oso.GCSServiceAccountKeyPath) > 0 || len(oso.S3CredentialsFilePath) > 0 || len(oso.ObjectStorageBucket) > 0 {
+			errors = append(errors, fmt.Errorf("worker-* flags cannot be used with single bucket flags"))
```
rzetelskik, line 64: What's the exact reasoning behind this? You could technically specify buckets for both the control plane cluster and worker clusters. In fact it might even make sense for manager itself, although admittedly it wouldn't make much sense with our CI tooling now (due to lack of multi-region buckets).

test/e2e/framework/framework.go
```diff
@@ -263,6 +263,24 @@ func (f *Framework) GetS3CredentialsFile() []byte {
...
+
+func NewWorkerClusterObjectStorageSettings(bucketName string, credentials []byte) WorkerClusterObjectStorageSettings {
+	return WorkerClusterObjectStorageSettings{
+		BucketName:  bucketName,
+		Credentials: credentials,
+	}
+}
+
+func (f *Framework) GetObjectStorageSettingsForWorkerCluster(workerName string) WorkerClusterObjectStorageSettings {
+	settings, ok := TestContext.WorkerObjectStorageSettings[workerName]
```
rzetelskik: can we make it an attribute of the framework instead of using the global TestContext here?

test/e2e/framework/framework.go
```diff
@@ -263,6 +263,24 @@ func (f *Framework) GetS3CredentialsFile() []byte {
 	return TestContext.S3CredentialsFile
 }
 
+type WorkerClusterObjectStorageSettings struct {
```
rzetelskik: Imo this needs to be aligned (and shared) with how we handle the "global" object storage now, iiuc there is nothing different about worker clusters' object storage. See the framework methods directly above. I'd imagine the workflow being identical for the global vs worker object storage. Having said that, I'm fine with changing how we work with the global one now, so introducing an interface/struct is fine.

test/e2e/framework/framework.go
```diff
@@ -263,6 +263,24 @@ func (f *Framework) GetS3CredentialsFile() []byte {
...
+	BucketName  string
+	Credentials []byte // These are raw credentials needed by any of supported object storage backends (e.g., GCS or S3).
```
rzetelskik: I'm not sure making it generic is a good way forward - you might limit yourself in case additional methods are required for missing storage backends (e.g. iirc minio would require extra config changes for manager agent, so this can't be 100% provider agnostic I'm afraid) . See the existing methods for GCS and S3 and how they are called based on type.

pkg/cmd/tests/object_storage_options.go
```diff
@@ -0,0 +1,142 @@
...
+	"os"
+
+	"github.com/scylladb/scylla-operator/pkg/util/maps"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"github.com/spf13/cobra"
+	apimachineryutilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// ObjectStorageOptions holds the options for object storage configuration in tests.
+type ObjectStorageOptions struct {
```
rzetelskik, line 13: nit: could this be implemented similarly to `ClientConfig` and `MultiDatacenterClientConfig`? So instead of having maps of names/credentials we could have a map of structs and the embedded "control-plane" struct?

pkg/cmd/tests/object_storage_options.go
```diff
@@ -0,0 +1,142 @@
...
+		if len(oso.GCSServiceAccountKeyPath) > 0 || len(oso.S3CredentialsFilePath) > 0 || len(oso.ObjectStorageBucket) > 0 {
+			errors = append(errors, fmt.Errorf("worker-* flags cannot be used with single bucket flags"))
+		}
+		if len(oso.WorkerGCSServiceAccountKeyPaths) > 0 && len(oso.WorkerS3CredentialsFilePaths) > 0 {
+			errors = append(errors, fmt.Errorf("only one of worker-gcs-service-account-key-paths or worker-s3-credentials-file-paths can be set"))
+		}
+		if len(oso.WorkerObjectStorageBuckets) != 0 && len(oso.WorkerGCSServiceAccountKeyPaths) == 0 && len(oso.WorkerS3CredentialsFilePaths) == 0 {
+			errors = append(errors, fmt.Errorf("either worker-gcs-service-account-key-paths or worker-s3-credentials-file-paths must be set when worker-object-storage-buckets is provided"))
+		}
+		if len(oso.WorkerGCSServiceAccountKeyPaths) > 0 && !maps.SameKeys(oso.WorkerObjectStorageBuckets, oso.WorkerGCSServiceAccountKeyPaths) {
```
rzetelskik: nit: not a fan of adding a function with such narrow usage to our libs, personally I'd rather we started moving towards using actual std libs more. This could easily be achieved with a composition of functions from slices/maps. How about something along these lines?

```
workerObjectStorageBucketsSortedKeys := slices.Sorted(maps.Keys(oso.WorkerObjectStorageBuckets))
workerGCSServiceAccountKeyPathsSortedKeys := slices.Sorted(maps.Keys(oso.WorkerGCSServiceAccountKeyPaths))

slices.Equal(workerObjectStorageBucketsSortedKeys, workerGCSServiceAccountKeyPathsSortedKeys)
```

Having said that I don't mind adding local wrappers.

pkg/cmd/tests/options.go
```diff
@@ -235,13 +208,21 @@ func (o *TestFrameworkOptions) Validate(args []string) error {
 		}
 	}
 
+	if len(o.WorkerObjectStorageBuckets) > 0 {
+		if !maps.SameKeys(o.WorkerKubeconfigs, o.WorkerObjectStorageBuckets) {
```
rzetelskik: nit: not 100% sure about this, but shouldn't this fail only when you try to get a bucket for a worker cluster (on a framework level)? Even from ScyllaDB Manager standpoint you could define regional buckets only for a subset of DCs and then use a multi-region one for the others

pkg/cmd/tests/object_storage_options.go
czeslavo: Makes sense - I'll make them follow the same pattern as the `--worker-kubeconfigs` usage string does

> Map of worker cluster identifiers to kubeconfig paths. Used in multi-datacenter setups.

pkg/cmd/tests/object_storage_options.go
czeslavo: I believe that we should allow specifying only `Worker*` variants or the "global" ones to make the use cases/infra setups (multi-DC vs. single-DC) clearly separated. I'm assuming a "control plane" cluster is always included in the workers set, so our requirement for configuring a manager agent in every DC with its respective bucket is doable with the API I proposed.

test/e2e/framework/framework.go
czeslavo: I was trying not to break the convention we have with using the global test context. The Framework is created in every test case, and IMO it doesn't make sense to populate it every time separately for such settings that should be set only once.

rzetelskik: it doesn't make sense for the framework's methods to serve as accessors for the global struct's attributes either imo, but I do understand you're expanding on what's already there unfortunately

pkg/cmd/tests/object_storage_options.go
czeslavo: I don't agree. I prefer to have a well-suited helper function that does exactly what I need in such cases. It's easier to read as it directly says what the intention is (I want to check my maps have the `SameKeys`). Your proposal can achieve a similar effect, but IMO it's less readable +  less performant (not a big deal, but still). This helper was needed in 3 places, which I think perfectly justifies extracting such a piece of code. Its maintenance cost in the future should be close to zero (given it has no external dependencies).

rzetelskik: > less performant

Not even asymptotically in the general case and definitely not in practice for so few entries

> This helper was needed in 3 places

3 places in one function, in which case a local helper is enough 

> Its maintenance cost in the future should be close to zero

imo the maintenance cost, and the reason I'm reluctant to expand on these library funcs, is having a hundred different ways to do a simple thing

pkg/cmd/tests/object_storage_options.go
rzetelskik: > I'm assuming a "control plane" cluster is always included in the workers set,

This is more about artificially limiting our options here - with manager you could specify regional buckets for a subset of DCs and use a multi-region one for the others. But let's start with this and expand if we ever need to.

pkg/cmd/tests/options.go
czeslavo: Makes sense, let's open for more use cases.

rzetelskik: since it's closely related to https://github.com/scylladb/scylla-operator/pull/2758#discussion_r2195193218 - I'm fine with leaving as is as well, I see some benefit in starting with a stricter setup

pkg/cmd/tests/object_storage_options.go
czeslavo: I'll go for an explicit field for the control plane instead of embedding, but it's a good idea.

test/e2e/framework/framework.go
czeslavo: I'll unify worker/CP handling so this type is used in both cases (renamed).

test/e2e/framework/framework.go
czeslavo: Agree, will enrich this struct with `storageType` field and separate fields for provider-specific stuff.

pkg/util/maps/keys.go
```diff
@@ -0,0 +1,14 @@
+package maps
+
+// SameKeys returns true if both maps have the same set of keys.
+func SameKeys[K comparable, T1 any, T2 any](a map[K]T1, b map[K]T2) bool {
```
rzetelskik: nit: EqualKeys

pkg/cmd/tests/object_storage_options.go
czeslavo: It was 3 places in 2 functions. If it was 1 function in the first place, I wouldn't extract a helper, agree. After recent changes (dropping worker kubeconfigs vs. worker storage buckets validation), I'm happy to make this a local helper now.

pkg/cmd/tests/object_storage_options.go
rzetelskik: Should we stick to Cluster/WorkerCluster naming convention? ControlPlane makes more sense in the multi-dc setup, but it makes none in single-dc. Plus it's somewhat already established in the framework and scripts. I feel like explaining in the flag descriptions should be enough? Wdyt?

test/e2e/framework/storage.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// S3CredentialsFile returns the AWS credentials in JSON format.
+// It's up to the caller to ensure that this is only called when Type is ObjectStorageTypeS3.
+func (s ClusterObjectStorageSettings) S3CredentialsFile() []byte {
+	return s.s3Credentials
+}
+
+// LocationForScyllaManager returns a `<provider>:<location>` string for Scylla Manager configuration.
+func (s ClusterObjectStorageSettings) LocationForScyllaManager() string {
```
rzetelskik: Let's leave this to test helpers. It breaks separation of concerns and will require prefixing with dc name for v1alpha1.SC.

test/e2e/framework/testcontext.go
```diff
@@ -47,18 +47,16 @@ type TestContextType struct {
...
+	CleanupPolicy                     CleanupPolicyType
+	IngressController                 *IngressController
+	ScyllaClusterOptions              *ScyllaClusterOptions
+	ScyllaDBVersion                   string
+	ScyllaDBManagerVersion            string
+	ScyllaDBManagerAgentVersion       string
+	ScyllaDBUpdateFrom                string
+	ScyllaDBUpgradeFrom               string
+	WorkerObjectStorageSettings       map[string]ClusterObjectStorageSettings
+	ControlPlaneObjectStorageSettings *ClusterObjectStorageSettings
```
rzetelskik: Ditto about cluster/workerCluster naming convention

test/e2e/framework/storage.go
czeslavo: This is a kind of test helper (it's defined in `test/e2e/framework` package), no? Before my changes, this logic was scattered in `Framework.GetObjectStorageProvider` and inlined in test cases with `fmt.Sprintf("%s:%s")` . The intention was not to repeat ourselves and have a single place for getting a location in a manager format for a bucket. If it will require prefixing with a DC name, we can extend it once we face that particular need.

If you think this is too specific to make this a struct method, I'm fine with extracting this as a standalone test helper function.

pkg/cmd/tests/object_storage_options.go
czeslavo: Renamed to `ClusterObjectStorageSettings` for the CP/single-DC settings.

test/e2e/framework/testcontext.go
czeslavo: Renamed to `ClusterObjectStorageSettings`.

test/e2e/set/scylladbmanagertask/scylladbmanagertask_globalmanager_object_storage.go
```diff
@@ -53,24 +53,25 @@ var _ = g.Describe("ScyllaDBManagerTask integration with global ScyllaDB Manager
 
 		metav1.SetMetaDataLabel(&sourceSDC.ObjectMeta, naming.GlobalScyllaDBManagerRegistrationLabel, naming.LabelValueTrue)
 
-		objectStorageType := f.GetObjectStorageType()
-		switch objectStorageType {
+		objectStorageSettings, ok := f.GetClusterObjectStorageSettings()
+		if !ok {
+			g.Fail("control plane object storage settings are not configured")
+		}
```
rzetelskik:
```suggestion
		o.Expect(ok).To(o.BeTrue(), "cluster object storage settings must be configured")
```

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -51,23 +51,26 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
 			sourceSC.Spec.Version = e.scyllaVersion
 		}
 
-		objectStorageType := f.GetObjectStorageType()
-		switch objectStorageType {
+		objectStorageSettings, ok := f.GetClusterObjectStorageSettings()
+		if !ok {
+			g.Fail("cluster object storage settings are not configured")
+		}
```
rzetelskik, line 57: ditto

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -490,23 +493,26 @@ var _ = g.Describe("Scylla Manager integration", framework.RequiresObjectStorage
 		sc := f.GetDefaultScyllaCluster()
 		sc.Spec.Datacenter.Racks[0].Members = 1
 
-		objectStorageType := f.GetObjectStorageType()
-		switch objectStorageType {
+		objectStorageSettings, ok := f.GetClusterObjectStorageSettings()
+		if !ok {
+			g.Fail("control plane object storage settings are not configured")
+		}
```
rzetelskik: ditto

rzetelskik (commented):
Thanks for the updates. Last few comments, rest lgtm.

test/e2e/set/scylladbmanagertask/scylladbmanagertask_globalmanager_object_storage.go
rzetelskik: (this is mostly about s/control plane/cluster, fail vs assertion is a  nit)

PR #2760, opened 2025-07-09
Fix order of namespace collection in multidc boostrap e2e

zimnx:
Registration was executed before ScyllaDBCluster was created, meaning it's Namespace was empty which caused registration that depends on Namespace to register wrong name for collection.

Thanks @rzetelskik for noticing it!
/cc rzetelskik


rzetelskik (approved):
/lgtm

zimnx:
Correct namespaces are collected in touched e2e, so issue is fixed, suite is still broken.

PR #2763, opened 2025-07-10
Avoid leading hyphens in truncated hashed names generation

rzetelskik:
**Description of your changes:** Recently added `generateTruncatedHashedName` helper has two issues:
- truncating may lead to panics when maxLength == len(nameSuffix),
- created full name may have a leading hyphen, which is not a valid DNS subdomain (RFC 1123).

This PR fixes both issues.

Resolves #

/kind bug
/priority important-soon

### DIFF
--- a/pkg/naming/names.go
+++ b/pkg/naming/names.go
@@ -2,6 +2,7 @@ package naming
 
 import (
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/naming/names.go
```diff
@@ -303,19 +304,37 @@ func scyllaDBManagerClusterRegistrationName(kind, name string) (string, error) {
...
+		return "", fmt.Errorf("maxLength cannot be lower than the length of the name suffix hash: %d", len(nameSuffix))
+	}
+
+	truncatedMaxLength := maxLength - len(nameSuffix) - 1
+	truncatedName := ""
+	// Avoid returning a name with leading hyphen even if it means underutilizing max length.
+	// Return a valid DNS subdomain (RFC 1123).
+	if truncatedMaxLength > 0 {
+		name := strings.ToLower(strings.Join(nonEmptyParts, "-"))
+		truncatedName = name[:min(len(name), truncatedMaxLength)] + "-"
```
czeslavo: Nit: just to be fully precise it's about truncating the name part, not the whole "hashed name"
```suggestion
	truncatedNameMaxLength := maxLength - len(nameSuffix) - 1
	truncatedName := ""
	// Avoid returning a name with leading hyphen even if it means underutilizing max length.
	// Return a valid DNS subdomain (RFC 1123).
	if truncatedNameMaxLength > 0 {
		name := strings.ToLower(strings.Join(nonEmptyParts, "-"))
		truncatedName = name[:min(len(name), truncatedNameMaxLength)] + "-"
```

czeslavo: Nit: for clarity
```suggestion
	truncatedMaxLength := maxLength - len(nameSuffix) - 1 // 1 character reserved for a potential hyphen
	truncatedName := ""
	// Avoid returning a name with leading hyphen even if it means underutilizing max length.
	// Return a valid DNS subdomain (RFC 1123).
	if truncatedMaxLength > 0 {
		name := strings.ToLower(strings.Join(nonEmptyParts, "-"))
		truncatedName = name[:min(len(name), truncatedMaxLength)] + "-"
```

PR #2764, opened 2025-07-10
Generate ScyllaDB Manager agent auth tokens for ScyllaDBClusters

### LINKED ISSUE #2765: Generate ScyllaDB Manager agent auth tokens for ScyllaDBClusters
### Acceptance criteria
- [ ] Modify ScyllaDBDatacenter controller to allow for overriding auth token generation. User provided config should still take precedence.
- [ ] Modify ScyllaDBCluster controller to generate ScyllaDB Manager auth tokens and propagate them to underlying ScyllaDBDatacenters. 

/kind feature
/priority important-soon

rzetelskik:
**Description of your changes:** This PR adds auth token generation for ScyllaDBClusters and a mechanism for propagating it to ScyllaDBDatacenters.

Resolves #2765 

/kind feature
/priority important-soon

### DIFF
--- a/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_config.go
+++ b/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_config.go
@@ -123,7 +123,7 @@ var _ = g.Describe("Multi datacenter ScyllaDBCluster", framework.MultiDatacenter
 		allDCsRacksScyllaDBManagerAgentVolumesSecret := makeGenericSecret("all-dcs-racks-scylladbmanageragent-volumes")
 		allDCsRacksScyllaDBVolumesSecret := makeGenericSecret("all-dcs-racks-scylladb-volumes")
 
-		secretsToBeMirroredToAllDCs := []*corev1.Secret{
+		userManagedSecretsToBeMirroredToAllDCs := []*corev1.Secret{
 			allDCsScyllaDBManagerAgentCustomConfigSecret,
 			allDCsScyllaDBManagerAgentVolumesSecret,
 			allDCsRackTemplateScyllaDBManagerAgentCustomConfigSecret,
@@ -146,7 +146,7 @@ var _ = g.Describe("Multi datacenter ScyllaDBCluster", framework.MultiDatacenter
 		dcRacksScyllaDBManagerAgentVolumesSecret := makeGenericSecret("dc-racks-scylladbmanageragent-volumes")
 		dcRacksScyllaDBVolumesSecret := makeGenericSecret("dc-racks-scylladb-volumes")
 
-		secretsToBeMirroredToSpecificDC := []*corev1.Secret{
+		userManagedSecretsToBeMirroredToSpecificDC := []*corev1.Secret{
 			dcScyllaDBManagerAgentCustomConfigSecret,
 			dcScyllaDBManagerAgentVolumesSecret,
 			dcRackTemplateScyllaDBManagerAgentCustomConfigSecret,
@@ -158,12 +158,12 @@ var _ = g.Describe("Multi datacenter ScyllaDBCluster", framework.MultiDatacenter
 			dcRacksScyllaDBVolumesSecret,
 		}
 
-		secretsToBeMirrored := slices.Concat(
-			secretsToBeMirroredToAllDCs,
-			secretsToBeMirroredToSpecificDC,
+		userManagedSecretsToBeMirrored := slices.Concat(
+			userManagedSecretsToBeMirroredToAllDCs,
+			userManagedSecretsToBeMirroredToSpecificDC,
 		)
 
-		for _, secret := range secretsToBeMirrored {
+		for _, secret := range userManagedSecretsToBeMirrored {
 			_, err := userClient.KubeClient().CoreV1().Secrets(secret.Namespace).Create(ctx, secret, metav1.CreateOptions{})
 			o.Expect(err).NotTo(o.HaveOccurred())
 		}
@@ -255,6 +255,28 @@ var _ = g.Describe("Multi datacenter ScyllaDBCluster", framework.MultiDatacenter
 		scylladbclusterverification.Verify(ctx, sc, rkcClusterMap)
 		err = scylladbclusterverification.WaitForFullQuorum(ctx, rkcClusterMap, sc)
 		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Collecting operator-managed Secrets which should be mirrored into remote datacenters")
+		var operatorManagedScyllaDBManagerAgentSecrets []*corev1.Secret
+
+		scyllaDBManagerAgentAuthTokenSecretName, err := naming.ScyllaDBManagerAgentAuthTokenSecretNameForScyllaDBCluster(sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		operatorManagedScyllaDBManagerAgentSecrets = append(operatorManagedScyllaDBManagerAgentSecrets, &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      scyllaDBManagerAgentAuthTokenSecretName,
+				Namespace: sc.Namespace,
+			},
+		})
+
+		operatorManagedSecretsToBeMirrored := slices.Concat(
+			operatorManagedScyllaDBManagerAgentSecrets,
+		)
+
+		secretsToBeMirrored := slices.Concat(
+			userManagedSecretsToBeMirrored,
+			operatorManagedSecretsToBeMirrored,
+		)
 
 		for _, dc := range sc.Spec.Datacenters {
 			framework.By("Verifying if ConfigMaps and Secrets referenced by ScyllaDBCluster %q are mirrored in %q Datacenter", sc.Name, dc.Name)


pkg/controller/scylladbcluster/sync_secrets.go
```diff
@@ -68,3 +70,92 @@ func (scc *Controller) syncRemoteSecrets(
...
+
+	getOptionalExistingScyllaDBManagerAgentAuthTokenSecret := func() ([]metav1.Condition, *corev1.Secret, error) {
+		secret, err := getOptionalExistingScyllaDBManagerAuthTokenSecret(sc, localSecrets)
+		return nil, secret, err
+	}
+
+	progressingConditions, scyllaDBManagerAgentAuthTokenConfig, err := controllerhelpers.GetScyllaDBManagerAgentAuthTokenConfig(
+		getOptionalScyllaDBManagerCustomAgentConfigSecret,
+		getOptionalExistingScyllaDBManagerAgentAuthTokenSecret,
+		false,
```
czeslavo: Please add a comment similar to the other call of `GetScyllaDBManagerAgentAuthTokenConfig` to make this boolean parameter's purpose clear.

rzetelskik: added it in the struct itself

rzetelskik:
/cc czeslavo zimnx
Ready for review. @zimnx please have a look as well.
I'm not too happy with how complicated it came out, so If you have any ideas on how to make it less convoluted while also avoiding code duplication between the two controllers, I'll be happy to hear.

pkg/controller/scylladbdatacenter/sync_agent_config.go
```diff
@@ -2,81 +2,74 @@ package scylladbdatacenter
...
+		if !errors.As(err, &controllerhelpers.ScyllaDBManagerAgentCustomConfigError{}) {
+			return progressingConditions, fmt.Errorf("can't get ScyllaDB Manager agent auth token config: %w", err)
 		}
-	}
 
-	// If we still don't have the token, we generate a random one.
-	if len(token) == 0 {
-		token = apimachineryutilrand.String(128)
+		// Return the custom agent config error at the end.
+		sdcc.eventRecorder.Eventf(sdc, corev1.EventTypeWarning, "InvalidManagerAgentConfig", "Can't gent agent token: %s", err.Error())
```
czeslavo:
```suggestion
		sdcc.eventRecorder.Eventf(sdc, corev1.EventTypeWarning, "InvalidManagerAgentConfig", "Can't get agent token: %s", err.Error())
```

pkg/controllerhelpers/scylladbmanager.go
```diff
@@ -34,3 +39,139 @@ func GetScyllaDBManagerClient(_ context.Context, _ *scyllav1alpha1.ScyllaDBManag
...
+
+		customConfigError = err
+	}
+	if len(progressingConditions) > 0 {
+		return progressingConditions, nil, customConfigError
+	}
+
+	authTokenConfig, err := helpers.GetAgentAuthTokenConfig(authToken)
+	if err != nil {
+		return nil, nil, fmt.Errorf("can't get ScyllaDB Manager agent auth token config: %w", err)
```
czeslavo: Shouldn't we return an aggregated error here? If `customConfigError != nil`, it won't be surfaced.

rzetelskik: That's how this works now - the error from config is only returned if there are no subsequent errors. I added unit tests for this specific behaviour. I'd prefer returning an error aggregate, but then `errors.As()` will also return true. I couldn't figure out how to do this nicely.

pkg/controller/scylladbdatacenter/sync_agent_config.go
rzetelskik: done

pkg/controllerhelpers/scylladbmanager.go
rzetelskik: agreed internally to modify this slightly

rzetelskik: done

rzetelskik:
Addressed all comments, ptal @czeslavo

rzetelskik:
> 
> ci/prow/e2e-gke-serial	[7fb8758](https://github.com/scylladb/scylla-operator/commit/7fb875881850aee11504e60dc137bfeae892362a)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2764/pull-scylla-operator-master-e2e-gke-serial/1945094751183179776)	true	`/test e2e-gke-serial`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2764). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

known flake
/retest

pkg/controller/scylladbdatacenter/sync_agent_config.go
```diff
@@ -6,69 +6,89 @@ import (
...
-) ([]metav1.Condition, error) {
-	var progressingConditions []metav1.Condition
+) (progressingConditions []metav1.Condition, err error) {
+	// We allow the controller to continue on a custom agent config error,
+	// so that a misconfigured or missing custom agent config does not block the controller from creating the auth token secret.
+	// In case the returned error is a custom agent config error, we continue and return the error at the end of the function.
+	var customAgentConfigError error
+	defer func() {
+		err = apimachineryutilerrors.NewAggregate([]error{err, customAgentConfigError})
+	}()
```
czeslavo: I believe named returns paired with defer make code harder to read (especially if the function is not a 5-liner or so). 

How about being more explicit and just aggregating the errors like below:

```go
func (sdcc *Controller) syncAgentToken(
	ctx context.Context,
	sdc *scyllav1alpha1.ScyllaDBDatacenter,
	secrets map[string]*corev1.Secret,
) ([]metav1.Condition, error) {
	var progressingConditions []metav1.Condition
	var allErrors []error

	getScyllaDBManagerAgentAuthTokenConfigOptions := controllerhelpers.GetScyllaDBManagerAgentAuthTokenConfigOptions{
		GetOptionalAgentAuthTokenFromCustomConfig: func(extractFunc func(*corev1.Secret) (string, error)) ([]metav1.Condition, string, error) {
			customProgressing, customAuthToken, customErr := controllerhelpers.GetScyllaDBManagerAgentAuthTokenFromSecret(
				getOptionalAgentCustomConfigSecretFunc(sdc, sdcc.secretLister),
				extractFunc,
			)
			if customErr != nil {
				// Emit an event, continue, and report as an error later.
				sdcc.eventRecorder.Eventf(sdc, corev1.EventTypeWarning, "InvalidManagerAgentConfig", "Can't get agent token: %s", customErr.Error())
				allErrors = append(allErrors, customErr)
				return customProgressing, "", nil
			}
			return customProgressing, customAuthToken, nil
		},
		GetOptionalAgentAuthTokenFromExisting: func(extractFunc func(*corev1.Secret) (string, error)) ([]metav1.Condition, string, error) {
			return controllerhelpers.GetScyllaDBManagerAgentAuthTokenFromSecret(
				getOptionalExistingAgentAuthTokenSecretFunc(sdc, secrets, sdcc.secretLister),
				extractFunc,
			)
		},
	}

	agentProgressing, agentAuthTokenConfig, err := controllerhelpers.GetScyllaDBManagerAgentAuthTokenConfig(getScyllaDBManagerAgentAuthTokenConfigOptions)
	progressingConditions = append(progressingConditions, agentProgressing...)
	if err != nil {
		allErrors = append(allErrors, fmt.Errorf("can't get ScyllaDB Manager agent auth token config: %w", err))
		// Nothing more to do, return the progressing conditions and errors.
		return progressingConditions, apimachineryutilerrors.NewAggregate(allErrors)
	}

	secret, err := makeAgentAuthTokenSecret(sdc, agentAuthTokenConfig)
	if err != nil {
		allErrors = append(allErrors, fmt.Errorf("can't make auth token secret: %w", err))
		// Nothing more to do, return the progressing conditions and errors.
		return progressingConditions, apimachineryutilerrors.NewAggregate(allErrors)
	}

	// TODO: Remove forced ownership in v1.5 (#672)
	_, changed, err := resourceapply.ApplySecret(ctx, sdcc.kubeClient.CoreV1(), sdcc.secretLister, sdcc.eventRecorder, secret, resourceapply.ApplyOptions{
		ForceOwnership: true,
	})
	if err != nil {
		allErrors = append(allErrors, fmt.Errorf("can't apply secret %q: %w", naming.ObjRef(secret), err))
		// Nothing more to do, return the progressing conditions and errors.
		return progressingConditions, apimachineryutilerrors.NewAggregate(allErrors)
	}
	if changed {
		controllerhelpers.AddGenericProgressingStatusCondition(&progressingConditions, agentTokenControllerProgressingCondition, secret, "apply", sdc.Generation)
	}

	if len(allErrors) > 0 {
		return progressingConditions, apimachineryutilerrors.NewAggregate(allErrors)
	}
	return progressingConditions, nil
}
```

I'll leave the decision up to you - it kind of is just a matter of taste here, as the result is the same.

czeslavo (commented):
lgtm (leaving the decision regarding dropping/leaving named returns to you).

pkg/controller/scylladbdatacenter/sync_agent_config.go
rzetelskik: It feels more repetitive having to do this on every error imo, I'd rather leave it as is. I'd generally agree with defer and named errors being harder to read, but we often declare err and progressing conditions at the beginning of the func anyway, so this doesn't seem too difficult to read to me.

czeslavo: I always feel `defer` used for something else than cleaning up resources makes code harder to reason about. It's less repetitive, but has a higher "magic" factor that implies a higher cognitive load on the reader. Still, I do not require changes if you believe it's better to keep going, I'm ok with that. Feel free to resolve this discussion.

pkg/controller/scylladbcluster/controller.go
```diff
@@ -213,7 +213,15 @@ func NewController(
...
-	// Local ConfigMap and Secret handlers are skipped to optimize number of syncs which doesn't do anything.
+	secretInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    scc.addSecret,
+			UpdateFunc: scc.updateSecret,
+			DeleteFunc: scc.deleteSecret,
+		},
+	)
+
+	// Handlers for local ConfigMaps and Secrets referenced by ScyllaDBClusters are skipped to optimize number of syncs which doesn't do anything.
```
zimnx, line 224:
```suggestion
	// Handlers for local ConfigMaps referenced by ScyllaDBClusters are skipped to optimize number of syncs which doesn't do anything.
```

pkg/controller/scylladbcluster/sync.go
```diff
@@ -110,6 +110,26 @@ func (scc *Controller) sync(ctx context.Context, key string) error {
...
+		scyllav1alpha1.ScyllaDBClusterGVK,
+		scLocalSelector,
+		controllerhelpers.ControlleeManagerGetObjectsFuncs[localCT, *corev1.Secret]{
+			GetControllerUncachedFunc: scc.scyllaClient.ScyllaV1alpha1().ScyllaDBClusters(sc.Namespace).Get,
+			ListObjectsFunc:           scc.secretLister.Secrets(sc.Namespace).List,
+			PatchObjectFunc:           scc.kubeClient.CoreV1().Secrets(sc.Namespace).Patch,
+		},
+	)
+	if err != nil {
+		localObjectErrs = append(localObjectErrs, fmt.Errorf("can't get secrets: %w", err))
```
zimnx:
```suggestion
		localObjectErrs = append(localObjectErrs, fmt.Errorf("can't get local secrets: %w", err))
```

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -72,6 +72,9 @@ var (
 		// object created on migration from scyllav1.ScyllaCluster object. Setting it shouldn't trigger a rollout as it
 		// doesn't affect the ScyllaDB cluster itself.
 		naming.ScyllaDBManagerClusterRegistrationNameOverrideAnnotation,
+		// This annotation is set by controllers to override the ScyllaDB Manager auth token generated for a particular ScyllaDBDatacenter.
+		// Setting it shouldn't trigger a rollout as the value is propagated to a Secret used by ScyllaDB Manager agent, which should be reloaded dynamically.
+		naming.ScyllaDBManagerAgentAuthTokenOverrideSecretRefAnnotation,
```
zimnx: This Secret won't be reloaded dynamically, as auth token volume has a subPath: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically (it works the same for Secrets) 
https://github.com/scylladb/scylla-operator/blob/ff6bc3937c4cf8b34f16674a98dd29b7b2010409/pkg/controller/scylladbdatacenter/resource.go#L1265-L1270

So we need to propagate it to trigger a rollout upon change.

pkg/controller/scylladbdatacenter/sync_agent_config.go
```diff
@@ -6,69 +6,89 @@ import (
...
+
+			return customAgentConfigProgressingConditions, customAgentConfigAuthToken, nil
+		},
+		GetOptionalAgentAuthTokenFromExisting: func(extractFunc func(*corev1.Secret) (string, error)) ([]metav1.Condition, string, error) {
+			return controllerhelpers.GetScyllaDBManagerAgentAuthTokenFromSecret(
+				getOptionalExistingAgentAuthTokenSecretFunc(sdc, secrets, sdcc.secretLister),
+				extractFunc,
+			)
+		},
 	}
```
zimnx: As discussed offline this should clearly see what are potential sources of agent config.

rzetelskik:
I'll rebase on top of https://github.com/scylladb/scylla-operator/pull/2782 once it is merged

rzetelskik:
We agreed with @zimnx privately to get rid of the custom error handling. This and other requested changes require a total overhaul of the proposed implementation, so I'm changing this to draft.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: ~Wouldn't it be better to just change the path so that we don't have to use a subpath then? It seems wrong to roll out the entire cluster just to reload the config for a sidecar container. That is unless manager agent is not capable of reloading the auth token dynamically - need to check.~ Edit: it isn't so we do have to roll out the entire cluster :cry:.

pkg/controller/scylladbcluster/controller.go
rzetelskik: I did that before but reverted it - there are still no handlers for local secrets **referenced** by ScyllaDBClusters (the user-managed ones), only the ones owned by the object. Correct me if my understanding of the comment before my changes is wrong, but I believe this is what it was referring to.

rzetelskik:
rebased on top of current master

pkg/controller/scylladbcluster/sync.go
rzetelskik: ok

pkg/controller/scylladbdatacenter/sync_agent_config.go
rzetelskik: 1. removed the delayed error propagation which was earlier made a requirement
2. increased code duplication for the sake of clarity, the helper now takes a slice of functions providing auth tokens and tries to get one greedily in the original order

zimnx (approved):
Thanks for updates, imo code is much clearer now, and I'm glad we eliminated that odd error suppression while still keeping the secured agent intact.

Issue #2696, opened 2025-06-17
Drop mentions of standalone ScyllaDB Manager controller from documentation


rzetelskik:
After https://github.com/scylladb/scylla-operator/issues/2559 has been merged, any mentions of the standalone ScyllaDB Manager controller should be dropped from documentation.

/kind documentation
/assign

PR #2770, opened 2025-07-11
Do not mention standalone Manager Controller

czeslavo:
Removes remaining mentions of Scylla Manager Controller from docs.

Resolves https://github.com/scylladb/scylla-operator/issues/2696.

/kind documentation
/priority important-soon


rzetelskik:
We should also drop it from any images such as https://github.com/scylladb/scylla-operator/blob/521f969a33cacf76b86091786879dc94f79a3974/docs/source/architecture/components-cluster_scoped.svg.

PR #2771, opened 2025-07-14
Apply e2e workarounds on all clusters

zimnx:
E2E workarounds - aio-max-nr tuning and ephemeral containers RBAC - was applied only on control plane clusters, causing Scylla running on worker clusters to fail due to too little aio-max-nr.


hack/.ci/lib/e2e.sh
```diff
@@ -181,6 +181,22 @@ function gracefully-shutdown-e2es {
...
+      worker_as_control_plane="${name}"
+      continue
+    fi
+
+    KUBECONFIG="${WORKER_KUBECONFIGS[$name]}" apply-e2e-workarounds
+  done
+
+  if [[ -z "${worker_as_control_plane+x}" ]]; then
+    KUBECONFIG="${KUBECONFIG}" apply-e2e-workarounds
+  fi
```
rzetelskik:
```suggestion
  apply-e2e-workarounds
  
  for name in "${!WORKER_KUBECONFIGS[@]}"; do
    if [[ "${WORKER_KUBECONFIGS[$name]}" == "${KUBECONFIG}" ]]; then
      # Skip if the control plane cluster is also among the worker clusters.
      continue
    fi

    KUBECONFIG="${WORKER_KUBECONFIGS[$name]}" apply-e2e-workarounds
  done
```
can be simplified a bit, similarly to the existing funcs

zimnx: good idea, done

PR #2781, opened 2025-07-15
Consider ScyllaDBDatacenter having no Status Conditions as not rolled out

zimnx:
Function used to determine whether given ScyllaDBDatacenter is rolled out didn't consider case when provided object is nil, or whether it has any Status Conditions, returning wrong value.

### DIFF
--- a/pkg/controllerhelpers/scylla.go
+++ b/pkg/controllerhelpers/scylla.go
@@ -296,6 +296,10 @@ func GetRackNodeCount(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string) (
 }
 
 func IsScyllaDBDatacenterRolledOut(sdc *scyllav1alpha1.ScyllaDBDatacenter) (bool, error) {
+	if sdc == nil || len(sdc.Status.Conditions) == 0 {
+		return false, nil
+	}
+
 	if !helpers.IsStatusConditionPresentAndTrue(sdc.Status.Conditions, scyllav1alpha1.AvailableCondition, sdc.Generation) {
 		return false, nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/controllerhelpers/scylla.go
```diff
@@ -296,6 +296,10 @@ func GetRackNodeCount(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string) (
 }
 
 func IsScyllaDBDatacenterRolledOut(sdc *scyllav1alpha1.ScyllaDBDatacenter) (bool, error) {
+	if sdc == nil || len(sdc.Status.Conditions) == 0 {
+		return false, nil
+	}
+
 	if !helpers.IsStatusConditionPresentAndTrue(sdc.Status.Conditions, scyllav1alpha1.AvailableCondition, sdc.Generation) {
```
rzetelskik, line 303: isn't this used incorrectly then? The negation should be equivalent to `is status condition unknown or false` - it seems like this should match for an empty slice? Shall we unit test and fix this instead?

zimnx: negation of it is `is status condition not present or false`, either way, for both cases it correctly returns true for empty slice as condition is not present.

pkg/controllerhelpers/scylla.go
```diff
@@ -296,6 +296,10 @@ func GetRackNodeCount(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string) (
 }
 
 func IsScyllaDBDatacenterRolledOut(sdc *scyllav1alpha1.ScyllaDBDatacenter) (bool, error) {
+	if sdc == nil || len(sdc.Status.Conditions) == 0 {
```
czeslavo, line 299: The condition checks below will make this function return `false` if there are no conditions, no?
```suggestion
	if sdc == nil {
```

zimnx:
PEBKAC

PR #2786, opened 2025-07-16
Await for previous DCs to finish progressing before creating next DC

zimnx:
Instead awaiting only for previous DC SDC rollout, we await until entire DC finished progressing before proceeding to create new one. In addition it fixes an issue where stale status and wrong object was considered for other conditions we await for.


rzetelskik (approved):
/lgtm

PR #2789, opened 2025-07-17
Mirror ScyllaDBCluster ConfigMaps referred from Volumes

### LINKED ISSUE #2784: ConfigMaps used in volumes in ScyllaDBCluster are not mirrored to remote clusters
### What happened?

When creating a `ScyllaDBCluster` with Volumes and `VolumeSource.ConfigMap != nil` defined in any of `ScyllaDBClusterDatacenterTemplate.{ScyllaDB, ScyllaDBManagerAgentTemplate}`, `RackTemplate.{ScyllaDB, ScyllaDBManagerAgent}`), ConfigMaps referred to in the volume source are not mirrored to remote Kubernetes clusters.

/priority important-soon

All mentioned ConfigMaps should be automatically mirrored to the respective remote cluster.

Create `ScyllaDBCluster` with a DC with `RackTemplate.ScyllaDBManagerAgent.Volumes` having a Volume referring to a ConfigMap.

### Scylla Operator version

v1.17

not relevant

not relevant

_No response_

zimnx:

Makes all ConfigMaps referred from Volumes defines in ScyllaDBCluster to be mirrored to the respecive DC's remote clusters.

Resolves #2784

/cc rzetelskik


zimnx:
@rzetelskik ready to review

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1070,14 +1056,143 @@ func makeMirroredRemoteConfigMaps(sc *scyllav1alpha1.ScyllaDBCluster, dc *scylla
...
+	err = apimachineryutilerrors.NewAggregate(errs)
 	if err != nil {
 		return progressingConditions, requiredRemoteConfigMaps, fmt.Errorf("can't make mirrored ConfigMaps: %w", err)
 	}
 
 	return progressingConditions, requiredRemoteConfigMaps, nil
 }
 
+// getConfigMapsToMirrorForAllDCs collects the names of ConfigMaps to mirror for all Datacenters in the ScyllaDBCluster.
+func getConfigMapsAndSecretsToMirrorForAllDCs(sc *scyllav1alpha1.ScyllaDBCluster) ([]string, []string, error) {
```
rzetelskik, line 1068: why have a common func for both secrets and configmaps? It sems like the flow would be more natural in both the implementation (no need to ignore returned values) and testing if they were separate?

pkg/controller/scylladbcluster/resource.go
```diff
@@ -1011,34 +1011,20 @@ func MakeRemoteConfigMaps(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alpha1
...
-
-	if dc.ScyllaDB != nil && dc.ScyllaDB.CustomConfigMapRef != nil {
-		configMapsToMirror = append(configMapsToMirror, *dc.ScyllaDB.CustomConfigMapRef)
-	}
-
-	if dc.RackTemplate != nil && dc.RackTemplate.ScyllaDB != nil && dc.RackTemplate.ScyllaDB.CustomConfigMapRef != nil {
-		configMapsToMirror = append(configMapsToMirror, *dc.RackTemplate.ScyllaDB.CustomConfigMapRef)
+	configMapsToMirrorForAllDCs, _, err := getConfigMapsAndSecretsToMirrorForAllDCs(sc)
+	if err != nil {
+		return progressingConditions, requiredRemoteConfigMaps, fmt.Errorf("can't make mirrored configmaps: %w", err)
```
rzetelskik, line 1020: (tiny) nit
```suggestion
		return progressingConditions, nil, fmt.Errorf("can't make mirrored configmaps: %w", err)
		```

pkg/controller/scylladbcluster/resource.go
zimnx: To avoid duplication of two long functions. Look at how getConfigMapsAndSecretsToMirrorForDC look like, it used to look like this but for just secrets, we would have to have a duplicated one, but collecting configmaps from the same spec paths.

rzetelskik: that's fair enough

PR #2790, opened 2025-07-18
Update ScyllaDB Manager to 3.5.1

rzetelskik:
**Description of your changes:** This PR updates ScyllaDB Manager and ScyllaDB Manager agent container image references to 3.5.1. Most notably, this patch release brings a fix to frequent snapshot name collisions in a single Manager instance (https://github.com/scylladb/scylla-manager/issues/3873), which we've often hit in our CI (https://github.com/scylladb/scylla-operator/issues/2370).

Resolves https://github.com/scylladb/scylla-operator/issues/2370

/kind dependency-bump
/priority important-soon

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -52,6 +52,9 @@ todo_include_todos = True
 
 myst_enable_extensions = ["colon_fence", "attrs_inline", "substitution"]
 myst_heading_anchors = 6
+
+# MyST substitutions are broken with multiversion docs. Versions specified in the main branch are used for all versions.
+# DO NOT USE OR CHANGE THESE VERSIONS!
 myst_substitutions = {
   "productName": "Scylla Operator",
   "repository": "scylladb/scylla-operator",
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -53,8 +53,10 @@ todo_include_todos = True
 myst_enable_extensions = ["colon_fence", "attrs_inline", "substitution"]
 myst_heading_anchors = 6
 
-# MyST substitutions are broken with multiversion docs. Versions specified in the main branch are used for all versions.
-# DO NOT USE OR CHANGE THESE VERSIONS!
+# DEPRECATION NOTICE
+# MyST substitutions work counterintuitively with multiversion docs. Versions specified in the main branch are used for all versions.
+# These variables have no effect if set on branches other than master.
+# https://github.com/scylladb/scylla-operator/issues/2795
 myst_substitutions = {
   "productName": "Scylla Operator",
   "repository": "scylladb/scylla-operator",
--- a/docs/source/resources/scyllaclusters/multidc/multidc.md
+++ b/docs/source/resources/scyllaclusters/multidc/multidc.md
@@ -110,7 +110,7 @@ metadata:
   name: scylla-cluster
   namespace: scylla
 spec:
-  agentVersion: 3.5.0
+  agentVersion: 3.5.1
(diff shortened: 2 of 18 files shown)


rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[b0cf53a](https://github.com/scylladb/scylla-operator/commit/b0cf53ad5ee7a574f64ce8d6752537d1b03e5c06)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2790/pull-scylla-operator-master-e2e-gke-parallel/1946117002066137088)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2790). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

failed to resolve image
/test images
/retest

docs/source/resources/scyllaclusters/multidc/multidc.md
```diff
@@ -110,7 +110,7 @@ metadata:
   name: scylla-cluster
   namespace: scylla
 spec:
-  agentVersion: 3.5.0
+  agentVersion: {{agentVersion}}
```
mflendrich: I am not sure about using agentVersion.

This variable is set across all Operator versions in `conf.py`, while the manager version is pinned only in that specific version. This will cause drift that I am not sure is acceptable.

Suggestion: keep hardcoding until we have a better solution for variable versioning 😢

rzetelskik: Should we remove agentVersion substitutions everywhere then? It's used in some docs and hardcoded in others, whatever we do it's probably best to make it consistent.

mflendrich: Yup, to avoid future confusion I would remove the variable from myst_substitutions altogether.

rzetelskik: >  I would remove the variable from myst_substitutions altogether.

That would break other branches. I'll remove the uses of manager-related ones in this PR and hardcode the new version without updating the one in myst_substitution, not to break other versions. We can handle the rest separately.

rzetelskik: @mflendrich https://github.com/scylladb/scylla-operator/compare/b0cf53ad5ee7a574f64ce8d6752537d1b03e5c06..73b7ac6bcc8b8d9834096292dcd5c8901f1a5b49

docs/source/conf.py
```diff
@@ -52,6 +52,9 @@
 
 myst_enable_extensions = ["colon_fence", "attrs_inline", "substitution"]
 myst_heading_anchors = 6
+
+# MyST substitutions are broken with multiversion docs. Versions specified in the main branch are used for all versions.
+# DO NOT USE OR CHANGE THESE VERSIONS!
```
mflendrich:
```suggestion
# DEPRECATION NOTICE
# MyST substitutions work counterintuitively with multiversion docs. Versions specified in the main branch are used for all versions.
# These variables have no effect if set on branches other than master.
# <link to a GH issue to get rid of these across all branches>
```

mflendrich:
/cc czeslavo

(because I've added a commit and would like a third party to review)

czeslavo (approved):
/lgtm

Issue #2788, opened 2025-07-17
Add an option to opt-out of ScyllaDB Manager integration for ScyllaClusters


rzetelskik:
### What should the feature do?

Allow for opting out of ScyllaDB Manager integration for ScyllaClusters. When opted-out, ScyllaClusters should not be registered with the global ScyllaDB Manager instance.

### What is the use case behind this feature?

WIth the standalone manager-controller, users were able to completely or temporarily opt out of registering their ScyllaClusters with the global ScyllaDB Manager instance by simply removing the manager-controller deployment. Moving this component into the operator removed this workaround. In certain cases it might be desired to turn this off.

### Anything else we need to know?

A custom annotation seems like an acceptable way to approach this.

PR #2793, opened 2025-07-21
Allow opting out of v1.ScyllaCluster global manager registration

czeslavo:
Allows opting out of `v1.ScyllaCluster` global ScyllaDB Manager integration by annotating it explicitly with `scylla-operator.scylladb.com/disable-global-scylladb-manager-integration: true`.

Resolves https://github.com/scylladb/scylla-operator/issues/2788.

/kind feature
/priority important-soon

### DIFF
--- a/pkg/controller/scyllacluster/migrate.go
+++ b/pkg/controller/scyllacluster/migrate.go
@@ -323,8 +323,12 @@ func MigrateV1ScyllaClusterToV1Alpha1ScyllaDBDatacenter(sc *scyllav1.ScyllaClust
 				}
 			}
 
-			// Register the underlying ScyllaDBDatacenter with global ScyllaDB Manager instance for backward compatibility.
-			objectMeta.Labels[naming.GlobalScyllaDBManagerRegistrationLabel] = naming.LabelValueTrue
+			if isGlobalScyllaDBManagerIntegrationDisabled(sc) {
+				objectMeta.Labels[naming.GlobalScyllaDBManagerRegistrationLabel] = naming.LabelValueFalse
+			} else {
+				// If the ScyllaDBManager integration is not disabled explicitly, we set the label to true to keep backward compatibility.
+				objectMeta.Labels[naming.GlobalScyllaDBManagerRegistrationLabel] = naming.LabelValueTrue
+			}
 
 			// Override the ScyllaDB Manager cluster name for backward compatibility.
 			objectMeta.Annotations[naming.ScyllaDBManagerClusterRegistrationNameOverrideAnnotation] = naming.ManagerClusterName(sc)
@@ -476,6 +480,11 @@ func migrateV1Alpha1ScyllaDBDatacenterStatusToV1ScyllaClusterStatus(sdc *scyllav
 func MigrateV1ScyllaClusterToV1Alpha1ScyllaDBManagerTasks(sc *scyllav1.ScyllaCluster) ([]*scyllav1alpha1.ScyllaDBManagerTask, error) {
 	var scyllaDBManagerTasks []*scyllav1alpha1.ScyllaDBManagerTask
 	var errs []error
+
+	// If the annotation disabling global manager integration is set to true, we skip migration of tasks.
+	if sc.Annotations != nil && sc.Annotations[naming.DisableGlobalScyllaDBManagerIntegrationAnnotation] == naming.LabelValueTrue {
+		return scyllaDBManagerTasks, nil
+	}
 
 	for i := range sc.Spec.Backups {
 		smt, err := migrateV1BackupTaskSpecToV1Alpha1ScyllaDBManagerTask(sc, &sc.Spec.Backups[i])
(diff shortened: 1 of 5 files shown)


rzetelskik:
```
Allows opting out of v1.ScyllaCluster by labeling it explicitly with scylla-operator.scylladb.com/register-with-global-scylladb-manager: false.
```

We shouldn't use the same label as for ScyllaDBDatacenter/ScyllaDBCluster registration, it has different semantics. ScyllaClusters were always registered and this mechanism is only supposed to serve as a failsafe comparable to removing the entire standalone manager-controller deployment. Let's use a dedicated annotation.

test/e2e/set/scyllacluster/scyllamanager.go
```diff
@@ -291,18 +309,64 @@ var _ = g.Describe("Scylla Manager integration", func() {
 		sc, err = controllerhelpers.WaitForScyllaClusterState(taskDeletionCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, repairTaskDeletedCond)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		framework.By("Verifying that repair task deletion was synchronized")
-		// Progressing conditions of the underlying ScyllaDBManagerTasks are not propagated to ScyllaClusters by the migration controller,
-		// neither does the controller await the deletion of ScyllaDBManagerTasks.
-		// Task deletion from ScyllaDB Manager state is asynchronous to ScyllaCluster state update.
```
rzetelskik: why delete this? I feel like it provides important context

test/e2e/set/scyllacluster/scyllamanager.go
```diff
@@ -116,9 +94,58 @@ var _ = g.Describe("Scylla Manager integration", func() {
 			WithTimeout(utils.SyncTimeout).
 			WithPolling(5 * time.Second).
 			Should(o.Succeed())
-	})
-
-	g.It("should register cluster and sync repair tasks", func(ctx g.SpecContext) {
+	},
+		g.Entry("when manager integration disabled", func(ctx context.Context, sc *scyllav1.ScyllaCluster) {
```
rzetelskik: nit: is disabled

test/e2e/set/scyllacluster/scyllamanager.go
```diff
@@ -116,9 +94,58 @@ var _ = g.Describe("Scylla Manager integration", func() {
...
+			sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace).Patch(
+				ctx,
+				sc.Name,
+				types.MergePatchType,
+				patch,
+				metav1.PatchOptions{},
+			)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}),
+		g.Entry("when deleted", func(ctx context.Context, sc *scyllav1.ScyllaCluster) {
```
rzetelskik: nit: is deleted

pkg/controller/scyllacluster/migrate.go
```diff
@@ -323,8 +323,12 @@ func MigrateV1ScyllaClusterToV1Alpha1ScyllaDBDatacenter(sc *scyllav1.ScyllaClust
...
 			}
 
-			// Register the underlying ScyllaDBDatacenter with global ScyllaDB Manager instance for backward compatibility.
-			objectMeta.Labels[naming.GlobalScyllaDBManagerRegistrationLabel] = naming.LabelValueTrue
+			// If no disable annotation is set, or it's explicitly set to `false`, register the underlying ScyllaDBDatacenter with global ScyllaDB Manager instance for backward compatibility.
+			if v, ok := sc.Annotations[naming.DisableGlobalScyllaDBManagerIntegrationAnnotation]; !ok || v == naming.LabelValueFalse {
+				objectMeta.Labels[naming.GlobalScyllaDBManagerRegistrationLabel] = naming.LabelValueTrue
+			} else {
+				objectMeta.Labels[naming.GlobalScyllaDBManagerRegistrationLabel] = naming.LabelValueFalse
+			}
```
rzetelskik: nit: why not just leave it out?

pkg/controller/scyllacluster/migrate.go
```diff
@@ -477,6 +481,11 @@ func MigrateV1ScyllaClusterToV1Alpha1ScyllaDBManagerTasks(sc *scyllav1.ScyllaClu
 	var scyllaDBManagerTasks []*scyllav1alpha1.ScyllaDBManagerTask
 	var errs []error
 
+	// If the annotation disabling global manager integration is set to true, we skip migration of tasks.
+	if sc.Annotations != nil && sc.Annotations[naming.DisableGlobalScyllaDBManagerIntegrationAnnotation] == naming.LabelValueTrue {
+		return scyllaDBManagerTasks, nil
```
rzetelskik, line 486: (unrelated to this but I don't have a better place to leave this comment)
What about status? We print `Info` logs when SMCR/SMT objects are not found - should we not do it when integration is disabled? I don't have a strong opinion, but I'd like to make sure this is a deliberate decision.

pkg/controller/scyllacluster/migrate.go
czeslavo: `resourceapply.ApplyScyllaDBDatacenter` doesn't remove labels if they were present in the previous object's version. Setting the label explicitly allows handling the case where we create a `ScyllaCluster` with the integration enabled first, and then we want to disable the integration.

test/e2e/set/scyllacluster/scyllamanager.go
czeslavo: Unintentionally got deleted when moving code around. Brought it back.

pkg/controller/scyllacluster/migrate.go
czeslavo: That's a good point - I missed these logs, and it makes sense to handle that situation more gracefully. I've added a check that skips tasks' status migration when the integration is disabled and logs what happened.

Issue #2792, opened 2025-07-21
Release and Publish v1.18


mflendrich:
- [ ] Follow the release procedure. We're phasing out [the old release procedure](https://github.com/scylladb/scylla-operator/blob/master/release.md) in favor of [the new one](https://github.com/scylladb/scylla-operator-release/pull/418).
    - Please test-drive the release procedure draft from the PR and feel free to push changes where appropriate.
- [ ] When we have a first tagged pre-release build, let Q/A know by pinging @abramche with a container tag to be verified. Repeat for subsequent builds.

### Blockers
- https://github.com/scylladb/scylla-operator/issues/2572
- https://github.com/scylladb/scylla-operator/issues/2772
- https://github.com/scylladb/scylla-operator/issues/2788
- https://github.com/scylladb/scylla-operator/pull/2794

PR #2799, opened 2025-07-22
Bump scylla-monitoring submodule to 4.11.0

czeslavo:
**Description of your changes:** Bumps scylla-monitoring submodule to v4.11.0 and aligns Grafana and Prometheus in `config.yaml` so they match versions from the scylla-monitoring stack. Also updates prometheus-operator in the examples to the latest one that supports our Prometheus version.

Part of https://github.com/scylladb/scylla-operator/issues/2792.

/kind dependency-bump
/priority important-soon


examples/third-party/prometheus-operator/50_operator.deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
         env:
         - name: GOGC
           value: "30"
-        image: quay.io/prometheus-operator/prometheus-operator:v0.80.0
+        image: quay.io/prometheus-operator/prometheus-operator:v0.84.0
```
zimnx, line 34: Looks like you have bumped only the image version, while manifests are still taken from 0.80.0. Lets make sure manifests are also updated.

czeslavo: Yes, I didn't realize there was something to be manually copied over. PTAL now.

zimnx: Looks better now, thanks

PR #2805, opened 2025-07-23
Set up dependabot for docs

rzetelskik:
**Description of your changes:** This PR configures dependabot for poetry used in our docs. I'm following https://github.com/scylladb/scylladb/blob/master/.github/dependabot.yml. ~I suspect this won't work too well with poetry v1.* (https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-) or version pinning (https://github.com/dependabot/dependabot-core/issues/1556), but let's give this a go.~ edit: bumped poetry to v2, hopefully that suffices.

Resolves #

/kind machinery
/priority important-longterm

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,3 +21,11 @@ updates:
   labels:
   - priority/important-longterm
   - kind/dependency-bump
+- package-ecosystem: pip
+  directory: /docs
+  schedule:
+    interval: daily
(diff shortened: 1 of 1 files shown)


zimnx (commented):
lgtm
/assign mflendrich

mflendrich (approved):
/lgtm

PR #2818, opened 2025-07-24
Fix release workflow to not mark old patches as latest

czeslavo:
Fixes GH release workflow to not mark releases as latest if they're not the newest in semver terms. `make_latest` is a parameter that's set to `true` ([API spec](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release)).

Tested locally, giving the following requests:

- v1.17.1 (current latest)
```json
{
  "tag_name": "v1.17.1",
  "name": "v1.17.1",
  "draft": false,
  "prerelease": false,
  "body": "Hardcoded release notes",
  "make_latest": true
}
```
- v1.16.3 (older patch)
```json
{
  "tag_name": "v1.16.3",
  "name": "v1.16.3",
  "draft": false,
  "prerelease": false,
  "body": "Hardcoded release notes",
  "make_latest": false
}
```

/kind machinery
/priority important-longterm

### DIFF
--- a/.github/actions/release-notes/action.yaml
+++ b/.github/actions/release-notes/action.yaml
@@ -87,6 +87,16 @@ runs:
       release_name="${current_tag#v}"
       previous_release_name="${previous_tag#v}"
       
+      # Determine if current tag is the latest version among proper releases (excluding pre-releases).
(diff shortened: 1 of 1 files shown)


zimnx (approved):
/lgtm

PR #2819, opened 2025-07-24
Add verify-docs to root Makefile's verify target

rzetelskik:
**Description of your changes:** Add docs' Makefile `verify` target to the root's Makefile `verify` to have a single target verifying both the code and the docs.

Resolves #

/kind machinery
/priority important-longterm


rzetelskik:
:facepalm: poetry is not in the image

Issue #2816, opened 2025-07-23
[Flake] Scylla Manager integration should register cluster, sync repair tasks, and delete them [It] when manager integration is disabled


rzetelskik:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2811/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1947984117425508352#1:test-build-log.txt%3A3222

### Snippet of what failed.

```
  [FAILED] Timed out after 600.001s.
  The function passed to Eventually failed at github.com/scylladb/scylla-operator/test/e2e/utils/verification/scylladbmanager.go:26 with:
  Repair task is not done, status: "3fce3cbc-64c8-4b06-9fa9-1f6c58aa872f", cause: "ERROR".%!(EXTRA string=see more errors in logs: master 10.73.250.46 keyspace tqnk7g4z table test command 0: get tablet repair task status: agent [HTTP 400] task with id a62e2540-67bc-11f0-ab5f-eac489cfe6c2 not found)
  Expected
      <string>: ERROR
  to equal
      <string>: DONE
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllamanager.go:294 @ 07/23/25 12:10:37.817
```

rzetelskik:
```
2025-07-23T12:00:37.219529642Z {"L":"INFO","T":"2025-07-23T12:00:37.219Z","N":"scheduler","M":"PutTask","task":"repair/3fce3cbc-64c8-4b06-9fa9-1f6c58aa872f","schedule":{"cron":"{\"spec\":\"\",\"start_date\":\"0001-01-01T00:00:00Z\"}","window":null,"timezone":"","start_date":"0001-01-01T00:00:00Z","interval":"","num_retries":3,"retry_wait":""},"properties":{"fail_fast":false,"intensity":1,"parallel":2,"small_table_threshold":1073741824},"create":true,"_trace_id":"IM6FsJ9WTcyZ2e9WFyQUzQ"}
...
[...]

PR #2834, opened 2025-07-28
Explicitly set and unify the number of retries and backoff duration in ScyllaDB Manager tasks in E2E tests

rzetelskik:
**Description of your changes:** This PR explicitly sets and unifies `NumRetries` and `RetryWait` parameters in ScyllaDB Manager tasks created in our e2e tests. The default value for the latter is 10m, which results in transitive failures in task scheduling causing our test suites to fail, as the value is often larger than the corresponding timeouts, as can be observed in https://github.com/scylladb/scylla-operator/issues/2816.

Resolves https://github.com/scylladb/scylla-operator/issues/2816

/kind flake
/priority important-soon

### DIFF
--- a/test/e2e/set/scylladbmanagertask/scylladbmanagertask_scylladbdataceter_globalmanager.go
+++ b/test/e2e/set/scylladbmanagertask/scylladbmanagertask_scylladbdataceter_globalmanager.go
@@ -67,7 +67,10 @@ var _ = g.Describe("ScyllaDBManagerTask and ScyllaDBDatacenter integration with 
 				Type: scyllav1alpha1.ScyllaDBManagerTaskTypeRepair,
 				Repair: &scyllav1alpha1.ScyllaDBManagerRepairTaskOptions{
 					ScyllaDBManagerTaskSchedule: scyllav1alpha1.ScyllaDBManagerTaskSchedule{
-						NumRetries: pointer.Ptr[int64](1),
+						NumRetries: pointer.Ptr[int64](3),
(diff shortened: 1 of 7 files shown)


test/e2e/set/scylladbmanagertask/scylladbmanagertask_scylladbdataceter_globalmanager.go
```diff
@@ -67,7 +67,10 @@ var _ = g.Describe("ScyllaDBManagerTask and ScyllaDBDatacenter integration with
 				Type: scyllav1alpha1.ScyllaDBManagerTaskTypeRepair,
 				Repair: &scyllav1alpha1.ScyllaDBManagerRepairTaskOptions{
 					ScyllaDBManagerTaskSchedule: scyllav1alpha1.ScyllaDBManagerTaskSchedule{
-						NumRetries: pointer.Ptr[int64](1),
+						NumRetries: pointer.Ptr[int64](3),
+						RetryWait: &metav1.Duration{
+							Duration: 30 * time.Second,
+						},
```
czeslavo, line 73: How about extracting consts to a package that could be used from all test packages, and allow avoiding hardcodes like here?

rzetelskik: done, also for timeouts, which are a bit context-dependent but I suppose it's better to give it a higher shared value than fail prematurely

czeslavo (approved):
/lgtm

PR #2835, opened 2025-07-29
Extend timeout in DynamicSharedInformerFactory UT

czeslavo:
**Description of your changes:** Extends a `drainingTimeout` from `1s` to `10s/3 - 500ms ~ 3.83s` so that the test has more time to read all events in every test case. It's intentional that we read from a channel for a fixed duration, as we not only want to ensure we get the expected events, but also that we _do not_ receive unexpected ones.

`1s` proved to be not enough in CI (at least once): https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2810/pull-scylla-operator-master-unit/1949850720530337792

/kind flake
/priority important-longterm


pkg/remoteclient/informers/factory_test.go
```diff
@@ -272,7 +273,10 @@ func TestDynamicSharedInformerFactory_EventHandles_MultipleClusters(t *testing.T
 			t.Error(err)
 		}
 
-		observedObjects := drainChannel(ctx, informerUpdateObjCh, time.Second)
+		// Set drainingTimeout to be a bit less than the total test timeout divided by the number of test cases to ensure
+		// we still have time left for other operations to complete.
+		drainingTimeout := time.Duration(int(testTimeout)/len(tss)) - (time.Millisecond * 500)
```
zimnx: I've got a feeling that leaving 500ms for 6 updates, 6 waits for cache sync and entire test setup is going to introduce more flakiness. Lets leave more space for it.

czeslavo: Sounds reasonable. Bumped the total test timeout to 20s instead of 10, and ensured we have 10s for other operations spared.

zimnx (approved):
/lgtm

czeslavo:
timeout 
/test images
/retest

PR #2836, opened 2025-07-29
Grant Gateway API RBAC to haproxy ingress controller

zimnx:
Example haproxy installation was extended with necessary Gateway API RBAC roles.

After bumping OKD version on Openshift jobs, Gateway API CRDs are installed by default which presence is verified by Haproxy Ingress Controller.
Because we haven't granted necessary RBAC, it's failing to bootstrap breaking our e2e setup.
There's no way to disable it explicitly, haproxy only checks CRD presence.

### DIFF
--- a/examples/third-party/haproxy-ingress.yaml
+++ b/examples/third-party/haproxy-ingress.yaml
@@ -78,6 +78,25 @@ rules:
   - get
   - list
   - watch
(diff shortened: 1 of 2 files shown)


czeslavo (commented):
/lgtm

Issue #2828, opened 2025-07-28
Upgrade procedure for Operator 1.17 -> 1.18 path is documented


czeslavo:
The procedure for upgrading the Operator from v1.17 to v1.18 should be explicitly documented in our official documentation, as it includes a unique step of deleting the global ScyllaDB Manager deployment (both for Helm and GitOps).

An issue that verified the upgrade procedure: https://github.com/scylladb/scylla-operator/issues/2695

Issue #2716, opened 2025-06-25
Upgrade via Helm instructions do not mention manager upgrade


czeslavo:
### What happened?

Instructions for [upgrading the Operator with Helm](https://operator.docs.scylladb.com/stable/installation/helm.html#upgrade-via-helm) do not mention that the Manager's Helm installation should also be upgraded. 

It will be relevant when upgrading from `v1.17` to `v1.18` as the Manager's chart drops the standalone manager-controller deployment and its related resources.

### What did you expect to happen?

"Upgrade via Helm" section in https://operator.docs.scylladb.com/stable/installation/helm.html#upgrade-via-helm should say there's another step required to upgrade the manager's Helm installation. It should be done _before_ upgrading the Operator installation.

### How can we reproduce it (as minimally and precisely as possible)?

Follow Helm installation instructions, and try to upgrade from v1.17 to v1.18 (master) - you'll get the following error in the upgraded Operator logs:

```text
Error: Standalone ScyllaDB Manager controller Deployment "scylla-manager/scylla-manager-controller" should not be running alongside Scylla Operator. Delete it before starting Scylla Operator.
```

### Scylla Operator version

master

### Kubernetes platform name and version

irrelevant

### Please attach the must-gather archive.

irrelevant

PR #2837, opened 2025-07-29
Add a dedicated upgrade guide

czeslavo:
**Description of your changes:** Adds a dedicated upgrade guide to docs, including instructions specific to v1.17->v1.18 upgrade path.

Resolves https://github.com/scylladb/scylla-operator/issues/2828 
Resolves https://github.com/scylladb/scylla-operator/issues/2716

/kind documentation
/priority important-soon


docs/source/installation/gitops.md
```diff
@@ -92,7 +92,7 @@ spec:
           value: docker.io/scylladb/scylla-operator:1.14.0@sha256:8c75c5780e2283f0a8f9734925352716f37e0e7f41007e50ce9b1d9924046fa1
 :::
 The {{productName}} image value and the `SCYLLA_OPERATOR_IMAGE` shall always match.
-Be careful not to use a rolling tag for any of them to avoid an accidental skew!
+Be careful not to use a rolling tag (i.e., `{major}.{minor}`) for any of them to avoid an accidental skew!
```
rzetelskik: `e.g.` instead of `i.e.` if anything, as `latest` would probably the most notorious rolling tag. But I'm not sure this requires specification altogether.

docs/source/installation/upgrade.md
```diff
@@ -0,0 +1,74 @@
...
+all ScyllaClusters that also run using the image that's being updated. ({{productName}} injects it as a sidecar to help run and manage ScyllaDB.)
+
+We value the stability of our APIs and all API changes are backwards compatible.
+
+## Upgrade via GitOps (kubectl)
+
+A typical upgrade flow using GitOps (kubectl) requires re-applying the manifests using ones from the release you want to upgrade to.
+Please refer to the [GitOps installation instructions](./gitops.md) for details.
+
+If any additional steps are required for a specific version upgrade, they will be documented in the [Upgrade steps for specific versions](#upgrade-steps-for-specific-versions) section below.
```
rzetelskik: I'd make this more visible with a caution/warning and perhaps move it to the top of the block, given the steps might have to be performed before the usual apply

docs/source/installation/upgrade.md
```diff
@@ -0,0 +1,73 @@
...
+2. Update Scylla Operator
+   ```
+   helm upgrade --version <version> <release_name> scylla/scylla-operator
+   ```
+
+## Upgrade steps for specific versions
+
+### 1.17 to 1.18
+
+Upgrading from v1.17.x requires extra actions due to the removal of the standalone ScyllaDB Manager Controller.
```
rzetelskik: nit: s/Controller/controller

docs/source/installation/upgrade.md
```diff
@@ -0,0 +1,73 @@
...
+The controller becomes an integral part of the Operator. As a result, the ScyllaDB Manager Controller deployment and related resources
```
rzetelskik:
```suggestion
The controller becomes an integral part of the Operator. As a result, the standalone ScyllaDB Manager controller deployment and its related resources
```

docs/source/installation/upgrade.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Upgrade ScyllaDB Manager
+
+Replace `<release_name>` with the name of your Helm release for ScyllaDB Manager and replace `<version>` with the version number you want to install:
+
+```
+helm upgrade --version <version> <release_name> scylla/scylla-manager
+```
+
+### Upgrade Scylla Operator
```
rzetelskik:
```suggestion
### Upgrade {{productName}}
```
nit: the use of {{productName}} is a bit inconsistent in this file

docs/source/installation/upgrade.md
```diff
@@ -0,0 +1,73 @@
...
+    poddisruptionbudgets.policy/scylla-manager-controller \
+    serviceaccounts/scylla-manager-controller \
+    clusterrolebindings.rbac.authorization.k8s.io/scylladb:controller:manager-controller \
+    deployments.apps/scylla-manager-controller
+```
+
+#### Upgrade via Helm
+
+ScyllaDB Manager Helm installation has to be upgraded before upgrading Scylla Operator Helm installation, following
+the standard [Helm upgrade procedure](#upgrade-via-helm). This will ensure that the ScyllaDB Manager Controller
```
rzetelskik, line 76: isn't `#upgrade-via-helm` conflicting? This subsection has the same title

docs/source/installation/gitops.md
czeslavo: I was under the impression that we lack an explanation of what rolling tags are in our case (as having `{major}.{minor}` rolling is not that common). You're right, `latest` is also the case. WDYT about putting this explanation in the top `Disclaimer` section instead?

docs/source/installation/upgrade.md
czeslavo: Will move it as a caution to the top-level section as it applies equally to Helm/GitOps scenarios.

docs/source/installation/upgrade.md
czeslavo: Good catch. Sphinx handles that by assigning conflicting anchors a unique integer ID, but it will be better to assign them explicitly. I had to add `attrs_block` option to the `myst` parser config to make it possible.

docs/source/installation/upgrade.md
czeslavo: Done.

docs/source/installation/upgrade.md
czeslavo: Done.

docs/source/installation/upgrade.md
czeslavo: Replaced all occurrences of hardcoded `Scylla Operator` with the variable.

docs/source/installation/gitops.md
rzetelskik: It won't hurt to explain that both X.Y and latest are rolling tags in our case

czeslavo: Done.

rzetelskik (approved):
/lgtm

PR #2838, opened 2025-07-29
Fix broken links in docs

czeslavo:
**Description of your changes:** It fixes all the existing broken links in docs either by changing them to the correct ones in sources, or ignoring them if we have no control over them (i.e., when they originate from the Kubernetes types that we import).

/kind documentation
/priority important-longterm


docs/source/conf.py
```diff
@@ -191,3 +191,19 @@
...
+    # Ignore links that originate from the Kubernetes objects' documentation we have no control over.
+    'http://kubernetes.io/docs/user-guide/annotations',
+    'http://kubernetes.io/docs/user-guide/labels',
+    'https://examples.k8s.io/mysql-cinder-pd/README.md',
+    'https://examples.k8s.io/volumes/cephfs/README.md',
+    'https://examples.k8s.io/volumes/glusterfs/README.md',
+    'https://examples.k8s.io/volumes/iscsi/README.md',
+    'https://examples.k8s.io/volumes/rbd/README.md',
+    'https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources',
+    'https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds',
```
rzetelskik, line 205: why list all of those explicitly instead of using shared prefixes?

czeslavo: IMO it's better to be explicit and ignore only specific links than ignore more than intended accidentally. 

Perhaps for `examples/k8s.io/volumes` we could use a prefix, as we can see `volumes` were completely removed. Still, I'd leave it as is due to the reason I mentioned (I can imagine `volumes` will be brought back, and we could ignore something we didn't intend).

PR #2839, opened 2025-07-29
Bump patch version of third-party haproxy-ingress example

zimnx:
Version we used didn't have proper check if gateway API is available. It checked whether one particular resource is installed, but it hasn't validated whether given Gateway API spec is supported, nor whether other resources are in expected apiVersions. Latest patch release contains the fix for this logic which should solve issues we observe on Openshift. Only patch release is bumped because there're a lot of new major versions avaiable, doing a big step may uncover more gaps and I would like to do smaller steps eliminating one issue at a time.

/cc @czeslavo

### DIFF
--- a/examples/third-party/haproxy-ingress.yaml
+++ b/examples/third-party/haproxy-ingress.yaml
@@ -12,6 +12,9 @@ metadata:
   name: haproxy-ingress
 
 ---
+# Sources:
+# https://github.com/haproxytech/kubernetes-ingress/blob/v1.10.16/deploy/haproxy-ingress.yaml
+# https://github.com/haproxytech/kubernetes-ingress/blob/v1.10.16/documentation/gateway-api.md
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
@@ -65,6 +68,23 @@ rules:
   - core.haproxy.org
   resources:
   - '*'
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+- apiGroups:
+  - "discovery.k8s.io"
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - "apiextensions.k8s.io"
+  resources:
+  - customresourcedefinitions
   verbs:
   - get
   - list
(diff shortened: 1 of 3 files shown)


zimnx:
> @zimnx: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/e2e-openshift-aws-parallel 	[6c3b83c](https://github.com/scylladb/scylla-operator/commit/6c3b83c4dc10f40cbb849450028cca10ff7dad3f) 	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2839/pull-scylla-operator-master-e2e-openshift-aws-parallel/1950273996024975360) 	false 	`/test e2e-openshift-aws-parallel`
> 
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2839). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:zimnx).

It went through haproxy setup which this PR intends to fix, the remaining issue of missing permissions for setting up finalizers (https://github.com/scylladb/scylla-operator/issues/2841) I'm going to fix in separate PR.

examples/third-party/haproxy-ingress.yaml
```diff
@@ -78,6 +78,23 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
czeslavo, line 84: Out of curiosity: where are these manifests taken from? Can we create an issue to add automation updating them? For starters, we could leave a yaml comment pointing to source manifests.

zimnx: I took it from https://github.com/haproxytech/kubernetes-ingress/blob/v1.10.16/deploy/haproxy-ingress.yaml + https://github.com/haproxytech/kubernetes-ingress/blob/v1.10.16/documentation/gateway-api.md
There're manual steps required to get them, there're also additional manifests created by hand + custom values. I'm not sure automation is viable here nor worth pursuing as these manifests are rarely updated.

czeslavo: Let's at least document the manual steps in a comment so the next person updating this doesn't have to reinvent the wheel.

zimnx: done

Issue #2827, opened 2025-07-28
[Flake] ScyllaDBManagerTask and ScyllaDBCluster integration with global ScyllaDB Manager [It] should synchronise a backup task and support a manual restore procedure [MultiDatacenter]


rzetelskik:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-latest-e2e-gke-multi-datacenter-parallel/1949711530924707840#1:test-build-log.txt

### Snippet of what failed.

```  
[FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000c87120>: 
      can't insert data: Operation timed out for kh7dblvf.test - received only 6 responses from 9 CL=ALL. (potentially executed: false)
      {
          msg: "can't insert data: Operation timed out for kh7dblvf.test - received only 6 responses from 9 CL=ALL. (potentially executed: false)",
          err: <*gocql.QueryError | 0xc0005d22b8>{
              err: <*gocql.RequestErrWriteTimeout | 0xc000bed280>{
                  errorFrame: {
                      frameHeader: {version: 132, flags: 0, stream: 1472, op: 0, length: 104, warnings: nil},
                      code: 4352,
                      message: "Operation timed out for kh7dblvf.test - received only 6 responses from 9 CL=ALL.",
                  },
                  Consistency: 5,
                  Received: 6,
                  BlockFor: 9,
                  WriteType: "SIMPLE",
              },
              potentiallyExecuted: false,
              isIdempotent: false,
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/utils/verification/cql.go:33 @ 07/28/25 06:40:51.094
  Full Stack Trace
    github.com/scylladb/scylla-operator/test/e2e/utils/verification.InsertAndVerifyCQLDataUsingDataInserter({0x7f4cbd111000, 0xc000538900}, 0xc000413740)
    	github.com/scylladb/scylla-operator/test/e2e/utils/verification/cql.go:33 +0x78
    github.com/scylladb/scylla-operator/test/e2e/utils/verification.InsertAndVerifyCQLDataByDC({0x7f4cbd111000, 0xc000538900}, 0xc000b60328?)
    	github.com/scylladb/scylla-operator/test/e2e/utils/verification/cql.go:26 +0x79
[...]

PR #2842, opened 2025-07-30
Await schema agreement after keyspace creation by E2E's data inserter

rzetelskik:
**Description of your changes:** Schema agreement should be awaited after schema-altering statements. This PR adds it after keyspace and table creation in E2E's data inserter.

I'm taking a stab at minimising instability in https://github.com/scylladb/scylla-operator/issues/2827 here, although I'm not expecting it to be a root cause.

/kind failing-test
/priority important-soon

### DIFF
--- a/test/e2e/utils/datainserter.go
+++ b/test/e2e/utils/datainserter.go
@@ -108,7 +108,7 @@ func (di *DataInserter) SetClientEndpoints(hosts []string) error {
 	return nil
 }
 
-func (di *DataInserter) Insert() error {
+func (di *DataInserter) Insert(ctx context.Context) error {
 	ss := make([]string, 0, len(di.replicationFactor))
 	for dc, rf := range di.replicationFactor {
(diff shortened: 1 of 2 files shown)


test/e2e/utils/datainserter.go
```diff
@@ -134,6 +134,12 @@ func (di *DataInserter) Insert() error {
 		return fmt.Errorf("can't create table: %w", err)
 	}
 
+	framework.Infof("Awaiting schema agreement")
+	err = di.session.AwaitSchemaAgreement(ctx)
+	if err != nil {
+		return fmt.Errorf("can't await schema agreement: %w", err)
+	}
```
zimnx, line 141: ~~Since schema is managed by raft awaiting for schema agreement is a noop. I doubt it's the root cause of the flake.~~
Raft will guarantee Q nodes have the schema, so it's not a noop. While we would want to execute CL=ALL queries I think it's desired to have it here.

Issue #2732, opened 2025-06-30
Test latest 2025.2.0-rc with Scylla Operator


zimnx:
We need to make sure there're no regressions and 1.18 is going to support 2025.2.

mflendrich:
Blocked on the availability of a 2025.2.0 RC that includes the fix to the missing executables problem.

mflendrich:
> and 1.18 is going to support 2025.2.

For posterity - as this issue got closed hinting that 1.18 supports 2025.2 at the time of release:

1.18.0 will not support 2025.2. Support (for >=2025.2.1) )will (probably) be introduced in a patch release.

PR #2850, opened 2025-08-04
Explicitly use root user on tuning containers

zimnx:
So far we have used user originating from the default of Scylla container image. Instead relying on what is defaulted (and which was changed to non-root), we should explicitly use root user for tuning containers.

Part of #2732


rzetelskik (approved):
/lgtm

PR #2851, opened 2025-08-04
Bump ScyllaDB to 2025.2.1

### LINKED ISSUE #2732: Test latest 2025.2.0-rc with Scylla Operator

zimnx:

Bumps default version of ScyllaDB to 2025.2.1.

Requires:

- [ ] #2850 

Resolves #2732


assets/config/config.yaml
```diff
@@ -13,6 +13,6 @@ operator:
   prometheusVersion: "v3.5.0" # Tracks scylla-monitoring/versions.sh PROMETHEUS_VERSION
 operatorTests:
   scyllaDBVersions:
-    updateFrom: "2025.1.4" # One patch lower than .operator.scyllaDBVersion
-    upgradeFrom: "6.2.3" # One minor lower than .operator.scyllaDBVersion
+    updateFrom: "2025.2.0" # One patch lower than .operator.scyllaDBVersion
```
rzetelskik, line 16: Shouldn't we use an rc here? Since we don't support 2025.2.0 (even if it's just due to the utils)

zimnx: There were none 2025.2.1-rc's. 2025.2.1 is the first version from 2025.2.x we support, neither .0 nor any rc is supported.

Issue #2803, opened 2025-07-23
Operator: Preflight passes locally


mflendrich:
## Acceptance Criteria
- [ ] A passing run documented by a `preflight` report for a specific commit

Issue #2802, opened 2025-07-23
Operator bundle have a place in the repository


mflendrich:
## Acceptance criteria

- [ ] The bundle is committed scylladb/scylla-operator (or some other place)

zimnx:
Extracted catalog part to separate issue https://github.com/scylladb/scylla-operator/issues/2905

PR #2885, opened 2025-08-14
Add OLM Bundle

zimnx:
Adds OLM Bundle a prerequisite toward OpenShift certification for our Operator.
With this, the Operator can now be installed via OLM.
The bundle is autogenerated from different sources like deploy manifests and manual metadata (like logo, project description, capabilities, minimum supported versions, etc.)

Along the way, I also made a few webhook fixes that came up during preflight testing.
I recommend reviewing this commit by commit - each one has notes on what changed and why.

Fixes scylladb/scylla-operator#2803
Fixes scylladb/scylla-operator#2802
Fixes scylladb/scylla-operator-release#481


Makefile
```diff
@@ -635,10 +638,79 @@ verify-links:
...
+# $2 - path to file we merge patch to
+define merge-yaml-files
+	$(YQ) -i -e '. *= load("$(2)")' $(1)
+endef
+
+# $1 - path to patch file (the target being patched)
+# $2 - glob pattern for patch files
+define patch-bundle
+	for f in $(2); do \
+		[ -e "$$f" ] || continue; \
```
rzetelskik: what's the idea behind `[ -e "$$f" ]`? How can it *not* exists if it's being listed? And why continue then, is it expected?

Makefile
```diff
@@ -635,10 +638,79 @@ verify-links:
...
+		$(YQ) -i -e '. *= load("'"$$f"'")' $(1); \
```
rzetelskik, line 645: isn't it possible to use `merge-yaml-files` defined above? Or use https://mikefarah.gitbook.io/yq/operators/multiply-merge#merging-all-files to merge all files?

Makefile
```diff
@@ -635,10 +638,79 @@ verify-links:
 	fi;
 .PHONY: verify-links
 
-verify: verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-lint verify-helm-lint verify-links verify-examples verify-docs-api verify-monitoring
+# $1 - path to patch file
+# $2 - path to file we merge patch to
+define merge-yaml-files
```
rzetelskik: seems unused

Makefile
```diff
@@ -635,10 +638,79 @@ verify-links:
...
+endef
+
+# $1 - path to bundle
+define apply-bundle-fixes
+	$(call patch-bundle,$(1)/manifests/scylla-operator.clusterserviceversion.yaml,$(1)/patches/*.clusterserviceversion.yaml)
+
+	# Workaround for https://github.com/operator-framework/operator-registry/issues/1741
+	$(call fix-bundle-manifests-filenames, $(1)/manifests)
+endef
+update-bundle:
```
rzetelskik: nit: add empty line

Makefile
```diff
@@ -635,10 +638,79 @@ verify-links:
...
+# $1 - path to manifests directory
+define fix-bundle-manifests-filenames
+	find $(1) -type f -name '*:*' | while read file; do \
+		newfile=$$(echo "$$file" | sed 's/:/-/g'); \
+		mv "$$file" "$$newfile"; \
+	done
+endef
+
+# $1 - logo path
+# $2 - logo path path
```
rzetelskik: logo path path?

assets/metadata/metadata.yaml
```diff
@@ -0,0 +1,4 @@
+operator:
+  minOpenShiftVersion: 4.19
+  maxOpenShiftVersion: 4.19
+  minKubernetesVersion: 1.33
```
rzetelskik: we claim `>=1.30 && <=1.33` in master, shouldn't this reflect it (or vice-versa, the docs should reflect >=1.33 because of it, unless it's just for openshift)?

assets/metadata/metadata.yaml
```diff
@@ -0,0 +1,4 @@
+operator:
```
rzetelskik, line 3: can we have a comment explaining what this file is and how it differs from the config file? Is it only related to openshift?

Makefile
```diff
@@ -635,10 +638,79 @@ verify-links:
...
+define update-bundle-patches
+	$(call update-bundle-patches-logo,./logo.svg,$(1)/patches/logo.clusterserviceversion.yaml)
+	$(call update-bundle-patches-metadata,assets/metadata/metadata.yaml,$(1)/patches/metadata.clusterserviceversion.yaml)
+endef
+
+# $1 - path to bundle
+define apply-bundle-fixes
+	$(call patch-bundle,$(1)/manifests/scylla-operator.clusterserviceversion.yaml,$(1)/patches/*.clusterserviceversion.yaml)
+
+	# Workaround for https://github.com/operator-framework/operator-registry/issues/1741
```
rzetelskik, line 690: do we have a tracker to know when this should be removed? The referenced PR has already been merged so I suppose we wait for a release?

pkg/cmd/operator/webhooks.go
```diff
@@ -91,9 +93,37 @@ type Validator interface {
 	GetWarningsOnUpdate(obj, oldObj runtime.Object) []string
 }
 
+// Container runtime may set env var (SCYLLA_OPERATOR_PORT) to either number or "tcp://<service-host>:<service-port>"
+var portFlagRe = regexp.MustCompile(`^(?:tcp://[^:]+:|)(\d+)$`)
```
rzetelskik: nit: `^(?:tcp://[^:]+:)?(\d+)$` seems clearer to me (optional group marked with `?`), but that's up to you

helm/scylla-operator/templates/webhookserver.deployment.yaml
```diff
@@ -37,6 +37,7 @@ spec:
         - --loglevel=2
         - --tls-cert-file=/tmp/k8s-webhook-server/serving-certs/tls.crt
         - --tls-private-key-file=/tmp/k8s-webhook-server/serving-certs/tls.key
+        - --port=5000
```
rzetelskik, line 40: question: what's the purpose of the `SCYLLA_OPERATOR_PORT` env var? Why is it done on the application level and not in a k8s' service?

Makefile
zimnx: Added https://github.com/scylladb/scylla-operator/issues/2912, yes, we need to wait until it's released.

Makefile
zimnx: it's a leftover from initial implementation, removed

Makefile
zimnx: merge-yaml-files is gone. The solution suggested merges file in memory, we need to do it inplace so I left as is

Makefile
zimnx: removed

Makefile
zimnx: done

Makefile
zimnx: done

assets/metadata/metadata.yaml
zimnx: changed to 1.30

assets/metadata/metadata.yaml
zimnx: done

pkg/cmd/operator/webhooks.go
zimnx: done

rzetelskik (approved):
\lgtm
I left two more comments but I won't hold the PR as I'm not in tomorrow - please consider sending a follow-up unless you disagree with the approach.

bundle/patches/metadata.clusterserviceversion.yaml
```diff
@@ -0,0 +1,53 @@
...
+    features.operators.openshift.io/fips-compliant: "false"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "false"
+    features.operators.openshift.io/csi: "false"
+    com.redhat.openshift.versions: v4.9-v4.9
```
rzetelskik: I'm not a fan of this file being only partially autogenerated - I was confused by these values being later updated in the "Update autogenerated" commit. Could this serve as a source file separate from the one updated by the makefile target, i.e. the "autogenerated" one? That way the manual changes would always be separate from the generated ones.
Alternatively, this could use a comment explaining that this value will be overwritten by our machinery.

zimnx: good idea extracted values which are autogenerated to separate file

zimnx:
> \lgtm
I left one more comment but I won't hold the PR as I'm not in tomorrow - please consider sending a follow-up unless you disagree with the approach.

Applied so I will apply lgtm manually

zimnx:
Something is wrong with CI, it's still using old job definition giving a false positive
/hold

zimnx:
> Something is wrong with CI, it's still using old job definition giving a false positive

https://github.com/scylladb/scylla-operator-release/pull/480 wasn't merged, job is updated now. 

I also fixed one issue, apparently 1.30 specified as minKubeVersion is invalid, changed to 1.30.0 which is a valid one.

zimnx:
Requires: https://github.com/scylladb/scylla-operator-release/pull/484

Makefile
```diff
@@ -635,10 +638,73 @@ verify-links:
...
+	select(fi==0).spec.minKubeVersion = ( select(fi==1) | .operator.minKubernetesVersion ) | \
+	select(fi==0).metadata.annotations."com.redhat.openshift.versions" = ( select(fi==1) | "v" + .operator.minOpenShiftVersion + "-v" + .operator.maxOpenShiftVersion ) | \
+	select(fi==0)' \
+	'$(2)' '$(1)'
+endef
+
+# $1 - bundle path
+define update-bundle-patches
+	$(call update-bundle-patches-logo,./logo.svg,$(1)/patches/logo.clusterserviceversion.yaml)
+	$(call update-bundle-patches-versions,assets/metadata/metadata.yaml,$(1)/patches/versions.clusterserviceversion.yaml)
```
rzetelskik, line 683: I think the two files that are updated automatically could still use a comment saying this, otherwise it can only be discovered by going through the Makefile targets. But it can be done separately, without retriggering CI.

rzetelskik (approved):
/lgtm

PR #2890, opened 2025-08-15
chore: Update cert-manager to 1.17.4

### LINKED ISSUE #2883: chore: Update cert-manager to a recent version
### Problem
By default, our examples deploy the cert-manager of a [very old version](https://github.com/scylladb/scylla-operator/blob/v1.18/examples/third-party/cert-manager.yaml#L5400) that does not support current k8s version ranges.

### Solution
Update the cert-manager to at least [1.17.4 LTS](https://cert-manager.io/docs/releases/release-notes/release-notes-1.17#v1174) which will support k8s up to 1.33. 

v1.18.0 is more recent, but it includes some breaking changes with signing process so a bit of a bigger effort.

abramche:
Bump Kubernetes cert-manager to 1.17.4 LTS to support k8s range up to 1.33

Resolves #2883

### DIFF
--- a/examples/third-party/cert-manager.yaml
+++ b/examples/third-party/cert-manager.yaml
@@ -27,7 +27,7 @@ metadata:
     app.kubernetes.io/name: 'cert-manager'
     app.kubernetes.io/instance: 'cert-manager'
     # Generated labels
-    app.kubernetes.io/version: "v1.14.0"
+    app.kubernetes.io/version: "v1.17.4"
 spec:
   group: cert-manager.io
   names:
@@ -225,7 +225,7 @@ metadata:
     app.kubernetes.io/name: 'cert-manager'
     app.kubernetes.io/instance: 'cert-manager'
     # Generated labels
-    app.kubernetes.io/version: "v1.14.0"
+    app.kubernetes.io/version: "v1.17.4"
 spec:
   group: cert-manager.io
   names:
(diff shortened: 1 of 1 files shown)


abramche:
> What's the source of these changes? There are many more differences between this and https://github.com/cert-manager/cert-manager/releases/download/v1.17.4/cert-manager.yaml.

Current cert manager does not support two of our three k8s versions (1.32 and 1.33) so it needs to be updated. I have successfully deployed a ScyllaDB cluster using the configuration in this PR. Do you think more changes are necessary?

rzetelskik:
> Current cert manager does not support two of our three k8s versions (1.32 and 1.33) so it needs to be updated. I have successfully deployed a ScyllaDB cluster using the configuration in this PR.

Let me rephrase: what is the source of the updated manifests? If you just changed the versions in the existing ones, did you verify the CRDs don't need updating and there are no new components we're missing that way?

>  Do you think more changes are necessary?

I think our third-party manifests need to be aligned with the upstream source to avoid reasoning about whether the differences will cause any issues.

abramche:
> Let me rephrase: what is the source of the updated manifests? If you just changed the versions in the existing ones, did you verify the CRDs don't need updating and there are no new components we're missing that way?

> I think our third-party manifests need to be aligned with the upstream source to avoid reasoning about whether the differences will cause any issues.

I only bumped the version and changed nothing else. It worked for me as is on both GKE and EKS, but I see what you mean. Still, leaving it as is is even more dangerous as two of three k8s versions are unsupported and we can only hope the user decides to themselves they need a fresh version of cert-manager.

I do not have more time to work on this right now (there are pressing problems with SCT), but you are welcome to take this over; alternatively we can close this for now/convert to draft, if we accept the aforementioned risk and ill guidance from our side.

rzetelskik:
@abramche this boils down to updating the manifests by copying the upstream ones instead of manually editing the versions but ok, I can take this. Closing in favour of https://github.com/scylladb/scylla-operator/pull/2900.

PR #2891, opened 2025-08-15
chore: Update example operator tag in gitops.md to 1.18

abramche:
Update the example operator YAML to use the current Operator release version

-

### DIFF
--- a/docs/source/installation/gitops.md
+++ b/docs/source/installation/gitops.md
@@ -85,11 +85,11 @@ spec:
       containers:
       - name: scylla-operator
         # ...
-        image: docker.io/scylladb/scylla-operator:1.14.0@sha256:8c75c5780e2283f0a8f9734925352716f37e0e7f41007e50ce9b1d9924046fa1
+        image: docker.io/scylladb/scylla-operator:1.18
         env:
           # ...
         - name: SCYLLA_OPERATOR_IMAGE
-          value: docker.io/scylladb/scylla-operator:1.14.0@sha256:8c75c5780e2283f0a8f9734925352716f37e0e7f41007e50ce9b1d9924046fa1
+          value: docker.io/scylladb/scylla-operator:1.18
 :::
 The {{productName}} image value and the `SCYLLA_OPERATOR_IMAGE` shall always match.
 Be careful not to use a rolling tag for any of them to avoid an accidental skew!


docs/source/installation/gitops.md
```diff
@@ -85,11 +85,11 @@ spec:
       containers:
       - name: scylla-operator
         # ...
-        image: docker.io/scylladb/scylla-operator:1.14.0@sha256:8c75c5780e2283f0a8f9734925352716f37e0e7f41007e50ce9b1d9924046fa1
+        image: docker.io/scylladb/scylla-operator:1.18
```
rzetelskik: It says `Be careful not to use a rolling tag for any of them to avoid an accidental skew!` just below.
Although this is just an example, I see some value in keeping this up to date, but we have to make it correct by either using a patch release tag or a digest. Either way this should be updated in the source branch or, preferably, use a substitution, provided we find a way to automate updating the tag/digest, otherwise we'd have to update it on every patch release.

abramche: Indeed, updated to a patch tag. I agree automation would be even better in the long run, but for now I suppose we can update the latest release only?

docs/source/installation/gitops.md
```diff
@@ -85,11 +85,11 @@ spec:
       containers:
       - name: scylla-operator
         # ...
-        image: docker.io/scylladb/scylla-operator:1.18
+        image: docker.io/scylladb/scylla-operator:1.18.0
```
rzetelskik:
```suggestion
        image: docker.io/scylladb/scylla-operator:1.18.0@sha256:ea1d287c2699efdfdb040b4a8c35be74a12d19b292541d8a4204f7a079731ca6
```

docs/source/installation/gitops.md
```diff
@@ -85,11 +85,11 @@ spec:
...
         env:
           # ...
         - name: SCYLLA_OPERATOR_IMAGE
-          value: docker.io/scylladb/scylla-operator:1.18
+          value: docker.io/scylladb/scylla-operator:1.18.0
```
rzetelskik: ditto

docs/source/installation/gitops.md
rzetelskik: patch version tags are not rolling but let's keep the digest as well as a good practice example

abramche: done in both suggestions

rzetelskik:
/approve
/lgtm
Thanks!

Issue #2579, opened 2025-04-09
Node setup fails on filesystem creation in `v1.32.3-gke.1170000`


rzetelskik:
### What happened?

Node setup fails on filesystem creation in GKE `v1.32.3-gke.1170000`.

```
E0409 10:57:50.806556       1 nodesetup/controller.go:192] "Unhandled Error" err="syncing key 'scylladb-nodepool-1' failed: [can't sync filesystems: failed to create filesystems: can't create filesystem \"xfs\" on device \"/dev/md/nvmes\" at \"/dev/md127\": can't run mkfs with args [-t xfs -b size=4096 -K /dev/md127]: exit status 1, stdout: \"\", stderr: \"mkfs: failed to execute mkfs.xfs: No such file or directory\\n\", can't sync mounts: can't create mounts: can't ensure units: unit \"var-lib-persistent\\\\x2dvolumes.mount\" is in a \"failed\" active state]" logger="UnhandledError"
...
I0409 10:57:51.295491       1 record/event.go:389] "Event occurred" object="scylladb-nodepool-1" fieldPath="" kind="NodeConfig" apiVersion="scylla.scylladb.com/v1alpha1" type="Warning" reason="CreateFilesystemFailed" message="Failed to create filesystem xfs on /dev/md/nvmes device: can't run mkfs with args [-t xfs -b size=4096 -K /dev/md127]: exit status 1, stdout: \"\", stderr: \"mkfs: failed to execute mkfs.xfs: No such file or directory\\n\""
```

### What did you expect to happen?

Node setup should be able to create the filesystem.

### How can we reproduce it (as minimally and precisely as possible)?

1. Create the GKE cluster with `v1.32.3-gke.1170000` cluster version (the current latest, **as advised by GKE quickstart in our documentation**).
2. Follow the GitOps installation guide until the NodeConfig step: https://operator.docs.scylladb.com/master/installation/gitops.html.
3. Verify that the NodeConfig is in a degraded state.

### Scylla Operator version

master

### Kubernetes platform name and version

<details>
```console
$ kubectl version
Client Version: v1.32.3
Kustomize Version: v5.5.0
Server Version: v1.32.3-gke.1170000
```

Kubernetes platform info:
GKE with v1.32.3-gke.1170000 cluster-version.
</details>

### Please attach the must-gather archive.

[must-gather.tar.gz](https://github.com/user-attachments/files/19665666/must-gather.tar.gz)

PR #2892, opened 2025-08-18
Add GKE (Ubuntu) >= 1.32 to unsupported platforms

czeslavo:
**Description of your changes:** Adds GKE >= 1.32 with Ubuntu OS image to unsupported platforms due to #2579.

### DIFF
--- a/docs/source/support/releases.md
+++ b/docs/source/support/releases.md
@@ -98,7 +98,8 @@ The following platforms are known **not to work correctly** at this time.
 :::{table}
 | Platform         | OS Image     | Details |
 |:-----------------|:-------------| :------ |
-| GKE              | Container OS |         |
+| GKE              | Container OS | Lack of XFS support |
+| GKE >= 1.32      | Ubuntu       | See [#2579](https://github.com/scylladb/scylla-operator/issues/2579) |
 | EKS              | Bottlerocket | Suspected kernel/cgroups issue that breaks available memory detection for ScyllaDB |
 :::
 :::


docs/source/support/releases.md
```diff
@@ -98,7 +98,8 @@ The following platforms are known **not to work correctly** at this time.
 :::{table}
 | Platform         | OS Image     | Details |
 |:-----------------|:-------------| :------ |
-| GKE              | Container OS |         |
+| GKE              | Container OS | Lack of XFS support |
+| GKE >= 1.32      | Ubuntu       | See [#2579](https://github.com/scylladb/scylla-operator/issues/2579) |
```
rzetelskik, line 102: Shouldn't this be mentioned in the support matrix (https://operator.docs.scylladb.com/stable/support/releases.html#support-matrix) instead of here?

czeslavo: Support matrix is platform-agnostic, isn't it? I understand this table gathers exceptions for particular platforms, and the GKE >= 1.32 case fits here.

rzetelskik: You're right. But there's also [Supported Kubernetes Platforms](https://operator.docs.scylladb.com/master/support/releases.html#supported-kubernetes-platforms) where we claim support for Ubuntu. Doesn't it look weird to say >=1.32 is unsupported just below, also making it harder to see? Maybe we could just use a warning block above the `Supported Kubernetes platforms` table? Or anything that would make it stand out more and seems less self-contradictory?

rzetelskik: on the second look the unsupported table is already a warning block, so I'm leaving this up to you

PR #2897, opened 2025-08-19
Handle missing xfsprogs on GKE >= 1.32

czeslavo:
**Description of your changes:** Starting with GKE version 1.32.1-gke.1002000, the default OS image for Ubuntu is updated from Ubuntu 22.04 to Ubuntu 24.04 ([source](https://cloud.google.com/kubernetes-engine/docs/release-notes#January_23_2025)). This OS image does not have the `xfsprogs` package preinstalled, which is required for the Operator to create XFS filesystems. Therefore, we have to manually install the missing dependency.

This PR documents that from the perspective of a user in our docs, and adds an installation step to our GKE CI scripts.

**Which issue is resolved by this Pull Request:**
Resolves https://github.com/scylladb/scylla-operator/issues/2579.

PR #2902, opened 2025-08-20
Update supported Kubernetes versions

czeslavo:

- Removes a note about unsupported GKE 1.32 as it's been addressed already by https://github.com/scylladb/scylla-operator/pull/2897
- Adds upper bounds for all Operator versions' supported Kubernetes (the upper bound should be the version we test against in CI)


zimnx (approved):
/lgtm

PR #2909, opened 2025-08-21
Add ScyllaDBMonitoring external Prometheus support enhancement proposal

czeslavo:
**Description of your changes:** Introduces an enhancement proposal for extending the `v1alpha1.ScyllaDBMonitoring` CRD with support for an external Prometheus instance.

Design for https://github.com/scylladb/scylla-operator/issues/2490.

### DIFF
--- a/enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
+++ b/enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
@@ -1,0 +1,627 @@
+# Extending ScyllaDBMonitoring with external Prometheus support
+
+## Summary
+
+This proposal aims to extend the `ScyllaDBMonitoring` CRD to support external Prometheus instances. Currently, there's
+no possibility to configure the monitoring stack using this CRD
+other way than letting it create managed Prometheus and Grafana instances. The goal is to allow users to configure
+`ScyllaDBMonitoring` to _not_ deploy managed Prometheus, and to allow configuring managed Grafana instance with an
+external Prometheus as a data source.
+
+## Motivation
+
+It is common for users to already operate a Prometheus instance (e.g.,
+OpenShift [user-workload monitoring](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/monitoring/configuring-user-workload-monitoring),
+a global corporate Prometheus stack, etc.). Running a second Prometheus solely for ScyllaDB introduces unnecessary
+duplication, resource overhead, and management
+complexity.
+
+By providing knobs to skip deploying Prometheus and to configure Grafana with an external Prometheus as a data source,
+we give users the flexibility to integrate with their existing monitoring stack.
+
+### Goals
+
+- Allow users to disable the managed Prometheus component.
+- Allow users to configure the managed Grafana instance to use an external Prometheus instance as a data source.
+- Maintain backwards compatibility: default behavior remains unchanged (managed Prometheus enabled, Grafana pointing at
+  it).
+- Allow secure connection to the external Prometheus instance using TLS (CA cert, client cert/key) and bearer token
+  authentication.
+
+### Non-Goals
+
+- Support Prometheus instances installed other way than by the Prometheus Operator (to keep integration with Prometheus
+  via `ServiceMonitor` and `PrometheusRule` resources).
+- Add support for non-Prometheus Grafana data sources.
+- Support for disabling the managed Grafana instance.
+- Support all possible authentication methods.
+
+## Proposal
+
+I propose to extend the `v1alpha1.ScyllaDBMonitoring` CRD with new fields that will allow users to disable the managed
+Prometheus instance and configure the managed Grafana instance to use an external Prometheus instance as a data source.
+
+The operator will keep creating scraping configuration for an external Prometheus instance via `ServiceMonitor` and
+`PrometheusRule`
+resources as it does now.
+
+Grafana data source provisioning configuration will be provided inline in the CR spec, using a strongly-typed API.
+It will support only a single Prometheus data source, as ScyllaDB monitoring stack dashboards are expecting the
+Prometheus data source to be named `prometheus`.
+
+It will allow configuring connection to the external Prometheus instance using TLS (CA cert, client cert/key) and bearer
+token auth.
+
+### User Stories
+
+#### Story 1
+
+As a cluster admin on OpenShift, I want to monitor Scylla Operator-managed ScyllaDB cluster, and I already have a
+Prometheus instance running in my cluster. I want the Operator to configure its managed Grafana instance to use my
+existing Prometheus instance as a data source.
+
+#### Story 2
+
+As a cluster admin on Kubernetes, I want to monitor Scylla Operator-managed ScyllaDB cluster, and I already have a
+Prometheus instance running in my cluster (deployed with Prometheus Operator). I want the Operator to configure its
+managed Grafana instance to use my existing Prometheus instance as a data source.
+
+### Notes
+
+- We assume that the "external Prometheus" means a Prometheus instance that is managed by
+  a [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator). This is because the ScyllaDB
+  Operator uses `ServiceMonitor` and `PrometheusRule` resources to define scraping and alerting rules, which are
+  reconciled by the Prometheus Operator.
+
+### Risks and Mitigations
+
+1. Strongly-typed Grafana datasource specification will not include all possible fields supported by Grafana.
+
+   We consciously decided to use a strongly-typed API to make sure all supported configuration options are validated
+   and tested on our side. If an extension is needed in the future, we can add more fields as necessary based on user
+   feedback.
+
+## Design Details
+
+### Proposed API changes
+
+`ScyllaDBMonitoring` CRD will be extended with the following changes:
+
+1. Add a new `spec.components.prometheus.mode` field to allow users to choose the mode of the Prometheus component.
+   The field will be an enum, defaulting to `Managed` (i.e., Operator-managed Prometheus is enabled by default).
+   When set to `External`, the Operator will not deploy a managed Prometheus instance. Only the scraping and alerting
+   resources will be created (`ServiceMonitor`s and `PrometheusRule`s) to allow the external Prometheus instance to
+   scrape metrics and alerts from the ScyllaDB cluster.
+
+    ```go
+    // PrometheusMode describes the mode of the Prometheus instance.
+    // +kubebuilder:validation:Enum="Managed";"External"
+    type PrometheusMode string
+
+    const (
+        // PrometheusModeManaged defines a mode where a `Prometheus` object is created as a child of a `ScyllaDBMonitoring`
+        // object. `ServiceMonitor` and `PrometheusRule` resources are also created to configure scraping and alerting.
+		// This mode requires a Prometheus Operator to be installed in the cluster.
+        PrometheusModeManaged PrometheusMode = "Managed"
+   
+        // PrometheusModeExternal defines a mode where no `Prometheus` child object is created, but `ServiceMonitor` and
+        // `PrometheusRule` objects are still created to configure scraping and alerting.
+        // This mode requires a Prometheus Operator to be installed in the cluster, along with a `Prometheus` instance
+        // configured to reconcile `ServiceMonitor` and `PrometheusRule` resources.
+        PrometheusModeExternal PrometheusMode = "External"
+    )
+    
+    type PrometheusSpec struct {
+        // mode defines the mode of the Prometheus instance. It's an immutable field.
+        // +kubebuilder:default:="Managed"
+        // +optional
+        Mode PrometheusMode `json:"mode,omitempty"`
+        ...
+    }
+    ```
+
+   Externally managed Prometheus instances can be configured to reconcile `ServiceMonitor` and `PrometheusRule`
+   resources by setting `ruleSelector` and `serviceMonitorSelector` fields in
+   the [Prometheus](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.Prometheus) spec.
+   We will document that these fields have to match the
+   `scylla-operator.scylladb.com/scylladbmonitoring-name: "{{ .scyllaDBMonitoringName }}"`
+   label in order to work, as that's the label the Operator adds to the `ServiceMonitor` and `PrometheusRule` resources
+   it creates.
+
+   The added `mode` field will be immutable to prevent users from switching between managed and external Prometheus
+   modes. Users will be asked to delete and recreate the `ScyllaDBMonitoring` resource if they want to switch modes.
+
+2. Add new `datasources` field to `spec.components.grafana` to allow users to configure the managed Grafana instance to
+   use an external Prometheus instance as a data source.
+
+   Because ScyllaDB monitoring stack dashboards are expecting the Prometheus data source to be named `prometheus`, we
+   will support only 1 data source in this field for now, but we will make it a list to allow adding more data sources
+   in the future if needed.
+
+   ```go
+   type GrafanaSpec struct {
+        ...
+       // datasources is a list of Grafana datasources to configure.
+       // It's expected to be set when using Prometheus component in `External` mode.
+       // At most one datasource is allowed for now (only Prometheus is supported).
+       // +kubebuilder:validation:MaxItems=1
+       // +optional
+       Datasources []GrafanaDatasourceSpec `json:"datasources,omitempty"`
+   }
+   ```
+
+   We will set the datasource name to `prometheus` by default and the type to `Prometheus`, so users
+   don't have to specify them explicitly.
+
+   To securely provide sensitive information, all sensitive fields in the Grafana data source configuration (like tokens
+   or TLS certificates) will be expected to be provided via `Secret` references.
+
+   ```go
+   type GrafanaDatasourceSpec struct {
+       // name is the name of the datasource as it will appear in Grafana.
+       // Only "prometheus" is supported for now as that's the datasource name expected by the ScyllaDB monitoring stack dashboards.
+       // +kubebuilder:validation:Enum="prometheus"
+       // +kubebuilder:default:="prometheus"
+       Name string `json:"name,omitempty"`
+
+       // type is the type of the datasource. Only "prometheus" is supported.
+       // +kubebuilder:default:="Prometheus"
+       // +optional
+       Type GrafanaDatasourceType `json:"type,omitempty"`
+
+       // url is the URL of the datasource.
+       // +kubebuilder:validation:MinLength=1
+       URL string `json:"url"`
+
+       // prometheus defines Prometheus-specific options.
+       // +optional
+       PrometheusOptions *GrafanaPrometheusDatasourceOptions `json:"prometheusOptions,omitempty"`
+   }
+   ```
+
+   Strongly-typed enum will be used for the datasource type:
+
+   ```go
+   // GrafanaDatasourceType defines the type of Grafana datasource.
+   // +kubebuilder:validation:Enum="Prometheus"
+   type GrafanaDatasourceType string
+
+   const (
+       // GrafanaDatasourceTypePrometheus is the Prometheus datasource type.
+       GrafanaDatasourceTypePrometheus GrafanaDatasourceType = "Prometheus"
+   )
+   ```
+
+   `GrafanaPrometheusDatasourceSpec` will define Prometheus-specific options, including TLS and authentication.
+
+   ```go
+   type GrafanaPrometheusDatasourceOptions struct {
+      // tls holds TLS configuration for connecting to Prometheus over HTTPS.
+      // +optional
+      TLS *GrafanaDatasourceTLSSpec `json:"tls,omitempty"`
+
+      // auth holds authentication options for connecting to Prometheus.
+      // +optional
+      Auth *GrafanaPrometheusDatasourceAuthSpec `json:"auth,omitempty"`
+   }
+   ```
+
+   `GrafanaDatasourceTLSSpec` will define TLS configuration options for server certificate verification and client
+   authentication.
+
+   ```go
+   // LocalObjectKeySelector selects a key of a ConfigMap or Secret in the same namespace.
+   type LocalObjectKeySelector struct {
+      // name of the selected object.
+      // +kubebuilder:validation:MinLength=1
+      Name string `json:"name"`
+
+      // key within the selected object.
+      // +kubebuilder:validation:MinLength=1
+      Key string `json:"key"`
+   }
+
+   // LocalObjectReference contains a reference to an object in the same namespace.
+   // It can be used to reference a Secret, ConfigMap, or any other namespaced resource.
+   type LocalObjectReference struct {
+      // Name of the referent.
+      Name string `json:"name"`
+   }
+
+   type GrafanaDatasourceTLSSpec struct {
+      // caCert is a reference to a key within the CA bundle ConfigMap. The key should hold the CA cert in PEM format.
+      // When not specified, system CAs are used.
+      // +optional
+      CACertConfigMapRef *LocalObjectKeySelector `json:"caCertConfigMapRef,omitempty"`
+
+      // insecureSkipVerify controls whether to skip server certificate verification.
+      // +kubebuilder:default:=false
+      // +optional
+      InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+
+	  // clientTLSKeyPairSecretRef is a reference to a Secret holding client TLS certificate and key for mTLS authentication.
+	  // It's expected to be a standard Kubernetes TLS Secret with `tls.crt` and `tls.key` keys.
+      // +optional
+      ClientTLSKeyPairSecretRef *LocalObjectReference `json:"clientTLSKeyPairSecretRef,omitempty"`
+   }
+   ```
+
+   `GrafanaPrometheusDatasourceAuthSpec` will define authentication options, starting with bearer token auth. It will
+   be a separate struct to allow adding more auth methods in the future if needed (e.g., basic auth).
+
+   ```go
+   type GrafanaPrometheusDatasourceAuthType string
+   
+   const (
+      // GrafanaPrometheusDatasourceAuthTypeNoAuthentication means no authentication.
+      GrafanaPrometheusDatasourceAuthTypeNoAuthentication GrafanaPrometheusDatasourceAuthType = "NoAuthentication"
+
+      // GrafanaPrometheusDatasourceAuthTypeBearerToken means Bearer token authentication.
+      GrafanaPrometheusDatasourceAuthTypeBearerToken GrafanaPrometheusDatasourceAuthType = "BearerToken"
+    )
+   
+   type GrafanaPrometheusDatasourceAuthSpec struct {
+	  // type is the type of authentication to use.
+      // +kubebuilder:default:="NoAuthentication"
+      // +optional
+      Type GrafanaPrometheusDatasourceAuthType `json:"type,omitempty"`
+   
+      // bearerToken holds options for Bearer token authentication.
+      // +optional
+      BearerTokenOptions *GrafanaPrometheusDatasourceBearerTokenAuthOptions `json:"bearerTokenOptions,omitempty"`
+   }
+
+   type GrafanaPrometheusDatasourceBearerTokenAuthOptions struct {
+      // secretRef is a reference to a key in a Secret holding a Bearer token to use to authenticate with Prometheus.
+      // +optional
+      SecretRef *LocalObjectKeySelector `json:"secretRef,omitempty"`
+   }
+   ```
+
+On every change to the `datasources` field, the Grafana deployment will be rerolled to apply the new
+configuration. It will be achieved by leveraging the existing `scylla-operator.scylladb.com/inputs-hash` annotation on 
+the Grafana deployment's Pod template. Among other already used objects and fields, hash will be extended to take into account
+content of the referenced `Secret`/`ConfigMap`s as well.
+
+`Secret` and `ConfigMap` references in the data source configuration will be tracked, and changes to them will also
+trigger a `ScyllaDBMonitoring` reconciliation. `ScyllaDBMonitoring` controller will watch `Secret` and `ConfigMap` resources, and
+will trigger a reconciliation of `ScyllaDBMonitoring` resources that reference them in their Grafana data source
+configuration. Whether a `Secret` or `ConfigMap` is referenced will be determined by querying an index created by the
+controller. The index will be created by using `ScyllaDBMonitoring`'s informer's `cache.SharedIndexInformer.AddIndexers`
+method. Queries to the index will be performed using `cache.Indexer.ByIndex` method.
+
+### Example (verbose)
+
+```yaml
+apiVersion: scylla.scylladb.com/v1alpha1
+kind: ScyllaDBMonitoring
+metadata:
+  name: example
+spec:
+  type: Platform
+  endpointsSelector:
+    matchLabels:
+      app.kubernetes.io/name: scylla
+      scylla-operator.scylladb.com/scylla-service-type: member
+      scylla/cluster: replace-with-your-scyllacluster-name
+  components:
+    prometheus:
+      mode: External
+    grafana:
+      datasources:
+        - name: prometheus
+          url: "https://external-prometheus-address:9090"
+          type: Prometheus
+          prometheusOptions:
+            auth:
+              type: BearerToken
+              bearerTokenOptions:
+                secretRef:
+                  name: prometheus-bearer-token
+                  key: token
+            tls:
+              caCertConfigMapRef:
+                name: prometheus-serving-ca
+                key: ca.crt
+              insecureSkipVerify: false
+              clientTLSKeyPairSecretRef:
+                name: prometheus-client-certs
+```
+
+### Handling error conditions
+
+If the `ScyllaDBMonitoring` resource is misconfigured (e.g., missing required fields, invalid references to
+`Secret`/`ConfigMap`s, etc.), the Operator will set proper conditions on the `ScyllaDBMonitoring` resource to inform the user
+about the issue.
+
+When the configured external Prometheus instance is not reachable or returns errors, the managed Grafana instance will
+show errors when trying to use the Prometheus data source. The Operator will not try to validate the connectivity to the
+external Prometheus instance, as it would introduce unnecessary complexity and potential issues (e.g., network policies,
+firewalls, etc. preventing the Operator from reaching the Prometheus instance).
+
+### Validation
+
+#### Admission Webhook
+
+The Operator will validate in its admission webhook the `ScyllaDBMonitoring` objects to comply with the following rules:
+
+- If `spec.components.prometheus.mode` is set to `External`, `spec.components.grafana.datasources` must be set
+- If `spec.components.grafana.datasources` is set, it must contain exactly one datasource of type `Prometheus`.
+- Immutability of `spec.components.prometheus.mode` field.
+
+To achieve this:
+
+- The webhook server will be extended with a new validator for `ScyllaDBMonitoring` resources.
+- `ValidatingWebhookConfiguration` definition will be extended to track `ScyllaDBMonitoring` creations and updates.
+- `ClusterRole` used by the webhook server will be extended to allow `get` and `list` operations on
+  `ScyllaDBMonitoring` resources.
+
+### Test Plan
+
+- Validation rules will be covered in unit tests.
+- Functions generating `ScyllaDBMonitoring` child resources will be covered in unit tests.
+- A regular E2E test will be added to verify that it's possible to use `ScyllaDBMonitoring` with `External`
+  Prometheus mode and configure the managed Grafana instance to use an external Prometheus instance as a data source
+  (with mTLS).
+- An OpenShift E2E test will be added that will verify that the `ScyllaDBMonitoring` CRD can be used to integrate
+  with a Prometheus instance managed by OpenShift's user-workload monitoring stack (with bearer token).
+- A E2E test will be added to verify that create and update operations to `ScyllaDBMonitoring` resources trigger
+  validating
+  webhook and that invalid resources are rejected.
+
+### Upgrade / Downgrade Strategy
+
+No specific upgrade or downgrade strategy is required.
+
+### Version Skew Strategy
+
+To create `ServiceMonitor` and `PrometheusRule` resources, Operator uses
+`github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring` Go module for their schema. In an OpenShift
+cluster, mentioned CRDs are already installed (along with the Prometheus Operator) therefore we can only support
+OpenShift clusters with the Prometheus Operator supporting the version of the `ServiceMonitor` and `PrometheusRule`
+resources we use in the Operator. Their API is stable enough to not expect any breaking changes in the foreseeable
+future, but it's worth noting that in our documentation (i.e., in
+the [Support matrix](https://operator.docs.scylladb.com/stable/support/releases.html#support-matrix)).
+
+### Documentation plan
+
+- Add a guide that will walk users through the process of configuring the `ScyllaDBMonitoring` CRD. It should contain
+  the "standard" example with Scylla Operator-managed Prometheus, and a detailed example with an externally managed
+  Prometheus instance.
+- Add a guide that will explain how to configure `ScyllaDBMonitoring` to work with OpenShift's user-workload
+  monitoring stack (how to bind the right `ClusterRole` to the Grafana service account, how to get the bearer token, etc.).
+- Update the `ScyllaDBMonitoring` CRD documentation to reflect the new fields and their usage.
+- (Option) Write a blog post that will show how to use the externally managed Prometheus mode in a real-world
+  scenario on OpenShift. Rationale for making it a blog post is that it will not fit well in our official documentation
+  and could get outdated quickly due to OpenShift changes, but it could be useful for users.
+
+## Implementation History
+
+- 2025-08-26: First review iteration.
+- 2025-08-21: Enhancement proposal introduced.
+
+## Alternatives
+
+### Inline `apiextensionsv1.JSON` for datasource configuration
+
+_Note: This was the initially proposed approach. It was rejected due to being too low level in terms of API. More
+user-friendly
+approach was desired._
+
+Grafana data source provisioning configuration will be provided inline in the CR spec. It will be expected to contain a
+YAML configuration compliant with
+the [Grafana provisioning API](https://grafana.com/docs/grafana/latest/administration/provisioning/#datasources).
+
+To support TLS/bearer auth to external Prometheus, users can mount Secrets/ConfigMaps into Grafana and reference them
+from provisioning data source config via `$__file{}` directive.
+
+`ScyllaDBMonitoring` CRD will be extended with the following changes:
+
+1. Add a new `spec.components.prometheus.mode` field to allow users to choose the mode of the Prometheus component.
+   The field will be an enum, defaulting to `Managed` (i.e., Operator-managed Prometheus is enabled by default).
+   When set to `External`, the Operator will not deploy a managed Prometheus instance. Only the scraping and alerting
+   resources will be created (`ServiceMonitor`s and `PrometheusRule`s) to allow the external Prometheus instance to
+   scrape metrics and alerts from the ScyllaDB cluster.
+
+    ```go
+    // PrometheusMode describes the mode of the Prometheus instance.
+    // +kubebuilder:validation:Enum="Managed";"External"
+    type PrometheusMode string
+
+    const (
+        // PrometheusModeManaged defines a Prometheus instance that is managed by the operator.
+        PrometheusModeManaged PrometheusMode = "Managed"
+        // PrometheusModeExternal defines a Prometheus instance that is managed externally by a Prometheus Operator.
+        PrometheusModeExternal PrometheusMode = "External"
+    )
+
+    type PrometheusSpec struct {
+        // mode defines the mode of the Prometheus instance. It's an immutable field.
+        // +kubebuilder:default:="Managed"
+        // +optional
+        Mode PrometheusMode `json:"mode,omitempty"`
+        ...
+    }
+    ```
+
+   Externally managed Prometheus instances can be configured to reconcile `ServiceMonitor` and `PrometheusRule`
+   resources by setting `ruleSelector` and `serviceMonitorSelector` fields in
+   the [Prometheus](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.Prometheus) spec.
+   We will document that these fields have to match the
+   `scylla-operator.scylladb.com/scylladbmonitoring-name: "{{ .scyllaDBMonitoringName }}"`
+   label in order to work, as that's the label the Operator adds to the `ServiceMonitor` and `PrometheusRule` resources
+   it creates.
+
+   The added `mode` field will be immutable to prevent users from switching between managed and external Prometheus
+   modes. Implementation of switching modes would require additional work to ensure the managed resources are
+   properly cleaned up, and it's not worth the effort at this point.
+
+
+2. Add new fields to `spec.components.grafana` to allow users to configure the managed Grafana instance to use an
+   external Prometheus instance as a data source.
+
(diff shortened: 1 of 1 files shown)


enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+        PrometheusDatasource *apiextensiosnv1.JSON `json:"prometheusDatasource,omitempty"`
+
+        // volumes is a list of volumes to be added to the Grafana pod.
+        // This is useful for providing additional configuration files or data to Grafana.
+        Volumes []corev1.Volume `json:"volumes,omitempty"`
+
+        // volumeMounts is a list of volume mounts to be added to the Grafana container.
+        // This is useful for mounting additional configuration files or data into Grafana.
+        VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
+    ```
```
zimnx: I’m not a big fan of keeping a raw JSON blob as a field in our API. This approach introduces a few drawbacks:

* The API is no longer fully defined, which means users cannot easily see which fields inside the JSON are required or optional when reading our *API Reference* or using `kubectl explain`.
* The fields inside the blob are not part of the OpenAPI schema, so validation of the format has to be moved to the webhook. This requires us to manually check required keys, reject additional properties, etc.
* Because we don’t know what’s inside the JSON, we end up exposing additional fields like `Volume` and `VolumeSource` just to give users a way to specify extra volumes. This shifts too much of the configuration burden to the user.

I believe our API should explicitly define the fields we expect datasources to have.
That way, instead of requiring users to manage low-level details, we can define a clear and extensible API. For example:

```yaml
kind: ScyllaDBMonitoring
metadata:
  name: example
spec:
  type: Platform
  endpointsSelector:
    matchLabels:
      scylla-operator.scylladb.com/scylla-service-type: member
      scylla/cluster: replace-with-your-scyllacluster-name
  components:
    prometheus:
      mode: External
    grafana:
      datasources:
      - type: Prometheus
        auth:
          type: TLS
          tls:
            caBundleRef: ca-bundle-configmap-name
            certificateRef: client-cert-secret-name
        name: my-external-prometheus
        address: https://my-external-prometheus.foo.svc:9090
        isDefault: true
```

Grafana specific fields such as `isDefault`, `scrapeInterval` should be either clearly defined or omitted (like `readOnly`).
For those super datasource specific - like `incrementalQueryOverlapWindow` or `exemplarTraceIdDestinations`, I think we should delay adding them until we have a clear use case, while still allow for specifying them but only via a raw JSON blob coming from annotation.

With this structure, the user only needs to provide the address and secrets/configmaps for authentication, while the operator takes care of things like volumes and volume mounts.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+complexity.
+
+By providing knobs to skip deploying Prometheus and to configure Grafana with an external Prometheus as a data source,
+we give users the flexibility to integrate with their existing monitoring stack.
+
+### Goals
+
+- Allow users to disable the managed Prometheus component.
+- Allow users to configure the managed Grafana instance to use an external Prometheus instance as a data source.
+- Keep creating scraping configuration for an external Prometheus instance via `ServiceMonitor` and `PrometheusRule`
```
rzetelskik: not to be pedantic, but this seems like a design decision rather than a goal of this change?

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+    ```go
+    // PrometheusMode describes the mode of the Prometheus instance.
+    // +kubebuilder:validation:Enum="Managed";"External"
+    type PrometheusMode string
+
+    const (
+        // PrometheusModeManaged defines a Prometheus instance that is managed by the operator.
+        PrometheusModeManaged PrometheusMode = "Managed"
+        // PrometheusModeExternal defines a Prometheus instance that is managed externally by a Prometheus Operator.
+        PrometheusModeExternal PrometheusMode = "External"
```
rzetelskik: I'd suggest `OperatorManaged` and `UserManaged`/`ExternallyManaged` to follow the naming already in use for TLS certs

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+   `prometheusDatasource` field will be added to the `GrafanaSpec`. It will use an `apiextensionsv1.JSON` type, allowing
+   users to
+   specify the Grafana data source configuration inline. Thanks to allowing arbitrary fields, it's open to Grafana
+   schema changes. Operator will create a `ConfigMap` automatically and mount it in the Grafana pod under the expected
+   directory.
+
+   Because ScyllaDB monitoring stack dashboards are expecting the Prometheus data source to be named `prometheus`, we
+   will set the datasource name to `prometheus` by default and will ignore the name provided by the user in the
+   `prometheusDatasource` field. We will also set the type to `prometheus` by default, so users don't have to specify
+   it explicitly.
```
rzetelskik, line 472: This might be a bit far-fetched, but adding a `prometheusDatasource` field doesn't map too well to grafana's provisioning API and makes potential extensions very awkward

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+OpenShift clusters with the Prometheus Operator supporting the version of the `ServiceMonitor` and `PrometheusRule`
+resources we use in the Operator. Their API is stable enough to not expect any breaking changes in the foreseeable
+future, but it's worth noting that in our documentation (i.e., in
+the [Support matrix](https://operator.docs.scylladb.com/stable/support/releases.html#support-matrix)).
+
+### Documentation plan
+
+- Add a guide that will walk users through the process of configuring the `ScyllaDBMonitoring` CRD to disable the
+  managed Prometheus instance and configure the managed Grafana instance to use an external Prometheus instance as a
+  data source on the OpenShift platform.
```
rzetelskik: I'd mention how this plays with the existing monitoring documentation which doesn't follow the guide/tutorial/reference split atm.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+- Update the `ScyllaDBMonitoring` CRD documentation to reflect the new fields and their usage.
+
+## Implementation History
+
+- 2025-08-21: Enhancement proposal introduced.
+
+## Alternatives
+
+### Using `ConfigMap` reference for data source API
```
rzetelskik, line 553: What are the benefits of the proposed approach as compared to this one? I'd personally prefer this to having an unstructured field in the API. It seems to map to grafana's config better as well.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,344 @@
...
+        tlsClientKey: "$__file{/var/run/secrets/prometheus-client-certs/tls.key}"
+```
+
+### Richer Grafana data source API
+
+This option involves extend the `ScyllaDBMonitoring` CRD with a richer strongly-typed API for configuring Grafana data
+sources (similar to the `GrafanaDatasource` CRD from
+the [Grafana Operator](https://grafana.github.io/grafana-operator/docs/datasources/)). I decided to not pursue this
+option because it would require a significant amount of work to implement and maintain the API we have no control over (
+schema is defined by Grafana).
```
rzetelskik: is that actually that big of an issue that we don't have control over it if we provide our own, limited abstraction? It mostly depends on how much flexibility we want to give to the users. Given the assumptions made in this doc and the user stories listed, I'd say there might be more arguments for a strongly-typed API

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I was wondering what the real potential is for any extensions in the future, as the ScyllaDB monitoring stack supports only Prometheus datasources (that's the datasource hardcoded in dashboards). Anyway, I'm ok with making this API ready for changes - let's make it a slice, and for now expect exactly 1 or none datasources.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Leaving a `ConfigMap` to be created by a user can be seen as a drawback - that's an additional resource to create/manage. With the proposed inline spec, everything is defined in one object.

czeslavo:
@zimnx @rzetelskik PTAL. I've changed the API to be strongly-typed and not reliant on verbose low-level volume/volume mounts definitions.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: I do not think that `OperatorManaged` is a good choice (even if consistent). It is introducing confusion because in either case the Prom instance is "operator managed" (by prometheus operator). With certs we don't have this problem, because the operator term isn't that much overloaded in that case.

rzetelskik: @mflendrich that's a good point, I agree

mflendrich: Also, "UserManaged" makes little sense in the case where there's a central instance not managed by the user (but e.g. existing as part of the platform).

Another naming idea: name the field `PrometheusInstallationType` and name the enum values `Owned` and `External`.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,602 @@
...
+      // +optional
+      Client *GrafanaDatasourceClientTLSSpec `json:"client,omitempty"`
+   }
+
+   type GrafanaDatasourceServerTLSSpec struct {
+      // caCertConfigMapName is a name of a CA bundle ConfigMap that should be used to validate server certificates.
+      // It's expected to have a single key `ca.crt` holding the CA cert in PEM format.
+      // When not specified, system CAs are used.
+      // +optional
+      CACertConfigMapName string `json:"caCertConfigMapName,omitempty"`
```
czeslavo: We could make this more flexible by allowing specifying a key that should be used to extract a value from a `ConfigMap`.

```go
type ConfigMapValueReference struct {
  Name string `json:"name"`
  Key string `json:"key"`
}
```
The same applies to Secret references. I'm leaning towards changing this, happy to hear reviewers' opinions, though.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,602 @@
...
+cluster, mentioned CRDs are already installed (along with the Prometheus Operator) therefore we can only support
+OpenShift clusters with the Prometheus Operator supporting the version of the `ServiceMonitor` and `PrometheusRule`
+resources we use in the Operator. Their API is stable enough to not expect any breaking changes in the foreseeable
+future, but it's worth noting that in our documentation (i.e., in
+the [Support matrix](https://operator.docs.scylladb.com/stable/support/releases.html#support-matrix)).
+
+### Documentation plan
+
+- Add a guide that will walk users through the process of configuring the `ScyllaDBMonitoring` CRD to disable the
+  managed Prometheus instance and configure the managed Grafana instance to use an external Prometheus instance as a
```
mflendrich: > the process of configuring the `ScyllaDBMonitoring` CRD to disable the managed Prometheus instance

This wording is doing some heavy-lifting. In order to _disable_ the managed Prom instance it would be building on top of something that has it _enabled_ in the first place. Even if that's how we see it (that we're introducing a knob to disable parts of our functionality), I don't think that such a mental model will work for the user.

I think that we need to think the other way around: reach a state in which we have a "set up monitoring" guide that has a fork on the road: "spin up Prometheus" and "connect to an external Prometheus".

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I wouldn't focus that much on an enum naming - it can be self-descriptive, but we can also get rid of ambiguity in its godoc. To address your points, maybe let's go with something more verbose in naming, and focus on unambiguous comments as well:

```go
    // PrometheusMode describes the mode of the Prometheus instance.
    // +kubebuilder:validation:Enum="ScyllaOperatorManaged";"ExternallyManaged"
    type PrometheusMode string

    const (
        // ScyllaOperatorManaged defines a Prometheus that is managed by the Scylla Operator. In this mode Scylla Operator 
		// will:
        //  - create a Prometheus resource to be reconciled by Prometheus Operator installed in the cluster.
        //  - create ServiceMonitor and PrometheusRule resources to configure scraping and alerting.
        PrometheusModeScyllaOperatorManaged PrometheusMode = "ScyllaOperatorManaged"
   
        // ExternallyManaged defines a Prometheus instance that is created and managed externally by Prometheus Operator.
        // Scylla Operator will not create a Prometheus resource in this mode, but will still create ServiceMonitor and
        // PrometheusRule resources to configure scraping and alerting.
        PrometheusModeExternallyManaged PrometheusMode = "ExternallyManaged"
    )
```

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Yeah, I like the idea of a fork for bringing your own `ExeternallyManaged` Prometheus in a single guide explaining the `ScyllaDBMonitoring` CRD. We can discuss the exact structure under the actual documentation PR. I've changed the wording accordingly.

I also added an optional blog post idea that would describe using the externally managed Prometheus on OpenShift, to avoid the risk of our docs becoming outdated on OpenShift changes.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: The last version scratches my itch. :+1:

mflendrich: Actually, wait, it should be `ScyllaDBOperatorManaged`, not `ScyllaOperatorManaged` :sweat_smile:

rzetelskik: @mflendrich I'm not disagreeing, but have we actually ever made the switch? Product name in our docs is still ScyllaOperator (and so is e.g. ScyllaOperatorConfig CRD)

czeslavo:
Pushed changes with a more flexible way of referencing secret/configmaps keys (instead of assuming keys, let the user specify them explicitly).

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+   type GrafanaDatasourceClientTLSSpec struct {
+      // certRef is a reference to a key within a Secret holding the client certificate in PEM format.
+      // When it's defined, keyRef must also be defined.
+      // +optional
+      CertSecretRef *LocalObjectKeySelector `json:"certSecretRef,omitempty"`
+
+      // keyRef is a reference to a key within a Secret holding the client private key in PEM format.
+      // When it's defined, certRef must also be defined.
+      // +optional
+      KeySecretRef *LocalObjectKeySelector `json:"keySecretRef,omitempty"`
```
zimnx: What's the point of splitting those into two separate fields? There're standardized TLS Secrets having both client and key with standardized names. I see little value in supporting opaque secrets for client certificates.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+   }
+
+   type GrafanaDatasourceTLSSpec struct {
+      // server holds options for TLS server authentication.
+      // +optional
+      Server *GrafanaDatasourceServerTLSSpec `json:"server,omitempty"`
+
+      // client holds options for TLS client authentication.
+      // +optional
+      Client *GrafanaDatasourceClientTLSSpec `json:"client,omitempty"`
```
zimnx: I would recommend to inline these. This TLS configuration is obviously client-side TLS, having "server" in this context is ambiguous to me. I understand the motivation, but CA bundle and insecure flag is still something clients are setting, not servers.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+      // keyRef is a reference to a key within a Secret holding the client private key in PEM format.
+      // When it's defined, certRef must also be defined.
+      // +optional
+      KeySecretRef *LocalObjectKeySelector `json:"keySecretRef,omitempty"`
+   }
+   ```
+
+On every change to the `datasources` field, the Grafana deployment will be rerolled to apply the new
+configuration. Also, Secret and ConfigMap references in the data source configuration will be tracked, and changes to
+them will also trigger a Grafana deployment rollout.
```
zimnx: How it's achieved?

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+
+   ```go
+   type GrafanaPrometheusDatasourceSpec struct {
+      // tls holds TLS configuration for connecting to Prometheus over HTTPS.
+      // +optional
+      TLS *GrafanaDatasourceTLSSpec `json:"tls,omitempty"`
+
+      // auth holds authentication options for connecting to Prometheus.
+      // +optional
+      Auth *GrafanaPrometheusDatasourceAuthSpec `json:"auth,omitempty"`
```
zimnx, line 205: I can't find it's definition

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+          url: "https://external-prometheus-address:9090"
+          prometheus:
+            auth:
+              bearerTokenSecretName: prometheus-bearer-token
+            tls:
+              server:
+                caCertConfigMapName: prometheus-serving-ca
+                insecureSkipVerify: false
+              client:
+                secretName: prometheus-client-certs
```
zimnx: it's out of sync with API format proposal proposes

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+    // PrometheusMode describes the mode of the Prometheus instance.
+    // +kubebuilder:validation:Enum="OperatorManaged";"UserManaged"
+    type PrometheusMode string
+
+    const (
+        // ScyllaDBOperatorManaged defines a Prometheus that is managed by the ScyllaDB Operator. In this mode ScyllaDB Operator 
+		// will:
+        //  - create a Prometheus resource to be reconciled by Prometheus Operator installed in the cluster.
+        //  - create ServiceMonitor and PrometheusRule resources to configure scraping and alerting.
+        PrometheusModeScyllaDBOperatorManaged PrometheusMode = "ScyllaDBOperatorManaged"
```
zimnx: We never migrated to new name anywhere. I think we should think it through at global scale, prepare migration plan and execute it, not incrementally move towards it without any plan.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+                insecureSkipVerify: false
+              client:
+                secretName: prometheus-client-certs
+```
+
+### Validation
+
+#### Admission Webhook
+
+The Operator will validate in its admission webhook the `ScyllaDBMonitoring` objects to comply with the following rules:
```
zimnx, line 346: Currently we are not handling ScyllaDBMonitoring in webhook, if we want to add it, proposal should mention what RBAC and what additional resources will be added, as well mention how it's going to be tested. (e2e checking if ScyllaDBMonitoring is validated by webhook is not mentioned in Test Plan)

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+
+### Documentation plan
+
+- Add a guide that will walk users through the process of configuring the `ScyllaDBMonitoring` CRD. It should contain
+  the "standard" example with Scylla Operator-managed Prometheus, and a detailed example with an externally managed
+  Prometheus instance.
+- Update the `ScyllaDBMonitoring` CRD documentation to reflect the new fields and their usage.
+- (Option) Write a blog post that will show how to use the externally managed Prometheus mode in a real-world
+  scenario on OpenShift. Rationale for making it a blog post is that it will not fit well in our official documentation
+  and could get outdated quickly due to OpenShift changes, but it could be useful for users.
```
zimnx, line 396: Having a blog post wouldn't hurt. Even though they eventually get outdated, they show up in search engines and users are following them.
I think we should have a dedicated example for openshift in documentation page as well. It can link to External mode documentation and should cover from where to get URL, certs etc which are specific to Openshift and probably won't get outdated.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: As we discussed offline, the motivation was to keep consistency with the other secret/configmap references in this CRD. As we agreed, we'll rely on the Kubernetes standard TLS secret type, sacrificing consistency a bit.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I assumed it's obvious that both are set on the client side, but for different purposes (server verification vs. client auth against server). If you find it ambiguous, I'll make the structure flat then.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,633 @@
...
+
+1. Add a new `spec.components.prometheus.mode` field to allow users to choose the mode of the Prometheus component.
+   The field will be an enum, defaulting to `OperatorManaged` (i.e., Operator-managed Prometheus is enabled by default).
+   When set to `UserManaged`, the Operator will not deploy a managed Prometheus instance. Only the scraping and alerting
+   resources will be created (`ServiceMonitor`s and `PrometheusRule`s) to allow the external Prometheus instance to
+   scrape metrics and alerts from the ScyllaDB cluster.
+
+    ```go
+    // PrometheusMode describes the mode of the Prometheus instance.
+    // +kubebuilder:validation:Enum="OperatorManaged";"UserManaged"
```
rzetelskik: Do we need the kubebuilder validation? We'll do this in the admission webhook either way.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I added a more elaborate explanation touching on implementation details.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Added missing definition.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Aligned.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Rolled back to `ScyllaOperatorManaged`.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Addressed that.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,606 @@
...
+       // +optional
+       Type GrafanaDatasourceType `json:"type,omitempty"`
+
+       // url is the URL of the datasource.
+       // +kubebuilder:validation:MinLength=1
+       URL string `json:"url"`
+
+       // prometheus defines Prometheus-specific options.
+       // +optional
+       Prometheus *GrafanaPrometheusDatasourceSpec `json:"prometheus,omitempty"`
```
rzetelskik:
```suggestion
      PrometheusOptions *GrafanaPrometheusDatasourceSpec `json:"prometheusOptions,omitempty"`
```
to follow the naming convention in other APIs, see e.g. https://github.com/scylladb/scylla-operator/blob/51b9ea58176a689a428cbb57a4ad869a3e931c80/pkg/api/scylla/v1alpha1/types_scylladbdatacenter.go#L246-L257.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I wanted to ensure we're not overwhelmed with keeping documentation specific to a platform up to date, thus proposing a blog post form that doesn't require updating.

We can try to prepare an OpenShift-specific set of instructions page and make it as abstract as possible, to make sure it doesn't get outdated in the next release. I added a point about it.

czeslavo:
@zimnx I hope I've addressed all of your comments. PTAL

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: Merging https://github.com/scylladb/scylla-operator/pull/2909#discussion_r2303409939 into this thread.

The naming of this enum has become very surprisingly contentious. I do not disagree with any opinion voiced in this thread, but we've had very strong and opinionated voices in favor of using `Scylla**DB**` (for branding reasons) vs keeping `Scylla` (for consistency / migration plan reasons).

I feel that a productive way forward (that addresses voices that we should think a migration through, which I agree with to some extent, but let's choose the right-sized sledgehammer to crack this little nut) is to sidestep the problem altogether by adding neither `Scylla` nor `ScyllaDB` here (and avoid branding in our APIs going forward).

Suggestion: rename this enum value to `Owned`. Or cut our losses and pick any other naming that doesn't paint us into a corner branding-wise. And solve the branding conundrum in a separate, targeted conversation.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: Rolling this thread into https://github.com/scylladb/scylla-operator/pull/2909#discussion_r2303780470.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: The problem with `Owned` for me is that it's similarly ambiguous as `Managed`. Owned (or managed) by whom? It's hard to mention a specific product/component without explicitly calling its name. :sweat_smile:

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,608 @@
...
+   ```
+
+   `GrafanaPrometheusDatasourceAuthSpec` will define authentication options, starting with bearer token auth. It will
+   be a separate struct to allow adding more auth methods in the future if needed (e.g., basic auth).
+
+   ```go
+   type GrafanaPrometheusDatasourceAuthSpec struct {
+      // bearerToken holds options for Bearer token authentication.
+      // +optional
+      BearerToken *GrafanaPrometheusDatasourceBearerTokenAuthSpec `json:"bearerToken,omitempty"`
```
rzetelskik: ditto about options

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+	  // clientTLSKeyPairSecretName is a name of a Secret holding client TLS certificate and key for mTLS authentication.
+	  // It's expected to be a standard Kubernetes TLS Secret with `tls.crt` and `tls.key` keys.
+      // +optional
+      ClientTLSKeyPairSecretName string `json:"clientTLSKeyPairSecretName,omitempty"`
+   }
+   ```
+
+   `GrafanaPrometheusDatasourceAuthSpec` will define authentication options, starting with bearer token auth. It will
+   be a separate struct to allow adding more auth methods in the future if needed (e.g., basic auth).
```
rzetelskik, line 250: doesn't this call for a `type` field? Or are these non-exclusive?

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Pushed `Managed`/`Headless` as we agreed offline.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Good call, added that.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I've got mixed feelings about this - I see your point about keeping this consistent in our APIs, but at the same time, it feels like an unnecessary addition (the sole `Options` word doesn't bring any value to the field name). From the API perspective, I feel that 

```yaml
      datasources:
        - name: prometheus
          type: Prometheus
          url: "https://external-prometheus-address:9090"
          prometheus:
            auth:
              bearerToken:
                secretRef:
                  name: prometheus-bearer-token
                  key: token
```

looks better than

```yaml
      datasources:
        - name: prometheus
          type: Prometheus
          url: "https://external-prometheus-address:9090"
          prometheusOptions:
            auth:
              bearerToken:
                secretRef:
                  name: prometheus-bearer-token
                  key: token
```

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: I believe that if we can validate something on the schema level, we should do it. Do you foresee any problems with such an approach?

rzetelskik: It just seems redundant, but other than that no, so treat it as a nit

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
rzetelskik: This is entirely subjective and we can keep having these discussions over and over again as long as it's up to the person proposing the API. The problem that it leads to, which I'd like to avoid, is multiple different ways of doing the same thing across resources (or even the same resource) in one API version.
As we discussed internally, the choice is yours here (and I'm fine with either, treat this as a separate discussion), but it becomes very clear that we need an api convention (or a style guide) for stuff that's not covered by [k8s api convention](https://github.com/kubernetes/community/blob/d3c6c06c43b882c8aca692dd92e1adedad1dfab8/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field). I think the `Type` + `<Name>`/`<Name>Options` pattern would be a good candidate.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,624 @@
...
+
+      // insecureSkipVerify controls whether to skip server certificate verification.
+      // +kubebuilder:default:=false
+      // +optional
+      InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+
+	  // clientTLSKeyPairSecretName is a name of a Secret holding client TLS certificate and key for mTLS authentication.
+	  // It's expected to be a standard Kubernetes TLS Secret with `tls.crt` and `tls.key` keys.
+      // +optional
+      ClientTLSKeyPairSecretName string `json:"clientTLSKeyPairSecretName,omitempty"`
```
rzetelskik: nit: doesn't the name go against https://github.com/kubernetes/community/blob/d3c6c06c43b882c8aca692dd92e1adedad1dfab8/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field?

czeslavo: Good point. Let's also go with `corev1.LocalObjectReference` for the reference type instead of a string.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Let's switch to the consistent approach and codify it for the future: https://github.com/scylladb/scylla-operator/pull/2919

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,611 @@
...
+        // PrometheusModeManaged defines a mode where a `Prometheus` object is created as a child of a `ScyllaDBMonitoring`
+        // object. `ServiceMonitor` and `PrometheusRule` resources are also created to configure scraping and alerting.
+		// This mode requires a Prometheus Operator to be installed in the cluster.
+        PrometheusModeManaged PrometheusMode = "Managed"
+   
+        // PrometheusModeHeadless defines a mode where no `Prometheus` child object is created, but `ServiceMonitor` and
+        // `PrometheusRule` objects are still created to configure scraping and alerting.
+        // This mode requires a Prometheus Operator to be installed in the cluster, along with a `Prometheus` instance
+        // configured to reconcile `ServiceMonitor` and `PrometheusRule` resources.
+        PrometheusModeHeadless PrometheusMode = "Headless"
```
ylebi: The term "Headless" is confusing in this context. `PrometheusModeExternal` will have better clarity for me. 
`PrometheusModeExternal PrometheusMode = "External"`

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
rzetelskik: thanks!

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
rzetelskik: late to the party :stuck_out_tongue_closed_eyes: https://github.com/scylladb/scylla-operator/pull/2909#discussion_r2297514660

rzetelskik:
lgtm, no more comments from my side
/unassign

ylebi (commented):
What happens if external Prometheus is unreachable? I don't see any error scenarios or user feedback in this proposal.

czeslavo:
> How existing Prometheus resources are handled when switching modes?

It's addressed in the doc:

> The added `mode` field will be immutable to prevent users from switching between managed and external Prometheus modes. Implementation of switching modes would require additional work to ensure the managed resources are properly cleaned up, and it's not worth the effort at this point.

Happy to hear if you think we should spend more time on it.

ylebi:
> > How existing Prometheus resources are handled when switching modes?
> 
> It's addressed in the doc:
> 
> > The added `mode` field will be immutable to prevent users from switching between managed and external Prometheus modes. Implementation of switching modes would require additional work to ensure the managed resources are properly cleaned up, and it's not worth the effort at this point.
> 
> Happy to hear if you think we should spend more time on it.

Thanks. Looks good.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: Sharing for context that we've already considered `ExternallyManaged` here https://github.com/scylladb/scylla-operator/pull/2909#discussion_r2297514660. (edit: @rzetelskik beat me to it)

@czeslavo, with all the decisions made so far, what do you think?

czeslavo:
> What happens if external Prometheus is unreachable? I don't see any error scenarios or user feedback in this proposal.

If it's unreachable, Grafana dashboards won't load properly, giving the user feedback in the Grafana UI.

In theory, we could try to implement a probing mechanism from the Operator and propagate results in `ScyllaDBMonitoring`'s status conditions, but it still wouldn't be 100% guaranteed that what we probe is what Grafana sees due to possible networking differences. I wouldn't go this way.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: Unresolving - let's actually consider this seriously. It's a cheap change to make that can be impactful if it reduces the cognitive load.

We focused a lot on naming the "managed" aka "created" aka "owned" one but kinda left the "external" aka "headless" one behind.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,611 @@
...
+   type GrafanaPrometheusDatasourceBearerTokenAuthOptions struct {
+      // secretRef is a reference to a key in a Secret holding a Bearer token to use to authenticate with Prometheus.
+      // +optional
+      SecretRef *LocalObjectKeySelector `json:"secretRef,omitempty"`
+   }
+   ```
+
+On every change to the `datasources` field, the Grafana deployment will be rerolled to apply the new
+configuration. It will be achieved by adding an annotation to the Grafana deployment's Pod template with a hash of the
+data source configuration and referenced `Secret`/`ConfigMap` names and keys.
```
zimnx: https://github.com/scylladb/scylla-operator/blob/6d673e2aaba4070a21460f8534cf69d632d67508/pkg/controller/scylladbmonitoring/sync_grafana.go#L338-L339

We already have this functionality. In case of change in Grafana datasource, provisioning CM will change which is part of inputs hash of grafana deployment.
https://github.com/scylladb/scylla-operator/blob/9d36ac3d636fd29124ad9ebd2086829a43f2f3a7/assets/monitoring/grafana/v1alpha1/deployment.yaml#L14

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
mflendrich: Anyone feel strongly that `Headless` explains it better than `External`? I personally like `External` better which makes it 2 voices (@ylebi and myself).

czeslavo: Actually, that was my original proposal, but as @rzetelskik pointed out, we've already collectively agreed to stick to `Managed` and `Headless` after some back-and-forth changes. We can open the Pandora box one more time and consider switching to the original proposal, which was `Managed` and `External`, but I would prefer not. Naming is hard. :)

rzetelskik: I don't have a preference between the two. For the record, my comment wasn't meant as a resolution to this thread :wink:

mflendrich: Fair enough.

ylebi: > I don't have a preference between the two. For the record, my comment wasn't meant as a resolution to this thread 😉

I resolved it so we won't have duplicates :) 
Anyway, I think that it's better to user `External` instead of `Headless`. It is more clear to me.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,611 @@
...
+
+   `GrafanaPrometheusDatasourceAuthSpec` will define authentication options, starting with bearer token auth. It will
+   be a separate struct to allow adding more auth methods in the future if needed (e.g., basic auth).
+
+   ```go
+   type GrafanaPrometheusDatasourceAuthType string
+   
+   const (
+      // GrafanaPrometheusDatasourceAuthTypeNone means no authentication.
+      GrafanaPrometheusDatasourceAuthTypeNone GrafanaPrometheusDatasourceAuthType = "None"
```
zimnx: How about we call it Unauthorized or NoAuthentication to better reflect what is being set?

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,611 @@
...
+
+      // insecureSkipVerify controls whether to skip server certificate verification.
+      // +kubebuilder:default:=false
+      // +optional
+      InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+
+	  // clientTLSKeyPairSecretRef is a name of a Secret holding client TLS certificate and key for mTLS authentication.
+	  // It's expected to be a standard Kubernetes TLS Secret with `tls.crt` and `tls.key` keys.
+      // +optional
+      ClientTLSKeyPairSecretName *corev1.LocalObjectReference `json:"clientTLSKeyPairSecretRef,omitempty"`
```
zimnx: https://github.com/scylladb/scylla-operator/blob/9f1d676c34804a6bd794111b107ef6eb57f0c774/vendor/k8s.io/api/core/v1/types.go#L7090-L7099

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Given we have 3 voices for `External` (Yehuda, Michal, me) + 1 with no strong preferences (Kacper), I'm changing `Headless` to `External`.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,611 @@
...
+   the [Prometheus](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.Prometheus) spec.
+   We will document that these fields have to match the
+   `scylla-operator.scylladb.com/scylladbmonitoring-name: "{{ .scyllaDBMonitoringName }}"`
+   label in order to work, as that's the label the Operator adds to the `ServiceMonitor` and `PrometheusRule` resources
+   it creates.
+
+   The added `mode` field will be immutable to prevent users from switching between managed and external Prometheus
+   modes.
+   Implementation of switching modes would require additional work to ensure the managed resources are
+   properly cleaned up, and it's not worth the effort at this point.
```
zimnx: Operator is responsible for cleaning resources and every our controller does that. It's simple because they all match particular label selector. 
I do think this field should be immutable, but definitely not for cleaning burden reason.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: :+1:  It will have to be extended to take into account the referenced `Secret`s and `ConfigMap`s as well, though. Do you expect changes to the proposal? I believe it's still valid with this wording.

zimnx: I expect no annotation to be added but existing one to be extended, while proposal mentions it's going to be added.

czeslavo: Reworded this paragraph to make that clear.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: `NoAuthentication` it is. :+1:

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Switched to a local type.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
czeslavo: Reworded.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,627 @@
...
+
+## Design Details
+
+### Proposed API changes
+
+`ScyllaDBMonitoring` CRD will be extended with the following changes:
+
+1. Add a new `spec.components.prometheus.mode` field to allow users to choose the mode of the Prometheus component.
+   The field will be an enum, defaulting to `OperatorManaged` (i.e., Operator-managed Prometheus is enabled by default).
+   When set to `UserManaged`, the Operator will not deploy a managed Prometheus instance. Only the scraping and alerting
```
zimnx: Enum values are not updated.

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,627 @@
...
+        // PrometheusModeExternal defines a mode where no `Prometheus` child object is created, but `ServiceMonitor` and
+        // `PrometheusRule` objects are still created to configure scraping and alerting.
+        // This mode requires a Prometheus Operator to be installed in the cluster, along with a `Prometheus` instance
+        // configured to reconcile `ServiceMonitor` and `PrometheusRule` resources.
+        PrometheusModeExternal PrometheusMode = "External"
+    )
+    
+    type PrometheusSpec struct {
+        // mode defines the mode of the Prometheus instance. It's an immutable field.
+        // +kubebuilder:default:="OperatorManaged"
```
zimnx: Wrong default value

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,627 @@
...
+
+   Because ScyllaDB monitoring stack dashboards are expecting the Prometheus data source to be named `prometheus`, we
+   will support only 1 data source in this field for now, but we will make it a list to allow adding more data sources
+   in the future if needed.
+
+   ```go
+   type GrafanaSpec struct {
+        ...
+       // datasources is a list of Grafana datasources to configure.
+       // It's expected to be set when using Prometheus component in `UserManaged` mode.
```
zimnx: wrong enum

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,627 @@
...
+external Prometheus instance, as it would introduce unnecessary complexity and potential issues (e.g., network policies,
+firewalls, etc. preventing the Operator from reaching the Prometheus instance).
+
+### Validation
+
+#### Admission Webhook
+
+The Operator will validate in its admission webhook the `ScyllaDBMonitoring` objects to comply with the following rules:
+
+- If `spec.components.prometheus.mode` is set to `UserManaged`, `spec.components.grafana.datasources` must be set
```
zimnx: wrong enum

enhancements/proposals/2490-scylladbmonitoring-external-prometheus/README.md
```diff
@@ -0,0 +1,627 @@
...
+- The webhook server will be extended with a new validator for `ScyllaDBMonitoring` resources.
+- `ValidatingWebhookConfiguration` definition will be extended to track `ScyllaDBMonitoring` creations and updates.
+- `ClusterRole` used by the webhook server will be extended to allow `get` and `list` operations on
+  `ScyllaDBMonitoring` resources.
+
+### Test Plan
+
+- Validation rules will be covered in unit tests.
+- Functions generating `ScyllaDBMonitoring` child resources will be covered in unit tests.
+- A regular E2E test will be added to verify that it's possible to use `ScyllaDBMonitoring` with `UserManaged`
```
zimnx: wrong enum

czeslavo:
@zimnx Fixed the enums renaming leftovers.

zimnx (approved):
/lgtm

PR #2910, opened 2025-08-21
Disable manager's ScyllaCluster integration with the manager itself

### LINKED ISSUE #2807: Opt out of ScyllaDB Manager cluster registration with ScyllaDB Manager by default
At this point, the ScyllaDB cluster backing ScyllaDB Manager is registered with manager itself, which doesn't make much sense to my knowledge.
Once the migration to in-operator manager controller is completed and the ScyllaCluster backing ScyllaDB Manager itself is backed by the new CRD, we should consider adding the annotation to the manifest by default to disable the integration altogether.

/kind cleanup
/priority important-longterm

czeslavo:
**Description of your changes:** Stop registering the manager's ScyllaCluster in the manager itself - this is not needed.

Resolves #2807.


czeslavo:
@rzetelskik PTAL, I made a mistake of not enclosing the annotation value in quotes and had to fix this.

PR #2919, opened 2025-08-28
Add API conventions document

czeslavo:
**Description of your changes:** Introduces the API conventions document to ensure consistency in our APIs and to avoid relying on implicit rules. If there's any discussion regarding some convention in the review process, let's document a decision here so we can follow it in the future.

Follow up to https://github.com/scylladb/scylla-operator/pull/2909/files#r2303751011.


rzetelskik (approved):
lgtm, thanks for starting this :)

pkg/api/api-conventions.md
```diff
@@ -0,0 +1,43 @@
...
+
+Use this convention whenever a field's configuration options depends on a "type" field chosen by a user, and
+the options are mutually exclusive.
+
+- The "type" field should be named `Type`, and be of a string type (usually a custom string type with constants).
+- The configuration options for a given type should be grouped in a struct with a name suffixed with `<Type>Options`.
+- The configuration options fields' names should be suffixed with `<Type>Options` where `<Type>` is the value of the type field.
+- The configuration options fields should be pointers and marked as `+optional` in the comments, to allow them to be omitted when not used.
+- The configuration option fields should be marked with `omitempty` in the JSON tags, to avoid cluttering the serialized output.
+- The validating server should validate that only the configuration options corresponding to the chosen type are provided, and return an error if not.
```
zimnx: I don't think this requirement is reasonable. It should be ignored, validating it is a mess when number of possible Type value grows.

```suggestion
- The validating server should validate that the configuration options corresponding to the chosen type are provided.
```

czeslavo:
Moved it to the root so we can easily skip tests if only markdown changes (https://github.com/scylladb/scylla-operator-release/pull/485).

CONTRIBUTING.md
```diff
@@ -34,7 +34,7 @@ Follow these guidelines for writing and structuring your code:
 * [Go Code Review Comments](https://go.dev/wiki/CodeReviewComments)
 * [API changes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md)
 * [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)
-
+* [ScyllaDB Operator API design conventions](./pkg/api/api-conventions.md)
```
rzetelskik: path needs to be updated

API_CONVENTIONS.md
```diff
@@ -0,0 +1,43 @@
+# API design conventions
```
rzetelskik, line 1: nit: is there a reason for the all caps name (API_CONVENTIONS.md)? If not, I'd switch it to lowercase

czeslavo: The main reason is to match the casing of markdown files in the root directory.

PR #2898, opened 2025-08-19
Reorganize K8s prerequisites pages

czeslavo:
**Description of your changes:** Merges `Installation/Kubernetes/*` pages into one `Kubernetes prerequisites` for improved readability. The separate Generic, GKE, and EKS pages weren't descriptive enough.

Prerequisite for https://github.com/scylladb/scylla-operator/pull/2897.

PR #2938, opened 2025-09-03
[v1.18] Handle missing xfsprogs on GKE >= 1.32

czeslavo:
Backport of https://github.com/scylladb/scylla-operator/pull/2897 and #2898 (skipping CI scripts changes as they're required only on master).


rzetelskik (approved):
/lgtm

PR #2939, opened 2025-09-03
Add sysctl configuration guide to documentation

rzetelskik:
**Description of your changes:** This PR adds a how-to-guide for sysctl configuration with NodeConfig to our documentation. It covers the single and multi-datacenter deployments. 

Resolves #

/kind documentation
/priority important-soon

### DIFF
--- a/docs/source/management/sysctls.md
+++ b/docs/source/management/sysctls.md
@@ -1,0 +1,52 @@
+# Configuring kernel parameters (sysctls)
+
+This guide explains how to configure Linux kernel parameters (sysctls) for nodes running production-grade ScyllaDB deployments.
+
+{{productName}} enables you to manage sysctl settings in a Kubernetes-native way using the NodeConfig resource.
+
+:::{note}
+To configure sysctls for multi-datacenter ScyllaDB deployments, simply perform the steps outlined in this guide in each worker cluster.
+:::
+
+## Configuring the recommended sysctls for ScyllaDB nodes
+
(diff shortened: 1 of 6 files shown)


docs/source/management/sysctls.md
```diff
@@ -0,0 +1,46 @@
...
+kubectl apply --server-side -f=https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/common/configure-sysctls.nodeconfig.yaml
+:::
+
+After applying the manifest, wait for the NodeConfig to be reconciled.
+:::{include} ./../.internal/wait-for-status-conditions.nodeconfig.code-block.md
+:::
+
+### Configuring sysctls via Helm
+
+Configuring sysctls via Helm is currently unsupported.
```
czeslavo: I'm not sure if we really need this split. I wouldn't say we're expected to provide a Helm chart integration for every possible CRD we support. :thinking:

czeslavo (approved):

docs/source/management/sysctls.md
mflendrich: I think it's a matter of framing:

"If you've installed ScyllaDB using Helm, you still need to create the NodeConfig by hand"

would be a more useful message IMO than "unsupported" which leaves the user out in the cold.

rzetelskik: addressed

czeslavo (approved):
/lgtm

Issue #2928, opened 2025-09-02
Implement ScyllaDBMonitoring external Prometheus mode


czeslavo:
This issue covers the implementation of the external Prometheus mode for `ScyllaDBMonitoring` according to the design accepted in https://github.com/scylladb/scylla-operator/pull/2909.

It includes:
- [x]  Handle External mode (create only required resources)
- [ ]  Handle Grafana datasources
    - Configure volumes/volumeMounts
    - Extend Grafana deployment template
    - Extend Grafana provisioning template
        - Bearer token auth
        - mTLS
- [ ]  Handle Grafana restart hash to include Secrets and ConfigMaps
- [ ]  Add admission webhook validation rules
- [ ]  Add ConfigMap/Secret indexers and handlers in `ScyllaDBMonitoring` controller
- [ ]  Add unit tests covering:
    - Generating resources required for Prometheus
    - Generating resources required for Grafana

PR #2922, opened 2025-08-29
Refactor ScyllaCluster's admission webhook E2E to test operations independently

rzetelskik:
**Description of your changes:** As requested in https://github.com/scylladb/scylla-operator/pull/2901#discussion_r2306579144, this PR refactors the E2E test for ScyllaCluster's admission webhook to test operations independently.

**Which issue is resolved by this Pull Request:**
Resolves #

/kind cleanup
/priority important-soon
/cc czeslavo

PR #2901, opened 2025-08-20
Extend node tuning with kernel parameter configuration (sysctls)

rzetelskik:
**Description of your changes:** Following the [enhancement proposal](https://github.com/scylladb/scylla-operator/tree/master/enhancements/proposals/2457-nodeconfig-sysctls), this PR extends the NodeConfig API with support for node-level kernel parameter configuration (sysctls). This bridges the feature gap between v1.ScyllaCluster and v1alpha1.ScyllaDBCluster APIs. `ScyllaCluster.scylla.scylladb.com/v1.spec.sysctls` is simultaneously deprecated.

**Which issue is resolved by this Pull Request:**
Resolves https://github.com/scylladb/scylla-operator/issues/2457

/kind feature
/priority important-soon

Issue #2722, opened 2025-06-26
Operator creates invalid names for managed resources


rzetelskik:
### What happened?

Creating a v1.ScyllaCluster object with `.metadata.name`, `.spec.datacenter.name` and `.spec.datacenter.racks[].name` such that |`.metadata.name`| + |`.spec.datacenter.name`| and |`.spec.datacenter.racks[].name`| > 63 results in an invalid deployment.

```yaml
  - lastTransitionTime: "2025-06-26T13:08:47Z"
    message: |-
      StatefulSetControllerDegraded: can't create StatefulSet(s): can't create missing statefulset: StatefulSet.apps "simple-pxkn8-remotekubernetescluster-europe-west1-europe-west1-a" is invalid: metadata.name: Invalid value: "simple-pxkn8-remotekubernetescluster-europe-west1-europe-west1-a": must be no more than 63 characters
      ServiceControllerDegraded: Service "simple-pxkn8-remotekubernetescluster-europe-west1-europe-west1-a-0" is invalid: [metadata.name: Invalid value: "simple-pxkn8-remotekubernetescluster-europe-west1-europe-west1-a-0": must be no more than 63 characters, spec.selector: Invalid value: "simple-pxkn8-remotekubernetescluster-europe-west1-europe-west1-a-0": must be no more than 63 characters]
    observedGeneration: 1
    reason: Error,Error
    status: "True"
    type: Degraded
```

### What did you expect to happen?

Operator shouldn't produce invalid names for managed resources. We could generate deterministic names with a hash, similarly to how we do for ScyllaDBManagerClusterRegistrations.

### How can we reproduce it (as minimally and precisely as possible)?

```sh
kubectl apply --server-side -f=- <<EOF
apiVersion: scylla.scylladb.com/v1
kind: ScyllaCluster
metadata:
  name: simple-pxkn8-remotekubernetescluster
spec:
  agentVersion: 3.5.0
  version: 2025.1.2
  developerMode: true
  automaticOrphanedNodeCleanup: true
  sysctls:
    - fs.aio-max-nr=30000000
  datacenter:
    name: europe-west1
    racks:
      - name: europe-west1-a
        members: 1
        storage:
          capacity: 1Gi
          storageClassName: scylladb-local-xfs
        resources:
          requests:
            cpu: 10m
            memory: 100Mi
          limits:
            cpu: 1
            memory: 1Gi

EOF
```

### Scylla Operator version

master

### Kubernetes platform name and version

[...]

PR #2943, opened 2025-09-05
[WIP] Add support for external Prometheus mode to ScyllaDBMonitoring

czeslavo:

Resolves https://github.com/scylladb/scylla-operator/issues/2928.


czeslavo:
It requires https://github.com/scylladb/scylla-operator/pull/2954 to fix the data race detected in unit tests.

/hold

czeslavo:
> It requires #2954 to fix the data race detected in unit tests.

Despite failing unit tests, making it ready for review to start iterating as I expect some review rounds anyway.

pkg/controller/scylladbmonitoring/sync_prometheus.go
```diff
@@ -584,3 +576,63 @@ func (smc *Controller) syncPrometheus(
...
+	Service                *corev1.Service
+	Ingress                *networkingv1.Ingress
+	Prometheus             *monitoringv1.Prometheus
+	LatencyPrometheusRule  *monitoringv1.PrometheusRule
+	AlertsPrometheusRule   *monitoringv1.PrometheusRule
+	TablePrometheusRule    *monitoringv1.PrometheusRule
+	ScyllaDBServiceMonitor *monitoringv1.ServiceMonitor
+}
+
+func prepareRequiredPrometheusResources(sm *scyllav1alpha1.ScyllaDBMonitoring, soc *scyllav1alpha1.ScyllaOperatorConfig) (requiredPrometheusResources, error) {
```
rzetelskik: nit: makeRequiredPrometheusResources
as this is an aggregate of `make<Resource>` functions, which grew to be somewhat of a pattern in our codebase

pkg/controller/scylladbmonitoring/sync_prometheus.go
```diff
@@ -584,3 +576,63 @@ func (smc *Controller) syncPrometheus(
...
+
+		resources.Service, _, err = makePrometheusService(sm)
+		renderErrors = append(renderErrors, err)
+
+		resources.Ingress, _, err = makePrometheusIngress(sm)
+		renderErrors = append(renderErrors, err)
+
+		resources.Prometheus, _, err = makePrometheus(sm, soc)
+		renderErrors = append(renderErrors, err)
+	}
```
rzetelskik: nit: using a switch with "enums" is much clearer to me, especially here that the external mode is an explicit no-op and unsupported types should result in an error

pkg/naming/names.go
```diff
@@ -363,3 +363,11 @@ func RemoteNamespaceName(sc *scyllav1alpha1.ScyllaDBCluster, dc *scyllav1alpha1.
...
 	return fmt.Sprintf("%s-%s", sc.Namespace, suffix), nil
 }
+
+func ManagedPrometheusClientGrafanaSecretName(sm *scyllav1alpha1.ScyllaDBMonitoring) string {
+	return sm.Name + "-prometheus-client-grafana"
+}
+
+func ManagedPrometheusServingCAConfigMapName(sm *scyllav1alpha1.ScyllaDBMonitoring) string {
+	return sm.Name + "-prometheus-serving-ca"
+}
```
rzetelskik: optional: perhaps we could make these names non-overflowing while we're at it? (see https://github.com/scylladb/scylla-operator/issues/2722)
This might require some changes to the documentation as well unfortunately

pkg/controller/scylladbmonitoring/grafana_datasource.go
```diff
@@ -0,0 +1,104 @@
...
+func makePrometheusDatasourceSpec(sm *scyllav1alpha1.ScyllaDBMonitoring) *scyllav1alpha1.GrafanaDatasourceSpec {
+	if spec := getGrafanaSpec(sm); spec != nil && len(spec.Datasources) > 0 {
+		// We only support one datasource for now.
+		ds := spec.Datasources[0]
+		if ds.Type == scyllav1alpha1.GrafanaDatasourceTypePrometheus {
+			return &ds
+		}
+	}
+	return nil
+}
```
rzetelskik: nit: shall we stick with the (mostly) existing file structure and move it to `resource.go`?

pkg/controller/scylladbmonitoring/grafana_datasource.go
```diff
@@ -0,0 +1,104 @@
...
+
+func makePrometheusDatasourceSpec(sm *scyllav1alpha1.ScyllaDBMonitoring) *scyllav1alpha1.GrafanaDatasourceSpec {
+	if spec := getGrafanaSpec(sm); spec != nil && len(spec.Datasources) > 0 {
+		// We only support one datasource for now.
+		ds := spec.Datasources[0]
+		if ds.Type == scyllav1alpha1.GrafanaDatasourceTypePrometheus {
+			return &ds
+		}
+	}
+	return nil
```
rzetelskik, line 120: Despite the fact we validate this, for the sake of correctness and clarity I'd much rather this returned an error if `len(spec.Datasources) > 1` or `ds.Type != scyllav1alpha1.GrafanaDatasourceTypePrometheus` instead of returning nil

pkg/controller/scylladbmonitoring/grafana_datasource.go
```diff
@@ -0,0 +1,104 @@
...
+		InsecureSkipVerify: false,
+	}
+}
+
+func makeGrafanaDatasourceURL(sm *scyllav1alpha1.ScyllaDBMonitoring) string {
+	if spec := getGrafanaSpec(sm); spec != nil && len(spec.Datasources) > 0 {
+		// We only support one datasource for now.
+		ds := spec.Datasources[0]
+		return ds.URL
+	}
```
rzetelskik, line 106: ditto

pkg/controller/scylladbmonitoring/sync_grafana.go
```diff
@@ -336,6 +345,7 @@ func (smc *Controller) syncGrafana(
 
 	var requiredDeployment *appsv1.Deployment
 	// Trigger restart for inputs that are not live reloaded.
+	// TODO: add custom TLS certs and auth token Secrets for Prometheus datasource.
```
rzetelskik: left a todo

pkg/controller/scylladbmonitoring/grafana_datasource.go
czeslavo: There are two valid cases here:
1. There is 1 datasource with Prometheus type defined -> return the datasource.
2. There's no such datasource defined -> return nil.

I intended not to duplicate the validation logic here, but instead rely on it.

pkg/controller/scylladbmonitoring/grafana_datasource.go
czeslavo: I see your concern, but I'm still not convinced that we should duplicate validation logic in `make*` functions. The passed `ScyllaDBMonitoring` is validated, and we can safely expect it to be correct already.

pkg/controller/scylladbmonitoring/sync_prometheus.go
czeslavo: Renamed.

pkg/controller/scylladbmonitoring/sync_prometheus.go
czeslavo: Converted into a switch.

pkg/controller/scylladbmonitoring/grafana_datasource.go
czeslavo: Renamed.

pkg/controller/scylladbmonitoring/sync_grafana.go
czeslavo: Removed it.

pkg/naming/names.go
czeslavo: Used `generateTruncatedHashedName` in both functions to handle that. I didn't spot any required docs updates. Did you have anything specific on your mind?

rzetelskik: Looks like only grafana ones are referenced in the user facing docs, thanks

pkg/controller/scylladbmonitoring/grafana_datasource.go
rzetelskik: It's fine to rely on it, we do so in many cases. To me the biggest benefit here would be the readability and maintainability - e.g. when you allow for more types and more entries in validation, this will start failing, instead of silently returning nil. The intent is also clearer then, now the implied logic would be that other entires/types are simply ignored. But, as the actual outcome is the same, let's call this a preference. I'll leave this up to you.

pkg/controller/scylladbmonitoring/sync_grafana.go
rzetelskik: Isn't it still required though? Are the tokens/certificates live-reloaded?

assets/monitoring/grafana/v1alpha1/provisioning.cm.yaml
```diff
@@ -20,16 +20,36 @@ data:
...
+        tlsCACert: "$__file{/var/run/configmaps/prometheus-serving-ca/{{ .prometheusDatasource.TLS.ServingCAConfigMapRef.Key }}}"
+          {{- end }}
+          {{- if $hasClientPair }}
+        tlsClientCert: "$__file{/var/run/secrets/prometheus-client-certs/tls.crt}" 
         tlsClientKey: "$__file{/var/run/secrets/prometheus-client-certs/tls.key}"
+          {{- end }}
+          {{- if $hasBearer }}
+        httpHeaderValue1: "Bearer $__file{/var/run/secrets/prometheus-bearer-token/{{ .prometheusDatasource.Auth.BearerTokenSecretRef.Key }}}"
+          {{- end }}
+        {{- end }}
```
rzetelskik:
```suggestion
      {{- if or $hasCACert $hasClientPair $hasBearer }}
      secureJsonData:
        tlsCACert: "$__file{/var/run/configmaps/prometheus-serving-ca/ca-bundle.crt}"
        tlsClientCert: "$__file{/var/run/secrets/prometheus-client-certs/tls.crt}"
        {{- if $hasCACert }}
        tlsCACert: "$__file{/var/run/configmaps/prometheus-serving-ca/{{ .prometheusDatasource.TLS.ServingCAConfigMapRef.Key }}}"
        {{- end }}
        {{- if $hasClientPair }}
        tlsClientCert: "$__file{/var/run/secrets/prometheus-client-certs/tls.crt}" 
        tlsClientKey: "$__file{/var/run/secrets/prometheus-client-certs/tls.key}"
        {{- end }}
        {{- if $hasBearer }}
        httpHeaderValue1: "Bearer $__file{/var/run/secrets/prometheus-bearer-token/{{ .prometheusDatasource.Auth.BearerTokenSecretRef.Key }}}"
        {{- end }}
      {{- end }}
```
What's going on with the indentation here? Can't we align the curly brackets with conditional parts?

pkg/controller/scylladbmonitoring/sync_grafana.go
czeslavo: It's been handled in https://github.com/scylladb/scylla-operator/pull/2943/commits/ca10f43bc8dd851a43da5a823b79a1ac886c2e66.

pkg/controller/scylladbmonitoring/sync_grafana_test.go
```diff
@@ -723,3 +971,278 @@ spec:
...
+      editable: false
+      jsonData:
+        timeInterval: "5s"
+        tlsSkipVerify: true
+  notifiers.yaml: ""
+  plugins.yaml: ""
+`, "\n"),
+		},
+		{
+			name: "custom prometheus datasource without TLS",
```
rzetelskik: nit: without TLS and no auth?

pkg/controller/scylladbmonitoring/sync_grafana_test.go
```diff
@@ -723,3 +971,278 @@ spec:
...
+      secureJsonData:
+        tlsCACert: "$__file{/var/run/configmaps/prometheus-serving-ca/ca-bundle.crt}"
+        tlsClientCert: "$__file{/var/run/secrets/prometheus-client-certs/tls.crt}" 
+        tlsClientKey: "$__file{/var/run/secrets/prometheus-client-certs/tls.key}"
+  notifiers.yaml: ""
+  plugins.yaml: ""
+`, "\n"),
+		},
+		{
+			name: "custom prometheus datasource with mTLS and bearer token",
```
rzetelskik, line 1038: we could use another case for no TLS, but a bearer token?

pkg/api/scylla/validation/scylladbmonitoring_validation.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func GetWarningsOnScyllaDBMonitoringUpdate(_, _ *scyllav1alpha1.ScyllaDBMonitoring) []string {
+	return nil
+}
+
+func validateScyllaDBMonitoringSpecUpdate(new, old *scyllav1alpha1.ScyllaDBMonitoringSpec, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if new.Components != nil && old.Components != nil {
```
rzetelskik, line 37: there's a lot of unnecessary nil-checking which doesn't seem to have any purpose, @czeslavo as discussed internally, let's add a point to API promotion tracker about making it a required non-nil field or removing it altogether

pkg/api/scylla/validation/scylladbmonitoring_validation.go
```diff
@@ -0,0 +1,252 @@
...
+		} else {
+			datasources := components.Grafana.Datasources
+			if len(datasources) == 0 {
+				allErrs = append(allErrs, field.Required(fldPath.Child("grafana").Child("datasources"), "exactly one datasource must be specified when Prometheus is in External mode"))
+			} else if len(datasources) > 1 {
+				allErrs = append(allErrs, field.TooMany(fldPath.Child("grafana").Child("datasources"), len(datasources), 1))
+			}
+		}
+	}
+
```
rzetelskik, line 120: shouldn't `Datasources` be required to be empty when `components.Prometheus.Mode` is `Managed`?

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -135,7 +135,7 @@ var _ = g.Describe("ScyllaDBMonitoring", func() {
 
 	// Disabled on OpenShift because of https://github.com/scylladb/scylla-operator/issues/2319#issuecomment-2643287819
 	g.DescribeTable("should setup monitoring stack TESTCASE_DISABLED_ON_OPENSHIFT", func(e *entry) {
-		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Minute)
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
```
rzetelskik: should we remove the test timeout altogether? It's still limited by the entire suite. Alternatively use `g.SpecContext` and set timeout on the ginkgo node

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
```diff
@@ -0,0 +1,76 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/storage/names"
+)
+
+var _ = g.Describe("ScyllaDBMonitoring webhook", func() {
+	f := framework.NewFramework("scylladbmonitoring")
+
+	g.It("should forbid invalid requests", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
```
rzetelskik: nit: use `g.SpecContext` and set timeout on the ginkgo node
nit: I'd also remove the test timeout altogether, it's limited by the suite anyway

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: I'd leave the timeouts on individual test cases. It frees up resources quicker, and there's no point in waiting for the whole suite when we know an individual case's reasonable time limit.

assets/monitoring/grafana/v1alpha1/provisioning.cm.yaml
czeslavo: To be fair, I don't know why I picked this indentation. Your suggestion looks way better. :sweat_smile:

pkg/controller/scylladbmonitoring/sync_grafana_test.go
czeslavo: Fixed.

pkg/controller/scylladbmonitoring/sync_grafana_test.go
czeslavo: Added a test case for it.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
czeslavo: Good catch. Added such a validation rule.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: `context.Background()` :arrow_right:  `specContext` applied.

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
czeslavo: `context.Background()` :arrow_right:  `specContext` applied.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
czeslavo: Created a tracker issue and added this as a point that has to be completed before the promotion: https://github.com/scylladb/scylla-operator/issues/2959.

pkg/controller/scylladbmonitoring/grafana_datasource.go
czeslavo: > when you allow for more types and more entries in validation, this will start failing, instead of silently returning nil

In both cases, to provide support for new types, one will have to touch this method and handle them here. Double-checking conditions that are already handled in the validating code sounds to me like an anti-pattern, as that would mean we have to always mirror them here. I prefer to leave it as is if you don't mind.

pkg/controller/scylladbmonitoring/sync_grafana_test.go
```diff
@@ -723,3 +971,345 @@ spec:
...
+`, "\n"),
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			_, objString, err := makeGrafanaProvisionings(tc.sm)
+			if err != nil {
+				t.Fatalf("unexpected error: %v\nRendered object:\n%s", err, objString)
```
rzetelskik: why print the returned object on an error? Isn't the returned value considered undefined in "status return" pattern?

pkg/api/scylla/validation/scylladbmonitoring_validation.go
```diff
@@ -0,0 +1,256 @@
...
+func validateScyllaDBMonitoringGrafanaDatasource(ds scyllav1alpha1.GrafanaDatasourceSpec, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if !oslices.ContainsItem(allowedDatasourceTypes, string(ds.Type)) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("type"), string(ds.Type), allowedDatasourceTypes))
+	}
+	if !oslices.ContainsItem(allowedDatasourceNames, ds.Name) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("name"), ds.Name, allowedDatasourceNames))
+	}
+
```
rzetelskik, line 168: URL validation is missing (only present on the schema level)

pkg/api/scylla/validation/scylladbmonitoring_validation.go
```diff
@@ -0,0 +1,256 @@
...
+	string(scyllav1alpha1.GrafanaPrometheusDatasourceAuthTypeNoAuthentication),
+	string(scyllav1alpha1.GrafanaPrometheusDatasourceAuthTypeBearerToken),
+}
+
+func validateScyllaDBMonitoringGrafanaPrometheusDatasource(ds *scyllav1alpha1.GrafanaDatasourceSpec, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if ds.URL == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("url"), "must be specified for Prometheus datasource"))
+	}
```
rzetelskik: isn't this redundant? Schema defines `MinLength=1` regardless of the type, so it feels like it shouldn't be validated for a specific type

pkg/api/scylla/validation/scylladbmonitoring_validation.go
```diff
@@ -0,0 +1,256 @@
...
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("type"), string(ds.Type), allowedDatasourceTypes))
+	}
+	if !oslices.ContainsItem(allowedDatasourceNames, ds.Name) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("name"), ds.Name, allowedDatasourceNames))
+	}
+
+	if ds.Type == scyllav1alpha1.GrafanaDatasourceTypePrometheus {
+		allErrs = append(allErrs, validateScyllaDBMonitoringGrafanaPrometheusDatasource(&ds, fldPath)...)
+	}
+
```
rzetelskik, line 175: nit: it should also be verified that PrometheusOptions are not set with other DS types (there are no other valid types now, but better safe than sorry)

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
rzetelskik: nit: Ok for individual timeouts. As for the specContext, consider also using ginkgo to configure the timeout then. Instead of
```
g.It("should forbid invalid requests", func(specCtx g.SpecContext) {
		ctx, cancel := context.WithTimeout(specCtx, testTimeout)
		defer cancel()
		...
})
```
You can do
```
g.It("should forbid invalid requests", func(ctx g.SpecContext) {
		...
}).WithTimeout(testTimeout)
```

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
```diff
@@ -0,0 +1,76 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/storage/names"
+)
+
+var _ = g.Describe("ScyllaDBMonitoring webhook", func() {
+	f := framework.NewFramework("scylladbmonitoring")
+
+	g.It("should forbid invalid requests", func(specCtx g.SpecContext) {
```
rzetelskik: let's use the test case tree we recently worked out for ScyllaCluster webhook validation, it should be very easy to adopt:  https://github.com/scylladb/scylla-operator/pull/2922, the warnings part is still unmerged in https://github.com/scylladb/scylla-operator/pull/2901, but it's not required here

pkg/controller/scylladbmonitoring/controller.go
```diff
@@ -590,3 +653,11 @@ func (smc *Controller) Run(ctx context.Context, workers int) {
 
 	<-ctx.Done()
 }
+
+func combineEnqueueFuncs(funcs ...controllerhelpers.EnqueueFuncType) controllerhelpers.EnqueueFuncType {
+	return func(depth int, obj kubeinterfaces.ObjectInterface, op controllerhelpers.HandlerOperationType) {
+		for _, fn := range funcs {
+			fn(depth, obj, op)
```
rzetelskik:
```suggestion
			fn(depth+1, obj, op)
```

pkg/controller/scylladbmonitoring/controller.go
```diff
@@ -528,6 +553,44 @@ func (smc *Controller) deleteServiceMonitor(obj interface{}) {
...
+
+	indexedSDBMs, err := smc.scyllaDBMonitoringInformer.Informer().GetIndexer().ByIndex(scyllaDBMonitoringByConfigMapIndexName, name)
+	if err != nil {
+		apimachineryutilruntime.HandleError(fmt.Errorf("can't get ScyllaDBMonitoring for ConfigMap %q: %w", name, err))
+	}
+
+	for _, indexedSDBM := range indexedSDBMs {
+		sdbm, ok := indexedSDBM.(*scyllav1alpha1.ScyllaDBMonitoring)
+		if !ok {
+			apimachineryutilruntime.HandleError(fmt.Errorf("expected *scyllav1alpha1.ScyllaDBMonitoring, got %T", indexedSDBM))
```
rzetelskik: should we use scylladbMonitoringControllerGVK instead of hardcoding the type?

pkg/controller/scylladbmonitoring/controller.go
```diff
@@ -528,6 +553,44 @@ func (smc *Controller) deleteServiceMonitor(obj interface{}) {
 	)
 }
 
+func (smc *Controller) enqueueBySecretRef(depth int, obj kubeinterfaces.ObjectInterface, op controllerhelpers.HandlerOperationType) {
+	name := obj.GetName()
```
rzetelskik: should we verify the object's type (secret)?

pkg/controller/scylladbmonitoring/conditions.go
```diff
@@ -1,8 +1,10 @@
...
-	prometheusControllerProgressingCondition = "PrometheusControllerProgressing"
-	prometheusControllerDegradedCondition    = "PrometheusControllerDegraded"
-	grafanaControllerProgressingCondition    = "GrafanaControllerProgressing"
-	grafanaControllerDegradedCondition       = "GrafanaControllerDegraded"
+	prometheusControllerProgressingCondition       = "PrometheusControllerProgressing"
+	prometheusControllerDegradedCondition          = "PrometheusControllerDegraded"
+	grafanaControllerProgressingCondition          = "GrafanaControllerProgressing"
+	grafanaControllerDegradedCondition             = "GrafanaControllerDegraded"
+	grafanaControllerConfigMapReferenceProgressing = "GrafanaControllerConfigMapReferenceProgressing"
+	grafanaControllerSecretReferenceProgressing    = "GrafanaControllerSecretReferenceProgressing"
```
rzetelskik: I don't think this calls for separate condition types, it feels like trashing the status a bit. A custom message is enough imo. It's still aggregated to GrafanaControllerProgressing/GrafanaControllerDegraded through RunSync

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: Are you sure it's possible to set a timeout the way you suggest? As I can see, `WithTimeout` method is only available for `o.Eventually` (`AsyncAssertion` interface). `g.It` / `g.DescribeTable` return bool.

pkg/controller/scylladbmonitoring/sync_grafana_test.go
czeslavo: That's a good point, no need for that. Fixed.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
czeslavo: We need to decide whether we'd really like to do it. In https://github.com/scylladb/scylla-operator/pull/2919#discussion_r2307548678 I accepted the suggestion to drop such validations. We can have a safe contract that once `type` is set to `TypeX`, we expect `XOptions` to be set, and we ignore others, so even if they're set, they have no effect.

pkg/controller/scylladbmonitoring/controller.go
czeslavo: `scylladbMonitoringControllerGVK` wouldn't be exactly the type we expect. How about this instead:

```go
apimachineryutilruntime.HandleError(fmt.Errorf("expected %T, got %T", &scyllav1alpha1.ScyllaDBMonitoring{}, indexedSDBM))

```

pkg/api/scylla/validation/scylladbmonitoring_validation.go
czeslavo: Moved from Prometheus-specific function.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
czeslavo: Moved.

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
czeslavo: Adopted.

pkg/controller/scylladbmonitoring/controller.go
czeslavo: Fixed.

pkg/controller/scylladbmonitoring/controller.go
czeslavo: It's only used in `add/update/deleteSecret`  methods where we're sure it's a Secret, but I guess it won't hurt us to double-check. Added both for Secret and ConfigMap.

pkg/controller/scylladbmonitoring/conditions.go
czeslavo: I wasn't sure what the convention was in such cases. I noticed it's aggregated, but it felt weird to me to set the same condition multiple times. Removed.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
rzetelskik: You're right, sorry for the confusion. This is done by providing additional argument to ginkgo nodes, see: https://github.com/scylladb/scylla-operator/blob/daaccfb82aaa2e10ee0272acdfc0e7bfe3a6068f/test/e2e/set/nodeconfig/nodeconfig_disksetup.go#L702.
Either way this is a nit.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
rzetelskik: IMO cross-checking improves UX as we're being verbose about the options which won't be used. This comment surprises me a lot because we implemented similar cross-checking recently with ScyllaDBManagerTask and it doesn't seem like that big of a mess to me: https://github.com/scylladb/scylla-operator/blob/daaccfb82aaa2e10ee0272acdfc0e7bfe3a6068f/pkg/api/scylla/validation/scylladbmanagertask_validation.go#L189-L225? For every TypeX, you essentially do:
- if TypeX, then validate XOptions
- if not TypeX, then validate XOptions not set

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: Converted to use `g.NodeTimeout`.

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
```diff
@@ -0,0 +1,102 @@
...
+			smCopy := createdSM.DeepCopy()
+			for _, f := range e.modifierFuncs {
+				f(smCopy)
+			}
+			framework.By("Updating the ScyllaDBMonitoring")
+			_, err = f.ScyllaAdminClient().ScyllaV1alpha1().ScyllaDBMonitorings(f.Namespace()).Update(ctx, smCopy, metav1.UpdateOptions{})
+			o.Expect(err).To(e.expectedErrMatcherFunc(smCopy))
+		}, g.NodeTimeout(testTimeout))
+	},
+		g.Entry("with valid spec", &entry{
```
rzetelskik:
```suggestion
		g.Entry("with acceptance when a valid ScyllaDBMonitoring", &entry{
```
otherwise this will read as `ScyllaDBMonitoring webhook should respond with valid spec [It] is created`
as opposed to `ScyllaDBMonitoring webhook should respond with acceptance when a valid ScyllaDBMonitoring [It] is created`

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
```diff
@@ -0,0 +1,102 @@
...
+			modifierFuncs:          nil,
+			expectedErrMatcherFunc: func(sm *scyllav1alpha1.ScyllaDBMonitoring) o.OmegaMatcher { return o.Succeed() },
+		}),
+		g.Entry("with invalid spec", &entry{
```
rzetelskik: ditto

pkg/controller/scylladbmonitoring/conditions.go
rzetelskik: The aggregating function ignores the types altogether, so it just seems like unnecessary noise to me - alternatively we could have some private, internal types, but I wouldn't expose them in `conditions.go`.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
czeslavo: I agree, we can adjust the API convention.

czeslavo: Addressed. Posted https://github.com/scylladb/scylla-operator/pull/2961.

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
czeslavo: Fixed.

test/e2e/set/scylladbmonitoring/scylladbmonitoring_webhooks.go
czeslavo: Fixed.

pkg/api/scylla/validation/scylladbmonitoring_validation.go
```diff
@@ -0,0 +1,267 @@
...
+	if !oslices.ContainsItem(allowedDatasourceNames, ds.Name) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("name"), ds.Name, allowedDatasourceNames))
+	}
+	if ds.URL == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("url"), "must be specified"))
+	}
+
+	if ds.Type == scyllav1alpha1.GrafanaDatasourceTypePrometheus {
+		allErrs = append(allErrs, validateScyllaDBMonitoringGrafanaPrometheusDatasource(&ds, fldPath)...)
+	}
```
rzetelskik: why is this separate from `validateScyllaDBMonitoringGrafanaDatasourceOptions`?

czeslavo: I left it separate as my idea for `validateScyllaDBMonitoringGrafanaDatasourceOptions` was to only include these cross-validation rules. But after looking at it again, it makes more sense to put it all together. Updated.

rzetelskik:
/lgtm
Thanks for the updates!

czeslavo:
Rebased onto master and added E2E tests to this PR, as we're blocked from merging it anyway (the initial intention was to keep tests separate to streamline the review process).

Programmatically skipping the OpenShift-specific test case due to the lack of an OpenShift-only suite (https://github.com/scylladb/scylla-operator/pull/2972, which is also blocked from merging due to the unstable multi-DC suite). The test itself is passing when run on OpenShift (see https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2980/pull-scylla-operator-master-e2e-openshift-aws-parallel/1971553134241124352).

/hold for https://github.com/scylladb/scylla-operator/pull/2972

czeslavo:
Rebased onto master and aligned with the new OpenShift E2E suite. PTAL

czeslavo:
Upgrade flake in `e2e-gke-parallel`, cluster provisioning fail in `olm-bundle-openshift-preflight`. 
/retest

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -188,12 +244,12 @@ func awaitScyllaDBMonitoringRollout(ctx context.Context, f *framework.Framework,
 func verifyManagedPrometheus(ctx context.Context, f *framework.Framework, sm *scyllav1alpha1.ScyllaDBMonitoring) {
 	g.GinkgoHelper()
 
-	// We need to retry the prometheus and grafana assertion for several reasons, some of them are:
-	//  - ingress exposure is asynchronous and some controllers don't report back status to wait for
-	//  - prometheus configuration is asynchronous without any acknowledgement
-	//  - grafana configuration is asynchronous without any acknowledgement
-	// Some of these may be fixable by manually verifying it in the operator sync loop so it can also be
-	// consumed by clients, but it's a bigger effort.
```
rzetelskik, line 195: Why is this comment being removed? Shouldn't it move to `verifyPrometheusTargetsAndRules`? I don't see any code changes that would explain it (it's still polling), but I might have lost some context.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -436,3 +497,357 @@ func decodeGrafanaDashboardFromGZBase64String(s string) (*grafanaDashboard, erro
...
+	return svc
+}
+
+func createExternalPrometheusInstanceWithoutTLS(ctx context.Context, f *framework.Framework, sm *scyllav1alpha1.ScyllaDBMonitoring, sa *corev1.ServiceAccount, svc *corev1.Service) *monitoringv1.Prometheus {
+	g.GinkgoHelper()
+
+	framework.By("Creating a Prometheus instance")
+	prom, err := f.PrometheusOperatorAdminClient().MonitoringV1().Prometheuses(f.Namespace()).Create(ctx, &monitoringv1.Prometheus{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      prometheusNameForScyllaDBMonitoring(sm),
```
rzetelskik: nit: I found it very confusing how the external prometheus instance was wired into ScyllaDBMonitoring, i.e. that is used the same naming convention as the managed instance. Should it take a name (service name??) as parameter?

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -436,3 +497,357 @@ func decodeGrafanaDashboardFromGZBase64String(s string) (*grafanaDashboard, erro
...
+				ServiceAccountName: sa.Name,
+				ServiceName:        pointer.Ptr(svc.Name),
+				SecurityContext: &corev1.PodSecurityContext{
+					RunAsNonRoot: pointer.Ptr(true),
+					RunAsUser:    pointer.Ptr[int64](65534),
+					FSGroup:      pointer.Ptr[int64](65534),
+				},
+				ServiceMonitorSelector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"scylla-operator.scylladb.com/scylladbmonitoring-name": sm.Name,
```
rzetelskik: nit: I think the wiring would be much more understandable if the constructors were "scylladbmonitoring agnostic" - i.e. they took everything needed for wiring with SDBM as parameters

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -436,3 +497,357 @@ func decodeGrafanaDashboardFromGZBase64String(s string) (*grafanaDashboard, erro
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	return prom
+}
+
+func createExternalPrometheusInstanceWithTLS(ctx context.Context, f *framework.Framework, sm *scyllav1alpha1.ScyllaDBMonitoring, sa *corev1.ServiceAccount, svc *corev1.Service) *monitoringv1.Prometheus {
+	g.GinkgoHelper()
+
+	framework.By("Creating a Prometheus instance with TLS")
+	tlsSecretName := prometheusTLSSecretNameForScyllaDBMonitoring(sm)
```
rzetelskik: Again, lost a bit of context. Who creates this? Are we wiring the external prometheus instance with certs generated by the operator?

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -171,6 +174,41 @@ var _ = g.Describe("ScyllaDBMonitoring", func() {
...
+				sm.Spec.Components.Grafana.Datasources = []scyllav1alpha1.GrafanaDatasourceSpec{
+					{
+						Name: "prometheus",
+						Type: scyllav1alpha1.GrafanaDatasourceTypePrometheus,
+						URL:  "https://thanos-querier.openshift-monitoring.svc:9091",
+						PrometheusOptions: &scyllav1alpha1.GrafanaPrometheusDatasourceOptions{
+							TLS: &scyllav1alpha1.GrafanaDatasourceTLSSpec{
+								InsecureSkipVerify: false,
+								CACertConfigMapRef: &scyllav1alpha1.LocalObjectKeySelector{
+									Name: "openshift-service-ca.crt",
```
rzetelskik: Q: how is this created in the namespace?

rzetelskik (commented):
Solid :muscle: 
left some minor comments

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: As we've agreed offline, I split this into separate comments for Grafana and Prometheus.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: I reworked this so we do not accept a whole `ScyllaDBMonitoring` object where possible anymore.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: It's automatically populated by OpenShift's Service CA Operator. But that's a good catch, as it turns out, it's a [legacy behavior](https://github.com/openshift/service-ca-operator/blob/e5d65c6f6b1446b77e0e915d64825d97be9c06de/pkg/controller/cabundleinjector/configmap.go#L70).

I changed the test logic to create a ConfigMap in our namespace and inject the CA bundle into it by annotating it as advised in the current OpenShift [docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/security_and_compliance/certificate-types-and-descriptions#cert-types-service-ca-certificates).

PR #2950, opened 2025-09-10
[Documentation] Remove number-lines directive from GKE systemconfig.yaml

abramche:
This makes it easier to copy the snippet from the docs (otherwise the line numbers also get copied)

Other snippets also do not have line numbers so this change makes it consistent


czeslavo (approved):
Just a note: It's possible to copy content without line numbers using the top-right clipboard button:
image

Anyway, in this instance, there's no need for line numbers as we do not refer to them anywhere in this doc. :+1: 

/lgtm

czeslavo:
/assign @rzetelskik 
I have no power to approve.

PR #2955, opened 2025-09-12
Add an enhancement proposal for a failed replace recovery guide

mflendrich:
Defines a procedure to recover from a stuck failed node replace.
My intention is that if/when we approve this approach, we copy the procedure (almost) directly into our documentation.

Reviewers - please cross-check the correctness of this sequence of operations.

/cc rzetelskik czeslavo
/kind documentation
/priority important-longterm

### DIFF
--- a/enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
+++ b/enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
@@ -24,6 +24,8 @@ No well-defined recovery path in scenarios where Operator gets stuck for unhandl
 
 The proposal is to add a guide that builds upon the [failed membership change guide](https://docs.scylladb.com/manual/branch-2025.1/operating-scylla/procedures/cluster-management/handling-membership-change-failures.html#cleaning-up-after-a-failed-membership-change) that would employ the following steps:
 
+_**Note Box**: To recover from failed node replace for multi-datacenter ScyllaDB deployments, simply perform the steps outlined in this guide in the worker cluster where the failed node is located._
+
 #### Verify that there is indeed a node that failed to join the cluster
 
 Perform `nodetool status` on a functioning node of the cluster (different than the culprit node). You should see a node with status different than `UN` (for example `DN`, `?N`).


enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+UN 10.152.183.43  456.25 KB 256    ?    ac4e578d-cc82-4b71-9ba1-0f40aede9e8d examplerack
+```
+
+### Notes/Constraints/Caveats [Optional]
+
+Question to reviewers.
+- Should we make this tutorial target other use cases as well (failed bootstrap in general?)
+- Can we expect in the general case that this will guarantee a return of the data integrity conditions one would expect from a successful node replace?
+
+### Risks and Mitigations
```
rzetelskik, line 164: if there are no risks, please leave an explicit comment

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+
```
rzetelskik, line 165: test plan is missing - how will this procedure be covered?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+## Alternatives
+
+- Implement a "freeze" switch in Operator to mitigate the need to scale it down to zero.
+- Implement this sequence of operations in Operator directly.
+
```
rzetelskik, line 215: if the "Design details" are not needed, please keep it with an explicit comment for clarity, now it just looks like it was omitted

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+- Add more automation to Operator
+
+## Proposal
+
+The proposal is to add a guide that builds upon the [failed membership change guide](https://docs.scylladb.com/manual/branch-2025.1/operating-scylla/procedures/cluster-management/handling-membership-change-failures.html#cleaning-up-after-a-failed-membership-change) that would employ the following steps:
+
+#### Verify that there is indeed a node that failed to join the cluster
+
+Perform `nodetool status` on a functioning node of the cluster (different than the culprit node). You should see a node with status `DN` or `?N`.
```
rzetelskik: > You should see a node with status `DN` or `?N`.

I don't see this mentioned anywhere in the upstream document? Where's this coming from? Are we sure the node replace procedure can't fail and result in a different status/state? What if the node in `DN/?N` is not the node that failed to be replaced or join the cluster, i.e. not the node that should be removed? Any other node could temporarily be seen as DOWN during the check - you definitely wouldn't want to remove it by mistake. We actually know the host ids of the node that was meant to be replaced and the one replacing it, perhaps we could try using it in addition to the nodetool status?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+#### Verify healing results
+
+Sit back and see Operator:
+- recreate the (identical) `StatefulSet` for the rack under repair (`scylla-exampledc-examplerack`),
+- recreate the (identical) `Service` for the node that failed to replace previously (`scylla-exampledc-examplerack-1`),
+- create new `Pod` and `PersistentVolumeClaim` for a net new node in place of the culprit node (each named `scylla-exampledc-examplerack-1`),
+
+After that happens, run `nodetool status` to see that the new node has joined the cluster and is `UN`:
```
rzetelskik, line 143: what if it's not UN?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+-- Address        Load      Tokens Owns Host ID                              Rack
+UN 10.152.183.112 491.57 KB 256    ?    e7478c73-07a9-4fb2-a435-6603ccc9e6bd examplerack
+UN 10.152.183.234 493.66 KB 256    ?    NEW-UUID-DIFFERENT-THAN-BEFORE-abcde examplerack
+UN 10.152.183.43  456.25 KB 256    ?    ac4e578d-cc82-4b71-9ba1-0f40aede9e8d examplerack
+```
+
+### Notes/Constraints/Caveats [Optional]
+
+Question to reviewers.
+- Should we make this tutorial target other use cases as well (failed bootstrap in general?)
```
rzetelskik, line 161: The upstream doc isn't specific to node replacement, so in general this would be desired imo. It seems like some steps would be superfluous for a failed bootstrap (disabling the operator, removing the STS), so it might depend on if we want a single generic procedure or multiple specific ones. I'd opt for prioritising maintainability on our end, so tl;dr; yes. Although there could be specific steps depending on the origin of the issue: e.g. verifying the host id could differ for node replacement and a "regular" bootstrap.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+Datacenter: exampledc
+=====================
+Status=Up/Down
+|/ State=Normal/Leaving/Joining/Moving
+-- Address        Load      Tokens Owns Host ID                              Rack
+UN 10.152.183.112 491.57 KB 256    ?    e7478c73-07a9-4fb2-a435-6603ccc9e6bd examplerack
+DN 10.152.183.214 466.12 KB 256    ?    09d815de-6f6d-4394-8439-bd8d34231835 examplerack
+UN 10.152.183.43  456.25 KB 256    ?    ac4e578d-cc82-4b71-9ba1-0f40aede9e8d examplerack
+```
+
```
rzetelskik, line 47: are there no other prerequisites? Is it not possible that there is a different ongoing topology change? What if there are other down nodes?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,149 @@
...
+#### Obtain the Host ID of the culprit node, and the Host IDs of potential ghost nodes.
+
+Follow the [_Step One: Determining Host IDs of Ghost Members_](https://docs.scylladb.com/manual/branch-2025.1/operating-scylla/procedures/cluster-management/handling-membership-change-failures.html#step-one-determining-host-ids-of-ghost-members) guide and note down the Host IDs of any potential _ghost nodes_.
+
+#### Stop the culprit node
+
+**WARNING**: This deletes the node's data.
+
+```console
+$ # Note: This command will put the PVC in the "Terminating" state and block until the pod gets deleted by the later step.
```
czeslavo: If I'm not mistaken, given Pods are already orphaned at this point by the STS, `Terminating` state can be avoided if we delete the PVC first.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
czeslavo: That's a good point. We could have an E2E test tied in some way to this procedure/guide. We expose some implementation details (manipulating individual Pods, STS, etc.) in the procedure to users, therefore they kind of become our interface. If we decide to drop STS for any reason (as we discussed recently) in the future, such a test would have to be adjusted, telling us the guide/doc has to be adjusted as well.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: You meant _if we delete the ~PVC~ **pod** first_ - correct?

czeslavo: Yes, sorry, my mistake in writing.

My point is that we can simplify if we delete the Pod first; we wouldn't have to explain the `Terminating` state then. Nothing crucial, though.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: > We actually know the host ids of the node that was meant to be replaced and the one replacing it, perhaps we could try using it in addition to the nodetool status?

I see one issue with this now:
- the `"internal.scylla-operator.scylladb.com/host-id"` label is not a part of an external interface
- the host ID wasn't propagated to Pod's labels or the pod has since been deleted

We could just cover this with a blank statement along the lines of "verify that the Host ID matches the Host ID of the node that failed to join the cluster", but maybe there is a more robust way to do this?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Actually there are plenty of risks - added descriptions thereof, and introduced some more defensiveness into the process (by recommending a backup, expecting a must-gather before performing the operation, and putting big warnings around the statefulset deletion operation).

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Based on our conversations today I am not confident that the move away from STS will happen.

That said, I see the need for testing in two aspects:
- actually verify (manually) that this procedure works well in a typical scenario. My inclination is that a single-DC test would be sufficient; I do not see how a multi-DC setup would impact it (the operations are limited to a single rack, except for the pause of the operator).
- verify the known failure modes (including cascading deletion of the sts) - that they actually do not cause data loss and that the recovery option given in _Risks and Mitigations_ is correct.
- regression testing the procedure, as you're mentioning? @abramche I would like your input here. I'm wondering if we could cover it by an SCT scenario.

Updated the proposal. Adding the regression test as optional.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Added, and moved the Test Plan underneath.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: If one does not see UN, it means that something bad happened, and they need to investigate it. Do you think that we should cover that explicitly in this document? If so, what would you like to see here?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: > The upstream doc isn't specific to node replacement, so in general this would be desired imo.

It _is_ specific, but to a wide range of scenarios (node replacement, node decommission, and other classes of failures), and achieves this by making cross-references between those classes. I do not want to cover too much in this fashion because (1) it creates an exponentially growing testing commitment that I don't want to make in the first iteration, (2) if we instead make it very general, it will not be useful, because it won't be clear which situations it applies to.

For this reason: I would be open to suggestion as to how to cover a broader class of problems with similar degree of coverage, straight path allowing to easily pattern-match the situation to the solution (without many assumptions), and minimal additional testing commitment.

Or just start small with one real-world need we see in CX, and reconsider expansion/generalization later on.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Good question!

I don't think that node removal is illegal during topology changes or with downtime in other parts of the cluster. This may impact data integrity, though.

Added a non-goal and an explicit note _The guide assumes that the rest of the cluster is healthy._ to keep the problem space reasonably small.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Reworded this section in an attempt to:
- open it up for an unbounded class of statuses that one can see here,
- point at the node's logs. It is my understanding that the replaced / replacing host id should appear there. I avoided the reference to the service annotations (but I personally don't buy the idea of external/internal interface on declarative state - I walked around this problem here, though, so probably not the right place to discuss).

PTAL

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Good point. Updated. (I did not test it though - will need to be manually tested, but I can't see why it wouldn't work)

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
czeslavo: > regression testing the procedure, as you're mentioning? 

Yes, that's what I meant. It should be done _somewhere_. Could be in our E2E as well, I believe.

rzetelskik: >  My inclination is that a single-DC test would be sufficient; I do not see how a multi-DC setup would impact it (the operations are limited to a single rack, except for the pause of the operator).

The document itself should definitely mention how this should be carried out in multi-dc deployments, but I don't see the need for a separate test.

> Yes, that's what I meant. It should be done somewhere. Could be in our E2E as well, I believe.

E2E sounds reasonable given that's what we do for the node replacement procedure itself.

mflendrich: > The document itself should definitely mention how this should be carried out in multi-dc deployments, but I don't see the need for a separate test.

From my PoV the procedure looks exactly the same (scale down the local operator instance, but no need to do so in the control-plane), just execute it in the datacenter cluster.

I'm happy to follow your lead on what you'd like this information to appear like in the guide.

> E2E sounds reasonable given that's what we do for the node replacement procedure itself.

Fair enough. Relaxed the requirement to allow SCT or E2E. I would like to make the final decision when @abramche is back and can give his 3 eurocents.

mflendrich:
I have a bit more confidence in this procedure than before, since "orphan delete the statefulset" is already a part of an [established procedure in our docs](https://operator.docs.scylladb.com/stable/resources/scyllaclusters/nodeoperations/volume-expansion.html).

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: > I'm happy to follow your lead on what you'd like this information to appear like in the guide.

Imo a simple note along the lines of what you just said "In multi-datacenter deployments, execute this it in the worker cluster" would suffice - see a similar one in [sysctls guide](https://operator.docs.scylladb.com/master/management/sysctls.html).

mflendrich: done - reused much of the format of the sysctls guide notice you've hinted at. :+1: 
PTAL

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: What does "is healthy" mean though? Should we specify if it's all/quorum UN (similarly to the upstream procedures)? We could also specify what to do if it's not in the expected state, i.e. "restore the state first" (see e.g. https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/replace-dead-node.html).

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,215 @@
...
+
+Once scaled up to a positive number of replicas, Operator will restore the original declarative state (recreate the `StatefulSet`) automatically. In the event that Operator is non-functional for whatever reason (environmental, bug, unhandled case), the StatefulSet can be recreated manually from its configuration in the must-gather archive collected.
+
+## Design Details
+
+### Test Plan
+
+#### Acceptance Test
+
+Verify that the procedure yields the desired result of deleting a non-functioning node and creating one that is up, **in a cluster that has all other nodes up**.
```
rzetelskik, line 196: Should we also manually verify this in a multi-datacenter setup (to see if e.g. it's not required to also disable the control plane operator)?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: It's ok to start with this imo, what I'm worried about is maintenance - how do we ensure we generalise this document to fit more scenarios in the future, instead of adding more, mostly identical ones for each scenario?

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: The upstream procedure doesn't mention anything, and, well, this is already the troubleshooting document. The correct step here would be to follow https://docs.scylladb.com/manual/branch-2025.1/operating-scylla/procedures/cluster-management/handling-membership-change-failures.html#cleaning-up-after-a-failed-membership-change again, right? So it's debatable whether it's even worth mentioning.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+#### Resume Operator and let it heal the cluster
+
+Resume Operator by scaling back to the original number of replicas.
+
+```console
+$ # Replace NNN with the number of replicas from before the initial scale-down-to-0.
+$ # That number is typically 2.
+$ kubectl scale -n scylla-operator deploy/scylla-operator --replicas=NNN --timeout=5m
```
rzetelskik, line 133: nit: in the actual doc, the number of replicas could be saved in an env variable

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: > how do we ensure we generalise this document to fit more scenarios in the future, instead of adding more, mostly identical ones for each scenario?

Is it the end of the world if we do, though? If we identify a need to adjust our docs to fit a different class of problems, we won't be too tied and will have the freedom to adjust this guide (and the overall shape of troubleshooting docs) with the context we'll have then (and that we don't have now).

We know today that cleaning up after a node that failed to replace is a necessary tool. This is the tool. Once we identify more tools needed, we'll create them (or adjust the existing ones).

rzetelskik: > Question to reviewers.
> - Should we make this tutorial target other use cases as well (failed bootstrap in general?)

So it seems like you already know the answer to your question

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: This is human-executed, so I would rely on a "common sense" definition of "is healthy". We are not in a position to define the state of healthy, and the definition will drift away from what we try to specify here.

> We could also specify what to do if it's not in the expected state, i.e. "restore the state first"

I cannot imagine a scenario in which such guidance will be accurate enough to be useful.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: It could; I don't think that it makes any practical difference, though. Let's think about it in the actual doc.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Should not be necessary IMO because it's a different realm of responsibility -- I'm coming from the standpoint of desire not to make this a larger project than it needs to be.

Auto multi-DC is supposed to mirror the config and offload execution to the datacenter operator. If there are unwanted side effects then it's a bug in auto multi-DC that I would rather not get into here.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: > We are not in a position to define the state of healthy

Aren't we? We know the operation that is being carried out at the end of this document - it's adding a new node. So if we want it to succeed, shouldn't the reader make sure no other nodes are DN? Especially once we have the bootstrap sync feature?
Having said that, I understand the starting state is going to be broken in (possibly) unforeseen ways, so the umbrella term might be as good as it gets. I'm leaving this up to you - if you think this is a good enough prerequisite for the user not to make things even worse, i'm ok with this.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: I don't have confidence that we have a recommendation that is robust and general enough for "my cluster is not fully up, guess why". This is in the domain of Scylla troubleshooting (that we can't afford to try to own as part of the operator project). That's why I would focus on this being a tool for a well-defined need to use in certain circumstances. (i.e. leave as-is)

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
rzetelskik: > Auto multi-DC is supposed to mirror the config and offload execution to the datacenter operator. 

I know and I'm mostly positive that this will work as intended. But bear in mind the node replacement procedure is manual and carried out in datacenter clusters directly in multi-dc deployments.

> If there are unwanted side effects then it's a bug in auto multi-DC that I would rather not get into here.

What's the alternative to getting into it here though?

mflendrich: > What's the alternative to getting into it here though?

Bluntly speaking - not adding this scenario into the test plan.

And I think it's justified, because auto multi-DC is a provisioning mechanism that deploys a remote datacenter using the Operator CRDs (SDBDC to be specific) as the interface.

And, as part of the procedure, we disable this interface by pausing Operator.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Capture a must-gather archive
+
+Since recovery by manual removal of a node is a destructive operation, please collect a [must-gather](https://operator.docs.scylladb.com/stable/support/must-gather.html) archive before changing any Kubernetes state. Do not skip this step - the archive may be useful in configuration recovery if something goes wrong.
+
+#### Pause the components that might interfere with the procedure
+
+This assumes that the failed node's name is `scylla-exampledc-examplerack-1` in the namespace `examplens`
+
+Stop the Operator (keep the ScyllaDB instances running, but prevent reconciliation). In a future improvement we can add a field to `ScyllaCluster` and/or `ScyllaDBDatacenter` to prevent Operator from "seeing" it, instead of scaling down to 0.
```
abramche, line 62: This sounds like a hack tbh. In my (idealistic) view, Operator could actually drive this procedure - as well as all other (possibly stuck/broken) topology fixes - or at least monitor, detect it and emit signals for a third party controller to handle the node operations. I would like to explore what holds us off investigating this stream of thought - as opposed to designing workarounds to the current limitations.

In this picture of the world, Operator detects an unhealthy node, tries to replace it with a timeout, polls for node health and starts a new one/terminates the old one when necessary.

Do you think it is overreaching? Remember that we are trying to build architecture which actively self heals without user intervention.

abramche (commented):
I feel like the manual approach leaves a grand room for mistakes and generally is very cumbersome. I would like us to explore if we can delegate some of all of the workflow actions to Operator and/or another Controller.

enhancements/proposals/2955-failed-replace-recovery/failed-replace-recovery.md
mflendrich: Is stopping operator by scaling it down to zero a hack? Hell yes. But it is an established one (a very similar approach [in our docs](https://operator.docs.scylladb.com/stable/resources/scyllaclusters/nodeoperations/volume-expansion.html)).

> we are trying to build architecture which actively self heals without user intervention

> I would like to explore what holds us off investigating this stream of thought - as opposed to designing workarounds to the current limitations

Two reasons.

1. Putting the product manager hat on - such automation competes for resources with other priorities that have much better reasons to be achieved, while this is "good enough" for what we need (and what we need is defined as "unblock cluster operators with stuck state and nowhere to go").
2. Experience shows that corner cases in automated self-healing procedures (like the one you're proposing to explore) do not correctly account for the broad range of possible situations and require a human's eye. Think about it this way: instead of trying to minimize the amount of manual intervention, this is an attempt to instead minimize the opportunities for stuckness where the cluster operator doesn't know how to recover from a certain situation.

The purpose of this procedure helps to recover from situations where an underlying system fails for undocumented/wrong reasons (because node replaces are supposed to work and not fail!). If one fails, we can only have a guess about what residue intermediate state the cluster is in. Automating that guess without a proven track record (which we do not have today) would be premature optimization. I am totally open to automating this path if we prove (by real-world use) that 100% of the time this procedure yields the desired result if executed precisely to the letter (because that's how automation would work, and the user would be back to square 1 if this "automated recovery" procedure wouldn't work for them).

mflendrich: discussed this offline earlier this week with a conclusion as described above (I believe - anyway, the point was that we're good to go).

therefore resolving this outstanding comment

PR #2967, opened 2025-09-18
[v1.18] Fix potential data races in RenderTemplate function

czeslavo:
Backport of #2954.


czeslavo:
Looks like a temporary problem with CI controller-manager validating webhook availability. 
/retest

PR #2486, opened 2025-02-18
Disable ScyllaDBMonitoring tests in OpenShift E2E

mflendrich:
Disables the ScyllaDBMonitoring test on OpenShift. We have decided ([here](https://github.com/scylladb/scylla-operator/issues/2319#issuecomment-2643287819)) that we won't support ScyllaDBMonitoring in its current form (spinning up a Prometheus) on OpenShift and will instead implement #2490.

Resolves #2318
Resolves #2319

I wanted to avoid calling out test cases by name in the CI machinery and therefore introduced a "tag" concept as a simplistic approach at https://github.com/scylladb/scylla-operator/issues/2448.
I think that the right long-term solution will be to define a number of test "suites" https://github.com/scylladb/scylla-operator/issues/2448 and have a proper grid for test configurations.

One failure is expected in the parallel suite and will be handled separately: https://github.com/scylladb/scylla-operator/issues/2489

PR #2972, opened 2025-09-18
[WIP] Use native ginkgo labels to group tests

czeslavo:


test/e2e/set/test_labels.go
```diff
@@ -0,0 +1,8 @@
+package set
+
+// This file contains labels used to group and filter test cases.
+
+const (
+	// NotSupportedOnOpenShiftLabel is a label for tests that are not supported on OpenShift.
+	NotSupportedOnOpenShiftLabel = "not-supported-on-openshift"
```
rzetelskik: we already have labels defined in test/e2e/framework/meta.go - is there a need for separating the two?

czeslavo: Didn't notice that, thanks for pointing it out. I changed the approach - I think it makes sense to add a `parallel/openshift` test suite instead of hacking on different layers. PTAL

mflendrich (approved):
/lgtm

>  I assume there had to be a reason it was hacked that way - @mflendrich please verify.

When implementing https://github.com/scylladb/scylla-operator/pull/2486 we considered labels, but the assumption was that adding another dimension ("is openshift / is not openshift") would require reconstruction of existing labels and labelfilters. I think this would have passed our "elegant solution" filter; I cannot recall why we didn't consider this.

I see no reason not to unhack this. Thank you.

czeslavo:
Turning this into a draft due to the multi-DC suite flakiness to avoid a false impression that it can be merged. It has to be merged together with https://github.com/scylladb/scylla-operator-release/pull/493.

rzetelskik:
/lgtm
@czeslavo feel free to cancel the hold when e2e-openshift-aws-parallel passes

rzetelskik:
> /lgtm @czeslavo feel free to cancel the hold when e2e-openshift-aws-parallel passes

passed
/hold cancel

Issue #2718, opened 2025-06-25
Scylla should be in its own namespace (best practices? Document only?)


mykaul:
We should advise, document, provide an example, where Scylla cluster is in its own namespace. To the best of my knowledge (and I could be wrong here), this is generally a best practice in K8S.

PR #2976, opened 2025-09-22
Document namespace requirements and best practices

mflendrich:
Fixes https://github.com/scylladb/scylla-operator/issues/2718

Also makes a drive-by fix to the outdated k8s version requirement.

/kind documentation
/priority backlog

### DIFF
--- a/docs/source/.internal/namespaces.md
+++ b/docs/source/.internal/namespaces.md
@@ -1,0 +1,5 @@
+:::{note}
+You can run your ScyllaDB cluster in the Kubernetes **namespace of your choice** (you can change the namespace used in this guide to your preference). It is a best practice (but not strictly required) to run your ScyllaDB cluster in a namespace separate from other applications.
+
+However, **the ScyllaDB Operator and ScyllaDB Manager must run in namespaces `scylla-operator` and `scylla-manager`, respectively**. It is [not currently possible](https://github.com/scylladb/scylla-operator/issues/2563) to use different namespaces for these two components.
+:::
--- a/docs/source/installation/helm.md
+++ b/docs/source/installation/helm.md
@@ -12,8 +12,11 @@ We will use Minikube K8s cluster, but this could be any K8s cluster supported by
 
 ## Prerequisites
 
-- Kubernetes 1.16+
+- Kubernetes (see the supported version range in the [support matrix](../support/releases.md#support-matrix))
 - Helm 3+
+
+:::{include} ../.internal/namespaces.md
+:::
 
 ## TL;DR
 
--- a/docs/source/installation/gitops.md
+++ b/docs/source/installation/gitops.md
@@ -9,6 +9,8 @@ For production deployment, you should always replace the {{productName}} image i
 We'd encourage you to use a sha reference, although using full-version tags is also fine.
 :::
 
+:::{include} ../.internal/namespaces.md
+:::
 
 ## Installation
 


docs/source/installation/helm.md
```diff
@@ -12,9 +12,12 @@ We will use Minikube K8s cluster, but this could be any K8s cluster supported by
 
 ## Prerequisites
 
-- Kubernetes 1.16+
+- Kubernetes (see the supported version range in the [support matrix](../support/releases.md#support-matrix))
```
rzetelskik, line 15: is this even worth mentioning that kubernetes and helm are prerequisites? GitOps guide doesn't mention k8s or kubectl. Maybe we should just have a box pointing to the supported versions (which should include helm)?

docs/source/.internal/namespaces.md
```diff
@@ -0,0 +1,5 @@
+## Namespaces
```
rzetelskik: if this is meant to be included, I'd use a directive (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#admonitions-messages-and-warnings) instead of a heading. Otherwise its reusability will be limited by the heading structure. IMO it will also better fit the structure of the documents you're including it in.

docs/source/.internal/namespaces.md
```diff
@@ -0,0 +1,5 @@
...
+
+You can run your ScyllaDB cluster in the Kubernetes namespace of your choice (this guide creates the `scylla` namespace, but you can change it to your preference). It is a best practice (but not strictly required) to run your ScyllaDB cluster in a namespace separate from other applications.
```
rzetelskik: > this guide creates the `scylla` namespace, but you can change it to your preference

The GitOps guide and the generic `docs/source/resources/scyllaclusters/basics.md` and `docs/source/resources/scylladbclusters/scylladbclusters.md` guides don't use any namespace (purposefully, although I personally don't agree with the reasoning). I suspect making the commands consistently specify one is the actual intention behind https://github.com/scylladb/scylla-operator/issues/2718.

docs/source/.internal/namespaces.md
rzetelskik: It's also invalid now and would require an exclusion: https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2976/pull-scylla-operator-master-build-docs/1970084440734437376#1:build-log.txt%3A157

docs/source/installation/helm.md
mflendrich: > Maybe we should just have a box pointing to the supported versions

This is exactly what is happening here (for kubernetes).

And for Helm I don't think it would be useful; the version only matters if the reader somehow got lost a few years back and didn't upgrade Helm 2->3 (which feels near impossible today).

I think it is practical to have a link to the supported k8s version, and I'd rather not overthink Helm here.

docs/source/.internal/namespaces.md
mflendrich: Done - replaced with a `note` box directive. The warning is not there anymore.

docs/source/.internal/namespaces.md
mflendrich: > I suspect making the commands consistently specify one is the actual intention behind https://github.com/scylladb/scylla-operator/issues/2718.

The right thing to do (and the intention of #2718) is to have a well-defined recommendation (as opposed to just having the user unknowingly fall into whatever we choose). This PR provides a well-defined recommendation.

Also, updated the note box to account for the fact that the GitOps guide does _not_ use the `scylla` namespace.

docs/source/installation/helm.md
rzetelskik: Not sure I understand - it sounds like you're agreeing there's no point in keeping Helm in prerequisites yet you're leaving it there?

docs/source/.internal/namespaces.md
rzetelskik: Thanks

docs/source/installation/helm.md
mflendrich: My desired outcome is signalling "You will need helm" and not overthinking it any further.

In particular, not polluting the version matrix with a component this stable, where we only care about API compatibility and not much more.

Hence "You need Helm v3" is a message that I view as reasonable in this place.

docs/source/.internal/namespaces.md
rzetelskik: It's definitely most important to have a clear recommendation. I think what I was trying to say is we should also unify what we do in our guides and specify a chosen namespace for Scylla deployment to allow for copy-pasting commands. But this is more of an independent UX improvement so let's not block this PR.

docs/source/installation/helm.md
rzetelskik: Ok, I can agree with Helm not being in the support matrix, although in that case we could also remove it from these prerequisites altogether as it becomes painfully obvious you'll need "some version" of Helm.
My main point was why does Helm guide mention K8s/Helm versions as prerequisites while GitOps guide doesn't mention K8s or kubectl - hence the proposal to have a unified box mentioning the supported versions. I don't think this is overthinking but an attempt at unifying and uncluttering the docs. But I won't be blocking this PR over this as it's not something introduced by yourself.

mflendrich: I think your point about uncluttering and unifying is right. That said, a direction that isn't consistently followed in our docs (and is [in my view] a quality of installation docs in general) is having a well-defined and **correct** list of prerequisites to review before you hit the ground running with the installation steps -- so that by looking at the prerequisites you know if you'll be successful or you'll fail, as opposed to bumping into (sometimes frustrating, sometimes blocking) issues as you go.

Anyway - this is a digression, thanks for unblocking the (unrelated) change.

Issue #2929, opened 2025-09-02
ScyllaDBMoniotring E2E test: External Prometheus with no auth and no TLS


czeslavo:
This issue covers implementing a regular E2E test for the following scenario:

- Prometheus deployed externally using Prometheus Operator, configured to be accessible without any authentication and TLS
- ScyllaDBMonitoring configured to use the externally deployed Prometheus as its datasource

PR #2977, opened 2025-09-22
Refactor ScyllaDBMonitoring E2E suite

czeslavo:
**Description of your changes:** Changes ScyllaDBMonitoring E2E test structure by splitting it into helpers. Pluggable helpers are going to be needed for external Prometheus testing.

Part of https://github.com/scylladb/scylla-operator/issues/2929.

### DIFF
--- a/test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
+++ b/test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
@@ -25,6 +25,7 @@ import (
 	promeheusappv1api "github.com/prometheus/client_golang/api/prometheus/v1"
 	configassests "github.com/scylladb/scylla-operator/assets/config"
 	grafanav1alpha1assets "github.com/scylladb/scylla-operator/assets/monitoring/grafana/v1alpha1"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/pointer"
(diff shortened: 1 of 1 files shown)


test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -37,263 +38,260 @@ import (
...
-	homeDashboardFolder := grafanav1alpha1assets.GrafanaDashboardsPlatform.Get()[homeDashboardDir]
-	o.Expect(homeDashboardFolder).NotTo(o.BeEmpty())
-	o.Expect(homeDashboardFolder).To(o.HaveKey(homeDashboardFile))
-	homeDashboardString := homeDashboardFolder[homeDashboardFile]
-	o.Expect(homeDashboardString).NotTo(o.BeZero())
-	ghd, err := decodeGrafanaDashboardFromGZBase64String(homeDashboardString)
-	o.Expect(err).NotTo(o.HaveOccurred())
-	o.Expect(ghd).NotTo(o.BeZero())
-	o.Expect(ghd.UID).NotTo(o.BeZero())
+func createSingleNodeScyllaCluster(ctx context.Context, f *framework.Framework) *scyllav1.ScyllaCluster {
```
rzetelskik: What's the point of this helper? Other than setting `Members` to 1 (is this even necessary?), everything this func does is very standard and already quite compact - it gets the specification and creates the object using a client. I found it harder to read the test having to jump to the helper just to see it doesn't do anything out of the ordinary or particularly repetitive

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -37,263 +38,260 @@ import (
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return sc
+}
 
-	type entry struct {
-		Type                     scyllav1alpha1.ScyllaDBMonitoringType
-		ExpectedDashboards       *[]gapi.FolderDashboardSearchResponse
-		ExpectedHomeDashboardUID *string
-	}
+func createScyllaDBMonitoringForScyllaCluster(ctx context.Context, f *framework.Framework, sc *scyllav1.ScyllaCluster, modifierFn func(*scyllav1alpha1.ScyllaDBMonitoring)) *scyllav1alpha1.ScyllaDBMonitoring {
```
rzetelskik, line 139: nit: for readability I wouldn't make the client call a part of this func, but it's somewhat of a personal preference

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -37,263 +38,260 @@ import (
...
+		},
+	)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return sm
+}
 
-		sc := f.GetDefaultScyllaCluster()
-		o.Expect(sc.Spec.Datacenter.Racks).To(o.HaveLen(1))
-		sc.Spec.Datacenter.Racks[0].Members = 1
+func verifyScyllaClusterRollout(ctx context.Context, f *framework.Framework, sc *scyllav1.ScyllaCluster) {
```
rzetelskik: The name is misleading - it waits for a state and only then verifies it. And again, I'm not sure it calls for a helper

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
```diff
@@ -37,263 +38,260 @@ import (
...
 )
 
-type grafanaDashboard struct {
-	Title string   `json:"title"`
-	Tags  []string `json:"tags"`
-	UID   string   `json:"uid"`
+type scyllaDBMonitoringEntry struct {
+	WithDescription              string
+	ScyllaDBMonitoringModifierFn func(*scyllav1alpha1.ScyllaDBMonitoring)
+	PreparePrometheusFn          func(context.Context, *framework.Framework, *scyllav1alpha1.ScyllaDBMonitoring)
```
rzetelskik: what does "PreprarePrometheus" mean? There are no comments or usages

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: The main point of this helper (and the others as well) is to make the body of the test case remain concise (preferably, to make it fit on a single screen). I find it way easier to read as a whole this way. It's 12 lines vs. 1 to read for this single helper, and 256 vs. 25 for the test body. I don't think there's always a need to read the whole implementation to see what the test does - that's what helpers are useful for, no?

As for setting `members` to 1 - indeed, the default set in the template is 1 already. Dropped it.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: As I mentioned in the response to https://github.com/scylladb/scylla-operator/pull/2977#discussion_r2372528039, my goal was to make the test body as concise as possible. What would be the benefit of not making that client call in this function, but splitting them and, e.g., making the call in the test body?

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: Is it? I think it depends on the interpretation of "verify" and "rollout" terms, but I think it should describe the intention quite good from the high level (it verifies the ScyllaCluster has been rolled out correctly). In the end, someone can look up the implementation itself when a need appears.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
czeslavo: Added a comment. It's a hook that will allow preparing an external Prometheus installation (not managed by ScyllaDBMonitoring).

rzetelskik: nit: I don't want to get into the naming discussion again, but maybe `PrepareUnmanagedPrometheusFn` (external, headless, whatever) would make it clearer then?

czeslavo: :+1: Let's rename to `PrepareExternalPrometheusFn`.

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
rzetelskik: > In the end, someone can look up the implementation itself when a need appears.

I could name the function `foo` and say the same thing :wink: 

> I think it should describe the intention quite good from the high level (it verifies the ScyllaCluster has been rolled out correctly)

I simply wouldn't expect it to be blocking then. Would `awaitAndVerifyScyllaClusterRollout` work?

czeslavo: > I could name the function foo and say the same thing 😉

That's an extreme example :sweat_smile: 

> I simply wouldn't expect it to be blocking then. Would awaitAndVerifyScyllaClusterRollout work?

Sounds good to me. :+1:

test/e2e/set/scylladbmonitoring/scylladbmonitoring.go
rzetelskik: I simply prefer the test body to be more explicit, especially wrt kube client methods - here the abstraction simply makes it less readable to me as I need to see the helper's implementation to understand what the test really does.
I understand it's mostly a matter of personal preference though and I can be biased towards the existing patterns, so I'm resolving this.

PR #2981, opened 2025-09-25
Use supported Grafana OAPI client in E2Es

czeslavo:
**Description of your changes:** Drops usage of deprecated https://github.com/grafana/grafana-api-golang-client in favor of auto-generated https://github.com/grafana/grafana-openapi-client-go. Also adds a convenient `grafana.Client` wrapper to test utils to reduce noise in tests themselves.

Motivations:

- https://github.com/grafana/grafana-openapi-client-go supports `GET /api/datasources/{uid}/health` (required for external prometheus tests) and `GET /api/dashboard/home` (already required, handled by manually crafted requests).
- dropping deprecated/unmaintained dependency

Note: https://github.com/grafana/grafana-openapi-client-go doesn't have official releases/tags, but it's actively maintained. It's the client recommended by Grafana. There's an issue tracking an official release: https://github.com/grafana/grafana-openapi-client-go/issues/89. As we need it only for testing purposes, I believe we are fine with using master's tip, though.

Part of https://github.com/scylladb/scylla-operator/issues/2929.


test/e2e/utils/grafana/client.go
```diff
@@ -0,0 +1,83 @@
...
+		c: c,
+	}, nil
+}
+
+func (c *Client) Dashboards() ([]Dashboard, error) {
+	searchParams := grafanasearch.NewSearchParams()
+	searchParams.Limit = pointer.Ptr[int64](1000)
+	searchParams.Type = pointer.Ptr("dash-db")
+
+	resp, err := c.c.Search.Search(searchParams)
```
rzetelskik: shouldn't this be rerun with `searchParams.Page` set? if you hit the limit (`len(resp.GetPayload()) >= limit`)?

test/e2e/utils/grafana/client.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if m, ok := resp.GetPayload().Dashboard.(map[string]interface{}); ok {
+		if uid, ok := m["uid"].(string); ok {
+			return uid, nil
+		}
+		return "", fmt.Errorf("home dashboard does not have a uid")
+	}
+
+	return "", fmt.Errorf("unexpected type for dashboard payload")
+}
```
rzetelskik, line 97: nit: Shouldn't the UID be a part of the `Dashboard` struct? In which case this method could return a dashboard and we'd extract its UID?

czeslavo: We specifically need only the dashboard UID here. If we ever need more than that, we can revisit this implementation. What would be the benefit of returning more than we need?

test/e2e/utils/grafana/client.go
czeslavo: Technically, yes. Practically, we expect up to ~50 items, and we have control over them. I assumed limit of 1000 is more than enough in our case and ignored this. But we can handle this. :+1:

test/e2e/utils/grafana/client.go
rzetelskik: I was thinking it would be more generic to have this method, but I now see that this is purpose-made for the e2e tests so it's fine as is

PR #2989, opened 2025-10-06
Don't emit FilesystemCreated when device already formatted

czeslavo:
**Description of your changes:** It fixes a bug that caused the emission of `FilesystemCreated` events even when a device was already formatted.

[Here](https://github.com/scylladb/scylla-operator/blob/6d673e2aaba4070a21460f8534cf69d632d67508/pkg/controller/nodesetup/sync_filesystems.go#L38) is the only production call of `MakeFS` - it treats the returned boolean as "changed", not "completed".

### DIFF
--- a/pkg/disks/fs.go
+++ b/pkg/disks/fs.go
@@ -12,6 +12,8 @@ import (
 	"k8s.io/utils/exec"
 )
 
+// MakeFS creates a filesystem of type fsType on the given device with the specified block size.
+// It returns a boolean indicating whether there were any changes made (i.e., if the filesystem was created).
 func MakeFS(ctx context.Context, executor exec.Interface, device string, blockSize int, fsType string) (bool, error) {
 	existingFs, err := blkutils.GetFilesystemType(ctx, executor, device)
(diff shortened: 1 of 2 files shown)


pkg/disks/fs.go
```diff
@@ -12,14 +12,14 @@ import (
 	"k8s.io/utils/exec"
 )
 
-func MakeFS(ctx context.Context, executor exec.Interface, device string, blockSize int, fsType string) (bool, error) {
+func MakeFS(ctx context.Context, executor exec.Interface, device string, blockSize int, fsType string) (changed bool, err error) {
```
rzetelskik: I wouldn't resort to using named returns unless they're needed in a defer statement. It doesn't improve the visibility that much either IMO. Should we just have a comment explaining what the expected returned values are?

rzetelskik (commented):
lgtm overall, left one comment
/approve

pkg/disks/fs.go
czeslavo: Converted into a comment.

PR #2992, opened 2025-10-10
Ensure all nodes in the cluster see all as UP as a prerequisite for adding new nodes

### LINKED ISSUE #2947: Ensure all nodes in the cluster see all as UP as a prerequisite for adding new nodes
### Description

As a result of investigating issues found in multi-datacenter bootstrap, we concluded that Operator does not currently follow topology change / rolling restart procedures. It should always ensure that all nodes in the cluster see all nodes as UP before progressing. See https://github.com/scylladb/scylladb/issues/25410#issuecomment-3270793089.

Despite https://github.com/scylladb/scylladb/commit/28c0a27467a16eacfa17fa8692fb2e433671055e alleviating the issue on the happy path, it has become very prevalent in multi-datacenter E2E tests as the DCs are rolling restarted throughout the bootstrap of the entire cluster.

### Acceptance criteria

- [ ] A mechanism ensuring all nodes in the cluster see all as UP as a prerequisite for adding new nodes exists

### Notes

For more context, see:
- https://github.com/scylladb/scylladb/issues/25410
- https://github.com/scylladb/scylladb/issues/8275
- https://github.com/scylladb/scylladb/issues/16763
- https://github.com/scylladb/scylla-operator/issues/1077

A predicted implementational issue is that K8s uses the same readiness probe to determine if:
a) the node is ready to serve traffic,
b) the node can be brought down / added; the entire cluster is ready for a topology change / rolling restart if all nodes are ready. 

It seems like the two concepts should be separate, and we should try to implement it somehow. We currently implement the readiness probe to only answer a) by checking if a node 1. considers itself UP (by GET /gossiper/endpoint/live/ ) and 2. has local CQL port open (by GET /storage_service/native_transport). The other point is not addressed by the readiness probe - addressing it there would result in availability issues.

There is nothing preventing the probe from talking to other nodes. Although, if we ask the entire cluster for the status of a node for every single node with a set frequency, it will cause a lot of noise. This may become a huge problem in clusters with high numbers of nodes or in multi-datacenter clusters where intercluster node-to-node communication may be expensive.

/kind bug
/priority important-soon

rzetelskik:

In this PR, I'm addressing a long-standing issue of topology changes not being synchronised by ScyllaDB Operator.
As per the [core documentation](https://docs.scylladb.com/manual/master/operating-scylla/procedures/cluster-management/add-node-to-cluster.html), before adding a new node, it should be ensured that no existing nodes are down. In fact, as described in https://github.com/scylladb/scylla-operator/issues/2947, the recommendation we've received from the cluster team is that the precondition for performing a topology change is that all nodes in the cluster see all nodes as Up.
This PR implements a mechanism ensuring this precondition is satisfied before **any new node is added**, i.e. when we know it needS to be bootstrapped. What follows is that we consciously excluded any other topology operations and rolling restarts from this mechanism. For the time being, **node replace operations are not synchronised** either.

To implement the synchronisation mechanism:
- A new, internal CRD `ScyllaDBDatacenterNodesStatusReport` is introduced. This resource is designed to aggregate and propagate node status information within a ScyllaDB datacenter, enabling the operator to coordinate topology changes and similar operations. The resource is not intended for direct user interaction.
- The Operator's sidecar is extended to poll its corresponding ScyllaDB node for statuses of nodes in the cluster. The status report is saved in the Pod's annotation.
- The ScyllaDBDatacenter controller is extended to collect and aggregate reports from all expected nodes in the DC.
- The ScyllaDBCluster controller is extended to mirror the reports from its datacenters between the worker clusters.
- An init container implementing the synchronisation mechanism is added to ScyllaDB Pods.

The synchronisation mechanism is a barrier blocking the regular proceeding to the start of the ScyllaDB process until the precondition of "all nodes in the cluster see all as UP" is met.
Nodes that have already bootstrapped are exempt from waiting for this precondition. To check if the node has already been bootstrapped, a `scylla sstable query` is performed to check the system.local table for bootstrapped value.
Any node replacing an existing node, i.e. having a non-empty `internal.scylla-operator.scylladb.com/replacing-node-hostid` annotation on its corresponding Service, is also exempt from this process.

Addition of the init container is put behind a `BootstrapSynchronisation` feature gate in alpha stage and disabled by default.
The init container is not created for ScyllaDBDatacenters running a ScyllaDB tagged image matching the semver with version lower than 2025.2.

The regular behaviour of the blocking barrier can also be overridden using `"scylla-operator.scylladb.com/force-proceed-to-bootstrap": "true"` annotation. The purpose of the annotation is to serve as a safety valve in case a user wants the ScyllaDB Pod to proceed to running the ScyllaDB process regardless of the preconditions and the extracted bootstrap status.

Adding the synchronisation mechanism also led me to discovering a bug in ScyllaDBCluster orchestration where multiple DCs could be created at once as the controller only checked for progressing conditions being set by preceding DC's reconciliation loops. Therefore any errors would've gone unnoticed. Hence, I'm also fixing this issue by checking for degraded conditions.

Resolves #2947 

/kind feature
/priority important-soon


rzetelskik:
/cc czeslavo mflendrich
Let's discuss the API. Naming suggestions are welcome.

pkg/api/scylla/v1alpha1/scylla.scylladb.com_scylladbdatacenterstatusreports.yaml
```diff
@@ -0,0 +1,76 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.18.0
+  name: scylladbdatacenterstatusreports.scylla.scylladb.com
+spec:
+  group: scylla.scylladb.com
+  names:
+    kind: ScyllaDBDatacenterStatusReport
```
mflendrich: 1. _status report_ is very abstract - hand-wavy as to what "status" it refers to, and who we're reporting to (to the user?)

2. this will get in the way of all autocompletes if the user wants to look at the SDC. I would prefer that we keep this internal bit out of the user's path. Therefore I recommend against using the `ScyllaDBDatacenter` prefix.

suggestion: `ScyllaDBNodeAvailability`

pkg/api/scylla/v1alpha1/types_scylladbdatacenterstatusreport.go
```diff
@@ -0,0 +1,66 @@
...
+type NodeStatusReport struct {
+	// Ordinal is the ordinal of the node within its rack.
+	Ordinal int `json:"ordinal"`
+
+	// HostID is the ScyllaDB node's host ID.
+	// +optional
+	HostID *string `json:"hostID,omitempty"`
+
+	// ObservedNodes holds the list of node statuses as observed by this node.
+	// +optional
```
czeslavo: Nit to make comments more precise.
```suggestion
	// Ordinal is the ordinal of the reporter node within its rack.
	Ordinal int `json:"ordinal"`

	// HostID is the ScyllaDB reporter node's host ID.
	// +optional
	HostID *string `json:"hostID,omitempty"`

	// ObservedNodes holds the list of node statuses as observed by the reporter node.
	// +optional
```

docs/source/api-reference/groups/scylla.scylladb.com/scylladbdatacenternodesstatusreports.rst
mflendrich, line 1: Not for now, but we should list the CRDs that are outside of the user's path separately from ones that the user might want to instantiate.

pkg/api/scylla/v1alpha1/types_scylladbdatacenterstatusreport.go
```diff
@@ -0,0 +1,66 @@
...
+	ObservedNodes []ObservedNodeStatus `json:"observedNodes,omitempty"`
+}
+
+type RackStatusReport struct {
+	// Name is the name of the rack.
+	Name string `json:"name"`
+
+	// Nodes holds the list of node status reports for this rack.
```
czeslavo: Ditto.
```suggestion
	// Nodes holds the list of node status reports reported by nodes from this rack.
```

pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
```diff
@@ -0,0 +1,66 @@
...
+	Name string `json:"name"`
+
+	// Nodes holds the list of node status reports for this rack.
+	Nodes []NodeStatusReport `json:"nodes"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
czeslavo, line 48: I think it would make sense to leave a comment on the top-level structure, explaining the purpose of this CRD. It could also mention this is meant to be an internal interface, and it's not expected to be used by users directly.

pkg/api/scylla/v1alpha1/scylla.scylladb.com_scylladbdatacenterstatusreports.yaml
rzetelskik: > suggestion: ScyllaDBNodeAvailability

I see a couple of issues:
1. semantically, It limits us to this single information - if we wanted to extend it with, say, node state (and not just status) it would already be more than that
2. it kind of implies a relation to a single node, as opposed to the entire datacenter and makes adding a resource for a single node more difficult (in case wanted to move the report out of the Pod's annotations, which I hope we will at some point)

docs/source/api-reference/groups/scylla.scylladb.com/scylladbdatacenternodesstatusreports.rst
rzetelskik: Agreed. Until then, I'll add a comment explaining it's a resource not intended for direct user interaction.

pkg/api/scylla/v1alpha1/scylla.scylladb.com_scylladbdatacenterstatusreports.yaml
rzetelskik: Also, in both scylla and k8s terms this would be "liveness", not availability?

czeslavo: I believe the `ScyllaDBDatacenter` prefix makes sense in this context, as this is exactly the scope of this status - it aggregates node statuses for a single DC. Also, I think that "status" is already an established term in the Scylla world; "availability" is not used that widely, I'd say? Maybe `ScyllaDBDatacenterNodesStatusReport`?

> this will get in the way of all autocompletes if the user wants to look at the SDC. I would prefer that we keep this internal bit out of the user's path. Therefore I recommend against using the ScyllaDBDatacenter prefix.

I would argue whether we should base our naming strategy on a worry about users' autocompletion hitting this. :thinking: If that's a real problem for us, we should rather aim for separating internal interfaces (e.g., by moving such CRDs to a separate API group).

mflendrich: I won't hold the autocomplete opinion strongly (I do not see how moving to a different apigroup would solve that problem though - I digress, more than happy to leave this point undiscussed or move to slack as a curiosity out of the critical path). I am just sad about `ScyllaCluster` / `ScyllaDBCluster` and may be overcorrecting here.

`ScyllaDBDatacenterNodesStatusReport` works for me wrt. being less hand-wavy about "status" and I believe it addresses the extensibility concerns brought up by @rzetelskik. So it might be a good name (at least to me).

pkg/api/scylla/v1alpha1/types_scylladbdatacenterstatusreport.go
rzetelskik: done

pkg/api/scylla/v1alpha1/types_scylladbdatacenterstatusreport.go
rzetelskik: changed a bit to avoid "reports reported"

pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
rzetelskik: done

pkg/api/scylla/v1alpha1/scylla.scylladb.com_scylladbdatacenterstatusreports.yaml
rzetelskik: done

czeslavo:
lgtm :+1:

rzetelskik:
@mflendrich @czeslavo I'm sending the implementation for review.

There are two commits which should be go separately: bumping Scylla version and fixing the bug in ScyllaDBCluster controller where multiple DCs could be created simultaneously. I'm leaving them here as one is a prerequisite (ScyllaDB version bump) and the other one would contribute to failing tests.

This is still a WIP: I've tried to add FIXME/TODO comments wherever I'm expecting to modify something or add any additional tests - feel welcome to leave suggestions there. Everything else you can treat as in its final state.

This PR won't include any documentation or e2e tests.

pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
```diff
@@ -0,0 +1,68 @@
...
+
+type NodeStatus string
+
+const (
+	NodeStatusUp   NodeStatus = "UP"
+	NodeStatusDown NodeStatus = "DOWN"
+)
+
+type ObservedNodeStatus struct {
+	//  HostID is the ScyllaDB node's host ID.
```
mflendrich:
```suggestion
	// HostID is the ScyllaDB node's host ID.
```

pkg/cmd/operator/sidecar/sidecar.go
```diff
@@ -94,11 +101,12 @@ func NewSidecarCmd(streams genericclioptions.IOStreams) *cobra.Command {
 	cmd.Flags().StringSliceVar(&o.ExternalSeeds, "external-seeds", o.ExternalSeeds, "The external seeds to propagate to ScyllaDB binary on startup as \"seeds\" parameter of seed-provider.")
 	cmd.Flags().StringVarP(&o.NodesBroadcastAddressTypeString, "nodes-broadcast-address-type", "", o.NodesBroadcastAddressTypeString, "Address type that is broadcasted for communication with other nodes.")
 	cmd.Flags().StringVarP(&o.ClientsBroadcastAddressTypeString, "clients-broadcast-address-type", "", o.ClientsBroadcastAddressTypeString, "Address type that is broadcasted for communication with clients.")
+	cmd.Flags().IntVarP(&o.StatusReportPeriodSeconds, "status-report-period-seconds", "", o.StatusReportPeriodSeconds, "How often (in seconds) to poll the status.")
```
mflendrich: Use the duration type flag instead of int?

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -614,35 +616,142 @@ func StatefulSetForRack(rack scyllav1alpha1.RackSpec, sdc *scyllav1alpha1.Scylla
...
+					InitContainers: func() []corev1.Container {
+						cs := []corev1.Container{
+							{
+								Name:            naming.SidecarInjectorContainerName,
+								Image:           sidecarImage,
+								ImagePullPolicy: corev1.PullIfNotPresent,
+								Command: []string{
+									"/bin/sh",
+									"-c",
+									fmt.Sprintf("cp -a /usr/bin/scylla-operator %s", naming.SharedDirName),
```
mflendrich: quoting bug (likely preexisting) - if `naming.SharedDirName` has a space or any other "difficult" characters it will be incorrectly expanded by the shell. suggestion: wrap `%s` with single quotes (`'%s'`)

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -614,35 +616,142 @@ func StatefulSetForRack(rack scyllav1alpha1.RackSpec, sdc *scyllav1alpha1.Scylla
...
+										strings.TrimSpace(`
+/usr/bin/scylla sstable query \
+--system-schema \
+` + fmt.Sprintf("--scylla-data-dir=%s", path.Join(naming.DataDir, "/data")) + ` \
+--keyspace=system \
+--table=local \
+--output-format=json \
+--query="SELECT bootstrapped FROM scylla_sstable.local" \
+` + path.Join(naming.DataDir, "/data/system/local-*/*-Data.db") + ` >/mnt/shared/bootstrapped.json || touch /mnt/shared/bootstrapped.json
+`),
```
mflendrich: nit: how about an incremental step towards cleaning up this code by wrapping in a `func MakeBootstrappedQueryCommand(...)`? (applies across the PR; this is just a starting point, there's much more to be done in this package)

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -614,35 +616,142 @@ func StatefulSetForRack(rack scyllav1alpha1.RackSpec, sdc *scyllav1alpha1.Scylla
...
+									ImagePullPolicy: corev1.PullIfNotPresent,
+									Command: []string{
+										"/usr/bin/bash",
+										"-euEo",
+										"pipefail",
+										"-O",
+										"inherit_errexit",
+										"-c",
+										strings.TrimSpace(`
+if [ "$(jq 'all(.[]; .bootstrapped? == "COMPLETED")' /mnt/shared/bootstrapped.json)" = "true" ]; then
```
mflendrich: code smell - repetitive use of the `/mnt/shared/bootstrapped.json` literal

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -704,6 +813,7 @@ exec /mnt/shared/scylla-operator sidecar \
 										return string(scyllav1alpha1.ScyllaDBDatacenterDefaultClientsBroadcastAddressType)
 									}() + ` \
 --service-name=$(SERVICE_NAME) \
+--status-report-period-seconds=5 \
```
mflendrich: use a flag default instead of hardcoding like this?

pkg/controller/statusreport/controller.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return fmt.Errorf("can't prepare annotation patch: %w", err)
+	}
+
+	_, err = c.kubeClient.CoreV1().Pods(c.namespace).Patch(ctx, c.podName, types.StrategicMergePatchType, patch, metav1.PatchOptions{})
+	if err != nil {
+		return fmt.Errorf("can't patch pod %q: %w", naming.ObjRef(pod), err)
+	}
+
+	klog.V(4).InfoS("Finished patching Pod with new node status report annotation", "Pod", naming.ObjRef(pod))
```
mflendrich, line 107: I am rather unconvinced about extending RBAC rules to grant patch on pods... is there a reason why this isn't on the node service instead?

pkg/controller/scylladbdatacenter/resource.go
mflendrich: And another question. Is there a good reason why this condition lives in bash, and not in the run-bootstrap-barrier implementation?

pkg/cmd/operator/sidecar/statusreporter.go
```diff
@@ -0,0 +1,79 @@
...
+	"sync"
+	"time"
+
+	"github.com/scylladb/scylla-operator/pkg/controller/statusreport"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type StatusReporter struct {
+	statusreport.Controller
```
czeslavo: Nit: There's no need to embed it; I believe it's better to keep it private.

pkg/controller/scylladbdatacenter/resource.go
czeslavo: BTW, have we thought about making user tweaking this period possible?

pkg/controller/statusreport/controller.go
czeslavo: Isn't it something we agreed on? The reason behind moving annotations from services to pods was that we want to react as quickly as possible to pods being torn down (in the ScyllaDBDatacenter controller that aggregates these statuses). As I understand, we could have an annotation on services, but we still would have to check with each pod whether it still exists anyway when building the aggregated status, so it's easier to just operate exclusively on pods.

It's to make sure we do not aggregate outdated status from a service once a pod is down (and doesn't report its status).

pkg/controller/statusreport/controller.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return nil
+}
+
+func getNodeStatusReport(ctx context.Context) (*internalapi.NodeStatusReport, error) {
+	scyllaClient, err := controllerhelpers.NewScyllaClientForLocalhost()
+	if err != nil {
+		return nil, fmt.Errorf("can't create Scylla client for localhost: %w", err)
+	}
```
czeslavo, line 118: I don't know how likely it is for this branch to be hit, but wouldn't it make sense to report it in the `NodeStatusReport` the way we do for `NodesStatusInfo` returning an error?

pkg/controller/statusreport/controller.go
mflendrich: Fair enough -- the latency argument makes sense to me.

pkg/controller/statusreport/controller.go
```diff
@@ -0,0 +1,152 @@
...
+	defer scyllaClient.Close()
+
+	nodeStatuses, err := scyllaClient.NodesStatusInfo(ctx, localhost)
+	if err != nil {
+		return &internalapi.NodeStatusReport{
+			Error: pointer.Ptr(err.Error()),
+		}, nil
+	}
+
+	var observedNodeStatuses []scyllav1alpha1.ObservedNodeStatus
```
czeslavo: We could preallocate.

pkg/controller/statusreport/controller.go
rzetelskik: If extending RBAC is an issue we could always check for Pod's existence/liveness in the controller while aggregating the statuses from Services. It does make the logic a bit more convoluted though. But it is possible.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: > Is there a good reason why this condition lives in bash, and not in the run-bootstrap-barrier implementation?

I think it provides better separation of concerns, to an extent, i.e. the command is not aware of the whole sstable querying. Otherwise a file with the result of the query is provided to the run-bootstrap-barrier command and it needs to parse it:

```go
type BootstrappedQueryResult []struct {
	Bootstrapped string `json:"bootstrapped"`
}

type BootstrapBarrierOptions struct {
        ...

	BootstrappedPath                       string

	bootstrapped bool
}

func (o *BootstrapBarrierOptions) AddFlags(cmd *cobra.Command) {
        ...
        
	cmd.Flags().StringVarP(&o.BootstrappedPath, "bootstrapped-path", "", o.BootstrappedPath, "Path to containing the JSON result of scylla-sstable query.")
}

func (o *BootstrapBarrierOptions) Complete(args []string) error {
	var err error

	var rawData []byte
	rawData, err = os.ReadFile(o.BootstrappedPath)
	if err != nil {
		return fmt.Errorf("can't read boostrapped query result file %q: %w", o.BootstrappedPath, err)
	}

	var data BootstrappedQueryResult
	if len(rawData) != 0 {
		err = json.Unmarshal(rawData, &data)
		if err != nil {
			return fmt.Errorf("can't unmarshall bootstrapped query result file %q: %w", o.BootstrappedPath, err)
		}
	}

	for _, res := range data {
		if res.Bootstrapped == "COMPLETED" {
			o.bootstrapped = true
		}
	}
	
	...
}
```

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: We could add it to the operator config, but I don't really see any arguments behind making it configurable by the user?

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -2111,3 +2113,149 @@ func cloneMapExcludingKeysOrEmpty[M ~map[K]V, S ~[]K, K comparable, V any](m M,
...
+	rackStatusReport := &scyllav1alpha1.RackNodesStatusReport{
+		Name:  rackSpec.Name,
+		Nodes: nodeStatusReports,
+	}
+	return rackStatusReport, nil
+}
+
+// makeNodeStatusReport creates a NodeStatusReport for a specific node in a rack.
+// It returns an optional NodeStatusReport, a boolean indicating whether the NodeStatusReport is non-nil, and an error.
+func makeNodeStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec *scyllav1alpha1.RackSpec, ordinal int, services map[string]*corev1.Service, podLister corev1listers.PodLister, isNodeExpectedInK8sState bool) (*scyllav1alpha1.NodeStatusReport, bool, error) {
```
czeslavo, line 2310: Nit: Why not return just `*scyllav1alpha1.NodeStatusReport, error` and check for `nil` downstream?

pkg/controller/statusreport/controller.go
mflendrich: I think it's better not to expand RBAC if we don't need to. But to do it if we need to. Please use your judgment here.
If we can get equivalent results without writing sprawling our metadata over to pods, my strong opinion held lightly is that it's worth doing.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -2111,3 +2113,149 @@ func cloneMapExcludingKeysOrEmpty[M ~map[K]V, S ~[]K, K comparable, V any](m M,
...
+	nodeStatusReport.HostID = &hostID
+
+	podName := naming.PodNameFromService(svc)
+	pod, err := podLister.Pods(sdc.Namespace).Get(podName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, false, fmt.Errorf("can't get pod %q: %w", naming.ManualRef(sdc.Namespace, podName), err)
+		}
+
+		// Pod is missing, report an empty status.
```
czeslavo, line 2343: Thought: returning an empty status can happen in different cases that we do not differentiate on the aggregated status level (missing pod, status not reported yet, node reported an error). Won't that be problematic for debugging?

pkg/controller/scylladbdatacenter/resource.go
czeslavo: Why make it configurable at all then if we always set it to 5s anyway?

rzetelskik: That's a fair point. @mflendrich wdyt?

pkg/controller/bootstrapbarrier/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+func isBootstrapPreconditionSatisfied(scyllaDBDatacenterNodesStatusReports []*scyllav1alpha1.ScyllaDBDatacenterNodesStatusReport, selfDC string, selfRack string, selfOrdinal int) bool {
+	klog.V(4).InfoS("Verifying if bootstrap precondition is satisfied.", "Datacenter", selfDC, "Rack", selfRack, "Ordinal", selfOrdinal)
+
+	// allHostIDs is a set of host IDs of all nodes which appeared in the status report, including the reportees.
+	allHostIDs := map[string]bool{}
+	// reportingHostIDToObservedNodeStatusesMap maps a reporting node's host ID to a map of observed nodes' host IDs to their statuses as observed by the reporting node.
+	reportingHostIDToObservedNodeStatusesMap := map[string]map[string]bool{}
+
+	for _, sdcnsr := range scyllaDBDatacenterNodesStatusReports {
```
czeslavo: Naming nit: just `report` instead of `sdcnsr` would be easier to read, I think. :sweat_smile:

pkg/controller/bootstrapbarrier/controller.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return nil
+}
+
+func isBootstrapPreconditionSatisfied(scyllaDBDatacenterNodesStatusReports []*scyllav1alpha1.ScyllaDBDatacenterNodesStatusReport, selfDC string, selfRack string, selfOrdinal int) bool {
+	klog.V(4).InfoS("Verifying if bootstrap precondition is satisfied.", "Datacenter", selfDC, "Rack", selfRack, "Ordinal", selfOrdinal)
+
+	// allHostIDs is a set of host IDs of all nodes which appeared in the status report, including the reportees.
+	allHostIDs := map[string]bool{}
```
czeslavo: Nit: we could express this variable's semantics more directly with `sets.Set[string]` type.

pkg/controller/bootstrapbarrier/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+				reportingHostIDToObservedNodeStatusesMap[*node.HostID] = observedNodeHostIDToNodeStatusesMap
+			}
+		}
+	}
+
+	// In non-automated multi-datacenter deployments, we expect nodes from external DCs to appear in the status report as reportees only.
+	// It is required to check that ALL nodes, not just reporter nodes, are present and UP in each report.
+	allowNonReportingHostIDs := false
+	if len(scyllaDBDatacenterNodesStatusReports) == 1 && scyllaDBDatacenterNodesStatusReports[0].DatacenterName == selfDC {
```
czeslavo: Do I understand correctly that these conditions are an indirect way of determining whether we're either in a single-DC cluster or a manual multi-DC cluster context (as there's only 1 node status report matching)?

I'm wondering whether we could make this more specific so it only returns true for manual multi-DC clusters. Perhaps we'd have to look at the ScyllaDBDatacenter and see if it has external seeds set, right? I guess it might be tricky to do it in this function, so if it's not possible, the comment will suffice.

pkg/controller/bootstrapbarrier/controller.go
```diff
@@ -0,0 +1,210 @@
...
+			klog.V(4).InfoS("Required node's status report is missing. Bootstrap precondition is not satisfied.", "HostID", hostID)
+			return false
+		}
+
+		for otherHostID := range allHostIDs {
+			if !nodeStatuses[otherHostID] {
+				// The other node is either missing from this node's report or is considered DOWN.
+				klog.V(4).InfoS("Node's status report is missing another node or considers it DOWN. Bootstrap precondition is not satisfied.", "HostID", hostID, "MissingOrDownHostID", otherHostID)
+				return false
+			}
```
czeslavo: I believe we could distinguish between missing and down conditions by using `status, ok := nodeStatuses[otherHostId]`. It might be worth doing for debuggability's sake. 

Also, I think that instead of `bool` as values in the maps in this function, it would be cleaner to use a strongly typed `NodeStatus`.

pkg/controller/scylladbdatacenter/resource.go
czeslavo: I agree that we could improve this by having all the decision-making encapsulated in a Go command instead of using bash. This way it's easier to test its logic. Ideally, we'd have the `scylladb-sstable-bootstrap-extractor` merged there as well, but the image requirements are kind of blocking us here.

I think we do not have to make the parsing part of the command's `Complete` function. We can have a dedicated struct that will expose `IsBootstrapped() (bool, error)` interface, and its constructor will accept the expected boostrapped file path. It can be a dependency of `Controller`.

czeslavo:
I've reviewed all the commits. It's solid work. :1st_place_medal:  It would be perfect if we could spend some time on improving test coverage here, but apart from that, I think we're good.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: What I don't like about this is that it sort of breaks the point of running the bootstrap barrier (or at least its naming), i.e. why even run the "bootstrap barrier" if the node is already bootstrapped?

pkg/controller/bootstrapbarrier/controller.go
rzetelskik: done

pkg/controller/bootstrapbarrier/controller.go
rzetelskik: done

pkg/controller/bootstrapbarrier/controller.go
rzetelskik: done

pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
rzetelskik: done, squashed into original commit

pkg/controller/scylladbdatacenter/resource.go
mflendrich: I think we can have both the encapsulation/testability that @czeslavo and is bringing up (and I implicitly meant by my comment), and the separation that @rzetelskik is referring to, by reducing the command to a one-liner similar to:

`scylla-operator check-if-bootstrapped || scylla-operator wait-for-bootstrap`

Both can be subcommands of operator, and the semantics of `check-if-bootstrapped` can be defined so that it exits with 0 if bootstrapped and 1 otherwise.

Even better if this expression lives in a shell script embedded within the image (because I imagine that we will be adding some bash boilerplate around it, and maybe some kind of error handling).

pkg/controller/scylladbdatacenter/resource.go
mflendrich: We've had scenarios where a constant set to a reasonable value at implementation time proved to be blocking in the event of a production incident. I do not see how storing the const in a flag with a default value (that we never touch unless we do) hurts anything.

Even more so because these are tools that will generally stay out of anyone's way. Nobody will bump into this setting if they don't have a problem with this mechanism. And if they do, I make an educated guess that they'd appreciate the ability to tune it.

pkg/controller/bootstrapbarrier/controller.go
rzetelskik: As discussed internally, I added a flag to distinguish between single-dc and non-automated multi-dc. The testing is pretty awful so I welcome any suggestions on how to make this nicer.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: Isn't this a bit of an overkill though? Writing an entire command to do an equivalent of `[ "$(jq 'all(.[]; .bootstrapped? == "COMPLETED")' /mnt/shared/bootstrapped.json)" = "true" ]`? I find it hard to understand the benefit here, given the cost.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: I think that's the point @czeslavo is trying to make here though: there isn't much point to it unless we expose it to the user. As-is it's not configurable unless changed directly in a Pod spec, but that'd be overwritten by the controller.

pkg/controller/scylladbdatacenter/resource.go
mflendrich: The fact that "writing an entire command" is a cost is a problem of the project. It should not be difficult at all to write encapsulated logic, instead of hardcoding bash scripts in yaml.

From my PoV, a trivial one-pager cobra.Command with a single `Run` function would suffice - I wouldn't go into the Validate/Complete/Execute pattern.

In fact - I see a lot of value in establishing a pattern to create a trivial way to write stuff in a Go executable, not in inline bash.

One other alternative (that has the encapsulation feature, but not so much testability) is creating the bash script in the image, and calling it from yaml as an entrypoint.

I do not think that a (well written) new cobra command is more verbose than this, though. It might need us to give up on some of our existing cobra boilerplate - which I honestly see as superfluous.

pkg/controller/scylladbdatacenter/resource.go
czeslavo: Yup, I suggest that we either expose it as tunable or just drop this (internal) CLI option.

pkg/naming/constants.go
```diff
@@ -297,3 +304,14 @@ const (
...
+
+const (
+	// ScyllaDBDatacenterNodesStatusReportSelectorLabel is used to uniformly label nodes status reports created for a ScyllaDB cluster.
+	// It allows easy selection of all datacenter reports for a given cluster.
+	ScyllaDBDatacenterNodesStatusReportSelectorLabel = "scylla-operator.scylladb.com/scylladb-datacenter-nodes-status-report-selector"
+
+	// ForceProceedToBootstrapAnnotation allows to force the bootstrap barrier to proceed without waiting for normal preconditions to be satisfied.
+	// If set to "true", bootstrap barrier will proceed to bootstrap immediately.
+	// This can be used on a particular member Service to override the normal bootstrap precondition checks, or on the entire ScyllaDBDatacenter to override the default behavior for all members.
+	ForceProceedToBootstrapAnnotation = "scylla-operator.scylladb.com/force-proceed-to-bootstrap"
```
czeslavo, line 318: Does it have any other use than skipping the bootstrap sync in E2Es as a workaround for the limitation we have that prevents us from tweaking the feature gate from a test suite?

Couldn't we deduct the Scylla version at the level of the bootstrap barrier and decide based on that whether the feature is supported?

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -614,35 +616,131 @@ func StatefulSetForRack(rack scyllav1alpha1.RackSpec, sdc *scyllav1alpha1.Scylla
...
+									Name:            "scylladb-bootstrap-barrier",
+									Image:           sidecarImage,
+									ImagePullPolicy: corev1.PullIfNotPresent,
+									Command: []string{
+										"/usr/bin/scylla-operator",
+										"run-bootstrap-barrier",
+										"--service-name=$(SERVICE_NAME)",
+										"--sstable-bootstrapped-query-result-path=/mnt/shared/bootstrapped.json",
+										fmt.Sprintf("--selector-label-value=%s", naming.ScyllaDBDatacenterNodesStatusReportSelectorLabelValue(sdc)),
+										fmt.Sprintf("--single-report-allow-non-reporting-host-ids=%t", len(sdc.Spec.ScyllaDB.ExternalSeeds) > 0),
```
czeslavo: I believe it deserves a comment explaining why we deduct this value from `len(externalSeeds) > 0` (it's a heuristic that tells us if the DC is part of a manual multi-DC cluster), and why we want to allow non-reporting host IDs in this case.

test/e2e/utils/helpers.go
```diff
@@ -899,3 +902,14 @@ func GetContainerReadinessMap(pod *corev1.Pod) map[string]bool {
...
 	return res
 }
+
+func SetForceProceedToBootstrapAnnotationTrue(obj *metav1.ObjectMeta) {
+	g.GinkgoHelper()
+
+	if !utilfeature.DefaultMutableFeatureGate.Enabled(features.BootstrapSynchronisation) {
+		// Skip setting the annotation if the feature gate is disabled.
+		return
+	}
```
czeslavo: Should we check it this way? Doesn't it check the default feature gate value, not the actual value the Operator has set?

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: > The fact that "writing an entire command" is a cost is a problem of the project. It should not be difficult at all to write encapsulated logic, instead of hardcoding bash scripts in yaml.

It's not difficult. It's just inherently more time consuming than writing a one-liner with an existing tool that's already a part of our container image. It is an acceptable trade-off for readability and maintainability, though.

> In fact - I see a lot of value in establishing a pattern to create a trivial way to write stuff in a Go executable, not in inline bash.

I agree, the inline bash scripts are an eyesore and we've grown too accustomed to them.

> I do not think that a (well written) new cobra command is more verbose than this, though. It might need us to give up on some of our existing cobra boilerplate - which I honestly see as superfluous.

I personally like it, I find it quite readable and it's mostly copy-pasted. Perhaps we should deduplicate parts of it though.

Anyway, I moved everything to Go and added an embedded struct to bootstrap barrier command.

pkg/controller/statusreport/controller.go
rzetelskik: We've reiterated over this with @czeslavo and it seems that we both agree that while expanding RBAC is not perfect, the annotation is tied to Pod's lifetime because it reflects its state. Service became a de facto reflection of ScyllaDB node's persistent state and I don't think it fits it. Leaving as is.

pkg/controller/statusreport/controller.go
rzetelskik: Doesn't change much from the aggregation perspective but it probably provides better visibility. Done.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: I think it goes against the established status return pattern. For functions returning (*T, error) we expect *T to be non-nil when error is nil. Leaving as is.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: I wouldn't make errors part of the aggregated status report. I could add some debug logs though, wdyt?

pkg/controller/statusreport/controller.go
rzetelskik: done

pkg/cmd/operator/sidecar/statusreporter.go
rzetelskik: done

test/e2e/utils/helpers.go
rzetelskik: It checks the value that the testing binary has set, which determines what we should expect in tests imo? That's what we do in the existing E2Es btw.

czeslavo: Is it guaranteed that the testing binary will have the same set of flags set as the operator that's deployed separately?

pkg/naming/constants.go
rzetelskik: > Does it have any other use than skipping the bootstrap sync in E2Es as a workaround for the limitation we have that prevents us from tweaking the feature gate from a test suite?

Yes, it's a safety valve essentially. The bootstrap barrier is very disruptive and I believe there needs to be a way to override it. Support cases show that things sometimes go sideways and I think it's better to have an option to disable the barrier on a live cluster, especially for a particular node.

> Couldn't we deduct the Scylla version at the level of the bootstrap barrier and decide based on that whether the feature is supported?

We don't really have a robust way of checking a Scylla version at this point. We expect an image tag to match semver. We could maybe use semver.SupportFeatureUnsafe, in which case we would block ONLY if scylla container's image tag matches semver and is >=2025.2.Z. But I'm very hesitant. Wdyt?

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

test/e2e/utils/helpers.go
rzetelskik: It's not guaranteed, but it's controlled by an env var shared between our scripts. It's not perfect, but good enough imo I'd leave it as is unless we decide to do https://github.com/scylladb/scylla-operator/pull/2992#discussion_r2451229357

pkg/naming/constants.go
rzetelskik: @mflendrich in internal conversations Greg proposed to only create the bootstrap sync init containers when both the feature gate is enabled and scylla version is >= 2025.2. Wdyt? Notice my above comment about version extraction depending on the container image tag matching semver.

rzetelskik:
Update: all of your existing comments should've been addressed by now or are being discussed. I'm yet to add unit tests for `makeExternalScyllaDBDatacenterNodesStatusReports` in pkg/controller/scylladbcluster/resource.go and a regression test for the sstable bootstrapped query using testcontainers.
If I've missed anything, please ping me.

pkg/naming/constants.go
mflendrich: > Support cases show that things sometimes go sideways and I think it's better to have an option to disable the barrier on a live cluster, especially for a particular node.

This is a great point. An annotation is an easy to use, reliable and flexible way to bypass certain steps. And I think that in this regard it's superior to relying on an operator feature gate to recreate a pod with a different set of initcontainers.

> only create the bootstrap sync init containers when both the feature gate is enabled and scylla version is >= 2025.2

I cannot see how it would make sense to create them in the opposite case, so "yes of course".

> version extraction depending on the container image tag matching semver.

Sounds good if we have a reasonable mode of failure if the version check fails for whatever reason. I think "default enabled" + a warning print in the operatot logs might be a good starting point.

pkg/controller/scylladbdatacenter/resource.go
mflendrich: > there isn't much point to it unless we expose it to the user

there is - already discussed in another context

> Support cases show that things sometimes go sideways and I think it's better to have an option to disable [...]

If the world is on fire, having a knob to tune a mechanism can be worth its weight in gold. Even if not directly exposed; we have a track record of stopping operator to manually edit something. The point I'm trying to make is that a superfluous knob (defaulted and never touched in the happy path) is a lower cost than the risk of getting burned by an unnecessarily rigid automation.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -614,35 +616,122 @@ func StatefulSetForRack(rack scyllav1alpha1.RackSpec, sdc *scyllav1alpha1.Scylla
...
-						},
-					},
+								{
+									Name:            "scylladb-bootstrap-barrier",
+									Image:           sidecarImage,
+									ImagePullPolicy: corev1.PullIfNotPresent,
+									Command: []string{
+										"/usr/bin/scylla-operator",
+										"run-bootstrap-barrier",
+										"--service-name=$(SERVICE_NAME)",
```
mflendrich: TIL that one can expand `env` in `command`!!!

pkg/controller/scylladbdatacenter/resource.go
mflendrich: Thank @rzetelskik. The barrier command is looking good. I think it's an investment well worth making.
I am expecting the same for `makeScyllaDBSSTableBootstrappedQueryCommand` - the same arguments apply. I think it's in line with the observation that

> I agree, the inline bash scripts are an eyesore and we've grown too accustomed to them.

which I stand strongly behind :100:.

rzetelskik: > I am expecting the same for makeScyllaDBSSTableBootstrappedQueryCommand

What are you expecting there? It uses scylla image. Unless we want to do some weird injections we can't do much.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: I'm totally fine with leaving the defaulted flag there, but could you explain how it could be used in a live cluster? Edit the pod/statefulset manually?

mflendrich: Exactly like that (having paused operator as a prerequisite).
Of course this is a "warranty void" scenario, but experience shows that we need them sometimes.

pkg/controller/scylladbdatacenter/resource.go
mflendrich: The container already mounts the `shared` volume, so it has the operator executable therein. Can we not just call it directly? Isn't it the "injection" you're looking for?

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

pkg/cmd/operator/sidecar/sidecar.go
rzetelskik: done

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

pkg/naming/constants.go
rzetelskik: done, with the exception that this

> a warning print in the operatot logs might be a good starting point

is a debug log - it's logged on every reconciliation of a statefulset, so a warning would be polluting the logs quite a lot. The only reasonable way I see of making this a warning is to add it to the admission webhook, but it seems a bit over the top.

rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[4e5b5fc](https://github.com/scylladb/scylla-operator/commit/4e5b5fcc6bf255e03e97a57a11c6741771c7a3f0)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2992/pull-scylla-operator-master-e2e-gke-parallel/1982934877825142784)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=2992). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

Cleanup jobs bug. Unrelated to this PR.
```
  - lastTransitionTime: "2025-10-27T22:52:34Z"
    message: 'JobControllerProgressing: Waiting for Job "e2e-test-scyllacluster-cddp2-jtdhv/cleanup-basic-88vw5-us-east-1-us-east-1a-0"
      to complete.'
    observedGeneration: 2
    reason: WaitingForJobCompletion
    status: "True"
    type: JobControllerProgressing
```

rzetelskik:
@czeslavo @mflendrich please verify that all outstanding comments have been addressed.
I've added the PR description. It should be ready to merge now.
As agreed in internal discussions, we won't be making the regression tests for `scylla sstable` command a part of this PR.
I'll be adding trackers for:
- adding an integration test using testcontainers for `scylla sstable` command regression
- expanding the mechanism to node replacement procedure
- refactoring the `StatefulSetForRack` function and its unit test to split it into smaller, testable pieces and only keeping an "integration test" for the entire product

Expanding the mechanism to rolling restarts is already tracked in https://github.com/scylladb/scylla-operator/issues/1077.

Please let me know if I've missed anything.

pkg/cmd/operator/bootstrapbarrier/bootstrapbarrier.go
```diff
@@ -0,0 +1,255 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	apimachineryutilerrors "k8s.io/apimachinery/pkg/util/errors"
+	apimachineryutilvalidation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	cliflag "k8s.io/component-base/cli/flag"
+	"k8s.io/klog/v2"
+)
+
+type Options struct {
```
czeslavo, line 33: For posterity, it might be worth noting that this command has a dependency on the `scylla sstable` command; therefore, it has to be run in the context of a Scylla container.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -1163,22 +1136,77 @@ func containerPorts(sdc *scyllav1alpha1.ScyllaDBDatacenter) ([]corev1.ContainerP
...
+	if err != nil {
+		return nil, fmt.Errorf("can't make ScyllaDB bootstrap barrier init container: %w", err)
+	}
+	if ok {
+		initContainers = append(initContainers, *bootstrapBarrierContainer)
+	}
+
+	return initContainers, nil
+}
+
```
czeslavo, line 1164: I think it could benefit from an explicit comment on why we need to copy the operator binary to a shared directory.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -1198,7 +1226,80 @@ func sysctlInitContainer(sdc *scyllav1alpha1.ScyllaDBDatacenter, image string) (
...
+		klog.V(4).InfoS("Not including bootstrap barrier init container as ScyllaDB version does not support it", "ScyllaDBDatacenter", naming.ObjRef(sdc), "ScyllaDBVersion", scyllaDBVersion, "ScyllaDBVersionRequiredForBootstrapSynchronisation", semver.ScyllaDBVersionRequiredForBootstrapSynchronisation)
+		return nil, false, nil
+	}
+
+	c := &corev1.Container{
+		Name:            "scylladb-bootstrap-barrier",
+		Image:           sdc.Spec.ScyllaDB.Image,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Command: []string{
+			"/mnt/shared/scylla-operator",
```
czeslavo, line 1257: Again, we could comment here why we expect the scylla-operator binary to be present in this directory while we run the ScyllaDB image.

pkg/cmd/operator/bootstrapbarrier/bootstrapbarrier.go
rzetelskik: done, squashed into existing commits

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done, squashed into existing commits

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done, squashed into existing commits

czeslavo:
lgtm :rocket: 
/assign @mflendrich

Issue #2934, opened 2025-09-02
ScyllaDBMonitoring docs: OpenShift guide


czeslavo:
Add a guide that will explain how to configure `ScyllaDBMonitoring` to work with OpenShift's user-workload monitoring stack (how to bind the right `ClusterRole` to the Grafana service account, how to get the bearer token, etc.).

mflendrich:
Should this indeed be a guide? What is the starting point for this guide? Are we going to teach how start from scratch?

czeslavo:
> Should this indeed be a guide? What is the starting point for this guide? Are we going to teach how start from scratch?

The starting point would be that the user has an OpenShift cluster with a Scylla cluster deployed and wants to configure monitoring using UWM and our ScyllaDBMonitoring.

Issue #2933, opened 2025-09-02
ScyllaDBMonitoring docs: Extend resource page to reflect new fields


czeslavo:
Update the `ScyllaDBMonitoring` CRD documentation to reflect the new fields and their usage.

https://operator.docs.scylladb.com/stable/resources/scylladbmonitorings.html

PR #2993, opened 2025-10-10
Rework and split ScyllaDBMonitoring docs into granular articles

czeslavo:
**Description of your changes:** Reworks ScyllaDBMonitoring docs by splitting the monolithic article under `resources` into 3 new articles and moving them under `management/monitoring`:

- ScyllaDB Monitoring overview (explanation)
- ScyllaDB Monitoring setup (guide)
- Exposing Grafana (guide)

**Which issue is resolved by this Pull Request:**
Prerequisite for https://github.com/scylladb/scylla-operator/issues/2933.

PR #3005, opened 2025-10-16
Extend monitoring docs with external Prometheus mode

czeslavo:
**Description of your changes:** Extends ScyllaDBMonitoring documentation with External Prometheus mode.

Resolves https://github.com/scylladb/scylla-operator/issues/2934, https://github.com/scylladb/scylla-operator/issues/2933

### DIFF
--- a/docs/source/installation/gitops.md
+++ b/docs/source/installation/gitops.md
@@ -43,21 +43,6 @@ for i in {1..30}; do
 done
 :::
 
-#### Prometheus Operator
-
-:::{code-block} shell
-:substitutions:
-kubectl -n=prometheus-operator apply --server-side -f=https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/third-party/prometheus-operator.yaml
-:::
-
-:::{code-block} shell
-# Wait for CRDs to propagate to all apiservers.
-kubectl wait --for='condition=established' crd/prometheuses.monitoring.coreos.com crd/prometheusrules.monitoring.coreos.com crd/servicemonitors.monitoring.coreos.com
-
-# Wait for prometheus operator deployment.
-kubectl -n=prometheus-operator rollout status --timeout=10m deployment.apps/prometheus-operator
-:::
-
 ### {{productName}}
 
 Once you have the dependencies installed and available in your cluster, it is the time to install {{productName}}.
@@ -204,6 +189,10 @@ kubectl -n=scylla-manager apply --server-side -f=https://raw.githubusercontent.c
 # Wait for it to deploy.
 kubectl -n=scylla-manager rollout status --timeout=10m deployment.apps/scylla-manager
 :::
+
+### Monitoring stack
+
+Please refer to the [ScyllaDB Monitoring setup](../management/monitoring/setup.md) guide to learn how to configure the monitoring stack.
 
 ## Next steps
 
--- a/docs/source/management/monitoring/external-prometheus-on-openshift.md
+++ b/docs/source/management/monitoring/external-prometheus-on-openshift.md
@@ -1,0 +1,123 @@
+# Setting up ScyllaDB Monitoring on OpenShift
+
+This guide will walk you through setting up a monitoring stack for your ScyllaDB clusters using the
+[`ScyllaDBMonitoring`](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) custom resource and an
+external Prometheus instance that is already deployed in your Kubernetes cluster in an OpenShift cluster using [User Workload Monitoring (UWM)][user-workload-monitoring].
+
+The guide assumes you have read the [overview](overview.md) and [setup](setup.md) of ScyllaDB monitoring and are familiar with the concepts of Prometheus and Grafana.
+
+## Requirements
+
+This guide assumes you have {{productName}} and a `ScyllaCluster` already installed in your OpenShift cluster.
+For more information on how to deploy {{productName}}, see [the installation guide](../../installation/overview.md).
+
+:::{note}
+The {{productName}} installation process on OpenShift is the same as on vanilla Kubernetes. However, unlike Kubernetes,
+OpenShift includes a built-in Prometheus Operator and [User Workload Monitoring (UWM)][user-workload-monitoring] for user
+workloads. Therefore, instead of deploying Prometheus using ScyllaDBMonitoring, we configure it to use the external
+Prometheus instance provided by OpenShift UWM.
+:::
+
+We also assume you have the `oc` CLI tool installed and configured to access your OpenShift cluster, and have the necessary
+permissions to create `ServiceAccounts` and `ClusterRoleBindings`.
+
+## Enable User Workload Monitoring in OpenShift
+
+OpenShift provides a built-in Prometheus instance that can be used for monitoring user workloads.
+To use this Prometheus instance, you need to enable User Workload Monitoring in your OpenShift cluster.
+You can do this by following the [official OpenShift documentation][user-workload-monitoring].
+
+## Configure OpenShift metrics access for Grafana datasource
+
+### Create ServiceAccount and ClusterRoleBinding
+
+To allow `ScyllaDBMonitoring`-managed Grafana to access the OpenShift User Workload Monitoring Prometheus instance,
+you need to create a `ServiceAccount` and a `ClusterRoleBinding` that grants the necessary permissions. We will use its
+`ServiceAccount` token for configuring Grafana datasource. See the OpenShift's [Accessing metrics as a developer](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/monitoring/accessing-metrics#accessing-metrics-as-a-developer) 
+article for more details.
+
+:::{note}
+We assume you have `ScyllaCluster` deployed in `scylla` namespace/project. Replace `scylla` with your namespace/project name if it's different.
+:::
+
+You can create the `ServiceAccount` and `ClusterRoleBinding` using the following commands:
+
+```shell
+oc create serviceaccount scylla-grafana-monitoring-viewer
+oc create clusterrolebinding scylla-monitoring-grafana-cluster-monitoring-view --clusterrole=cluster-monitoring-view --serviceaccount=scylla:scylla-grafana-monitoring-viewer
+```
+
+### Create ServiceAccount token Secret
+
+Next, you need to create a `Secret` that contains the `ServiceAccount` token. The following manifest will create such a `Secret`:
+
+:::{literalinclude} ../../../../examples/monitoring/v1alpha1/openshift/sa-token.secret.yaml
+:language: yaml
+:::
+
+You can create it using the following command:
+
+:::{code-block} shell
+:substitutions:
+kubectl apply -n scylla -f https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/monitoring/v1alpha1/openshift/sa-token.secret.yaml
+:::
+
+The `Secret` will be populated with the token automatically by Kubernetes and should be available under the `token` key of this `Secret`. Verify it by running:
+
+```shell
+kubectl -n scylla get secret scylla-monitoring-grafana-token -o=jsonpath='{.data.token}'
+```
+
+You should see the encoded token printed to the console.
+
+## Create service CA certificate ConfigMap
+
+OpenShift uses a self-signed CA to sign the certificates for its internal services. To allow Grafana to trust the OpenShift UWM Prometheus instance,
+you need to create a `ConfigMap` that, when properly annotated, will be populated with the OpenShift service CA certificate.
+Please refer to the [OpenShift documentation](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/security_and_compliance/certificate-types-and-descriptions#cert-types-service-ca-certificates)
+for more details on this mechanism.
+
+The following manifest will create such a `ConfigMap`:
+
+:::{literalinclude} ../../../../examples/monitoring/v1alpha1/openshift/service-ca.configmap.yaml
+:language: yaml
+:::
+
+You can create it using the following command:
+
+:::{code-block} shell
+:substitutions:
+kubectl apply -n scylla -f https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/monitoring/v1alpha1/openshift/service-ca.configmap.yaml
+:::
+
+You can verify that the `ConfigMap` has been populated with the service CA certificate under the `service-ca.crt` key by running:
+
+```shell
+kubectl -n scylla get configmap example-openshift-service-ca -o=jsonpath='{.data.service-ca\.crt}'
+```
+
+You should see the PEM-encoded CA certificate printed to the console.
+
+## Deploy ScyllaDBMonitoring
+
+The following `ScyllaDBMonitoring` configuration will set up the monitoring stack to use the OpenShift UWM Prometheus instance
+as an external Prometheus datasource for Grafana:
+
+:::{literalinclude} ../../../../examples/monitoring/v1alpha1/openshift/uwm.scylladbmonitoring.yaml
+:language: yaml
+:::
+
+You can apply it using `kubectl`:
+
+:::{code-block} shell
+:substitutions:
+kubectl apply -n scylla --server-side -f=https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/monitoring/v1alpha1/openshift/uwm.scylladbmonitoring.yaml
+:::
+
+See the [Setting up ScyllaDBMonitoring](setup.md#deploy-scylladbmonitoring) guide for more details on deploying `ScyllaDBMonitoring`.
+
+## Verify the setup
+
+You can verify that configuration is correct by [accessing Grafana](exposing-grafana.md) and verifying you can see metrics from your ScyllaDB cluster.
+
+[user-workload-monitoring]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/monitoring/configuring-user-workload-monitoring
--- a/docs/source/management/monitoring/overview.md
+++ b/docs/source/management/monitoring/overview.md
@@ -2,55 +2,84 @@
 
 ## Architecture
 
-ScyllaDB [exposes](https://monitoring.docs.scylladb.com/stable/reference/monitoring-apis.html) its metrics in Prometheus format. 
-{{productName}} leverages this capability in Kubernetes environments to provide a comprehensive monitoring solution for your ScyllaDB clusters.
-{{productName}} supports deploying and managing a complete monitoring stack for your ScyllaDB clusters using the
-[ScyllaDBMonitoring](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) custom resource.
+ScyllaDB [exposes](https://monitoring.docs.scylladb.com/stable/reference/monitoring-apis.html) its metrics in the Prometheus format. 
+{{productName}} provides the [`ScyllaDBMonitoring`](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) custom resource
+that allows you to set up a complete monitoring stack for your ScyllaDB clusters based on the following components:
 
-:::{note}
-ScyllaDBMonitoring CRD is still experimental. The API is currently in version `v1alpha1` and may change in future versions.
-:::
-
-The monitoring stack that can be deployed with ScyllaDBMonitoring includes the following components:
-- Prometheus for metrics collection and alerting (along with scraping and alerting rules targeting ScyllaDB instances).
-- Grafana for metrics visualization (with pre-configured dashboards for ScyllaDB).
+- [**Prometheus**](https://prometheus.io) for metrics collection and alerting (along with scraping and alerting rules targeting ScyllaDB instances).
+- [**Grafana**](https://grafana.com) for metrics visualization (with pre-configured dashboards for ScyllaDB).
 
 :::{include} diagrams/monitoring-overview.mmd
 :::
 
-The monitoring stack is expected to be created separately for each ScyllaDB cluster, allowing you to have isolated monitoring environments.
+:::{note}
+The `ScyllaDBMonitoring` CRD is still in its `v1alpha1` version, yet it is considered stable and ready for production use, with
+the following caveats:
+- `spec.components.grafana.exposeOptions` and `spec.components.prometheus.exposeOptions` are deprecated and will be removed in the next API version,
+- the **Managed** mode is deprecated; the **External** mode for Prometheus is likely to be the only supported mode in the next API version.
+:::
 
-## Third-party dependencies
+## Prometheus
 
-{{productName}} builds on top of the stable and widely adopted open-source projects to provide a robust monitoring solution.
+For deploying and/or configuring Prometheus, {{productName}} relies on [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator).
 
-### Prometheus Operator
+{{productName}} supports two modes of operation for `ScyllaDBMonitoring` regarding Prometheus deployment:
+**Managed** and **External**. You can choose the mode that best fits your needs by setting the `spec.components.prometheus.mode` field in the `ScyllaDBMonitoring` resource.
 
-For deploying and/or configuring Prometheus, {{productName}} relies on the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator).
-Depending on configuration, {{productName}} may deploy and manage a Prometheus instance for you, or it can be configured to use
-an existing Prometheus instance (managed by the Prometheus Operator) in your cluster. 
+Depending on the mode chosen, {{productName}} may deploy and manage a Prometheus instance for you, or it can be configured to use
+an existing Prometheus instance (managed by the Prometheus Operator) in your cluster.
 
-The following Prometheus Operator resources are created by the {{productName}} when you deploy a ScyllaDBMonitoring resource:
+The following Prometheus Operator resources are created by {{productName}} when you deploy a `ScyllaDBMonitoring` resource:
 
-- [Prometheus](https://github.com/prometheus-operator/prometheus-operator/blob/e4c727291acc543dab531bc4aaf16637067c1b86/pkg/apis/monitoring/v1/prometheus_types.go#L1085) - the Prometheus instance itself.
-- [ServiceMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/e4c727291acc543dab531bc4aaf16637067c1b86/pkg/apis/monitoring/v1/servicemonitor_types.go#L41) - the resource that defines how to scrape metrics from ScyllaDB nodes.
-- [PrometheusRule](https://github.com/prometheus-operator/prometheus-operator/blob/e4c727291acc543dab531bc4aaf16637067c1b86/pkg/apis/monitoring/v1/prometheusrule_types.go#L37) - the resource that defines alerting rules for Prometheus.
-  
-Prometheus version used in the deployment is tied to the version of {{productName}}. You can find the exact version used in the
-[config.yaml](https://github.com/scylladb/scylla-operator/blob/master/assets/config/config.yaml) file under `operator.prometheusVersion` key.
+- [`Prometheus`](https://github.com/prometheus-operator/prometheus-operator/blob/e4c727291acc543dab531bc4aaf16637067c1b86/pkg/apis/monitoring/v1/prometheus_types.go#L1085) - the Prometheus instance itself (it may be omitted in External mode).
+- [`ServiceMonitor`](https://github.com/prometheus-operator/prometheus-operator/blob/e4c727291acc543dab531bc4aaf16637067c1b86/pkg/apis/monitoring/v1/servicemonitor_types.go#L41) - the resource that defines how to scrape metrics from ScyllaDB nodes.
+- [`PrometheusRule`](https://github.com/prometheus-operator/prometheus-operator/blob/e4c727291acc543dab531bc4aaf16637067c1b86/pkg/apis/monitoring/v1/prometheusrule_types.go#L37) - the resource that defines alerting rules for Prometheus.
 
-### Grafana
+The Prometheus version used in the deployment is tied to the version of {{productName}}. You can find the exact version used in the
+[`config.yaml`](https://github.com/scylladb/scylla-operator/blob/master/assets/config/config.yaml) file under `operator.prometheusVersion` key.
+
+### External
+
+The **External** mode plugs into an existing Prometheus that is managed by Prometheus Operator (and therefore can be configured by `ServiceMonitor`).
+
+In the **External** mode, {{productName}} will not deploy a `Prometheus` resource (will instead expect a running Prometheus in the cluster already), but it will still create the `ServiceMonitor` and `PrometheusRule` resources
+that an existing Prometheus Operator in your cluster will use to configure your Prometheus instance. This mode is useful if you already have a Prometheus instance
+deployed in your cluster, and you want to use it for monitoring your ScyllaDB clusters. If you don't have Prometheus deployed in your cluster, you need to deploy one to proceed.
+
+When using this mode, you need to ensure that the existing Prometheus instance is configured to discover and scrape the 
+`ServiceMonitor` and `PrometheusRule` resources created by {{productName}}. Please refer to the [Setting up ScyllaDB Monitoring](setup.md#deploy-prometheus-instance) guide for more details.
+
+Please note that in this mode, `ScyllaDBMonitoring` has to be configured so that Grafana can access the Prometheus instance.
+You can configure Grafana datasources in the `spec.components.grafana.datasources` field of the `ScyllaDBMonitoring` resource.
+Please refer to the [`ScyllaDBMonitoring` API reference](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) for details.
+
+### Managed
+
+:::{note}
+This mode is deprecated and will be removed in a future version. Instead, please deploy your own Prometheus and use **External**.
+:::
+
+In the **Managed** mode, {{productName}} will deploy a Prometheus instance for you. This is the default mode.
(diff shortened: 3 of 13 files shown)


docs/source/installation/gitops.md
```diff
@@ -58,6 +58,10 @@ kubectl wait --for='condition=established' crd/prometheuses.monitoring.coreos.co
 kubectl -n=prometheus-operator rollout status --timeout=10m deployment.apps/prometheus-operator
 :::
 
+:::{note}
+See the [ScyllaDB Monitoring setup](../management/monitoring/setup.md) guide to learn how to configure the monitoring stack.
```
czeslavo: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429193214.

docs/source/management/monitoring/overview.md
```diff
@@ -3,54 +3,81 @@
...
-## Third-party dependencies
+You should create one ScyllaDBMonitoring resource per ScyllaDB cluster.
 
-{{productName}} builds on top of the stable and widely adopted open-source projects to provide a robust monitoring solution.
+:::{note}
+ScyllaDBMonitoring CRD is still in `v1alpha1` version, yet it is considered stable and ready for production use, with
+the following caveats:
+- `spec.components.grafana.exposeOptions` and `spec.components.prometheus.exposeOptions` are deprecated and will be removed in the next API version.
+- **External** mode for Prometheus is likely to be the only supported mode in the next API version.
+:::
```
czeslavo, line 20: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429216842.

docs/source/management/monitoring/overview.md
```diff
@@ -3,54 +3,81 @@
 ## Architecture
 
 ScyllaDB [exposes](https://monitoring.docs.scylladb.com/stable/reference/monitoring-apis.html) its metrics in Prometheus format. 
-{{productName}} leverages this capability in Kubernetes environments to provide a comprehensive monitoring solution for your ScyllaDB clusters.
-{{productName}} supports deploying and managing a complete monitoring stack for your ScyllaDB clusters using the
-[ScyllaDBMonitoring](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) custom resource.
+{{productName}} provides [ScyllaDBMonitoring](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) custom resource
+that allows you to set up a complete monitoring stack for your ScyllaDB clusters based on the following components:
```
czeslavo: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429234981.

docs/source/management/monitoring/overview.md
```diff
@@ -3,54 +3,81 @@
...
 
 :::{include} diagrams/monitoring-overview.mmd
 :::
 
-The monitoring stack is expected to be created separately for each ScyllaDB cluster, allowing you to have isolated monitoring environments.
-
-## Third-party dependencies
+You should create one ScyllaDBMonitoring resource per ScyllaDB cluster.
 
-{{productName}} builds on top of the stable and widely adopted open-source projects to provide a robust monitoring solution.
```
czeslavo, line 25: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429251303.

docs/source/management/monitoring/overview.md
```diff
@@ -3,54 +3,81 @@
...
+
+Please note that in this mode, ScyllaDBMonitoring has to be configured so that Grafana can access the Prometheus instance.
+You can configure Grafana datasources in the `spec.components.grafana.datasources` field of the ScyllaDBMonitoring resource.
+Please refer to the [ScyllaDBMonitoring API reference](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst) for details.
+
+## Grafana
 
 For deploying Grafana, {{productName}} doesn't use any third-party operator. Instead, it manages the Grafana deployment
-directly. It makes sure that Grafana is pre-configured with dashboards from [scylla-monitoring](https://github.com/scylladb/scylla-monitoring/).
+directly. It preconfigures Grafana with dashboards from [scylla-monitoring](https://github.com/scylladb/scylla-monitoring/).
```
czeslavo, line 71: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429618066.

docs/source/management/monitoring/overview.md
```diff
@@ -3,54 +3,81 @@
...
 
-An Ingress Controller can be used to expose the Grafana outside the Kubernetes cluster.
-{{productName}} does not manage the Ingress Controller itself, but it creates an [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resource to expose Grafana
-according to the configuration in the ScyllaDBMonitoring resource. You can use any Ingress Controller of your choice.
+- Port forwarding using `kubectl port-forward` command for temporary access.
+- Using [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) or [Gateway API](https://gateway-api.sigs.k8s.io/)
+  resources (e.g., HTTPRoute) for production access.
 
-You can learn more about exposing Grafana in the [ScyllaDB Monitoring setup](exposing-grafana.md) guide.
+You can learn more about exposing Grafana in the [Exposing Grafana](exposing-grafana.md) guide.
```
czeslavo, line 85: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429642837.

I changed the "Ingress-centric" description to the one showing wider options, focusing on the `ClusterIP` Service that the Operator creates as an integration point.

docs/source/management/monitoring/setup.md
```diff
@@ -3,16 +3,22 @@
...
 
+The guide assumes you have read the [overview](overview.md) of ScyllaDB monitoring and are familiar with the concepts of Prometheus and Grafana.
+It doesn't cover every possible configuration option, but it will highlight the most important ones. For a complete reference of all configuration options, 
+see the [ScyllaDBMonitoring API reference](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst).
+
 ## Requirements
 
-Before you can set up your ScyllaDB monitoring, you need {{productName}} already installed in your Kubernetes cluster.
+Before you can set up your ScyllaDB monitoring, you need {{productName}} and a ScyllaCluster already installed in your Kubernetes cluster.
 For more information on how to deploy {{productName}}, see [the installation guide](../../installation/overview.md).
```
czeslavo: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429658680.

docs/source/management/monitoring/setup.md
```diff
@@ -3,16 +3,22 @@
...
+### Deploy Prometheus Operator
+
+:::{note}
 ScyllaDBMonitoring relies on the Prometheus Operator for managing Prometheus-related resources.
 You can deploy it in your Kubernetes cluster using the provided third-party examples. If you already have it deployed
-in your cluster, you can skip the below steps.
-
-### Deploy Prometheus Operator
+in your cluster, you can skip this step.
+:::
```
czeslavo, line 21: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429660811.

docs/source/management/monitoring/setup.md
```diff
@@ -27,11 +33,82 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
 kubectl -n prometheus-operator rollout status --timeout=5m deployments.apps/prometheus-operator
 ```
 
-## Deploy ScyllaDBMonitoring
+## Configure ScyllaDBMonitoring
 
 You can see an example of a complete ScyllaDBMonitoring in this {{'[example manifest](https://raw.githubusercontent.com/{}/{}/examples/monitoring/v1alpha1/scylladbmonitoring.yaml)'.format(repository, revision)}}. 
 
-Make a local copy of it and update the `endpointsSelector` in the example with a label matching your ScyllaCluster instance name.
+You can make a local copy of it (e.g., under `./scylladbmonitoring.yaml`) and modify it to fit your needs. We'll go through the most important options below.
```
czeslavo: Continuation of https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429665104.

docs/source/management/monitoring/overview.md
```diff
@@ -3,54 +3,79 @@
...
-:::
-
-The monitoring stack that can be deployed with ScyllaDBMonitoring includes the following components:
 - Prometheus for metrics collection and alerting (along with scraping and alerting rules targeting ScyllaDB instances).
 - Grafana for metrics visualization (with pre-configured dashboards for ScyllaDB).
 
 :::{include} diagrams/monitoring-overview.mmd
 :::
 
-The monitoring stack is expected to be created separately for each ScyllaDB cluster, allowing you to have isolated monitoring environments.
```
czeslavo, line 21: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429248379.

docs/source/management/monitoring/setup.md
```diff
@@ -42,12 +120,10 @@ Scylla Operator will notice the new ScyllaDBMonitoring object, and it will recon
 
 ### Wait for ScyllaDBMonitoring to roll out
 ```console
-kubectl wait --for='condition=Progressing=False' scylladbmonitorings.scylla.scylladb.com/example
-kubectl wait --for='condition=Degraded=False' scylladbmonitorings.scylla.scylladb.com/example
 kubectl wait --for='condition=Available=True' scylladbmonitorings.scylla.scylladb.com/example
 ```
 
-### Wait for Prometheus to roll out
```
czeslavo, line 50: As discussed on a call, this is required because ScyllaDBMonitoring controller doesn't wait for the underlying infra to be actually available. Added an explicit note on it.

docs/source/management/monitoring/setup.md
```diff
@@ -42,12 +120,16 @@ Scylla Operator will notice the new ScyllaDBMonitoring object, and it will recon
 
 ### Wait for ScyllaDBMonitoring to roll out
 ```console
-kubectl wait --for='condition=Progressing=False' scylladbmonitorings.scylla.scylladb.com/example
```
czeslavo, line 45: Addresses https://github.com/scylladb/scylla-operator/pull/2993#discussion_r2429671885.

Added timeout and dropped Progressing/Degraded waits.

docs/source/management/monitoring/setup.md
czeslavo: Tracking issue: https://github.com/scylladb/scylla-operator/issues/3007

docs/source/management/monitoring/setup.md
rzetelskik: Progressing/Degraded/Available is a common API shared between our resources - what's wrong with waiting for all three? I imagine service can be available but the controller can still be progressing (not sure about degrsded)?

docs/source/installation/gitops.md
mflendrich: 1. The `kubectl apply`  and `kubectl wait` are duplicated here and in the monitoring setup guide. Should we remove from the gitops guide and just point to what the monitoring one says?
2. This step should probably be clustered with other extra components (like manager and CSI driver) at the end  of the instructions, no? (suggestion: move closer to the aforementioned two)

mflendrich:
Pushed some (hopefully uncontroversial :crossed_fingers:) editorial fixes directly to the branch. Continuing the review.

docs/source/management/monitoring/external-prometheus-on-openshift.md
```diff
@@ -0,0 +1,111 @@
...
+external Prometheus instance that is already deployed in your Kubernetes cluster in an OpenShift cluster using User Workload Monitoring (UWM).
+
+The guide assumes you have read the [overview](overview.md) and [setup](setup.md) of ScyllaDB monitoring and are familiar with the concepts of Prometheus and Grafana.
+It will focus on the external Prometheus configuration options. For a complete reference of all configuration options,
+see the [`ScyllaDBMonitoring` API reference](../../api-reference/groups/scylla.scylladb.com/scylladbmonitorings.rst).
+
+## Requirements
+
+This guide assumes you have {{productName}} and a `ScyllaCluster` already installed in your OpenShift cluster.
+For more information on how to deploy {{productName}}, see [the installation guide](../../installation/overview.md).
```
abramche, line 12: Since there is no explicit mention of OpenShift [on the page](../../installation/overview.md), I would add a small section how OpenShift is different from barebones k8s (insignificant outside of monitoring, for practical use cases?).

czeslavo: It's a good point, and I think we won't be able to explicitly claim support for OpenShift until we get the certification. Also, we won't be able to claim such support on arbitrary machines, so it will have to be properly specified. @mflendrich, what do you think? Are we going to support OpenShift only on AWS (the configuration we actually tested)?

docs/source/management/monitoring/setup.md
czeslavo: Shouldn't `Available` be enough in this case, where we intend to wait for availability specifically?

docs/source/installation/gitops.md
czeslavo: Moved it to the bottom as the `Monitoring stack` section, dropping the duplicated instructions.

docs/source/management/monitoring/external-prometheus-on-openshift.md
abramche: Note that there are two issues: 
- comment inside the PR (OpenShift absent from "supported kubernetes platforms")
- and the inline comment (Requirement of having OpenShift+ScyllaDB set up but the reference to the installation guide is moot as it misses OpenShift entirely - even though it does not differ from regular kubernetes outside of monitoring stack setup)

While we may omit OpenShift from the supported platforms (but why? should we not add it as a "alpha/beta" support?), the second issue is relevant in any case - otherwise we refer to documentation which explicitly misses this particular instruction.

See it like this: "You need an OpenShift x ScyllaDB -> if you don't have it, see instructions -> [instructions missing for OpenShift]"

We do not need to hold this PR off with this - can decouple this as support is still experimental.

mflendrich: > While we may omit OpenShift from the supported platforms (but why? should we not add it as a "alpha/beta" support?)

We _implicitly_ include OpenShfit in the list of supported platforms:

> Supported Kubernetes platforms
> 
> We officially test and recommend to use the following platforms:
>
> ... (AWS, GKE)
> 
> While our APIs generally work on any Kubernetes conformant cluster, performance tuning and other pieces that need to interact with the host OS, kubelet, CRI, kernel, etc. might hit some incompatibilities.

I think that this is fair statement regarding the degree of support we have for OpenShift today (imply that it works because it's kubernetes but we do not do anything special for it; some testing exists, but is limited). It will make sense to advertise OpenShift as supported once two things happen:
- operator is certified and published via operator bundle,
- openshift as a platform is fully tested on par with EKS and GKE.

We are not there yet (however it's roadmapped), so the problem we're solving here is: knowing that someone got Operator and Scylla installed on OpenShift (based on the general kubernetes compatibility "promise"), we help them set up monitoring specifically where they have the setup that is specific to OpenShift.

I do not understand the installation point - the instructions are expected to work on openshift just as well as anywhere else.

That said - I think that the best thing to do for this PR is what @abramche suggested in the first comment: _add a small section how OpenShift is different from barebones k8s_ - locally within this guide, assuming that "the remainder", including installation, is nothing special for openshift and the general k8s instructions should be followed.

czeslavo: I've added a note explaining it.

docs/source/management/monitoring/setup.md
rzetelskik: It could, but given https://github.com/scylladb/scylla-operator/issues/3007 I suppose it's not the case?

docs/source/management/monitoring/overview.md
```diff
@@ -2,55 +2,86 @@
...
+
+:::{include} diagrams/monitoring-overview.mmd
+:::
 
 :::{note}
-ScyllaDBMonitoring CRD is still experimental. The API is currently in version `v1alpha1` and may change in future versions.
+The `ScyllaDBMonitoring` CRD is still in its `v1alpha1` version, yet it is considered stable and ready for production use, with
+the following caveats:
+- `spec.components.grafana.exposeOptions` and `spec.components.prometheus.exposeOptions` are deprecated and will be removed in the next API version,
+- the **Managed** mode is deprecated; the **External** mode for Prometheus is likely to be the only supported mode in the next API version.
```
rzetelskik, line 19: Huh? When has that been decided? Why is the managed prometheus supposed to be removed?

docs/source/management/monitoring/setup.md
```diff
@@ -27,27 +33,103 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
 kubectl -n prometheus-operator rollout status --timeout=5m deployments.apps/prometheus-operator
 ```
 
-## Deploy ScyllaDBMonitoring
+## Configure `ScyllaDBMonitoring`
```
rzetelskik: nit: The inlined code block in a heading looks very bad imo

image

docs/source/management/monitoring/overview.md
```diff
@@ -2,55 +2,86 @@
...
+The **External** mode plugs into an existing Prometheus that is managed by Prometheus Operator (and therefore can be configured by `ServiceMonitor`).
+
+In the **External** mode, {{productName}} will not deploy a `Prometheus` resource (will instead expect a running Prometheus in the cluster already), but it will still create the `ServiceMonitor` and `PrometheusRule` resources
+that an existing Prometheus Operator in your cluster will use to configure your Prometheus instance. This mode is useful if you already have a Prometheus instance
+deployed in your cluster, and you want to use it for monitoring your ScyllaDB clusters. If you don't have Prometheus deployed in your cluster, you need to deploy one to proceed.
+
+When using this mode, you need to ensure that the existing Prometheus instance is configured to discover and scrape the 
+`ServiceMonitor` and `PrometheusRule` resources created by {{productName}}. This requires setting the appropriate
+`serviceMonitorSelector` and `ruleSelector` in the `Prometheus` resource to match the labels used by {{productName}} (or
+leaving them empty to select all resources).
```
rzetelskik: ok, but how do I do this? It could use an example or a link to the guide

docs/source/management/monitoring/setup.md
```diff
@@ -1,19 +1,25 @@
 # ScyllaDB Monitoring setup
```
rzetelskik:
```suggestion
# Setting up ScyllaDB Monitoring
```
If it's a guide now, let's communicate it's a guide, otherwise I don't immediately understand the difference between this and the overview - setup/overview look roughly the same in the index

docs/source/management/monitoring/external-prometheus-on-openshift.md
```diff
@@ -0,0 +1,120 @@
+# ScyllaDB Monitoring on OpenShift
```
rzetelskik:
```suggestion
# Setting up ScyllaDB Monitoring on OpenShift
```

ditto

docs/source/management/monitoring/setup.md
```diff
@@ -27,27 +33,103 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
 kubectl -n prometheus-operator rollout status --timeout=5m deployments.apps/prometheus-operator
 ```
 
-## Deploy ScyllaDBMonitoring
+## Configure `ScyllaDBMonitoring`
+
+You can see an example of a complete `ScyllaDBMonitoring` in this {{'[example manifest](https://raw.githubusercontent.com/{}/{}/examples/monitoring/v1alpha1/scylladbmonitoring.yaml)'.format(repository, revision)}}.
```
rzetelskik: this example uses 1) managed mode and 2) expose options, which are supposedly being deprecated - should we update it?

docs/source/management/monitoring/setup.md
```diff
@@ -27,27 +33,103 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
 kubectl -n prometheus-operator rollout status --timeout=5m deployments.apps/prometheus-operator
 ```
 
-## Deploy ScyllaDBMonitoring
+## Configure `ScyllaDBMonitoring`
+
+You can see an example of a complete `ScyllaDBMonitoring` in this {{'[example manifest](https://raw.githubusercontent.com/{}/{}/examples/monitoring/v1alpha1/scylladbmonitoring.yaml)'.format(repository, revision)}}. 
+
+You can make a local copy of it (e.g., under `./scylladbmonitoring.yaml`) and modify it to fit your needs. We'll go through the most important options below.
```
rzetelskik: > You can make a local copy of it (e.g., under `./scylladbmonitoring.yaml`) and modify it to fit your needs.

it's kind of painfully obvious

docs/source/management/monitoring/setup.md
```diff
@@ -27,27 +33,103 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
...
+Depending on your Prometheus configuration, you may also need to configure TLS (under `tls`) and authentication (under `authentication`).
+
+::::
+:::::
+
+## Deploy `ScyllaDBMonitoring`
 
 Deploy the monitoring setup using kubectl:
 ```console
 kubectl apply -n scylla --server-side -f=./scylladbmonitoring.yaml
```
rzetelskik: if I didn't make the local copy, which this PR made optional, this file doesn't exist

docs/source/management/monitoring/setup.md
```diff
@@ -27,27 +33,103 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
...
+::::
+
+::::{group-tab} External mode
+
+In the External mode, you need to ensure that your existing Prometheus instance is configured to discover and scrape the
+ServiceMonitor and PrometheusRule resources created by `ScyllaDBMonitoring`. This requires setting the
+`serviceMonitorSelector` and `ruleSelector` in the `Prometheus` resource to match the labels used by `ScyllaDBMonitoring`
+(or leaving them empty to select all resources).
+
+You also need to configure Grafana so that it can access the `Prometheus` instance. This will be explained in the Grafana configuration section below.
```
rzetelskik: This feels more like the "overview" section repeated - it says what you need to do but doesn't exactly show how. It could use some examples (e.g. similar to the yaml snippet above, or just links for two different in-tree examples?). Without concrete examples or commands, it's basically redundant to overview/api reference and doesn't fit a guide.

docs/source/management/monitoring/setup.md
```diff
@@ -27,27 +33,103 @@ kubectl apply -n prometheus-operator --server-side -f=https://raw.githubusercont
...
+You can set resource requests and limits (under `resources`), persistency (under `storage`), and workload placement (under `placement`).
+
+::::
+
+::::{group-tab} External mode
+
+In the External mode, you need to ensure that your existing Prometheus instance is configured to discover and scrape the
+ServiceMonitor and PrometheusRule resources created by `ScyllaDBMonitoring`. This requires setting the
+`serviceMonitorSelector` and `ruleSelector` in the `Prometheus` resource to match the labels used by `ScyllaDBMonitoring`
+(or leaving them empty to select all resources).
```
rzetelskik: nit: this is an edge case of "match the labels", I'd remove it as it only makes it more convoluted imo

docs/source/management/monitoring/overview.md
czeslavo: I initially made this statement a bit more blurry, saying something along "we _might_ deprecate managed mode in the future". @mflendrich changed the wording in his review push to the current wording for the sake of direct communication with users (instead of a mild _might_ statement, we declare the mode as deprecated, external is the recommended way to go). 

I didn't want to deprecate it already, but after an offline discussion with Michal, I'm convinced it might make sense. We aim to simplify our product as much as possible. Having multiple ways of doing the same thing is not a desired state in the long run if we want to keep the project maintainable.

Let's discuss it and make a decision.

mflendrich: Discussed offline with both of you. My point is that the cost of making a mistake here isn't high; if users bump into issues with the external mode, they always have the option to go with the managed mode for the time being. And keeping one endorsed way as a clear default will leave the user with one less decision to make (=easier to use).

Also, we always have the option to un-deprecate.

@rzetelskik with your permission I suggest that we keep the direct "deprecation" wording.

docs/source/management/monitoring/setup.md
mflendrich: This is on me. I added this, I'll fix.

docs/source/management/monitoring/external-prometheus-on-openshift.md
```diff
@@ -0,0 +1,120 @@
...
+
+:::{note}
+The {{productName}} installation process on OpenShift is the same as on vanilla Kubernetes. However, unlike Kubernetes,
+OpenShift includes a built-in Prometheus Operator and [User Workload Monitoring (UWM)][user-workload-monitoring] for user
+workloads. Therefore, instead of deploying Prometheus using ScyllaDBMonitoring, we configure it to use the external
+Prometheus instance provided by OpenShift UWM.
+:::
+
+We also assume you have the `oc` CLI tool installed and configured to access your OpenShift cluster, and have the necessary
+permissions to create `ServiceAccount`s and `ClusterRoleBinding`s.
```
rzetelskik:
```suggestion
permissions to create `ServiceAccounts` and `ClusterRoleBindings`.
```
nit

docs/source/management/monitoring/external-prometheus-on-openshift.md
```diff
@@ -0,0 +1,120 @@
...
+spec:
+  components:
+    grafana:
+      datasources:
+        - type: Prometheus
+          url: "https://thanos-querier.openshift-monitoring.svc:9091" # OpenShift exposes UWM Prometheus via Thanos Querier.
+          prometheusOptions:
+            tls:
+              caCertConfigMapRef:
+                name: openshift-service-ca.crt                        # OpenShift provides a ConfigMap with the service CA certificate.
```
rzetelskik: following the changes to E2E tests this needs to be updated, no?

docs/source/management/monitoring/setup.md
czeslavo: It's been updated in the implementation PR (exposeOptions have been dropped). It's already on master: https://github.com/scylladb/scylla-operator/blob/master/examples/monitoring/v1alpha1/openshift-uwm.scylladbmonitoring.yaml

docs/source/management/monitoring/setup.md
mflendrich: Pushed. (@czeslavo please squash as you please - I didn't want to force push)

docs/source/management/monitoring/setup.md
czeslavo: The intention here was to have an anchor for later commands (an example file name). I've changed it so it explicitly commands to create a local copy for clarity and further commands' validity.

docs/source/management/monitoring/setup.md
czeslavo: As explained in https://github.com/scylladb/scylla-operator/pull/3005#discussion_r2469556555.

docs/source/management/monitoring/setup.md
czeslavo: Agreed. Fixed by making the overview section less specific and extending the guide with a YAML example and a note.

docs/source/management/monitoring/setup.md
czeslavo: I think it's still worth mentioning as an edge case. Moved it to a separate note to make the main section less convoluted.

docs/source/management/monitoring/overview.md
czeslavo: Added an example in the guide and linked to it.

docs/source/management/monitoring/setup.md
czeslavo: Fixed it along with other comments.

docs/source/management/monitoring/setup.md
czeslavo: Waiting for `Progressing` and `Degraded` wouldn't help either. That's why we have two separate `kubectl rollout status` for Prometheus and Grafana below.

czeslavo:
I hope I've addressed all the comments. PTAL @rzetelskik @mflendrich 

Edit: Please note that the `Exposing Grafana` guide is being reworked in https://github.com/scylladb/scylla-operator/pull/3014 (as a result of `exposeOptions` deprecation).

mflendrich (approved):
lgtm
/assign rzetelskik

docs/source/management/monitoring/setup.md
rzetelskik: The e2e tests still use all three conditions, how do we know we're not regressing here?

docs/source/management/monitoring/setup.md
rzetelskik: We should have a command for making the local copy in this case (afaiu the intention is that, when following a guide, a user can just copy-paste and run the commands) , although imo it would be preferred to use the url throughout the guide

docs/source/management/monitoring/setup.md
rzetelskik: The generic example is only updated in https://github.com/scylladb/scylla-operator/pull/3014, no?

docs/source/management/monitoring/setup.md
rzetelskik: let's resolve this in https://github.com/scylladb/scylla-operator/pull/3005#discussion_r2454663977

docs/source/management/monitoring/setup.md
rzetelskik: And even then, it still uses the managed type - should we have one with external type outside of openshift?

docs/source/management/monitoring/setup.md
rzetelskik: Btw this also fails with
```
kubectl wait --for='condition=Available=True' scylladbmonitorings.scylla.scylladb.com/example
Error from server (NotFound): scylladbmonitorings.scylla.scylladb.com "example" not found
```
We should use unify the namespace usage - previous command creates ScyllaDBMonitoring in `scylla` namespace. Same thing for prometheus STS and grafana deployment.

czeslavo: Fixed the namespace issue. We're checking Grafana deployment status anyway in the next step; we won't regress here.

docs/source/management/monitoring/setup.md
czeslavo: Reworked the guide to be external Prometheus-centric and updated the examples.

docs/source/management/monitoring/setup.md
czeslavo: Reworked the guide to allow the copy-paste flow.

docs/source/management/monitoring/external-prometheus-on-openshift.md
czeslavo: Fixed.

docs/source/management/monitoring/external-prometheus-on-openshift.md
czeslavo: Added missing steps.

examples/monitoring/v1alpha1/scylladbmonitoring.yaml
```diff
@@ -4,31 +4,23 @@ metadata:
   name: example
 spec:
   type: Platform
+# [endpointsSelector]
   endpointsSelector:
     matchLabels:
       app.kubernetes.io/name: scylla
       scylla-operator.scylladb.com/scylla-service-type: member
-      scylla/cluster: replace-with-your-scyllacluster-name
+      scylla/cluster: scylladb
```
rzetelskik: nit: should we change it to "scylla" to work with our ScyllaCluster example and unify with `openshift/uwm.scylladbmonitoring.yaml`?

mflendrich: -1 to this change - it will steer the attention away from the fact that this is a configurable field.

`scylla/cluster: scylla` means nothing to a user who doesn't already know this annotation.

rzetelskik: Yes, but with `scylla/cluster: scylla` the manifests work without editing by default. And that's the assumption in this guide, that it work by copy-pasting the commands.

rzetelskik: The alternative to this is to revert to adding a command making a local copy and applying it.

mflendrich: Compromise:

```yaml
    scylla/cluster: scylladb # replace with your ScyllaCluster object name
```

rzetelskik: This is explained in the guide. But my comment is about the workflow in the guide, not about this being unclear. The command says
```
Deploy the monitoring using kubectl:

kubectl apply -n scylla --server-side -f=https://raw.githubusercontent.com/{{repository}}/{{revision}}/examples/monitoring/v1alpha1/scylladbmonitoring.yaml
```

This won't work unless it's either set up to work with our examples or there is a command instructing the users to make a local copy.

czeslavo: > nit: should we change it to "scylla" to work with our ScyllaCluster example and unify with openshift/uwm.scylladbmonitoring.yaml?

Good catch. Let's make it so :+1: 

> -1 to this change - it will steer the attention away from the fact that this is a configurable field.

The main motivation behind it was making the manifests work by default, as we want to have a guide "copy-paste-execute" flow. I think we can satisfy both needs by leaving an explicit comment on the `scylla/cluster` annotation. It's also explained in the guide itself. PTAL

image

docs/source/management/monitoring/external-prometheus-on-openshift.md
rzetelskik: @abramche I'm assuming this is resolved. If not, we'll resolve this in follow-up PRs.

rzetelskik:
@abramche we'll merge this and resolve further comments, if any, in follow-up PRs.
/hold cancel
/lgtm

Issue #3001, opened 2025-10-14
Deprecate exposeOptions in ScyllaDBMonitoring components


czeslavo:
### Description

In the initial implementation, `ScyllaDBMonitoring`  got `exposeOptions` fields both for Prometheus and Grafana components. The only supported option in those is Ingress. When set, it results in Operator creating an `Ingress` resource :

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: "{{ .scyllaDBMonitoringName }}-grafana"
  annotations:
    {{- .ingressAnnotations | toYAML | nindent 4 }}
spec:
  ingressClassName: {{ or .ingressClassName "null" }}
  rules:
  {{- range $_, $dnsDomain := .dnsDomains }}
  - host: "{{ $dnsDomain }}"
    http:
      paths:
      - backend:
          service:
            name: "{{ $.scyllaDBMonitoringName }}-<grafana/prometheus>"
            port:
              number: 3000
        path: /
        pathType: Prefix
  {{- end }}
```

Its use is limited as there's no way to configure an `Ingress` that will terminate TLS, which likely is the most common approach in production.

The goal of this issue is to deprecate `exposeOptions` for Grafana and Prometheus, effectively reducing the ScyllaDBMonitoring controller's scope of responsibility, simplifying the product. A more pragmatic way forward will be to document that both components get Services created for them and leave the exposure of them up to the user preferences (Ingress, Gateway API, or any other way).

### Acceptance criteria

- `exposeOptions` fields are documented as deprecated and to be removed in the next API version
- Monitoring documentation (i.e., articles describing Grafana exposure) is updated to guide users towards exposing Services on their own using their preferred approach (Ingress or Gateway API)

czeslavo:
Updated https://github.com/scylladb/scylla-operator/issues/2959 to account for fields drop during promotion.

PR #3014, opened 2025-10-22
Deprecate exposeOptions in ScyllaDBMonitoring

czeslavo:
**Description of your changes:** Deprecates `exposeOptions` in Grafana and Prometheus components of `ScyllaDBMonitoring`.

Resolves https://github.com/scylladb/scylla-operator/issues/3001.

### DIFF
--- a/bundle/manifests/scylla.scylladb.com_scylladbmonitorings.yaml
+++ b/bundle/manifests/scylla.scylladb.com_scylladbmonitorings.yaml
@@ -171,8 +171,11 @@ spec:
                         maxItems: 1
                         type: array
                       exposeOptions:
-                        description: exposeOptions specifies options for exposing
-                          Grafana UI.
+                        description: |-
+                          exposeOptions specifies options for exposing Grafana UI.
+
+                          Deprecated: This field will be removed in the next version of the API. Support for it will be removed in the future
+                          versions of the operator. We recommend managing your own Ingress or HTTPRoute resources to expose Grafana if needed.
                         properties:
                           webInterface:
                             description: webInterface specifies expose options for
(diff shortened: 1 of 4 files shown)


docs/source/management/monitoring/exposing-grafana.md
```diff
@@ -26,65 +27,85 @@ Wait for HAProxy Ingress to roll out:
...
+    http:
+      paths:
+      - backend:
+          service:
+            name: "example-grafana"
+            port:
+              number: 3000
+        path: /
+        pathType: Prefix 
+```
```
rzetelskik: should we have an in-tree example so that we can provide a command with a url as well (and test it, I suppose - if we don't, this guide will sooner or later be broken)?

docs/source/management/monitoring/exposing-grafana.md
```diff
@@ -26,65 +27,85 @@ Wait for HAProxy Ingress to roll out:
...
 ```console
-$ INGRESS_IP="$( kubectl -n=haproxy-ingress get service/haproxy-ingress --template='{{ ( index .status.loadBalancer.ingress 0 ).ip }}' )"
+INGRESS_IP="$( kubectl -n haproxy-ingress get svc haproxy-ingress --template '{{ index .status.loadBalancer.ingress 0 "ip" }}' )"
+INGRESS_PORT="443"
 ```
 
-#### Ingress NodePort
-
-NodePort is slightly less convenient, but it's available in development clusters as well.
+Otherwise, if you're running this locally (e.g. using `minikube` or `kind`), you can port-forward the Ingress controller service to your local machine:
```
rzetelskik, line 78: I don't understand this change - how's that an improvement compared to the existing example? And how is it related to dropping expose options?

docs/source/management/monitoring/exposing-grafana.md
czeslavo: I've extracted this manifest to an example file and added an explicit command to apply it with a URL. Regarding testing strategy, I've added an agenda item for our next team meeting, as that's a wider topic to discuss.

docs/source/management/monitoring/exposing-grafana.md
czeslavo: As we've discussed offline, this is to make the guide simpler and not explain all the nitty-gritty details of ingress handling, as that's not our domain.

rzetelskik (approved):
/lgtm
Thanks!

Issue #3009, opened 2025-10-16
Gracefully handle missing HostID during bootstrap


mflendrich:
During bootstrap of a Scylla node, sidecar repeatedly bumps into

https://github.com/scylladb/scylla-operator/blob/86956a49178e91b8ed390ce63de4d461e36f50ed/pkg/controller/sidecar/sync.go#L202

that results in an ugly error-level print to the logs

```
E1015 15:46:07.333513       1 sidecar/controller.go:156] "Unhandled Error" err="syncing key 'NAMESPACE/NODENAME' failed: can't sync the HostID annotation: local host ID \"HOST_ID_UUID_GOES_HERE\" not found in IP to hostID mapping: map[]" logger="UnhandledError"
```
.

This situation is part of the expected flow as a node becomes bootstrapped, so logging it as unhandled (and especially at the error level) confuses users debugging other issues.

## Acceptance Criteria
- [ ] Either do not log this condition at all (because it's part of a normal workflow), or log it in a way that clearly states that this is expected before the node is bootstrapped.

mflendrich:
It is important to mention that it is the **inner error** `local host ID \"HOST_ID_UUID_GOES_HERE\" not found in IP to hostID mapping` that is expected as normal in the bootstrap flow. Probably any other reason is not expected and should still be loud.

PR #3018, opened 2025-10-23
Gracefully handle hostID missing in IP to HostID mapping

czeslavo:
**Description of your changes:** Instead of repeatedly logging scary "Unhandled error" logs, let's make this message more approachable and make it clear it's an expected transient state.

Resolves https://github.com/scylladb/scylla-operator/issues/3009.

### DIFF
--- a/pkg/controller/sidecar/sync.go
+++ b/pkg/controller/sidecar/sync.go
@@ -139,7 +139,11 @@ func (c *Controller) syncAnnotations(ctx context.Context, svc *corev1.Service) e
 	}
 
 	if len(localIP) == 0 {
-		return fmt.Errorf("local host ID %q not found in IP to hostID mapping: %v", hostID, ipToHostIDMap)
+		// It most likely means that the Scylla node the sidecar is running on has not joined the cluster yet. We need
+		// to requeue and try again later.
+		klog.V(2).InfoS("The node with the expected HostID has not joined the cluster yet, will retry in a bit", "HostID", hostID, "IPToHostIDMap", ipToHostIDMap)
+		c.queue.AddRateLimited(c.key)
+		return nil
 	}
 
 	nodeTokens, err := scyllaClient.GetNodeTokens(ctx, localhost, localIP)


pkg/controller/sidecar/controller.go
```diff
@@ -152,6 +154,13 @@ func (c *Controller) processNextItem(ctx context.Context) bool {
 	case apierrors.IsAlreadyExists(err):
 		klog.V(2).InfoS("Hit already exists, will retry in a bit", "Key", key, "Error", err)
 
+	case errors.As(err, &hostIDNotFoundErr):
+		// Let's make this message not scary as this is expected during node bootstrap.
+		klog.V(2).InfoS("The node with the expected HostID has not joined the cluster yet, will retry in a bit",
```
rzetelskik: are we sure this is always the reason? Instead of adding a special case on the controller level, why can't we handle it in the sync function? We already have access to the API at this point, can't we just check if the node has joined the cluster and NOT return an error if it's not a part of the cluster yet (not sure how - I guess state NORMAL is enough but I might be wrong, you'd have to ask the cluster team)? 

On a side note, I think we could also benefit from syncing the annotations independently, e.g. I don't see why we couldn't set the HostID annotation while we're still waiting for the node to become a part of the ring.

mflendrich: I wouldn't go as far as building an elaborate state of the world where what we're trying to do is just sync the HostID annotation that is an established mechanism that works.

The point of this change is to make it obvious to the reader that this is an expected transient state, and that it's only not okay if it's never left.

(I do not have an opinion about structuring the code one way or another)

czeslavo: I intended to make as few changes as possible here to not break the existing mechanism. The ideas for separating annotations syncing are great, but I think we should save them for later and focus here on fixing the problem, which are scary error logs.

I didn't think about calling `queue.AddRateLimited` from inside the `sync` function. It seems a bit hacky, and in general, I think we're missing a good way of early returning from `sync` with an intention to retry (a second return value could help here, similar to controller-runtime's [Result](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result)). Anyway, it indeed seems like a more direct way of fixing this, so I've updated the proposed fix to not use custom errors.

rzetelskik: As discussed internally, this is preferred given any changes to the logic could result in a regression and we're approaching a release. However, we've confirmed with the cluster team that the HostID shall have an IP mapping if 1) the API is available and 2) the node is in state normal. So we could consider reflecting this in the logic. Let's add trackers for both adjusting the logic and setting the annotations independently as nice-to-haves.

czeslavo: I've created trackers:
- https://github.com/scylladb/scylla-operator/issues/3020
- https://github.com/scylladb/scylla-operator/issues/3021

rzetelskik: Thanks!

Issue #3023, opened 2025-10-27
Fix logging of the version number


mflendrich:
## Actual Behavior

The "operator version" log upon startup prints the version tag in a misleading way.

For 1.18.1 it prints:

```
I1021 16:33:34.359041       1 operator/operator.go:272] operator version "v1.18.0-26-g071684a"
```

which means "commit `071684a` which is 26 commits after the`v1.18.0` tag". This is misleading because the printed version tag does not indicate the actual tag running - it always shows the closest ancestor tag annotated by additional commits appended on top of it, instead.

For 1.18.0 it prints `1.18.0-beta.2-6-ge48f5ce`.

## Expected Behavior

- For released builds, the prints should include the actual tag being run (`1.18.1` for 1.18.1, `1.18.0` for 1.18.0, etc.) and **never** include the predecessor tag.

## Proposed Solution

Sadly -- the existing release process (that reuses the same pre-release build and promotes it until GA without embedding a version number) is tying our hands here, so we need to work around it, or rebuild the process (which in turn would require reproducibility of builds).

Requirements:

- Drop the `operator version` print in its current form
- Instead, print the Git SHA and the correct semver for the build running.

Instead of relying on the semver from the "predecessor tag", operator could use the semver is available from the `OPERATOR_IMAGE` env var that the user is expected to set to match the `spec.containers[].image` value. (it's not possible to use the downward API for this field instead of a manually set env var, unfortunately). Operator already uses `OPERATOR_IMAGE` to provision the sidecar in the matching version, among other things.

For this reason, the proposal is that the log:
- prints the Git SHA of the build running
- prints the value of the `OPERATOR_IMAGE` env
- defends about a situation where the user sets `OPERATOR_IMAGE` wrong --  by logging it in a way that clearly states that it's a value from the env var, and it's on the user to ensure that it matches.

## Alternatives Considered
- Bake the semver constant into the image. This will how ever require a rebuild of the existing process that relies on promoting a pre-release image without recompiling.

rzetelskik:
> defends about a situation where the user sets OPERATOR_IMAGE wrong -- by logging it in a way that clearly states that it's a value from the env var, and it's on the user to ensure that it matches.

Besides it being incorrect, what's even the point of it then? if I understand correctly it wouldn't even be extracted from the image manifests' labels, but you would expect a tagged image reference with a tag matching semver. It's like asking a user to set an OPERATOR_VERSION env var and printing it with extra steps.

[...]

PR #3024, opened 2025-10-27
Log git commit on commands run

czeslavo:
**Description of your changes:** Replaces all logs printing git tag/version with an exact git commit sha. Drops parts of the build-time-injection machinery that are unused.

Resolves https://github.com/scylladb/scylla-operator/issues/3023


Makefile
```diff
@@ -13,7 +13,6 @@ MAKE_REQUIRED_MIN_VERSION:=4.2 # for SHELLSTATUS
 # Support container build from git worktrees where the parent git folder isn't available.
 GIT ?=git
 
-GIT_TAG ?=$(shell [ ! -d ".git/" ] || $(GIT) describe --long --tags --abbrev=7 --match 'v[0-9]*')$(if $(filter $(.SHELLSTATUS),0),,$(error $(GIT) describe failed))
 GIT_TAG_SHORT ?=$(shell [ ! -d ".git/" ] || $(GIT) describe --tags --abbrev=7 --match 'v[0-9]*')$(if $(filter $(.SHELLSTATUS),0),,$(error $(GIT) describe failed))
 GIT_COMMIT ?=$(shell [ ! -d ".git/" ] || $(GIT) rev-parse --short "HEAD^{commit}" 2>/dev/null)$(if $(filter $(.SHELLSTATUS),0),,$(error $(GIT) rev-parse failed))
 GIT_TREE_STATE ?=$(shell ( ( [ ! -d ".git/" ] || $(GIT) diff --quiet ) && echo 'clean' ) || echo 'dirty')
```
rzetelskik: Is `GIT_TREE_STATE` used anywhere after this change?

pkg/cmd/version/version.go
```diff
@@ -70,6 +70,6 @@ func (o *Options) Complete() error {
 }
 
 func (o *Options) Run(originalStreams genericclioptions.IOStreams, cmd *cobra.Command) error {
-	fmt.Printf("%s: %s\n", cmd.Name(), version.Get())
+	fmt.Printf("%s: GitCommit=%s\n", cmd.Name(), build.GitCommit())
```
mflendrich: Why is it sometimes `fmt.Printf` and sometimes `klog.InfoS`?

What about stdout/stderr?

pkg/build/build.go
```diff
@@ -0,0 +1,10 @@
+package build
+
+// commitFromGit is a constant representing the source version that
+// generated this build. It should be set during build via -ldflags.
+var commitFromGit string
```
mflendrich, line 5: Doesn't `-X` only work with exported variables? :thinking:

mflendrich: I mean -- I cannot find a notion of that in the documentation, but (1) that is my experience from the past, (2) I can see how the compiler could optimize away an unexported variable before the linker has its say.

pkg/cmd/version/version.go
czeslavo: It's always `klog.InfoS` with this one exception: "version" command. This is expected to print to stdout I believe.

Makefile
czeslavo: No, it isn't, overlooked this one. Thanks.

pkg/build/build.go
czeslavo: Apparently not, as it works as expected (and it's worked so far - I just reused what was used before). But indeed, looking for documentation on whether exported or non-exported variables are supported, I cannot find anything specific.

mflendrich: Okay. I think that the "it seems to work" conversation could be easily settled by a test, though 😅.

A test would also help find regressions in the future. My experience with such brittle settings is that any refactors (either on the build system side, or moving packages around) easily lead to silent failures.

Suggestion: if it's feasible, implement a regression test verifying that the variable contains something reasonable.

czeslavo: PTAL, I've added a `test-binary` Makefile target that should be enough of a regression catcher.

rzetelskik: for the record, the `test` target is not being run by any job

czeslavo: Good catch, let's add a job for that then: https://github.com/scylladb/scylla-operator-release/pull/509

czeslavo: I've also dropped the dead makefile targets that you spotted @rzetelskik.

Makefile
```diff
@@ -723,24 +718,18 @@ test-unit:
...
-	./hack/lib/tag-from-gh-ref.sh
-.PHONY: test-scripts
-
-test: test-unit test-scripts
+test: test-unit test-binary
 .PHONY: test
 
+test-binary: build
+	# Verify that the built binary runs and prints the expected version (commit hash):
+	./scylla-operator version | grep $(GIT_COMMIT)
```
rzetelskik: Should we compare the entire output instead of grepping?

rzetelskik: also a nit: test-version would probably be more fitting to what this target does now

czeslavo: We can :+1: PTAL

czeslavo: > also a nit: test-version would probably be more fitting to what this target does now

I think `test-binary` better reflects the high-level intention here - it's meant to test whether a binary we built is correct (e.g., it returns a correct git commit that's injected at build time). If you don't mind, I prefer to leave it this way.

Issue #3004, opened 2025-10-15
[Documentation] Dead links in README.md


abramche:
The GKE and EKS links in the ["ScyllaDB Installation:"](https://github.com/scylladb/scylla-operator/blob/master/README.md#quickstarts-and-training) section are dead - do we still need these sections?

rzetelskik:
> do we still need these sections?

No, the entire file is very much outdated. Most of the content could be replaced with a reference to the documentation to limit the maintenance surface (edit: it's actually already there, so we should just remove the duplicates and whatever is outdated).

/priority important-longterm
/triage accepted

mflendrich:
Added to the bag of small docs improvements in 1.19.

PR #3026, opened 2025-10-28
Refresh project readme to rely more on docs

czeslavo:
**Description of your changes:** Drops dead links, shifts to linking more to official docs.

Resolves https://github.com/scylladb/scylla-operator/issues/3004.


rzetelskik:
It's an opinionated statement, but I'd push it even further. IMO the README should aggregate our resources without going into details that will get outdated sooner or later. Should we just link to docs without listing particular subjects, which are already present in the docs' index? I'd also get rid of the "Features" section, we never update it either way.

czeslavo:
@rzetelskik I share your feelings about the direction we should take with our readme. I made more drastic cuts, PTAL if that's what you had on your mind.

rzetelskik:
/uncc @mflendrich
As he's OOO tomorrow. We can always revert.
/lgtm

Issue #2624, opened 2025-05-06
Procedure for upgrading prometheus-operator


mflendrich:
We've discovered that bumping prometheus-operator requires multiple manual steps, copying Go code & YAML from the prometheus-operator repo, including manual reorganization/changes. This is not efficient and not well documented.

The scope of this issue is to get this process in a good shape (at least documented as a list of steps, possibly automated with a script).

## Acceptance criteria
- [ ] Basic version: prometheus operator upgrade is a list of steps
- [ ] Stretch version: the list of steps is 1 step that is running a script 😄

Issue #2926, opened 2025-09-01
Restructure third-party dependencies in GitOps and Helm installation guides


rzetelskik:
To lower the maintenance burden, we should move away from maintaining in-tree examples of third-party dependencies and only link to external procedures and provide the versions we test our stack against.

### Prerequisites:
- [x] https://github.com/scylladb/scylla-operator/issues/2913

### Acceptance criteria:
- [ ] GitOps installation guide does not provide third-party dependency installation commands and manifests but links to external instructions instead. External links should have version injected from assets/metadata.yaml.
- [ ] Helm installation guide does not provide third-party dependency installation commands and manifests but links to external instructions instead. External links should have version injected from assets/metadata.yaml.
- [ ] The installation guides should mention the versions we test against.
- [ ] Supported external dependency versions should be specified in the support matrix:
  - https://github.com/scylladb/scylla-operator/pull/3333

/kind cleanup
/priority important-soon
/triage accepted

mflendrich:
Now tracked as [OPERATOR-40](https://scylladb.atlassian.net/browse/OPERATOR-40) in the internal Jira.

[OPERATOR-40]: https://scylladb.atlassian.net/browse/OPERATOR-40?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3032, opened 2025-10-29
Generate Prometheus Operator example manifests

rzetelskik:
**Description of your changes:** Updating in-tree third-party Prometheus Operator is currently an awfully tedious process requiring a lot of backtracking and manual file operations. From my experience it takes hours. I also see no point of keeping the file split.
This PR adds automation for upgrading it using a single source of truth and following the upstream instructions. It is a step towards https://github.com/scylladb/scylla-operator/issues/2926. I want to avoid discussion around the docs before 1.19 though, so I'm keeping the in-tree examples for now.

Resolves https://github.com/scylladb/scylla-operator/issues/2624

/kind machinery
/priority important-soon
/cc

### DIFF
--- a/examples/third-party/prometheus-operator.yaml
+++ b/examples/third-party/prometheus-operator.yaml
@@ -1,3 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: prometheus-operator
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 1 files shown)


rzetelskik:
/cc czeslavo
The verify target is going to fail until kustomize reaches the golang-1.25 image, but other than that this is ready to review.

rzetelskik:
/hold cancel
fixed

Issue #3042, opened 2025-10-31
Exclude secrets from must-gather


mflendrich:
The scope of this issue is to change must-gather's default behavior so that it does not collect `Secret` objects (whether it's name, metadata, key names or key values).

### Proposed solution

- delete the [`collectSecret()`](https://github.com/scylladb/scylla-operator/blob/b37b4a3ba214c2dda7fa3af2e3c042b742e40f22/pkg/gather/collect/collect.go#L111) mechanism altogether
- implement a (disabled by default) mechanism that only captures namespace/name and keys (no other metadata like `kubectl.kubernetes.io/last-applied-configuration` which may contain secret values!)
- add a (default false) flag `include-secrets` that enables the namespace/name/keys collection mechanism with a very visible warning in the logs
- remove the "secret redaction" mechanism description from the documentation and replace with the updated semantics

### Acceptance criteria
- [ ] must-gather not even attempting to read secrets unless `include-secrets` is set
- [ ] in `include-secrets`, no information other than namespace, name and key names is included
- [ ] an unmistakable warning is present in case of `include-secrets` being enabled

rzetelskik:
@mflendrich I think we could make the `include-secrets` flag name a bit clearer about only gathering their identifiers and a list of keys. Otherwise it makes it seem as if it would collect entire objects. I don't have a good alternative yet though.

rzetelskik:
Also, we should reconsider including the list of keys (or have two "modes"). AFAIU gathering only namespace/name vs also including keys requires different permissions (list vs get).

czeslavo:
I've edited acceptance criteria according to our discussion during the team sync meeting (we'll not collect secrets whatsoever).

PR #3046, opened 2025-11-03
Make must-gather not collect sensitive resources by default

czeslavo:
**Description of your changes:** We don't want to collect sensitive resources like Secrets or bitnami's SealedSecrets by default. It's possible to collect them by passing the `--include-sensitive-resources` flag. If a user needs to exclude more resources than the default ones, they can do so by passing `--exclude-resources`.

Covered in unit tests. Verified manually with a local cluster.

Resolves https://github.com/scylladb/scylla-operator/issues/3042.
Fixes: CUSTOMER-10

### DIFF
--- a/docs/source/support/must-gather.md
+++ b/docs/source/support/must-gather.md
@@ -1,12 +1,13 @@
 # Gathering data with must-gather
 
 `must-gather` is an embedded tool in Scylla Operator that helps collecting all the necessary info when something goes wrong.
-
 The tool talks to the Kubernetes API, retrieves a predefined set of resources and saves them into a folder in your current directory.
-By default, all collected Secrets are censored to avoid sending sensitive data.
-That said, you can always review the archive before you attach it to an issue or your support request.
 
-Given it needs to talk to the Kubernetes API, at the very least, you need to supply the `--kubeconfig` flag with a path to the kubeconfig file for your Kubernetes cluster, or set the `KUBECONFIG` environment variable.
+By default, sensitive resources (`Secrets` and `bitnami.com.SealedSecrets`) are omitted from the collection.
+If you come across a clear need to include them (not recommended unless the default settings are deemed insufficient for investigation), please refer to the [include sensitive resources](#including-sensitive-resources) section below.
+If you have other resources that you wouldn't like to include in the collection, refer to the [excluding resources](#excluding-resources) section below.
+
+Given that `must-gather` needs to talk to the Kubernetes API, at the very least, you need to supply the `--kubeconfig` flag with a path to the kubeconfig file for your Kubernetes cluster, or set the `KUBECONFIG` environment variable.
 
 ## Running must-gather
 
@@ -98,4 +99,26 @@ You can also request collecting every resource in the Kubernetes API, if the def
 
 ```bash
 scylla-operator must-gather --all-resources
+```
+
+### Including sensitive resources
+
+By default, `must-gather` omits sensitive resources like `Secrets` or `bitnami.com.SealedSecrets`. You can turn this
+behavior off by passing the `--include-sensitive-resources` flag.
+
+```bash
+scylla-operator must-gather --all-resources --include-sensitive-resources
+```
+
+### Excluding resources
+
+If you have resources that you don't want to include in the collection other than the default ones, you can exclude them
+by passing the `--exclude-resource` flag multiple times.
+
+:::{note}
+The format for the resource is `kind` (for core resources) or `kind.group`. Examples: `LimitRange` or `DeviceClass.resource.k8s.io`.
+:::
+
+```bash
+scylla-operator must-gather --all-resources --exclude-resource="LimitRange" --exclude-resource="DeviceClass.resource.k8s.io"
 ```


pkg/gather/collect/collect_predicates.go
```diff
@@ -0,0 +1,10 @@
+package collect
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// IgnoreSecretsResourcePredicateFunc is a ResourcePredicateFunc that ignores Secret resources.
+func IgnoreSecretsResourcePredicateFunc(gv string, r *metav1.APIResource) bool {
+	return !(r.Kind == "Secret" && gv == "v1")
```
mflendrich: Is there a reason why the ignored resources are not configurable by a flag?

I can see how this could easily be a StringSlice flag of `gv:k` strings.

For example:

`--exclude-types v1:Secret --exclude-types bitnami.com/v1alpha1:SealedSecret`

I have "invented" the syntax with `:` on the spot without much thought - there might be a better way.

pkg/gather/collect/collect_predicates.go
```diff
@@ -0,0 +1,10 @@
+package collect
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// IgnoreSecretsResourcePredicateFunc is a ResourcePredicateFunc that ignores Secret resources.
+func IgnoreSecretsResourcePredicateFunc(gv string, r *metav1.APIResource) bool {
```
mflendrich: I suggest that we log the exclusions (and store them within the archive).

In combination with my other comment: writing the command line arguments into the log might be an option. Or dumping the options object?

czeslavo: What do you mean by "store _them_ within the archive"? Would you like us to list secrets anyway and list their namespace/name? Or do you mean us to list resources that we skipped (i.e., `v1.Secret`)?

pkg/gather/collect/collect_predicates.go
mflendrich: I can see how there might be a problem with multiversion CRDs - so maybe filter only the G and K, and not on the V? What version does must-gather download today? The storage version?

pkg/gather/collect/collect_predicates.go
mflendrich: This:

> list resources that we skipped (i.e., v1.Secret)?

Precisely speaking - put in the must_gather.log file that is within the created archive (or whatever its name is) a remark equivalent to: `Resource kinds ignored: [core Secret] [bitnami.com SealedSecret]`

pkg/gather/collect/collect_predicates.go
czeslavo: > Is there a reason why the ignored resources are not configurable by a flag?

Not really - I understood the task was about making sure `corev1.Secrets` are not collected in any circumstances. If you believe we need to make this configurable, it's doable, of course.

If it comes to the API, how would you like this to behave specifically? I can see two possible modes:

1. `--exclude-resources` by default contains `Secret`, and there's no way to turn this off. The user can only add more types to exclude (pros: there's no way to make any mistake and collect secrets).
2. It's possible to turn the Secrets collection on by setting some explicit flag (I believe we agreed that we do not want to allow this?) 

> I can see how there might be a problem with multiversion CRDs - so maybe filter only the G and K, and not on the V? What version does must-gather download today? The storage version?

Yes, group and kind should be enough. Currently, we discover resources with `ServerPreferredResources`, which returns the supported resources with the version preferred by the server.

czeslavo:
I've posted a refactor that excludes `Secret` and `SealedSecret.bitnami.com` by default (not allowing collecting them under any circumstances), and allows filtering out arbitrary other resources if needed with `--exclude-resources`. I'm keeping it as a separate [commit](https://github.com/scylladb/scylla-operator/pull/3046/commits/61bf87d45460903f1261d53e68c799eeec4a62f2) for review purposes. PTAL. Sample run:

```shell
$ go run ./cmd/scylla-operator must-gather --kubeconfig ~/kind.kubeconfig --dest-dir ./artifacts --exclude-resources=ConfigMap,Deployment.apps --exclude-resources=NonExistingResource.notexistingroup
I1104 21:49:22.138650 1149471 operator/gatherbase.go:123] "Created destination directory" Path="./artifacts"
I1104 21:49:22.138686 1149471 cmdutil/helpers.go:106] "Starting must-gather" GitCommit=""
I1104 21:49:22.138968 1149471 operator/mustgather.go:193] "Gathering artifacts" DestDir="./artifacts"
I1104 21:49:22.146206 1149471 collect/discoverer.go:160] "Warning: Resource to exclude from collection not found in the cluster, ensure you provided correct group and kind" group="notexistingroup" kind="NonExistingResource"
I1104 21:49:22.146220 1149471 collect/discoverer.go:64] "Excluding resource from collection" group="" kind="Secret"
I1104 21:49:22.146224 1149471 collect/discoverer.go:64] "Excluding resource from collection" group="bitnami.com" kind="SealedSecret"
I1104 21:49:22.146226 1149471 collect/discoverer.go:64] "Excluding resource from collection" group="" kind="ConfigMap"
I1104 21:49:22.146229 1149471 collect/discoverer.go:64] "Excluding resource from collection" group="apps" kind="Deployment"
I1104 21:49:22.538031 1149471 rest/warnings.go:110] "Warning: v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice"
I1104 21:49:22.549529 1149471 operator/mustgather.go:247] "Namespace not found" Namespace="scylla-manager"
I1104 21:49:22.550164 1149471 operator/mustgather.go:247] "Namespace not found" Namespace="scylla-operator-node-tuning"
I1104 21:49:22.550177 1149471 operator/mustgather.go:195] "Finished gathering artifacts" Duration="411.213962ms"
```

/hold for squash before merging

pkg/cmd/operator/gatherbase.go
```diff
@@ -63,6 +66,7 @@ func (o *GatherBaseOptions) AddFlags(flagset *pflag.FlagSet) {
 	flagset.Int64VarP(&o.LogsLimitBytes, "log-limit-bytes", "", o.LogsLimitBytes, "Maximum number of bytes collected for each log file, 0 means unlimited.")
 	flagset.BoolVarP(&o.CollectManagedFields, "managed-fields", "", o.CollectManagedFields, "Controls whether metadata.managedFields should be collected in the resource dumps.")
 	flagset.BoolVarP(&o.KeepGoing, "keep-going", "", o.KeepGoing, "Controls whether the collection should proceed to other resources over collection errors, accumulating errors.")
+	flagset.StringSliceVarP(&o.ExcludeResources, "exclude-resources", "e", []string{}, "List of resources to exclude from collection. Format: <kind>.<group>")
```
rzetelskik: nit: I wouldn't provide the shorthand for an option that's not likely to be used frequently

pkg/cmd/operator/gatherbase.go
```diff
@@ -126,6 +130,10 @@ func (o *GatherBaseOptions) Complete() error {
 		return fmt.Errorf("can't create destination directory %q: %w", o.DestDir, err)
 	}
 
+	for _, exclude := range o.ExcludeResources {
+		o.resourcesToExclude = append(o.resourcesToExclude, schema.ParseGroupKind(exclude))
```
rzetelskik, line 174: Q: this is case-sensitive, right? Should we mention it in the flag description?

pkg/gather/collect/discoverer.go
```diff
@@ -0,0 +1,191 @@
...
+	// Built-in Secrets.
+	{
+		Group: "",
+		Kind:  "Secret",
+	},
+	// Bitnami SealedSecrets.
+	{
+		Group: "bitnami.com",
+		Kind:  "SealedSecret",
+	},
```
rzetelskik: Q: is this API enough of a thing to justify including it here? First time I'm seeing it.

czeslavo: It seems SealedSecrets are quite common in the wild, and I thought it might be worth taking them into account by default. We'd never want to see them in our archives, so maybe it's better to take care of that for our users instead of ping-ponging with them in case they bump into an issue. These are my thoughts, happy to discuss. :)

pkg/cmd/operator/gatherbase.go
czeslavo: Fair enough, I don't see a high chance for exhaustion of available shorthand letters for this command, though. Dropped it.

pkg/cmd/operator/gatherbase.go
```diff
@@ -59,10 +62,11 @@ func NewGatherBaseOptions(gathererName string, keepGoing bool) *GatherBaseOption
...
-	flagset.BoolVarP(&o.CollectManagedFields, "managed-fields", "", o.CollectManagedFields, "Controls whether metadata.managedFields should be collected in the resource dumps.")
-	flagset.BoolVarP(&o.KeepGoing, "keep-going", "", o.KeepGoing, "Controls whether the collection should proceed to other resources over collection errors, accumulating errors.")
+	flagset.StringVar(&o.DestDir, "dest-dir", o.DestDir, "Destination directory where to store the artifacts.")
+	flagset.Int64Var(&o.LogsLimitBytes, "log-limit-bytes", o.LogsLimitBytes, "Maximum number of bytes collected for each log file, 0 means unlimited.")
+	flagset.BoolVar(&o.CollectManagedFields, "managed-fields", o.CollectManagedFields, "Controls whether metadata.managedFields should be collected in the resource dumps.")
+	flagset.BoolVar(&o.KeepGoing, "keep-going", o.KeepGoing, "Controls whether the collection should proceed to other resources over collection errors, accumulating errors.")
+	flagset.StringSliceVar(&o.ExcludeResources, "exclude-resources", []string{}, "List of resources to exclude from collection. Format: <kind>.<group>. Kind is expected to be in CamelCase (it's case sensitive).")
```
rzetelskik: nit: one thing I missed before, should we use o.ExcludeResources as the default (instead of inlined []string{})? And initialize it as part of the options constructor. Otherwise it's a bit all over the place with how the existing flags are initializied.

pkg/cmd/operator/gatherbase.go
rzetelskik: Actually, is there a good reason for it to be case sensitive?

pkg/gather/collect/collect_predicates.go
mflendrich: > Not really - I understood the task was about making sure `corev1.Secrets` are not collected in any circumstances. If you believe we need to make this configurable, it's doable, of course.

For clarity - this was a perfectly valid understanding; the idea to parameterize it came to my mind later. You're in the right, I'm just throwing a wrench into the works :grin: 

> If it comes to the API, how would you like this to behave specifically? I can see two possible modes:
> 
> 1. `--exclude-resources` by default contains `Secret`, and there's no way to turn this off. The user can only add more types to exclude (pros: there's no way to make any mistake and collect secrets).
> 
> 2. It's possible to turn the Secrets collection on by setting some explicit flag (I believe we agreed that we do not want to allow this?)

I don't have a favorite way of doing this; what was on my mind when I was writing this comment was that we could have a list of types we exclude from collection by default - that the user could modify. And the vision in my mind was a `StringSlice` with a default value set.

The degree of defensiveness here isn't of primary concern to me. I am seeing the below options:
1. a `StringSlice` flag `exclude-types` populated with secret by default (so that the user is free to exclude or include anything they want, including secrets)
2. same as (1), but with a flag validator that disallows exclusion of secrets
3. same as (2), but the validator being skippable with a flag like `--force-sensitive-values`

I do not feel strongly in favor of any of these approaches.

> 
> 
> > I can see how there might be a problem with multiversion CRDs - so maybe filter only the G and K, and not on the V? What version does must-gather download today? The storage version?
> 
> Yes, group and kind should be enough. Currently, we discover resources with `ServerPreferredResources`, which returns the supported resources with the version preferred by the server.

pkg/cmd/operator/gatherbase.go
czeslavo: We can choose either the resource name or the kind for the identification of a resource. Conventionally, kind is case sensitive. But yeah, we could make it case-insensitive. I don't think there's anything blocking us from it.

rzetelskik: I'm leaving this up to you then. Leaving it case-sensitive is fine as long as it's documented and I see it already is.

mflendrich: Depends on what we'd want to achieve by choosing case sensitivity or insensitivity?

From my PoV:
- we should expect the user to be very strict about what they include or exclude (because the cost of making a mistake here is high)
- the mechanism we implement here should be here to protect the user from archiving something they might not want to

which leads me to a conclusion that we might also consider the following option (scope creeping even further)

- compare the list against the equivalent of `kubectl api-resources`
- if the exclusion list includes types not there: fail with a message like `excluded type Foo.bar.com not a resource known to this Kubernetes cluster`

But this would in turn be incompatible with adding `SealedSecret` by default (because we can't expect all clusters to have it)

I do not have a clearly optimal approach in mind, posting this as a discussion point moreso than a specific request.

pkg/cmd/operator/gatherbase.go
czeslavo: How about doing it a bit differently and putting all the defaults in flag definitions? It seems counterintuitive to me to use the same variable for a default, and then for loading actual values from the CLI. It's also easier to grasp what the specific CLI flag is set to by default, having it this way.

pkg/cmd/operator/gatherbase.go
czeslavo: >  compare the list against the equivalent kubectl api-resources
    if the exclusion list includes types not there: fail with a message like excluded type Foo.bar.com not a resource known to this Kubernetes cluster

It's already done this way. I made it so there's a distinction between user-provided and default exclusions, and log the warning only for the user-provided ones.

See the `I1104 21:49:22.146206 1149471 collect/discoverer.go:160] "Warning: Resource to exclude from collection not found in the cluster, ensure you provided correct group and kind" group="notexistingroup" kind="NonExistingResource"` line from the example in the PR description.

pkg/cmd/operator/gatherbase.go
rzetelskik: From correctness perspective, unless you initialise in the constructor,  the struct has uninitialised fields before the AddFlags method is called. So I personally prefer it this way.

pkg/cmd/operator/gatherbase.go
mflendrich: Oh, I missed that. Should have read the code before commenting. :grin: 

I guess the discussion is settled in that case. resolving

pkg/gather/collect/collect_predicates.go
czeslavo: I'd be opting for _never_ allow Secrets collection. If we ever find a case that really needs that, we can revisit this. 

Alternatively, your 3rd option seems the most reasonable as it explicitly states that something sensitive should be included. Perhaps we could name it `--include-sensitive-resources` for symmetry with `--exclude-resources` and include in its description what resources we consider sensitive.

mflendrich: Sp the decision is to not go with my option 3, but to keep the secret and sealedsecret excluded unconditionally, right?

Completely fine with this.

pkg/cmd/operator/gatherbase.go
czeslavo: Fair enough.

pkg/gather/collect/collect_predicates.go
czeslavo: I liked the 3rd option, will post an update shortly.

czeslavo: @mflendrich PTAL

pkg/cmd/operator/gatherbase.go
```diff
@@ -49,20 +54,24 @@ func NewGatherBaseOptions(gathererName string, keepGoing bool) *GatherBaseOption
...
-	flagset.BoolVarP(&o.CollectManagedFields, "managed-fields", "", o.CollectManagedFields, "Controls whether metadata.managedFields should be collected in the resource dumps.")
-	flagset.BoolVarP(&o.KeepGoing, "keep-going", "", o.KeepGoing, "Controls whether the collection should proceed to other resources over collection errors, accumulating errors.")
+	flagset.StringVar(&o.DestDir, "dest-dir", o.DestDir, "Destination directory where to store the artifacts.")
+	flagset.Int64Var(&o.LogsLimitBytes, "log-limit-bytes", o.LogsLimitBytes, "Maximum number of bytes collected for each log file, 0 means unlimited.")
+	flagset.BoolVar(&o.CollectManagedFields, "managed-fields", o.CollectManagedFields, "Controls whether metadata.managedFields should be collected in the resource dumps.")
+	flagset.BoolVar(&o.KeepGoing, "keep-going", o.KeepGoing, "Controls whether the collection should proceed to other resources over collection errors, accumulating errors.")
+	flagset.StringSliceVar(&o.ExcludeResources, "exclude-resources", o.ExcludeResources, "List of resources to exclude from collection. Format: `<kind>.<group>`.")
```
mflendrich: Stepping into the user's shoes, I see three issues with this flag description:
1. It does not say how to include more than one resource. Might also consider whether to use `StringSlice` or `StringArray` to make it easy to use.
2. The format in the flag desc is incorrect for resources from the core group (`--exclude-resources ConfigMap`)
3. It does not say how this flag interacts with `include-sensitive-resources`. How do I make it collect something from the list? Provide in one of these? Both? What if I only specify in one but not in the other?

pkg/cmd/operator/gatherbase.go
```diff
@@ -166,3 +179,12 @@ func (o *GatherBaseOptions) GetPrinters() []collect.ResourcePrinterInterface {
...
 }
+
+// defaultExcludedSensitiveResourcesAsString returns a string representation of the default excluded sensitive resources for use in flag help.
+func defaultExcludedSensitiveResourcesAsString() string {
+	var ss []string
+	for _, gk := range collect.DefaultExcludedSensitiveResources() {
+		ss = append(ss, gk.String())
+	}
+	return strings.Join(ss, ", ")
+}
```
mflendrich, line 228: What about just `fmt.Sprintf("%v", collect.DefaultExcludedSensitiveResources())` instead of iterating?

From my quick experiment it calls `String()` on the slice objects which should be enough I suppose?

pkg/gather/collect/collect.go
```diff
@@ -357,3 +228,48 @@ func (c *Collector) CollectResources(ctx context.Context, resourceInfo *Resource
 
 	return apimachineryutilerrors.NewAggregate(errs)
 }
+
+func (c *Collector) CollectAllResources(ctx context.Context, resources []*ResourceInfo) error {
```
mflendrich: Does the name `CollectAllResources` still make sense?

How about `CollectAllDiscoveredResources` ?

pkg/cmd/operator/gatherbase.go
czeslavo: It's picking between `Controls whether sensitive resources (i.e., [Secret SealedSecret.bitnami.com]) should be collected. (default false)` vs `Controls whether sensitive resources (i.e., Secret, SealedSecret.bitnami.com) should be collected. (default false)`. I personally prefer the latter as I find it more human-readable.

pkg/gather/collect/collect.go
czeslavo: I initially named this `CollectAllResources` to avoid a clash with the existing `CollectResources` method. I think we can rename the original `CollectResources` to `CollectResourceObjects`, as what it does is list all objects of a given resource and collect them. Following this pattern, we can rename the discussed method to `CollectResourcesObjects` (where resources are the ones passed as an argument).

pkg/cmd/operator/gatherbase.go
czeslavo: I modified the flag to use `StringArray` and adjusted help strings for both flags to thoroughly explain the interaction.

```
      --exclude-resources stringArray   Comma-separated list of Kubernetes resources to exclude, in the format <kind>[.group].
                                        Kind matching is case-insensitive; omit the group only for core resources (e.g., "Pod,ConfigMap").
                                        By default, the following are always excluded, despite the value of this flag: Secret, SealedSecret.bitnami.com
                                        
                                        Specifying this flag adds to the default exclusions. For example:
                                          --exclude-resources=Pod,Deployment.apps
                                        
                                        To force-include resources that are excluded by default, use --include-sensitive-resources
      --include-sensitive-resources     Controls whether sensitive resources (Secret, SealedSecret.bitnami.com) should be collected. (default false)
                                        When this flag is set, you can still exclude selected ones using --exclude-resources.
                                        For example:
                                          --include-sensitive-resources --exclude-resources=Secret
                                        will only exclude Secrets.
```

mflendrich: > Comma-separated list of Kubernetes resources to exclude

I'm afraid that `StringArray` works the other way around:

> The value of each argument will not try to be separated by comma. Use a StringSlice for that. 

https://pkg.go.dev/github.com/spf13/pflag#StringArray

Suggested fix: _Kubernetes resource to exclude, in the format (...). Can be specified multiple times._
Also might make sense to drop the plural for this flag name (`exclude-resource`) if only one resource will follow every occurence.

Otherwise looks good.

czeslavo: Good catch, I read this documentation yesterday, and for some reason remembered it differently ("only allow a single occurrence of the flag and split by a comma"). Now it should be good.

pkg/cmd/operator/gatherbase.go
```diff
@@ -24,22 +26,42 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	excludeResourceFlagHelpFmt = `Kubernetes resource to exclude, in the format <kind>[.group]. Can be specified multiple times.
```
mflendrich: nit: inconsistency in typography; please pick one (I like the latter more)

```suggestion
	excludeResourceFlagHelpFmt = `Kubernetes resource to exclude, in the format <kind>[.<group>]. Can be specified multiple times.
```

```suggestion
	excludeResourceFlagHelpFmt = `Kubernetes resource to exclude, in the format kind[.group]. Can be specified multiple times.
```

pkg/cmd/operator/gatherbase.go
```diff
@@ -24,22 +26,42 @@ import (
...
-
-	DestDir              string
-	CollectManagedFields bool
-	LogsLimitBytes       int64
-	KeepGoing            bool
+	restConfig         *rest.Config
+	kubeClient         kubernetes.Interface
+	dynamicClient      dynamic.Interface
+	discoveryClient    discovery.DiscoveryInterface
+	resourcesToExclude []schema.GroupKind
```
mflendrich, line 78: `GatherBaseOptions` feels unnecessarily denormalized to me - `resourcesToExclude` is functionally dependent on `IncludeSensitiveResources` and `ExcludeResources`. For this reason keeping both in the same object feels wrong.

Also it is incorrect to write `resourcesToExclude` in `Validate()` which semantically should be a validator and should not be changing state.

suggestion: keep only the flags in this object; write a pure function that translates them `resourcesToExclude`; call this function in `Validate()` if necessary, and call it for the second time where it's actually evaluated.

pkg/cmd/operator/gatherbase.go
```diff
@@ -24,22 +26,42 @@ import (
...
+	excludeResourceFlagHelpFmt = `Kubernetes resource to exclude, in the format <kind>[.group]. Can be specified multiple times.
+Kind matching is case-insensitive; omit the group only for core resources (e.g., Pod, ConfigMap").
+By default, the following are always excluded, despite the value of this flag: %s
+
+Specifying this flag adds to the default exclusions. For example:
+  --exclude-resource=Pod --exclude-resource=Deployment.apps
+
+To force-include resources that are excluded by default, use --include-sensitive-resources`
+
+	includeSensitiveResourcesFlagHelp = `Controls whether sensitive resources (%s) should be collected. (default false)
```
mflendrich: missing `Fmt` suffix
also: I believe that the code style across the codebase uses `Format` instead of `Fmt`

pkg/cmd/operator/mustgather.go
```diff
@@ -40,21 +38,23 @@ var (
 		
 		# Collect archive of all resources present in the Kubernetes cluster.
 		scylla-operator must-gather --all-resources
+
+		# Collect archive of all resources present in the Kubernetes cluster,
+		# excluding ConfigMaps and Deployments.
+		scylla-operator must-gather --all-resources --exclude-resource=ConfigMap --exclude-resource=Deployment.apps
```
mflendrich: Actually this may be an unfortunate example because configmaps and deployments are among the most relevant resources that a must-gather can have :sweat_smile: 

nit - to avoid implying that exclusion of cm and deploy might be a good idea, use something more abstract and irrelevant. Examples: `LimitRange` in `core` and `DeviceClass`

pkg/gather/collect/discoverer.go
```diff
@@ -0,0 +1,222 @@
...
+	var filters []discovery.ResourcePredicateFunc
+	for _, gk := range resourcesToExclude {
+		filters = append(filters, ignoreResourceByGroupKind(gk))
+		klog.InfoS("Excluding resource from collection", "group", gk.Group, "kind", gk.Kind)
+	}
+
+	return filters
+}
+
+func replaceIsometricResourceInfosIfPresent(resourceInfos []*ResourceInfo) ([]*ResourceInfo, error) {
```
mflendrich, line 132: I suspect that this is dead functionality now:

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#event-v125

I wouldn't necessarily make it part of this PR to pay off preexisting debt, but I suggest that we at least track some idea how to clean it up.

mflendrich (commented):
Looking good overall.

pkg/cmd/operator/gatherbase.go
mflendrich, line 1: Another question - how do you want to approach documentation updates?
I don't mind if it's a separate PR - in that case we shouldn't probably be closing the issue with this PR.

pkg/cmd/operator/gatherbase.go
rzetelskik: isn't this actually following a convention for usage messages? Angled brackets indicate required arguments and square brackets indicate optional ones. In which case `[<arg>]` wouldn't make sense. But given that the usage message convention doesn't exactly fit here, I also think `kind[.group]` is fine.

mflendrich: My interpretation is that `<arg>` means an argument, `[ ... ]` means an optional part, and anything else is a literal.

pkg/cmd/operator/gatherbase.go
czeslavo: `resourcesToExclude` is written in `Complete`, which is what it is designed for, I believe. It's following the pattern `*Options` structs established. There are other values/objects that are computed based on the CLI flags and stored as a field of `GatherBaseOptions` in `Complete` method.

I agree it's not perfect, as it's not clear which values can be consumed directly from the CLI flag value. Ideally, we would have a getter for every value that's to be consumed from `GatherBaseOptions` and keep this decision encapsulated, but I haven't seen a precedent for such behavior. Also, I don't know whether we would like to introduce another way of doing things.

I suggest we extract a separate struct for clear separation of the CLI flags and keep the compute-step as is.

Side note: another step for refactoring must-gather will be merging `GatherBaseOptions` with `MustGatherOptions` as they're no longer needed - there's nothing else to share a base with after dropping `gather` command.

pkg/cmd/operator/gatherbase.go
czeslavo: >  also: I believe that the code style across the codebase uses Format instead of Fmt

I found occurrences of both styles, but yeah, it seems like `Format` is more common. Fixed.

pkg/cmd/operator/gatherbase.go
czeslavo: Fixed.

pkg/cmd/operator/mustgather.go
czeslavo: Fair enough. Updated.

pkg/gather/collect/discoverer.go
czeslavo: I haven't dove into it to be fair, didn't want to make this PR touch too much, so decided to just replicate this.

The link you posted is about the deprecation of `events.k8s.io/v1beta1`, which we do not take into account in this code, though. I think we still need this to avoid duplicated data in the archive - `corev1.Events` are legacy, but they're still kept for backward-compatibility. `events.k8s.io.Events` are the new, better API, and they effectively are conveying the same data from our perspective, so we only keep `events.k8s.io.Events` (see [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/383-new-event-api-ga-graduation/README.md) for reference).

pkg/cmd/operator/gatherbase.go
czeslavo: Let's address it in this PR already. I've updated the article about the must-gather command.

docs/source/support/must-gather.md
```diff
@@ -1,12 +1,16 @@
...
 
 `must-gather` is an embedded tool in Scylla Operator that helps collecting all the necessary info when something goes wrong.
-
 The tool talks to the Kubernetes API, retrieves a predefined set of resources and saves them into a folder in your current directory.
-By default, all collected Secrets are censored to avoid sending sensitive data.
-That said, you can always review the archive before you attach it to an issue or your support request.
 
-Given it needs to talk to the Kubernetes API, at the very least, you need to supply the `--kubeconfig` flag with a path to the kubeconfig file for your Kubernetes cluster, or set the `KUBECONFIG` environment variable.
+:::{note}
+By default, sensitive resources like `Secrets` or `bitnami.com.SealedSecrets` are omitted from the collection.
```
mflendrich:
```suggestion
By default, sensitive resources (`Secrets` and `bitnami.com.SealedSecrets`) are omitted from the collection.
```

docs/source/support/must-gather.md
```diff
@@ -1,12 +1,16 @@
...
+If you ever need to include them (ScyllaDB support will explicitly ask for that if a need arises), please refer to the [include sensitive resources](#including-sensitive-resources) section below.
```
mflendrich: The reference to ScyllaDB Support is problematic because operator per se is OSS. Suggested rewording.

```suggestion
If you come across a clear need to include them (not recommended unless the default settings are deemed insufficient for investigation), please refer to the [include sensitive resources](#including-sensitive-resources) section below.
```

docs/source/support/must-gather.md
```diff
@@ -1,12 +1,16 @@
...
+
+If you have other resources that you wouldn't like to include in the collection, refer to the [excluding resources](#excluding-resources) section below.
+:::
+
+Given `must-gather` needs to talk to the Kubernetes API, at the very least, you need to supply the `--kubeconfig` flag with a path to the kubeconfig file for your Kubernetes cluster, or set the `KUBECONFIG` environment variable.
```
mflendrich:
```suggestion
Given that `must-gather` needs to talk to the Kubernetes API, at the very least, you need to supply the `--kubeconfig` flag with a path to the kubeconfig file for your Kubernetes cluster, or set the `KUBECONFIG` environment variable.
```

docs/source/support/must-gather.md
mflendrich: On a second thought - is this note box necessary given that this page is not that long and is likely to be read in one go?

docs/source/support/must-gather.md
```diff
@@ -99,3 +103,25 @@ You can also request collecting every resource in the Kubernetes API, if the def
...
+scylla-operator must-gather --all-resources --include-sensitive-resources
+```
+
+### Excluding resources
+
+If you have resources that you don't want to include in the collection other than the default ones, you can exclude them
+by passing the `--exclude-resource` flag multiple times.
+
+:::{note}
+Format for the resource is `kind[.group]`, e.g., `LimitRange` or `DeviceClass.resource.k8s.io`. Group should be omitted for core resources.
```
mflendrich: More readable this way?

```suggestion
The format for the resource is `kind` (for core resources) or `kind.group`. Examples: `LimitRange` or `DeviceClass.resource.k8s.io`.
```

mflendrich (approved):
/lgtm

some nits attached, feel free to handle in a separate PR

pkg/gather/collect/discoverer.go
mflendrich: My intention  behind the 1.25 API deprecation guide was that `events.k8s.io/v1` is already well established as an API, and it feels reasonable to restrict event collection to just that.

However, that's just a drive-by observation outside of the scope of this PR.

czeslavo:
@mflendrich PTAL, comments addressed.

mflendrich (approved):
/lgtm

rzetelskik:
> @czeslavo: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> ci/prow/e2e-gke-parallel	[c44db8e](https://github.com/scylladb/scylla-operator/commit/c44db8e2368f5ddfbc8a095796c146b009c0e474)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3046/pull-scylla-operator-master-e2e-gke-parallel/1986722979794391040)	unknown	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3046). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:czeslavo).

~@mflendrich can we override it? The suite passed, only the artifacts collection failed.~ Scratch that, this PR affects artifact collection :grin:

rzetelskik:
> @czeslavo: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> ci/prow/e2e-gke-parallel	[c44db8e](https://github.com/scylladb/scylla-operator/commit/c44db8e2368f5ddfbc8a095796c146b009c0e474)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3046/pull-scylla-operator-master-e2e-gke-parallel/1986722979794391040)	unknown	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3046). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:czeslavo).

/retest

Issue #2913, opened 2025-08-25
Pin third-party dependency versions in a single source of truth and automate their updates


rzetelskik:
Currently, we keep in-tree copies of third-party dependencies in the repository. At the same time we don't track clear explanations of how they were created. In case of an update, the source of the manifests often needs to be traced back. Even then it's not immediately obvious if these manifests were modified. It makes updating these third-party manifests a lengthy, manual process.

Instead, I suggest we keep the dependency versions used in our CI pinned in a single source of truth and create Makefile targets for updating them.

mflendrich:
Now tracked as [OPERATOR-35](https://scylladb.atlassian.net/browse/OPERATOR-35) in the internal Jira.

[OPERATOR-35]: https://scylladb.atlassian.net/browse/OPERATOR-35?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3062, opened 2025-11-10
Update release schedule

rzetelskik:
**Description of your changes:** Update the release schedule after v1.19 GA.

Resolves #

/kind documentation
/priority important-soon
/cc
/hold for v1.19 GA

### DIFF
--- a/docs/source/support/releases.md
+++ b/docs/source/support/releases.md
@@ -6,7 +6,7 @@ We are aiming to ship a new release approximately every 2 months. The following 
 :::{table}
 | Release | Code freeze | General availability |
 |:-------:|:-----------:|:--------------------:|
-|  1.19   | 2025-11-10  |      2025-11-19      |
+|  1.20   |  2026-02-09 |      2026-02-16      |
 :::
 
 ## Supported releases
@@ -15,8 +15,9 @@ We support the latest 2 releases of Scylla Operator to give everyone time to upg
 :::{table}
 | Release | General availability |  Support ends   |
 |:-------:|:--------------------:|:---------------:|
+|  1.19   |      2025-11-19      | Release of 1.21 |
 |  1.18   |      2025-08-11      | Release of 1.20 |
-|  1.17   |      2025-05-12      | Release of 1.19 |
+|  1.17   |      2025-05-12      |   2025-11-19    |
 |  1.16   |      2025-03-03      |   2025-08-11    |
 |  1.15   |      2024-12-19      |   2025-05-12    |
 |  1.14   |      2024-09-19      |   2025-03-03    |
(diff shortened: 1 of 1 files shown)


rzetelskik:
btw @mflendrich, not necessarily in this PR, but should we pick a single syntax of specifying dependency version requirements in the table? I think we're using every possible flavour in these 5 rows :stuck_out_tongue: And none of them seems standard and explicit.

mflendrich:
Absolutely. Would you mind creating an issue for it? We should bag the similar small documentation issues together and get them done at some point. Probably in the brewing "make operator easier to comprehend and use" theme.

rzetelskik:
Updated the compatibility matrix.
@mflendrich ptal

docs/source/support/releases.md
```diff
@@ -64,13 +65,13 @@ GA images aren't built from scratch but rather promoted from an existing release
...
-| ScyllaDB          | `2024.1 ... 2025.1`          | `2024.1 ... 2025.1`          | `>=6.0`, `2023.1 ... 2025.1`          |
-| Scylla Manager    | `>=3.5.0 && <=3.5.1`         | `>=3.5.0 && <=3.5.1`         | `>=3.3.3 && <=3.5`                    |
-| Scylla Monitoring | `(CRD)`                      | `(CRD)`                      | `(CRD)`                               |
+| Component           | master                    | v1.19                     | v1.18             |
+| ------------------- | ------------------------- | ------------------------- | ----------------- |
+| Kubernetes          | `1.31 - 1.34`             | `1.31 - 1.34`             | `1.30 - 1.33`     |
+| CRI API             | `v1`                      | `v1`                      | `v1`              |
+| ScyllaDB            | `2024.1, 2025.1 - 2025.3` | `2024.1, 2025.1 - 2025.3` | `2024.1 - 2025.1` |
+| ScyllaDB Manager    | `3.5, 3.7`                | `3.5, 3.7`                | `3.5`             |
+| ScyllaDB Monitoring | `(CRD)`                   | `(CRD)`                   | `(CRD)`           |
```
czeslavo, line 74: As we're already modifying this table, maybe we could make this dependency straighten up.

I think it's obvious that ScyllaDBMonitoring we're compatible with is the version we're bundling the operator with. I think the missing part here is the dependency on Prometheus CRDs. Maybe we could adjust it as below? 

```suggestion
| Prometheus Operator CRDs | `v1`                   | `v1`                   | `v1`           |
```

rzetelskik: I'm not sure it should even be a part of this table really, i.e. dependencies could be listed separately.
Let's handle this in https://github.com/scylladb/scylla-operator/issues/3063 and https://github.com/scylladb/scylla-operator/issues/2913

mflendrich: `v1` is not precise enough when working with CRDs- it is legal to add new fields to a CRD but keep the same apiVersion.

Suggestion: version the CRDs after the prometheus operator version that includes them.

mflendrich: (should we resolve this thread?)

czeslavo: Agreed, let's tackle this separately.

PR #3070, opened 2025-11-13
Document handling multiple clusters in a single Prometheus

czeslavo:
**Description of your changes:** Documents how to configure `Prometheus` so it scrapes metrics from multiple `ScyllaClusters`.

Resolves #

### DIFF
--- a/docs/source/management/monitoring/setup.md
+++ b/docs/source/management/monitoring/setup.md
@@ -106,11 +106,36 @@ You also need to ensure that your `Prometheus` instance is configured to discove
 :end-before: "# [/selectors]"
 :::
 
-:::{note}
-If you leave selectors empty, Prometheus will select all `ServiceMonitor` and `PrometheusRule` resources. It may be desirable in some cases.
-Please refer to the [Prometheus Operator documentation](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusSpec)
-for more details on how to configure `serviceMonitorSelector` and `ruleSelector`.
+#### Monitoring multiple ScyllaClusters with a single Prometheus
+
+If you have more than one `ScyllaCluster` in your Kubernetes cluster, and you want to monitor all of them using the same `Prometheus` instance,
+you can customize the selectors to match `ServiceMonitors` and `PrometheusRules` created for multiple `ScyllaClusters`.
+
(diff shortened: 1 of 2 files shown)


examples/monitoring/v1alpha1/multi-cluster.prometheus.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: monitoring.coreos.com/v1
```
rzetelskik, line 3: nit: rename to `multi-cluster.prometheus.yaml` - we don't have a convention for it now but this is more or less what we do with other examples

docs/source/management/monitoring/setup.md
```diff
@@ -106,12 +106,35 @@ You also need to ensure that your `Prometheus` instance is configured to discove
...
+:start-after: "# [selectors]"
+:end-before: "# [/selectors]"
+:::
+
+In this case, we're setting `serviceMonitorNamespaceSelector` and `ruleNamespaceSelector` to an empty selector, which means that `Prometheus`
+will look for `ServiceMonitors` and `PrometheusRules` in all namespaces.
+
+:::{warning}
+Make sure that the `ScyllaClusters` you want to monitor using a single `Prometheus` instance have distinct names across the entire Kubernetes cluster
+(not just within their namespaces). This is important to avoid metrics collisions.
```
rzetelskik:
```suggestion
If you want to monitor multiple `ScyllaClusters` using a single Prometheus instance, ensure all selected `ScyllaClusters`  have unique names across the entire Kubernetes cluster (not just within their namespaces). Our monitoring setup cannot currently distinguish the metrics collected from `ScyllaClusters` with colliding names, even if they reside in separate namespaces.
```
Doesn't have to be this exact wording but I think we should put more emphasis here, I'm not sure how to say it but the way you put it it sounds like something expected? While this is definitely a modeling issue on our side and we need to make sure this is understandable and requires precautions.

examples/monitoring/v1alpha1/multi-cluster.prometheus.yaml
czeslavo: Fixed.

docs/source/management/monitoring/setup.md
czeslavo: Reworded.

rzetelskik:
@czeslavo let's backport to 1.19

PR #3071, opened 2025-11-13
Allow for setting additional arguments for Scylla Operator containers through Helm

rzetelskik:
**Description of your changes:** We currently have no way of providing additional arguments to Scylla Operator when deploying via Helm. This PR fixes it. The existing `logLevel` option remains, but `additionalArguments` take precedence.

This also allows for enabling/disabling feature gates. For this reason we'll also have to backport it to 1.19.

Resolves #

/kind bug
/priority critical-urgent
/cc czeslavo

### DIFF
--- a/helm/scylla-operator/values.yaml
+++ b/helm/scylla-operator/values.yaml
@@ -12,6 +12,8 @@ image:
 
 # Scylla Operator log level, 0-9 (higher number means more detailed logs)
 logLevel: 2
+# Additional arguments to pass to Scylla Operator container.
+additionalArgs: [ ]
 # Resources allocated to Scylla Operator pods
 resources:
   requests:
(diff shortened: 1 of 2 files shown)


helm/scylla-operator/values.yaml
```diff
@@ -12,6 +12,8 @@ image:
 
 # Scylla Operator log level, 0-9 (higher number means more detailed logs)
 logLevel: 2
+# Additional arguments to pass to Scylla Operator container.
+additionalArguments: [ ]
```
czeslavo: Nit: for consistency with upstream, we could name it `additionalArgs` (as we pass it to the container's `args` field).

rzetelskik: renamed

PR #3084, opened 2025-11-18
Add documentation for feature gates and bootstrap synchronisation

rzetelskik:
**Description of your changes:** This PR adds the missing documentation for feature gates and a separate document for bootstrap synchronisation to explain the reasoning behind it and the safety valve mechanism.
It also moves the generated API reference to a common "Reference" section.

Resolves #

/kind documentation
/priority important-soon
/cc

### DIFF
--- a/docs/source/management/bootstrap-sync.md
+++ b/docs/source/management/bootstrap-sync.md
@@ -1,0 +1,62 @@
+# Synchronising bootstrap operations in ScyllaDB clusters
+
+This document explains the automated bootstrap synchronisation in {{productName}}.
+
+## Overview
+
+[ScyllaDB requires that when a new node is added to a cluster, no existing nodes are down](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-node-to-cluster.html#check-the-status-of-nodes).
+Precisely speaking, no node in the cluster can consider any other node to be down when a topology change operation is performed.
+If this is not the case and a non-idempotent operation of bootstrapping a new node is started, the coordinator will deny the node joining the cluster and leave it in a state which is difficult to recover from automatically.
+
+To address this issue, by extending ScyllaDB `Pods` with init containers, {{productName}} implements a barrier which observes the statuses of all nodes in the cluster in order to verify that no node in the cluster is down before allowing a new ScyllaDB node to bootstrap.
+Any restarting nodes which have already been bootstrapped are exempt from meeting this precondition.
+The status of the nodes is propagated in the cluster using the internal `ScyllaDBDatacenterNodeStatuses` CRD. Please refer to the [ScyllaDBDatacenterNodeStatuses API reference](../reference/api/groups/scylla.scylladb.com/scylladbdatacenternodesstatusreports.rst) for more information.
+
+:::{caution}
+Bootstrap synchronisation does not currently extend to nodes replacing dead nodes in the cluster (i.e. the nodes which are being added to the cluster by a [Node replacement operation](../resources/scyllaclusters/nodeoperations/replace-node/)).
+Ensure [the applicable prerequisites](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/replace-dead-node.html#prerequisites) are met before initializing a node replacement operation.
+:::
+
+:::{caution}
+Bootstrap synchronisation does not currently extend to [multi-datacenter ScyllaDB clusters configured manually using ScyllaCluster API](../resources/scyllaclusters/multidc/multidc.md). This is due to the fact that node statuses can only be propagated within a single ScyllaDB datacenter.
+Before adding a new datacenter or a new node to an existing datacenter in a multi-datacenter ScyllaDB cluster, [check the status of nodes](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-node-to-cluster.html#check-the-status-of-nodes) in all datacenters.
+:::
+
+## Enable bootstrap synchronisation
+
+:::{include} ../.internal/bootstrap-sync-min-scylladb-version-caution.md
+:::
+
+Due to the fact that {{productName}} does not currently have a reliable way of determining the effective ScyllaDB version used in a `ScyllaCluster` and the required ScyllaDB version for this feature is 2025.2 or later,
+this feature can only be enabled via the `BootstrapSynchronisation` feature gate. This feature will remain opt-in while versions prior to 2025.2 are supported by {{productName}}.
+
+Please refer to [](../reference/feature-gates.md) for instructions on how enable or disable a feature gate.
+
+## Override the precondition check
+
+By default, when the `BootstrapSynchronisation` feature gate is enabled, bootstrap synchronisation is enforced for all `ScyllaClusters` in the cluster.
+However, in scenarios where you might need to circumvent the precondition for a specific `ScyllaCluster` or a selected ScyllaDB node, you can do so by setting a dedicated annotation.
+When the annotation is set on a `ScyllaCluster` or a ScyllaDB node `Service`, the corresponding entity will be forced to proceed with the bootstrap operation unconditionally.
+
+You can override the precondition check by running the following command:
+:::::{tabs}
+
+::::{group-tab} ScyllaCluster
+
+:::{code-block} bash
+kubectl annotate -n scylla scyllacluster <scyllacluster-name> scylla-operator.scylladb.com/force-proceed-to-bootstrap=true
+:::
+
+::::
+
+::::{group-tab} Selected ScyllaDB node
+
+:::{code-block} bash
+kubectl annotate -n scylla service <scylladb-node-service-name> scylla-operator.scylladb.com/force-proceed-to-bootstrap=true
+:::
+
+::::
+
+:::::
+
+
--- a/docs/source/management/index.md
+++ b/docs/source/management/index.md
@@ -4,6 +4,7 @@
 :maxdepth: 1
 
 sysctls
+bootstrap-sync
 upgrading/index
 monitoring/index
 :::
--- a/docs/source/reference/feature-gates.md
+++ b/docs/source/reference/feature-gates.md
@@ -1,0 +1,94 @@
+# Feature Gates
+
+{{productName}} lets you enable or disable features using feature gates. This document provides an overview of the available feature gates and instructions on how to use them.
+
+## Available feature gates
+
+The following feature gates are available in {{productName}}:
+
+:::{list-table}
+:widths: 60 20 20
+:header-rows: 1
+
+* - Feature Gate
+  - Default
+  - Since
+* - AutomaticTLSCertificates
+  - `true`
+  - v1.11
+* - BootstrapSynchronisation
+  - `false`
+  - v1.19
+:::
+
+- The "Default" indicates if the feature is enabled when you don't set it explicitly.
+- The "Since" column indicates the {{productName}} version in which the feature gate was introduced or its default was changed.
+
+### AutomaticTLSCertificates
+
+`AutomaticTLSCertificates` enables mTLS client connections to ScyllaDB. 
+When this feature is enabled, {{productName}} automatically generates and rotates serving and client TLS certificates.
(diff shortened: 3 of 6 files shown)


rzetelskik:
/cc @czeslavo
please have a look as well

docs/source/reference/feature-gates.md
```diff
@@ -0,0 +1,91 @@
...
+  - v1.19
+:::
+
+- The "Stage" column indicates the maturity level of the feature. {{productName}} follows the feature stages defined in the [upstream Kubernetes Feature Gates documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages).
+- The "Since" column indicates the {{productName}} version in which the feature gate was introduced or its stage was changed.
+
+### AutomaticTLSCertificates
+
+:::{caution}
+mTLS is not configured automatically for node-to-node communication.
```
czeslavo: This wording could suggest that one can manually configure node-to-node communication, while it's not supported as of now, according to the recent experiments. Let's ensure it's explicitly stated and link to https://github.com/scylladb/scylla-operator/issues/2434 as a tracker.

docs/source/reference/feature-gates.md
```diff
@@ -0,0 +1,91 @@
...
+:::
+
+`AutomaticTLSCertificates` enables automatic management of TLS certificates for ScyllaDB nodes.
+{{productName}} will automatically generate and rotate TLS certificates for ScyllaDB nodes and clients.
+It will also configure ScyllaDB nodes to use these certificates for secure client-to-node communication.
```
czeslavo: We could link to https://operator.docs.scylladb.com/v1.19/resources/scyllaclusters/clients/cql.html#remote-cqlsh as that shows how to actually make use of this feature on the client side.

docs/source/management/index.md
```diff
@@ -6,4 +6,5 @@
 sysctls
 upgrading/index
 monitoring/index
+bootstrap-sync
```
czeslavo: Please consider bundling grouped/standalone articles together (e.g., in order: sysctls, bootstrap-sync, upgrading, monitoring). IMO it just looks better this way on the menu.

docs/source/management/bootstrap-sync.md
```diff
@@ -0,0 +1,62 @@
+# Synchronising bootstrap operations in ScyllaDB clusters
+
+This document explains the automated bootstrap synchronisation in {{productName}}.
+
+## Overview
```
czeslavo, line 5: Should we document that it's essential/highly-recommended in automated multi-DC clusters? Perhaps we could link to this explanation somewhere from `/resources/scylladbclusters/scylladbclusters/`.

docs/source/management/bootstrap-sync.md
```diff
@@ -0,0 +1,62 @@
...
+
+:::{caution}
+Bootstrap synchronisation does not currently extend to nodes replacing dead nodes in the cluster (i.e. the nodes which are being added to the cluster by a [Node replacement operation](../resources/scyllaclusters/nodeoperations/replace-node/)).
+Ensure [the applicable prerequisites](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/replace-dead-node.html#prerequisites) are met before initializing a node replacement operation.
+:::
+
+:::{caution}
+Bootstrap synchronisation does not currently extend to [multi-datacenter ScyllaDB clusters](../resources/scyllaclusters/multidc/multidc.md). This is due to the fact that node statuses can only be propagated within a single ScyllaDB datacenter.
+Before adding a new datacenter or a new node to an existing datacenter in a multi-datacenter ScyllaDB cluster, [check the status of nodes](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-node-to-cluster.html#check-the-status-of-nodes) in all datacenters.
```
czeslavo: Should we make it more explicit that we're talking about manual multi-DC setup with `ScyllaCluster` here? It might sound like any multi-DC setup (even `ScyllaDBCluster`) is not supported.

docs/source/management/bootstrap-sync.md
```diff
@@ -0,0 +1,62 @@
+# Synchronising bootstrap operations in ScyllaDB clusters
+
+This document explains the automated bootstrap synchronisation in {{productName}}.
+
+## Overview
+
+ScyllaDB requires that when a new node is added to a cluster, no existing nodes are down.
```
czeslavo: How about backing this statement with a link to https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-node-to-cluster.html#check-the-status-of-nodes?

docs/source/reference/feature-gates.md
rzetelskik: done

docs/source/management/index.md
rzetelskik: ok, done

docs/source/reference/feature-gates.md
rzetelskik: done

docs/source/management/bootstrap-sync.md
rzetelskik: done

docs/source/management/bootstrap-sync.md
rzetelskik: Yes, but I believe our documentation doesn't make a distinction really, so I don't know how other than link to the specific document. There is not a single mention of "manual multi-dc" or "automated multi-dc". Both are just referred to as "multi-datacenter".

Michal added this description to the release announcement "(multiple ScyllaCluster objects bound together with external seeds)", but I don't really think it fits the documentation?

docs/source/management/bootstrap-sync.md
rzetelskik: thanks, I forgot to add it. Done.

rzetelskik:
> 
> ci/prow/verify-docs	[325c1ab](https://github.com/scylladb/scylla-operator/commit/325c1ab59d52049e2bdee2e38199e9d9485da246)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3084/pull-scylla-operator-master-verify-docs/1990782053112614912)	unknown	`/test verify-docs`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3084). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).

prometheus.io is down

docs/source/reference/feature-gates.md
```diff
@@ -0,0 +1,93 @@
...
+  - `true`
+  - Beta
+  - v1.11
+* - BootstrapSynchronisation
+  - `false`
+  - Alpha
+  - v1.19
+:::
+
+- The "Stage" column indicates the maturity level of the feature. {{productName}} follows the feature stages defined in the [upstream Kubernetes Feature Gates documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages).
```
mflendrich: It was decided not to use the alpha/beta language with bootstrap sync (to avoid confusion with quality level - it is a working feature, just not enabled by default). And in general I think that we will not be using the alpha/beta language for new feature graduation too frequently.

Also, if the the `AutomaticTLSCertificates` gate has been enabled since 1.11, then it's hardly a beta at this point.

For this reason I am asking that we drop the alpha/beta language altogether (just communicate the opt-in / opt-out) status, at least for the `BootstrapSynchronisation` gate.

As for `AutomaticTLSCertificates` - I would like to hear your opinion, but I feel that this may be time to drop the beta language. Thoughts?

docs/source/reference/feature-gates.md
```diff
@@ -0,0 +1,93 @@
...
+
+### AutomaticTLSCertificates
+
+:::{caution}
+mTLS for node-to-node communication is [not yet supported](https://github.com/scylladb/scylla-operator/issues/2434).
+:::
+
+`AutomaticTLSCertificates` enables automatic management of TLS certificates for ScyllaDB nodes.
+{{productName}} will automatically generate and rotate TLS certificates for ScyllaDB nodes and clients.
+It will also configure ScyllaDB nodes to use these certificates for secure client-to-node communication.
```
mflendrich:
```suggestion
`AutomaticTLSCertificates` enables certificate-based authentication for client connections to ScyllaDB.
{{productName}} will automatically generate and rotate TLS certificates for ScyllaDB nodes and clients.
It will also configure ScyllaDB nodes to use these certificates for secure client-to-node communication.
```

The word "management" doesn't help convey what this feature means to the user - suggested rewording to explain in terms of enabled functionality, not enabled responsibility.

docs/source/reference/feature-gates.md
rzetelskik: > It was decided not to use the alpha/beta language with bootstrap sync

When has that been decided? Feature stages are a part of the codebase: https://github.com/scylladb/scylla-operator/blob/07a0194361995fe002bea4a5b88a36c54b922cf9/pkg/features/features.go#L40.
Features can be enabled by setting `--feature-gate=AllAlpha=true,AllBeta=true` for example, so from the interface perspective it's also a thing.

Do you mean to just not communicate it in the docs?

> As for AutomaticTLSCertificates - I would like to hear your opinion, but I feel that this may be time to drop the beta language. Thoughts?

If we're certain that we won't make any incompatible changes then yes, I don't see why not. Although I remember the idea of dropping it and replacing it with an external dependency was floating around at some point?

docs/source/reference/feature-gates.md
rzetelskik: I think this would be even more misleading as it would suggest that this feature sets up `com.scylladb.auth.CertificateAuthenticator` (see https://docs.scylladb.com/manual/stable/operating-scylla/security/certificate-authentication.html), while it doesn't. Certs are not used for authentication in Scylla terms.

docs/source/management/bootstrap-sync.md
czeslavo: How about something similar to

```markdown
Bootstrap synchronisation does not currently extend to [multi-datacenter ScyllaDB clusters configured manually using ScyllaCluster API](../resources/scyllaclusters/multidc/multidc.md).
```

docs/source/reference/feature-gates.md
rzetelskik: dropped the stage concept from docs

docs/source/management/bootstrap-sync.md
rzetelskik: done

docs/source/reference/feature-gates.md
```diff
@@ -0,0 +1,90 @@
...
+  - Since
+* - AutomaticTLSCertificates
+  - `true`
+  - v1.11
+* - BootstrapSynchronisation
+  - `false`
+  - v1.19
+:::
+
+- The "Default" indicates if the feature gate is enabled by default.
```
mflendrich:
```suggestion
- The "Default" indicates if the feature is enabled when you don't set any value.
```

docs/source/reference/feature-gates.md
czeslavo: I suggest something like below:

```markdown
`AutomaticTLSCertificates` enables mTLS client connections to ScyllaDB. When this feature is enabled, {{productName}} automatically generates and rotates serving and client TLS certificates. 

:::{note}
Client certificates are validated by ScyllaDB nodes (the certificate chain must be trusted), but ScyllaDB does **not** perform client identity or authorization checks based on certificate contents ([com.scylladb.auth.CertificateAuthenticator](https://docs.scylladb.com/manual/stable/operating-scylla/security/certificate-authentication.html) is not used).
:::
```

rzetelskik: updated as agreed internally

docs/source/reference/feature-gates.md
rzetelskik: updated with a minor change

rzetelskik:
updated, @mflendrich @czeslavo ptal

czeslavo:
lgtm 
/assign @mflendrich

Issue #2813, opened 2025-07-23
Automate Prometheus Operator manifests updates


czeslavo:
### What should the feature do?

After https://github.com/scylladb/scylla-operator/pull/2810 is merged, we should ensure that example manifests (under `examples/third-party/prometheus-operator`) for the Prometheus Operator are automatically updated to follow the version we rely on in go.mod

/kind machinery
/priority important-longterm

### What is the use case behind this feature?

When releasing our Operator, we need to ensure that the example manifests are in sync with the version of the Go API we rely on.

rzetelskik:
@czeslavo do you think this has been sufficiently covered by https://github.com/scylladb/scylla-operator/pull/3032?

czeslavo:
> [@czeslavo](https://github.com/czeslavo) do you think this has been sufficiently covered by [#3032](https://github.com/scylladb/scylla-operator/pull/3032)?

It mostly covers that, but during #3032 review, I missed that we may now have two distinct Prometheus Operator versions in the project (one, used for examples, defined in `metadata.yaml`; and one for Go API, defined in `go.mod`). How about changing this [like so](https://github.com/scylladb/scylla-operator/compare/master...czeslavo:scylla-operator:use-prom-operator-version-from-gomod) and closing this issue?

Issue #2949, opened 2025-09-10
Updating Scylla-Operator to version 1.18 results in Prometheus not starting


pogzie:
### What happened?

Upgrading from 1.17 to 1.18 throws the following error: 

```
I0909 14:21:50.078945 1 record/event.go:389] "Event occurred" object="shared-services/scylladb-monitoring-scylladb" fieldPath="" kind="ServiceMonitor" apiVersion=" monitoring.coreos.com/v1( http://monitoring.coreos.com/v1 ) " type="Warning" reason="CreateServiceMonitorFailed" message="Failed to c │
│ reate ServiceMonitor shared-services/scylladb-monitoring-scylladb: ServiceMonitor in version "v1 \ " cannot be handled as a ServiceMonitor: strict decoding error: unknown field \"status\""
```

In version 1.17 all is working. 
Scylla manager version: 3.5.0
Scylla version: scylla-enterprise:2025.1.2

Additional information: 
It is running in Kubernetes and prometheus pod is not creating because of the error. The log is from scylla-operator.
Here is the log from scylladbmonitoring:

```
 Status:                                                                                                                                                                                                                                                                    │
│   Conditions:                                                                                                                                                                                                                                                              │
│     Last Transition Time:  2025-09-09T14:21:50Z                                                                                                                                                                                                                            │
│     Message:               can't apply [monitoring.coreos.com/v1](http://monitoring.coreos.com/v1), Kind=ServiceMonitor: ServiceMonitor in version "v1" cannot be handled as a ServiceMonitor: strict decoding error: unknown field "status"                                                                  │
│     Observed Generation:   1                                                                                                                                                                                                                                               │
│     Reason:                Error                                                                                                                                                                                                                                           │
│     Status:                True                                                                                                                                                                                                                                            │
│     Type:                  PrometheusControllerDegraded
```

### What did you expect to happen?

Scylla Monitor stack to launch in Operator 1.18 with no issues.

### How can we reproduce it (as minimally and precisely as possible)?

-- To follow -- 

### Scylla Operator version

1.18 (works on 1.17)

### Kubernetes platform name and version

-- To follow --

<details>
```console
$ kubectl version
# paste output here
```

Kubernetes platform info:
</details>


### Please attach the must-gather archive.


-- To follow --

### Anything else we need to know?

-- To follow --

mflendrich:
@czeslavo can this be fallout from https://github.com/scylladb/scylla-operator/pull/2810 ?

[...]

PR #3088, opened 2025-11-19
Sync prom-operator go.mod versions with metadata.yaml

czeslavo:
**Description of your changes:** To make sure we always keep Prometheus Operator dependencies versions in sync (to avoid issues like https://github.com/scylladb/scylla-operator/issues/2949#issuecomment-3274003808), let's make the version in `go.mod` sync with the one defined in `metadata.yaml`.

Note adjustments in Makefile reflect `local-csi-driver` changes we made recently: https://github.com/scylladb/local-csi-driver/pull/106

Resolves https://github.com/scylladb/scylla-operator/issues/2813.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -231,3 +231,10 @@ tool (
 	k8s.io/code-generator
 	sigs.k8s.io/controller-tools/cmd/controller-gen
 )
+
+// prometheus-operator modules are intentionally kept in sync with the version defined in `assets/metadata/metadata.yaml`.
+// Synchronization is performed automatically by the `make update-go-mod-replace` target, do not edit manually.
+replace (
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -231,3 +231,9 @@ tool (
 	k8s.io/code-generator
 	sigs.k8s.io/controller-tools/cmd/controller-gen
 )
+
+// prometheus-operator modules are intentionally kept in sync with the version defined in `assets/metadata/metadata.yaml`.
```
rzetelskik, line 235: I think this comment should indicate that it's automated and shouldn't be edited manually

czeslavo: Added an explicit comment.

rzetelskik:
/lgtm
Thanks!

PR #3091, opened 2025-11-21
Add missing fields to OLM bundle patch for fields overwritten during publishing

rzetelskik:
**Description of your changes:** This PR adds missing `.metadata.name` field to the OLM bundle patch containing fields which are supposed to be overwritten during publishing.

Resolves #

/kind cleanup
/priority important-soon


mflendrich (approved):
/lgtm

rzetelskik:
Turns out there was a reason for the `.spec.Version` to be at `1.0.0` - preflight considers `0.0.0` empty. I had to revert it.
@mflendrich ptal

mflendrich (approved):
/lgtm

PR #3093, opened 2025-11-21
Align metadata's min kubernetes version with compatibility matrix

rzetelskik:
**Description of your changes:** As per the compatibility matrix, min kube version is 1.31 now. This PR updates the metadata to reflect this. Ideally, the metadata should be reflected in the compatibility matrix, but we don't have the automation yet.

Resolves #

/kind cleanup
/priority important-soon

### DIFF
--- a/bundle/patches/versions.clusterserviceversion.yaml
+++ b/bundle/patches/versions.clusterserviceversion.yaml
@@ -4,4 +4,4 @@ metadata:
   annotations:
     com.redhat.openshift.versions: v4.19-v4.19
 spec:
-  minKubeVersion: 1.30.0
+  minKubeVersion: 1.31.0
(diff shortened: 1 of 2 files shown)


bundle/patches/versions.clusterserviceversion.yaml
```diff
@@ -4,4 +4,4 @@ metadata:
   annotations:
     com.redhat.openshift.versions: v4.19-v4.19
 spec:
-  minKubeVersion: 1.30.0
+  minKubeVersion: 1.31.0
```
mflendrich, line 7:
```suggestion
  minKubeVersion: 1.31.0 # tracks .operator.minKubernetesVersion in /assets/metadata/metadata.yaml
```

rzetelskik: This is automated in the Makefile target, i.e. the content of the patch is generated. Do you think it needs a comment?

mflendrich: I would expect a `GENERATED, DO NOT EDIT` style comment on the autogenerated file, or on a field, if only the field is affected.

rzetelskik: updated https://github.com/scylladb/scylla-operator/pull/3094 so I don't have to do this twice - ptal

PR #3094, opened 2025-11-21
Adjust bundle patches to set 'com.redhat.openshift.versions' in metadata annotations instead of CSV manifest

rzetelskik:
**Description of your changes:** The `com.redhat.openshift.versions` label is required in the bundle container image, not the CSV. Ref: https://redhat-connect.gitbook.io/certified-operator-guide/ocp-deployment/operator-metadata/bundle-directory.
It should be propagated from metadata on Dockerfile creation. This PR makes this adjustment.

Resolves #

/hold for https://github.com/scylladb/scylla-operator/pull/3090
/kind cleanup
/priority important-soon


rzetelskik:
/hold
Didn't notice that the comments propagate to the target file, let me fix this

rzetelskik:
/hold cancel
@mflendrich I added head comments to patches which are automatically populated, ptal.

PR #3101, opened 2025-11-24
Update ScyllaDB version to 2025.3.4

mflendrich:
Bumps ScyllaDB to the newest version.

Going forward, we should probably hook up Renovate to do this for us.

/kind cleanup
/priority important-longterm


rzetelskik:
@mflendrich lgtm but you need to update the generated files

mflendrich:
Superseded by https://github.com/scylladb/scylla-operator/pull/3132

Issue #3102, opened 2025-11-25
Cleanup after initial multi-node cluster provisioning


ylebi:
### What happened?

When a new ScyllaCluster is provisioned with multiple nodes (e.g., `members: 6`), the operator doesn't run cleanup jobs after the cluster is fully up.

The operator creates nodes sequentially (using `OrderedReadyPodManagement`), so:

Node 0 starts and owns all tokens
Node 1 joins → tokens redistributed, node 0 loses some
Node 2 joins → more redistribution
... continues until all 6 nodes are up
By the end, nodes 0-5 all have stale token ownership data (even the system tables that use vnodes), but cleanup never runs. Later when there is a try to decommission a node, ScyllaDB automatic cleanup runs first, which by that point is slow, blocking the decommission.

### What did you expect to happen?

After initial multi-node cluster provisioning completes, cleanup jobs should run on all nodes except the last one (nodes 0-4 in a 6-node cluster), matching the behavior we already have for scale-up operations.

This would be cheap at that point (minimal data), and prevent future decommission delays.

### How can we reproduce it (as minimally and precisely as possible)?

- Create a ScyllaCluster with `members: 6`
- Wait for all nodes to be up
- Check for cleanup jobs - none are created
- Later try to decommission a node - observe it waiting for cleanup

### Scylla Operator version

Latest master

### Kubernetes platform name and version

Any

### Please attach the must-gather archive.

.

### Anything else we need to know?

https://github.com/scylladb/scylla-operator/blob/master/test/e2e/set/scyllacluster/scyllacluster_cleanup.go#L62

```
framework.By("Validating no cleanup jobs were created")
jobEvents, err := jobObserver.Stop()
o.Expect(jobEvents).To(o.BeEmpty())
```
This is correct for single-node but wrong for multi-node.

This is a gap in the implementation of enhancement https://github.com/scylladb/scylla-operator/issues/1207. The enhancement correctly handles scale-up of existing clusters but missed the initial provisioning case.

The proposal even mentions this scenario ([lines 44-49](https://github.com/scylladb/scylla-operator/blob/master/enhancements/proposals/1207-cleanup-after-scaling/README.md?plain=1#L44)):

`When a cluster is created for the first time, tokens are reshuffled when new nodes are joining sequentially. This will cause the token hash mismatch and will require a cleanup once all nodes join the cluster.`

But it was dismissed as acceptable overhead rather than being properly implemented.

[...]

PR #3111, opened 2025-11-26
Add E2E verifying nodes cleanup after multi-node cluster provisioning

czeslavo:
**Description of your changes:** Adds an E2E test that verifies that cleanup jobs are created for multi-node clusters after their provisioning (not only after scaling up).

Related to https://github.com/scylladb/scylla-operator/issues/3102.

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_cleanup.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_cleanup.go
@@ -4,9 +4,11 @@ package scyllacluster
 
 import (
 	"context"
+	"fmt"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
 	"github.com/scylladb/scylla-operator/pkg/helpers"
 	oslices "github.com/scylladb/scylla-operator/pkg/helpers/slices"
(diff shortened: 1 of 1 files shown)


rzetelskik:
Why don't we just modify the existing test to deploy multiple members? Is there any value in having a dedicated test for a cluster scaled up from 1 node?

czeslavo:
> Why don't we just modify the existing test to deploy multiple members? Is there any value in having a dedicated test for a cluster scaled up from 1 node?

My rationale is that we'd like to keep the special 1 initial node case, because it should not run any cleanup job in such a case. Having only the case with 2 initial nodes, we'd verify it creates 1 job, which is a bit different than requiring none. WDYT?

test/e2e/set/scyllacluster/scyllacluster_cleanup.go
```diff
@@ -164,4 +164,66 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+
+		framework.By("Validating cleanup jobs were created for all nodes except last one")
+		jobEvents, err := jobObserver.Stop()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Expect(jobEvents).NotTo(o.BeEmpty())
+
+		tokenRingHash, err := utils.GetCurrentTokenRingHash(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		framework.Infof("Current token ring hash is %q", tokenRingHash)
```
mflendrich: nit - where it says "current token ring hash" what we actually mean is "the token ring hash in the fully scaled out cluster"

for the sake of making this unambiguous in the time domain, I suggest explicit wording `"Token ring hash of the fully rolled out cluster is %q"` - "current" is a moving target.

czeslavo: Reworded.

test/e2e/set/scyllacluster/scyllacluster_cleanup.go
```diff
@@ -164,4 +164,66 @@ var _ = g.Describe("ScyllaCluster", func() {
...
+				e.Obj.Labels[naming.NodeJobTypeLabel] == string(naming.JobTypeCleanup) &&
+				e.Obj.Annotations[naming.CleanupJobTokenRingHashAnnotation] == tokenRingHash
+		})
+
+		o.Expect(cleanupJobsCreated).To(o.HaveLen(2))
+		o.Expect(cleanupJobsCreated).To(o.ConsistOf(
+			o.Satisfy(nodeJobMatcher(naming.MemberServiceNameForScyllaCluster(sc.Spec.Datacenter.Racks[0], sc, 0))),
+			o.Satisfy(nodeJobMatcher(naming.MemberServiceNameForScyllaCluster(sc.Spec.Datacenter.Racks[0], sc, 1))),
+		))
+	})
```
rzetelskik: most of the test is repeating, let's abstract it away - using helpers or a test tree

czeslavo: Good point. PTAL now.

PR #3112, opened 2025-11-26
Reword platforms to environments

czeslavo:
**Description of your changes:** It rewords "platforms are known not to work correctly" to "combinations are known not to work correctly" to make it clear that we're not marking the whole platform as not supported.

### DIFF
--- a/docs/source/installation/overview.md
+++ b/docs/source/installation/overview.md
@@ -10,8 +10,8 @@ More over, {{productName}} implements additional features like performance tunin
 While we do our best to implement these routines as generically as possible, sometimes there isn't any low level API to base them on and they may work only on a subset of platforms.
 
 :::{caution}
-We *strongly* recommend using a [supported Kubernetes platform](../support/releases.md#supported-kubernetes-platforms).
-Issues on unsupported platforms are unlikely to be addressed.
+We *strongly* recommend using a [supported Kubernetes environment](../support/releases.md#supported-kubernetes-environments).
+Issues on unsupported environments are unlikely to be addressed.
 :::
 
 :::{note}
--- a/docs/source/support/releases.md
+++ b/docs/source/support/releases.md
@@ -78,15 +78,15 @@ The support matrix table shows version requirements for a particular **Scylla Op
 
 {{productName}} image is published as a manifest list to `docker.io/scylladb/scylla-operator:X.Y.Z` containing image build for `amd64` and `aarch64`.
 
-### Supported Kubernetes platforms
+### Supported Kubernetes environments
 
-We officially test and recommend to use the following platforms:
+We officially test and recommend to use the following environments:
 
 :::{table}
-| Platform         | OS Image     |
-|:-----------------|:-------------|
-| GKE              | Ubuntu       |
-| EKS              | Amazon Linux |
+| Platform | OS Image     |
+|:---------|:-------------|
+| GKE      | Ubuntu       |
+| EKS      | Amazon Linux |
 :::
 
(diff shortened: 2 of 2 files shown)


docs/source/support/releases.md
```diff
@@ -94,7 +94,7 @@ performance tuning and other pieces that need to interact with the host OS, kube
 
 
 :::{warning}
-The following platforms are known **not to work correctly** at this time.
+The following combinations are known **not to work correctly** at this time.
```
rzetelskik: We should be consistent in what we call a platform. A few lines above we call the same combination (platform x os image) simply a platform (`We officially test and recommend to use the following platforms`)

czeslavo: Maybe "platform combinations" then?

abramche: In the table below, a cloud provider is given the name of a `platform`; it is indeed inconsistent with the line you are referring to. We need to align them either way... personally, platform as a cloud provider alone makes more sense to me than a combination of provider and auxiliary variables in their use. WDYT?

czeslavo: Or we add 
```
:::{note}
The _platform_ is a combination of the Kubernetes distribution and the OS image used on the nodes.
:::
``` 

and leave the `platform` wording. https://github.com/scylladb/scylla-operator/pull/3112/commits/c0d3e9a6b356541f229fe857844265ea4aae5430

rzetelskik: I agree with Eugene. Platform is a common name for a kubernetes distribution. And to me the combination of platform x hostOS is known as simply an environment? Would that work?

czeslavo: PTAL now.

docs/source/installation/overview.md
```diff
@@ -10,8 +10,8 @@ More over, {{productName}} implements additional features like performance tunin
 While we do our best to implement these routines as generically as possible, sometimes there isn't any low level API to base them on and they may work only on a subset of platforms.
 
 :::{caution}
-We *strongly* recommend using a [supported Kubernetes platform](../support/releases.md#supported-kubernetes-platforms).
-Issues on unsupported platforms are unlikely to be addressed.
+We *strongly* recommend using a [supported Kubernetes environments](../support/releases.md#supported-kubernetes-environments).
```
rzetelskik:
```suggestion
We *strongly* recommend using a [supported Kubernetes environment](../support/releases.md#supported-kubernetes-environments).
```

docs/source/support/releases.md
```diff
@@ -78,28 +78,32 @@ The support matrix table shows version requirements for a particular **Scylla Op
 
 {{productName}} image is published as a manifest list to `docker.io/scylladb/scylla-operator:X.Y.Z` containing image build for `amd64` and `aarch64`.
 
-### Supported Kubernetes platforms
+### Supported Kubernetes environments
 
-We officially test and recommend to use the following platforms:
+:::{note}
+The _environment_ is a combination of the platform (Kubernetes distribution) and the OS image used on the nodes.
```
rzetelskik: nit: I don't think this note is necessary, the table makes it clear and we don't need surgical precision

rzetelskik (approved):
one typo and one nit, rest lgtm

docs/source/installation/overview.md
czeslavo: Fixed.

docs/source/support/releases.md
czeslavo: I don't have a strong opinion about that, dropped.

rzetelskik:
/lgtm
thanks for the updates!

PR #3114, opened 2025-12-01
List only Scylla nodes pods when getting nodes IPs in E2E

czeslavo:
**Description of your changes:** It fixes the test logic responsible for fetching ScyllaCluster nodes' Pod IPs. The label selector used is satisfied not only by STS replicas, but also by cleanup jobs' Pods that sometimes manage to get created before the assertion gets executed, resulting in unstable test results. 

Adds `scylla-operator.scylladb.com/pod-type` label to all managed workloads so they are set for created pods accordingly, identifying their type precisely. Uses that for fetching Scylla pods in E2E.

It will be backported to the supported release branches.

Resolves https://github.com/scylladb/scylla-operator/issues/2096

### DIFF
--- a/pkg/controllerhelpers/scylla.go
+++ b/pkg/controllerhelpers/scylla.go
@@ -252,11 +252,6 @@ func IsPodTunable(pod *corev1.Pod) bool {
 	return pod.Status.QOSClass == corev1.PodQOSGuaranteed
 }
 
-func IsNodeTuned(ncnss []scyllav1alpha1.NodeConfigNodeStatus, nodeName string) bool {
-	ns := FindNodeStatus(ncnss, nodeName)
-	return ns != nil && ns.TunedNode
-}
-
 func IsScyllaPod(pod *corev1.Pod) bool {
 	if pod.Labels == nil {
 		return false
--- a/pkg/naming/labels.go
+++ b/pkg/naming/labels.go
@@ -26,9 +26,12 @@ func ClusterLabelsForScyllaCluster(sc *scyllav1.ScyllaCluster) map[string]string
 	return labels
 }
 
-// ClusterSelector returns a labels selector for the given ScyllaCluster.
-func ClusterSelector(sdc *scyllav1alpha1.ScyllaDBDatacenter) labels.Selector {
-	return labels.SelectorFromSet(ClusterLabels(sdc))
+// ScyllaDBNodesPodsLabelsForScyllaCluster returns a map of label keys and values for ScyllaDB nodes pods limited
+// to the given ScyllaCluster.
+func ScyllaDBNodesPodsLabelsForScyllaCluster(sc *scyllav1.ScyllaCluster) map[string]string {
+	labels := ClusterLabelsForScyllaCluster(sc)
+	labels[PodTypeLabel] = string(PodTypeScyllaDBNode)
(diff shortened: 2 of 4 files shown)


pkg/controllerhelpers/scylla.go
```diff
@@ -258,17 +258,11 @@ func IsNodeTuned(ncnss []scyllav1alpha1.NodeConfigNodeStatus, nodeName string) b
 }
 
 func IsScyllaPod(pod *corev1.Pod) bool {
-	// TODO: use a better label, verify the container
```
rzetelskik, line 261: :wink:

pkg/naming/labels.go
```diff
@@ -60,8 +60,14 @@ func StatefulSetPodLabel(name string) map[string]string {
 	}
 }
 
-func ScyllaLabels() map[string]string {
+// ScyllaDBNodePodLabels returns a map of labels for ScyllaDB node pods.
```
rzetelskik: It might be worth specifying that this refers to unique labels, not all labels a ScyllaDB node pod should have. Also it's not obvious to me why we need this function along `ScyllaDBNodesPodsLabelsForScyllaCluster`.

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
```diff
@@ -344,7 +344,7 @@ var _ = g.Describe("NodeConfig Optimizations", framework.Serial, func() {
 		o.Expect(utils.IsScyllaClusterRolledOut(sc)).To(o.BeFalse())
 
 		framework.By("Verifying the containers are blocked and not ready")
-		podSelector := labels.Set(naming.ClusterLabelsForScyllaCluster(sc)).AsSelector()
+		podSelector := labels.Set(naming.ScyllaDBNodesPodsLabelsForScyllaCluster(sc)).AsSelector()
```
rzetelskik, line 347: why use the full selector here but only the single label in `IsScyllaPod`? Seems like we have a bit of a mess with the naming functions

pkg/naming/labels.go
czeslavo: I extended the comment to explicitly say these are labels for Scylla nodes for a given ScyllaCluster only.

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
czeslavo: Here, we want to limit the scope to the specific ScyllaCluster, while in the `IsScyllaPod` (used in nodeconfig controller), we only want to know whether it is a Scylla pod (belonging to any cluster).

pkg/controllerhelpers/scylla.go
czeslavo: Do you mean we should also verify the container? I assume we do not need that as the label is set only for Scylla pods.

rzetelskik: No, I just found "use a batter label" funny, nothing to do here

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
rzetelskik: ok, thanks

PR #3116, opened 2025-12-02
Install Operator through OLM in non-release deployment scripts for OpenShift

rzetelskik:
**Description of your changes:** This PR extends the existing deployment scripts with a possibility of installing the operator with OLM. I've also used this opportunity to reduce the number of ways in which we configure the operator deployment across the existing scripts.

I don't consider the refactoring work complete - for instance, the ci-deploy and ci-deploy-release scripts now ask for the operator image even if it's not necessary because the OLM installation mode is being used. Knowing that @czeslavo has also been working on improving the usage of kustomize across our manifests, I didn't want to go to heavy on the refactoring. I welcome your suggestions.

Resolves #

/kind machinery
/priority important-soon
/cc


rzetelskik:
/hold
We need to adjust the postsubmits/periodics to build and use the catalog image.

rzetelskik:
> /hold
> We need to adjust the postsubmits/periodics to build and use the catalog image.

After internal discussions with @mflendrich - I'm adjusting this PR to only deploy operator with OLM in non-release deployment scripts. The release deployment scripts for OpenShift will keep installing operator through the manifests.

Let's keep the discussion about what to do with the release jobs outside of scope of this PR.
That renders the refactoring of operator installation unrelated to the main purpose of this PR, so please let me know if you'd like me to split it.

hack/ci-deploy.sh
```diff
@@ -28,7 +31,7 @@ fi
 SO_DISABLE_SCYLLADB_MANAGER_DEPLOYMENT=${SO_DISABLE_SCYLLADB_MANAGER_DEPLOYMENT:-false}
 SO_INSTALL_XFSPROGS_ON_NODES="${SO_INSTALL_XFSPROGS_ON_NODES:-}"
 
-mkdir -p "${ARTIFACTS_DEPLOY_DIR}/"{operator,prometheus-operator,haproxy-ingress}
+mkdir -p "${ARTIFACTS_DEPLOY_DIR}/"{prometheus-operator,haproxy-ingress}
```
mflendrich, line 34: nit: one can get rid of the repetitive `prometheus-operator` and `haproxy-ingress` literals by defining variables

```bash
ARTIFACTS_DEPLOY_DIR_PROM_OPERATOR="${ARTIFACTS_DEPLOY_DIR}/prometheus-operator"
ARTIFACTS_DEPLOY_DIR_HAPROXY_INGRESS="${ARTIFACTS_DEPLOY_DIR}/haproxy-ingress"
```

and - going forward - to untangle the order of operations a bit - create the specific directory just before using it (that is -- create the prom operator one right before putting something there; and the same for haproxy)

hack/lib/install.sh
```diff
@@ -0,0 +1,122 @@
...
+source "$( dirname "${BASH_SOURCE[0]}" )/bash.sh"
+source "$( dirname "${BASH_SOURCE[0]}" )/kube.sh"
+
+# install-operator installs the Scylla Operator and its dependencies in the cluster through the specified method.
+# $1 - root source path to use. It can either be an absolute file path or a URL.
+function install-operator() {
+  if [[ "$#" -ne 1 ]]; then
+    echo "Missing arguments. Usage: ${FUNCNAME[0]} <source_root>" > /dev/stderr
+    exit 1
+  fi
```
mflendrich, line 18: arg validation pattern inconsistent with line 16 of `hack/ci-deploy.sh`

https://github.com/scylladb/scylla-operator/pull/3116/commits/7d745c3bd13915520bd1bfad89ab9870e54f7076#diff-b93a0830431b6e0685e8abcc8e25fb21197e6aefa9365a2bb35650c4dce414a6R15

suggestion: unify

hack/.ci/manifests/olm/00_scylladb-operator.catalogsource.yaml
mflendrich, line 1: Why a hidden directory `.olm`? Hiding the directory feels counterproductive(?)

suggestion: remove the leading dot

rzetelskik: done

hack/lib/install.sh
rzetelskik: done

hack/ci-deploy.sh
rzetelskik: I don't want to get into fixing these scripts in such tiny increments - I'd consider it busywork, especially trying to keep the two versions aligned. I see two options:
- only address operator installation in this PR (as-is)
- commit to sharing all installation steps between the two scripts and reduce them to wrappers providing source paths (significantly more work, but an actual improvement)

mflendrich: ~No strong opinion here, your call (feel free to resolve if that's what you decide).
My way of thinking was that a PR that creates a function is also responsible for keeping its argument handling.
And that sanitizing the inputs of a single function is not something that needs to happen in concert with other places in the code.~

mflendrich: Ah, the above was intended for a different comment - please disregard.

As-is sounds good.

PR #3121, opened 2025-12-04
Add a script for OLM bundle postprocessing

rzetelskik:
This PR adds a script for postprocessing the OLM bundle. It replaces the operator image references with a specified image reference. This can't be done in build time and is expected to happen in the CI job responsible for building the bundle image.
In presubmits, this is required to build the OLM bundle with the actual operator image being tested (instead of whatever placeholder is in the bundle manifests).

The script also updates `createdAt` and `containerImage` annotations in the CSV. It is not required as part of testing, but I'm hoping it can be used in the publishing procedure. 

Partially addresses https://github.com/scylladb/scylla-operator-release/issues/519.

Please review together with https://github.com/scylladb/scylla-operator-release/pull/523.

/kind machinery
/priority important-soon

### DIFF
--- a/hack/olm/postprocess-bundle.sh
+++ b/hack/olm/postprocess-bundle.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# Copyright (C) 2025 ScyllaDB
+#
+# This script applies post-processing to an OLM bundle located at the specified target directory.
+
+set -euxEo pipefail
+shopt -s inherit_errexit
+
+if [[ "$#" -ne 2 ]]; then
+    echo -e "Missing arguments.\nUsage: ${0} <target> <operator_image_ref>" >&2
+    exit 1
+fi
+
+target="${1}"
+operator_image_ref="${2}"
+
+# set_metadata_annotation_without_line_comment sets a metadata annotation without a line comment in the specified YAML file.
+# $1 - target file
+# $2 - annotation key
+# $3 - annotation value
+function set_metadata_annotation_without_line_comment() {
+    local target_file="${1}"
+    local key="${2}"
+    local value="${3}"
+
+    key="${key}" value="${value}" yq -e -i '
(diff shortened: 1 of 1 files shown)


hack/olm/postprocess-bundle.sh
```diff
@@ -0,0 +1,47 @@
...
+    local value="${3}"
+
+    key="${key}" value="${value}" yq -e -i '
+      .metadata.annotations[env(key)] = env(value) |
+      .metadata.annotations[env(key)] line_comment= ""
+    ' "${target_file}"
+}
+
+# Update all YAML manifests in the bundle to use the specified operator image reference as it can't be done in build time.
+find "${target}/manifests" -name "*.yaml" -exec sed -i "s|docker\.io/scylladb/scylla-operator\([:@][^[:space:]]*\)\?|${operator_image_ref}|g" {} +
```
mflendrich: Is there a reason to do it in a hacky way (text find replace) instead of editing a well-defined set of specific fields like we do for annotations?

rzetelskik: Not really, I aligned it with what we do in the deployment script but I'd like to change it there as well. Switched to using yq for specific references.

mflendrich: Thanks.

This is a taste thing, but I would rather have two simpler `yq` commands run one after another than one attempting to make changes in multiple places. I think it would be more maintainable this way, as opposed to this behemoth of an expression. WDYT?

rzetelskik: I don't think it's going to be particularly pretty in either case but I've split it between the operator and webhook-server deployments. Couldn't really come up with a function that would reduce code repetition and maintain readability. PTAL.

mflendrich: I think this way it's more logically isolated and hence requires less headache from the reader, looks good to me 👍

PR #3131, opened 2025-12-10
Extend the OLM bundle postprocessing script with modifications required for a release

rzetelskik:
**Description of your changes:** This PR adds a script which postprocesses the OLM bundle at the specified path to prepare it for a release. It is supposed to be used in tandem with the machinery in scylla-operator-release repository as part of the certification step of the release procedure. The split beetween the scripts is arbitrary so please review together with https://github.com/scylladb/scylla-operator-release/pull/526. 

Resolves #

/kind machinery
/priority important-soon
/cc

### DIFF
--- a/hack/olm/postprocess-bundle.sh
+++ b/hack/olm/postprocess-bundle.sh
@@ -2,36 +2,85 @@
 #
 # Copyright (C) 2025 ScyllaDB
 #
-# This script applies post-processing to an OLM bundle located at the specified target directory.
+# This script applies post-processing required for a release to an OLM bundle located at the specified target directory.
 
 set -euxEo pipefail
 shopt -s inherit_errexit
 
-if [[ "$#" -ne 2 ]]; then
-    echo -e "Missing arguments.\nUsage: ${0} <target> <operator_image_ref>" >&2
-    exit 1
-fi
-
-target="${1}"
-operator_image_ref="${2}"
-
-# set_metadata_annotation_without_line_comment sets a metadata annotation without a line comment in the specified YAML file.
+# set_yaml_field_without_line_comment sets a YAML field to a value without a line comment in the specified YAML file.
 # $1 - target file
-# $2 - annotation key
-# $3 - annotation value
-function set_metadata_annotation_without_line_comment() {
+# $2 - field path (e.g., .spec.version)
+# $3 - value
+function set_yaml_field_without_line_comment() {
     local target_file="${1}"
-    local key="${2}"
+    local field_path="${2}"
     local value="${3}"
 
-    key="${key}" value="${value}" yq -e -i '
-      .metadata.annotations[env(key)] = env(value) |
-      .metadata.annotations[env(key)] line_comment= ""
+    field_path="${field_path}" value="${value}" yq -e -i '
+      (. as $in | eval(strenv(field_path))) = env(value) |
+      (eval(strenv(field_path))) line_comment= ""
     ' "${target_file}"
 }
 
+readonly script_dir="$( realpath $( dirname "${BASH_SOURCE[0]}" ) )"
+source "${script_dir}/../lib/semver.sh"
+
+readonly semver_pattern='^([0-9]+)\.([0-9]+)\.([0-9]+)(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?(\+([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?$'
+
+if [[ "$#" -ne 3 ]]; then
+  echo -e "Missing arguments.\nUsage: ${0} <target_bundle_dir> <version> <channel>" >&2
+  exit 1
+fi
+
+target_bundle_dir="${1}"
+version="${2}"
+channel="${3}"
+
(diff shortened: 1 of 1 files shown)


rzetelskik:
/cc @czeslavo 
since you were selected for https://github.com/scylladb/scylla-operator-release/pull/526

hack/olm/postprocess-bundle-release.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+#
+# Copyright (C) 2025 ScyllaDB
+#
+# This script applies post-processing required for a release to an OLM bundle located at the specified target directory.
+
+set -euxEo pipefail
+shopt -s inherit_errexit
+
+source "$( dirname "${BASH_SOURCE[0]}" )/../lib/olm.sh"
```
czeslavo: Please consider assigning `dirname "${BASH_SOURCE[0]}` to a local `SCRIPT_DIR` variable, as it's used in two places already.

hack/olm/postprocess-bundle.sh
```diff
@@ -7,29 +7,16 @@
 set -euxEo pipefail
 shopt -s inherit_errexit
 
+source "$( dirname "${BASH_SOURCE[0]}" )/../lib/olm.sh"
+
 if [[ "$#" -ne 2 ]]; then
-    echo -e "Missing arguments.\nUsage: ${0} <target> <operator_image_ref>" >&2
-    exit 1
+  echo -e "Missing arguments.\nUsage: ${0} <target> <operator_image_ref>" >&2
```
czeslavo: Not strictly related to this PR, but `target` seems quite vague. How about making it an explicit `<target_bundle_dir>`?

hack/olm/postprocess-bundle-release.sh
```diff
@@ -0,0 +1,41 @@
...
+
+target="${1}"
+version="${2}"
+
+if [[ ! "${version}" =~ ${semver_pattern} ]]; then
+  echo "Error: Version '${version}' is not semver compliant"
+  exit 1
+fi
+
+OPERATOR_IMAGE_REF="${OPERATOR_IMAGE_REF:-docker.io/scylladb/scylla-operator:${version}}"
```
czeslavo: As a sanity check, we could add a step to `postprocess-bundle.sh` to verify the specified image exists. WDYT? It might save us time spent on running the pipeline in case of a typo.

hack/olm/postprocess-bundle-release.sh
czeslavo, line 1: As discussed offline, we could not create a separate variant of the `postprocess-bundle.sh` script (to avoid multiple scripts), but instead use the same one for both cases (creating a bundle for PR testing and creating a bundle for release). Let's see if this makes sense.

hack/olm/postprocess-bundle-release.sh
rzetelskik: is there a local variable concept in bash outside of a function? We have the `parent_dir` variable in some other scripts and it actually introduces bugs if it's overridden in a sourced file

czeslavo: We could use `readonly` then to make sure the variable is not changed in the sourced files, WDYT?

hack/olm/postprocess-bundle-release.sh
rzetelskik: At this point I think it's entirely a preference thing. We'd need to mock the version and override the operator image in CI. I believe we need to figure out how to deal with channels and upgrade graphs first to see if it won't make it more convoluted.

hack/olm/postprocess-bundle.sh
rzetelskik: done

hack/olm/postprocess-bundle-release.sh
rzetelskik: done

hack/olm/postprocess-bundle-release.sh
rzetelskik: done

rzetelskik:
@czeslavo I've addressed all your comments. I also extended the script with setting the upgrade-related fields. What do you think about merging the scripts after the changes? CI jobs would need to mock `version` and override the image. They would provide the "dev" channel so most of the postprocessing wouldn't apply either way. I don't feel strongly about either option.

hack/olm/postprocess-bundle-release.sh
czeslavo: I understand that we're concluding that we can merge the scripts and inject version in CI to be something like `0.0.0-<commit-hash>` and override the image to be a specific ref, not derived from the version. I'm ok with that, so if that's also your preference, let's do this.

rzetelskik: Done. https://github.com/scylladb/scylla-operator-release/pull/530

rzetelskik:
@czeslavo this is going to fail until the operator-manifest-tools binary is available in kube-tools, but the changes are ready for review

rzetelskik:
Setting the version in https://github.com/scylladb/scylla-operator-release/pull/530 broke installing by OLM in our e2e jobs. I'll try to address this separately by changing how we build the catalog image in our tests.
/test images
/retest
/hold cancel

PR #3138, opened 2025-12-15
Fix for Unused import

mykaul:
The best way to fix this issue is to simply remove the unused import statement `import sys` from line 4 of `docs/source/conf.py`. This does not affect the existing functionality, as the module is not referenced elsewhere in the provided snippet. No other imports or code need to be altered. The only change is to delete or comment out the line containing `import sys`.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

PR #3141, opened 2025-12-16
Add copilot instructions for commit messages

mflendrich:
This PR introduces [custom instructions for Copilot](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#enabling-or-disabling-repository-custom-instructions) in the repository.

The intention is to start with very minimal and very clear prompting targeted at a specific issue. This PR is an attempt at unconfusing Copilot wrt. our highly specific commit message format that was difficult to back out of in https://github.com/scylladb/scylla-operator/pull/3138.

Note: my intention with this PR is to gradually build a prompt roughly compatible with [CONTRIBUTING.md](https://github.com/scylladb/scylla-operator/blob/master/CONTRIBUTING.md) and other similar places, but with a different audience (and therefore format) in mind, focusing on prompt engineering techniques as opposed to verbatim feeding the non-prompt-engineered raw (and not as targeted and quite obsolete at times) rules.

/kind cleanup
/priority backlog

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -8,4 +8,6 @@ The commit message should always (no exceptions) be an imperative sentence witho
 
 Prefer commit message length under 72 characters.
 
+The commit message should never include "@mentions" or hashtag ("#") references. The commit message should never include [keywords that close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).
+
 The PR title should follow the same rules as the commit message.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,11 @@
+## Copilot Instructions for ScyllaDB Operator
+
+The commit message should always (no exceptions) be an imperative sentence without a trailing dot. Some examples:
```
rzetelskik, line 3: Should we explicitly mention no keywords (https://github.com/scylladb/scylla-operator/pull/3138#issuecomment-3655737049)? Or do you think it falls under "no exceptions"?

mflendrich: Great point. Added. PTAL

PR #3142, opened 2025-12-16
Add initial Renovate configuration for updating Scylla

czeslavo:
**Description of your changes:** Adds initial Renovate configuration that will make sure anytime a minor/patch release of ScyllaDB is released, we'll get it automatically updated the morning after. 

Note: It will still require a manual step of running `make update` and committing changes, but we can get to automating this as well: either by adding an automation that will be triggered on these automatic version updates relying on a GH label (e.g., `requires-regenerating`), or by offloading this to Copilot. Still, I believe that getting the automatic PRs is a step forward compared to current manual approach.

I've verified this config on my fork. Renovate created the following PR: https://github.com/czeslavo/scylla-operator/pull/11

### DIFF
--- a/assets/config/config.yaml
+++ b/assets/config/config.yaml
@@ -1,4 +1,5 @@
 operator:
+  # renovate: datasource=docker depName=scylladb packageName=docker.io/scylladb/scylla versioning=semver
   scyllaDBVersion: "2025.3.5"
   # scyllaDBEnterpriseVersionNeedingConsistentClusterManagementOverride sets enterprise version
   # that requires consistent_cluster_management workaround for restore.
--- a/hack/check-renovate.sh
+++ b/hack/check-renovate.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# This scripts runs renovate in dry-run mode against the local config.
+# It can be used to check if changes to the config result in the expected updates.
+
+# Note: Renovate CLI output is quote verbose, so it's recommended to redirect it to a file for easier analysis.
+
+set -euxEo pipefail
+shopt -s inherit_errexit
+
+npx --yes renovate --platform=local --dry-run=full --print-config=true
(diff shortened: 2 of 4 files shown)


assets/config/config.yaml
```diff
@@ -1,4 +1,5 @@
 operator:
+  # renovate: datasource=docker depName=scylladb packageName=docker.io/scylladb/scylla versioning=semver
```
rzetelskik, line 2: any reason not to add it for `scyllaDBEnterpriseVersionNeedingConsistentClusterManagementOverride` as well?

rzetelskik, line 2: how does it update `updateFrom` and `upgradeFrom`?

czeslavo: Isn't it basically a constant version?

czeslavo: It doesn't. We'd have to automate this. BTW couldn't this could be even removed from this config and deduced in runtime in tests?

rzetelskik: >BTW couldn't this could be even removed from this config and deduced in runtime in tests?

I think so, yes, although we'd need to inspect the available tags.

For the time being, without automation, I understand it would be up to the person making the manual adjustments to update it?

rzetelskik: I believe it still gets patches, but it's about to be sunsetted with the release of 2026.1 so maybe it's not worth the effort. I'll leave it up to you.

czeslavo: Yes, it would have to be this way for now.

rzetelskik: Ok, there are no existing protections so it's fine with me.

rzetelskik: It's also only used in a single test so there is not crucial to keep it up to date.

hack/check-renovate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
rzetelskik, line 1: nit: perhaps having a README-dev.md file with a section about renovate would be more fitting than having this script (or just have both)?

czeslavo: Added a section to `CONTRIBUTING.md`, PTAL.

czeslavo:
/hold cancel

Renovate has been installed for the repository.

rzetelskik (approved):
/lgtm

PR #3145, opened 2025-12-16
Add go cache mounts to dockerfile

czeslavo:
**Description of your changes:** It adds [cache mounts](https://docs.docker.com/build/cache/optimize/#use-cache-mounts) to the Dockerfile so that golang deps and builds are cached between image rebuilds, reducing the time to rebuild the image locally.

| Scenario | Original Dockerfile | Optimized Dockerfile | Speed Improvement |
| :--- | :--- | :--- | :--- |
| **Fresh Build**<br>*(No cache / First run)* | 1m 59s | 1m 59s | None |
| **Incremental Rebuild**<br>*(One line code change)* | 1m 59s | 42s | ~2.8x Faster |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,15 @@
 FROM quay.io/scylladb/scylla-operator-images:golang-1.25 AS builder
-WORKDIR /go/src/github.com/scylladb/scylla-operator
-COPY . .
 
-RUN --mount=type=cache,target=/root/.cache/go-build \
-    --mount=type=cache,target=/go/pkg/mod \
+RUN groupadd -g 1001 scylla && \
+    useradd -u 1001 -g scylla -m scylla
+
+USER scylla:scylla
+
+WORKDIR /go/src/github.com/scylladb/scylla-operator
+COPY --chown=1001:1001 . .
+
+RUN --mount=type=cache,target=/home/scylla/.cache/go-build,uid=1001,gid=1001 \
+    --mount=type=cache,target=/go/pkg/mod,uid=1001,gid=1001 \
     make build --warn-undefined-variables
 
 FROM quay.io/scylladb/scylla-operator-images:base-ubi-9.6-minimal


Dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM quay.io/scylladb/scylla-operator-images:golang-1.25 AS builder
 WORKDIR /go/src/github.com/scylladb/scylla-operator
 COPY . .
-RUN make build --warn-undefined-variables
+
+RUN --mount=type=cache,target=/root/.cache/go-build \
```
mflendrich: optional - the `root` in the path has exposed the fact that we're building as uid 0. I don't have a strong gut feeling as to how much of a problem this is, but it stood out to me as a "bad practice of working with containers"

Thinking out loud: maybe we can make a drive-by fix and change uid to a "normal user"?

mflendrich (approved):
Is there a viable path to make this work across CI runs, too? (maybe with a workflow engine of the future)

/lgtm
/hold for comment

czeslavo:
> Is there a viable path to make this work across CI runs, too? (maybe with a workflow engine of the future)

That's a good question. My first thought was that we could store the cache in a PV, but that would imply concurrency issues with multiple writers running. Did a brief research and it looks like it may be possible with [buildah](https://kaisenlinux.org/manpages/buildah-build.html)'s `--cache-to` and `--cache-from` options, which allow pushing/reading cache to/from a dedicated image repository.

rzetelskik:
> Is there a viable path to make this work across CI runs, too? 

Wouldn't that result in non-reproducible builds?

mflendrich:
> Wouldn't that result in non-reproducible builds?

Not necessarily - caching is based on the digest of its inputs. If we're not introducing non-determinism to the system (e.g. by installing a package from an external source 😿) it should result in an identical build as without caching.

Dockerfile
czeslavo: As long as it's the right point to consider doing this, I'm worried it's not the best change to be made as a "drive-by fix" as it may have user-facing consequences (e.g., if someone was setting `securityContext.runAsUser: 0` explicitly due to some compliance requirements). I believe this would have to be planned accordingly, providing users with a migration path.

mflendrich: But I am asking about running as root in the build container - the "not last" stage of the multistage dockerfile - it does not affect the end-user.

czeslavo: Ah, sorry, I misunderstood the intent. :facepalm: Pushed a commit that uses a non-root user in the build stage.

Dockerfile
```diff
@@ -1,7 +1,14 @@
 FROM quay.io/scylladb/scylla-operator-images:golang-1.25 AS builder
+
+RUN groupadd -g 1001 scylla && \
+    useradd -u 1001 -g scylla -m scylla
+
```
mflendrich, line 7:
```suggestion

USER scylla:scylla
```

Dockerfile
mflendrich: `USER` directive missing? (see other comment)

Dockerfile
czeslavo: I could swear I had it there. :sweat_smile: PTAL

mflendrich:
Something is broken:

```
github.com/scylladb/scylla-operator/cmd/gen-api-reference: go build github.com/scylladb/scylla-operator/cmd/gen-api-reference: copying /tmp/go-build2593065982/b001/exe/a.out: open gen-api-reference: permission denied
```

czeslavo:
> Something is broken:
> 
> ```
> github.com/scylladb/scylla-operator/cmd/gen-api-reference: go build github.com/scylladb/scylla-operator/cmd/gen-api-reference: copying /tmp/go-build2593065982/b001/exe/a.out: open gen-api-reference: permission denied
> ```

Worked locally for me with `podman build`. :thinking: But reproduces with `buildah build`. That was a matter of setting `USER` after `WORKDIR` - the `scylla` user couldn't create files in the working dir, because it wasn't the owner.

mflendrich (approved):
/lgtm

PR #3151, opened 2025-12-17
Exclude `vendor/` from `.gitattributes` generated files

mflendrich:
This PR proposes that we remove `vendor/**` from the list of "generated" files in `.gitattributes`.

As for the "why": reviewing dependency bump PRs, I want the possibility to skim through the code changes (to see the most basic of issues).

Rationale:
- vendored code is not "generated" - it is real code that we are checking into the project, and therefore subject to (at least the most basic) review
- vendored code can contain changes (like licensing) that require review (and therefore hiding them is counterproductive)
- 99% of the time, edits to vendor happen in a dependency bump PR, so there's no advantage in hiding the diff in those PRs -- there is literally nothing else in them than the dependency diff

Cons:
- it makes vendored code count into "repository language statistics" as per [docs](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)

/cc czeslavo
/kind machinery
/priority backlog

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -16,7 +16,6 @@ examples/haproxy-ingress.yaml linguist-generated=true
 values.schema.json linguist-generated=true
 
 # Path prefixes
-vendor/** linguist-generated=true
 docs/source/reference/api/** linguist-generated=true
 pkg/client/** linguist-generated=true
 pkg/externalclient/** linguist-generated=true


czeslavo:
I would rather leave it as is. I consider auto-hiding these files in PRs quite convenient - you can always click "Load diff" for files that you're interested in when reviewing deps bumps, but you don't have to look into every single one (I don't think reviewing all the changes made to all the dependencies is productive). Not including them in statistics is also something I like.

Happy to hear @rzetelskik's opinion on this.

mflendrich:
> you don't have to look into every single one (I don't think reviewing all the changes made to all the dependencies is productive)

That's where the difference of perspective appears to be coming from; I usually _do_ skim through everything to at least see if there isn't a local license header in a file, or something similar.

Clicking through 20-50+ files in a PR is unreasonable, in that case I need to resort to fetching and manually viewing the diff in my terminal.

mflendrich:
> Not including them in statistics is also something I like.

I feel the same way. But there's a problem if it gets in the way of quality of review.

rzetelskik:
I'm with @czeslavo on this one. I actually feel the opposite about opening the diffs one by one - it actually makes it possible to see it in the UI, otherwise, from my experience, the rendering tends to hang or even crash. So in that case you'd still most likely need to do this manually.

mflendrich:
After a thorough internal discussion the conclusion reached was that having the vendor deps auto-expanded is not productive (because it overwhelms the GH review UI); the degree of necessary review of vendor deps is a separate subject not being considered within this PR.

PR #3157, opened 2025-12-18
Add initial implementation of envtests

czeslavo:
Adds initial implementation of envtests (i.e., `Setup` function that prepares an environment with an etcd and apiserver behind), along with the first envtest for bootstrap barrier controller.

We're aiming to make use of such tests to fill the gap in lightweight testing of controllers' integration with the Kubernetes API.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,11 +5,18 @@ linters:
     - unused
     - govet
     - importas
+    - ginkgolinter
     # More linters can be added, see supported ones: https://golangci-lint.run/usage/linters/
   exclusions:
     presets:
       - common-false-positives
       - std-error-handling
+    rules:
+      # TODO: https://github.com/scylladb/scylla-operator/issues/3190
+      - path: test/e2e/
+        linters:
+          - ginkgolinter
+        text: "use BeforeEach\\(\\) to assign variable"
   settings:
     importas:
       no-unaliased: true
@@ -31,8 +38,15 @@ linters:
           alias: oslices
         - pkg: github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1
           alias: monitoringv1
+    ginkgolinter:
+      forbid-spec-pollution: true
+      forbid-focus-container: true
+      validate-async-intervals: true
 formatters:
   enable:
     - gofmt
     - goimports
     - gci
+run:
+  build-tags:
+    - envtest
--- a/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
+++ b/test/e2e/set/nodeconfig/nodeconfig_disksetup.go
@@ -50,7 +50,7 @@ var _ = g.Describe("Node Setup", framework.Serial, func() {
 		var err error
 		matchingNodes, err := utils.GetMatchingNodesForNodeConfig(ctx, f.KubeAdminClient().CoreV1(), ncTemplate)
 		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(matchingNodes).NotTo(o.HaveLen(0))
+		o.Expect(matchingNodes).NotTo(o.BeEmpty())
 		nodeUnderTest = matchingNodes[0]
 		framework.Infof("There are %d scylla nodes. %s will be used in tests.", len(matchingNodes), nodeUnderTest.GetName())
 	})
--- a/test/e2e/set/nodeconfig/nodeconfig_optimizations.go
+++ b/test/e2e/set/nodeconfig/nodeconfig_optimizations.go
@@ -49,7 +49,7 @@ var _ = g.Describe("NodeConfig Optimizations", framework.Serial, func() {
 		var err error
 		matchingNodes, err = utils.GetMatchingNodesForNodeConfig(ctx, f.KubeAdminClient().CoreV1(), ncTemplate)
 		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(matchingNodes).NotTo(o.HaveLen(0))
+		o.Expect(matchingNodes).NotTo(o.BeEmpty())
 		framework.Infof("There are %d scylla nodes", len(matchingNodes))
 	})
 
--- a/test/e2e/set/scyllacluster/scyllacluster_sa.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_sa.go
@@ -40,7 +40,7 @@ var _ = g.Describe("ScyllaCluster", func() {
 			},
 		}, metav1.CreateOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(sa.OwnerReferences).To(o.HaveLen(0))
+		o.Expect(sa.OwnerReferences).To(o.BeEmpty())
 
 		rb, err := f.KubeClient().RbacV1().RoleBindings(f.Namespace()).Create(ctx, &rbacv1.RoleBinding{
 			ObjectMeta: metav1.ObjectMeta{
@@ -61,7 +61,7 @@ var _ = g.Describe("ScyllaCluster", func() {
 			},
 		}, metav1.CreateOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(sa.OwnerReferences).To(o.HaveLen(0))
+		o.Expect(sa.OwnerReferences).To(o.BeEmpty())
 
 		framework.By("Creating a ScyllaCluster")
 		sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
--- a/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_config.go
+++ b/test/e2e/set/scylladbcluster/multidatacenter/scylladbcluster_config.go
@@ -294,7 +294,7 @@ var _ = g.Describe("Multi datacenter ScyllaDBCluster", framework.MultiDatacenter
 			},
 		}
 
-		o.Expect(len(sc.Spec.Datacenters)).To(o.BeNumerically(">", 0))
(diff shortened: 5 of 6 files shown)


test/envtest/setup.go
```diff
@@ -0,0 +1,115 @@
...
+// It returns an Environment struct that provides access to the Kubernetes and ScyllaDB clients, as well as the test namespace
+// for convenience.
+func Setup(t *testing.T) *Environment {
+	t.Helper()
+
+	testEnv := &envtest.Environment{
+		ControlPlaneStartTimeout:    time.Minute,
+		ControlPlaneStopTimeout:     time.Minute,
+		DownloadBinaryAssets:        true,
+		DownloadBinaryAssetsVersion: "v1.35.0", // TODO: move to assets
```
rzetelskik: TODO left

test/envtest/controllers/bootstrap_barrier_test.go
```diff
@@ -0,0 +1,226 @@
...
+			controllerOpts := controllerOpts{
+				nodeServiceName:                      nodeServiceName,
+				nodesStatusReportSelector:            nodesStatusReportSelector,
+				singleReportAllowNonReportingHostIDs: singleReportAllowNonReportingHostIDs,
+				bootstrapPreconditionSatisfiedCh:     make(chan struct{}),
+			}
+			runBoostrapBarrierController(t, env, controllerOpts)
+
+			if tc.expectPreconditionMet {
+				timeout := time.After(30 * time.Second)
```
rzetelskik: nit: should we use context.WithTimeoutCause instead of separate cases for timeout and finished context?

test/envtest/controllers/bootstrap_barrier_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+			controllerOpts := controllerOpts{
+				nodeServiceName:                      nodeServiceName,
+				nodesStatusReportSelector:            nodesStatusReportSelector,
+				singleReportAllowNonReportingHostIDs: singleReportAllowNonReportingHostIDs,
+				bootstrapPreconditionSatisfiedCh:     make(chan struct{}),
+			}
+			runBoostrapBarrierController(t, env, controllerOpts)
+
+			if tc.expectPreconditionMet {
```
rzetelskik: can't the two blocks be merged? We could just compare the expected and got values, no?

czeslavo:
> My only comment is - maybe we want to structure the tests around ginkgo, like it already is in e2es? I don't want to block this PR on it; a refactor can absolutely happen in a follow-up (before too many tests get written).

I intentionally didn't use Ginkgo as I feel we all do not like its structure -  it's rather annoying than helpful in most cases. The only thing that is lacking without it and might be useful is some kind of helper function that allows waiting for conditions to be true eventually. The gap could be filled by using https://github.com/stretchr/testify as an assertions library (I might be biased here as I've used it successfully over my career, and it's the one I'm most familiar with). Happy to hear your opinions here.

/hold for discussion

test/envtest/setup.go
czeslavo: Moved to assets config. Will try to configure a custom datasource so it's automatically updated by Renovate.

Edit: added Renovate config that uses https://raw.githubusercontent.com/kubernetes-sigs/controller-tools/HEAD/envtest-releases.yaml as a custom datasource. Verified on my fork it works as expected: https://github.com/czeslavo/scylla-operator/pull/16

test/envtest/controllers/bootstrap_barrier_test.go
czeslavo: They're separate because the logic for both is slightly different:
- when `expectPreconditionMet=true`: wait until bootstrap precondition channel yields (success) or timeout (fatal)
- when `expectPreconditionMet=false`: wait 10s and expect channel to not yield (success); if it yields or the test context gets cancelled - fatal.

test/envtest/controllers/bootstrap_barrier_test.go
czeslavo: Adjusted to use `context.WithTimeoutCause` :+1:

rzetelskik:
> I feel we all do not like its structure - it's rather annoying than helpful in most cases.

My experience is that ginkgo's DSL is somewhat overengineered: we don't know how to use it correctly and many of the rules (or good practices) are not enforced and can lead to bugs in parallel suites. I wouldn't discard it completely, but if we were to use it, we'd have to watch for not repeating the mistakes present in the "e2e framework" (e.g. variable initialisation in container nodes).

czeslavo:
> My experience is that ginkgo's DSL is somewhat overengineered: we don't know how to use it correctly and many of the rules (or good practices) are not enforced and can lead to bugs in parallel suites. I wouldn't discard it completely, but if we were to use it, we'd have to watch for not repeating the mistakes present in the "e2e framework" (e.g. variable initialisation in container nodes).

After giving it more thought, I think we could give Ginkgo a try instead of introducing another testing library for assertions (like testify) to avoid too much noise. We should just use it without customisations that are not necessary (e.g., I don't think we should have a dedicated runner binary handling parallelization on its own, but just use ginkgo CLI). I don't see the existing E2Es to be migrated out of Ginkgo anytime soon realistically so we might have cases where reusing helpers compatible with Ginkgo might be possible for E2Es and envtests which also brings some value.

Here's a version of this PR using ginkgo: https://github.com/scylladb/scylla-operator/compare/master...czeslavo:envtests-ginkgo-variant 

Note: we will need to add ginkgo CLI to a CI image.

czeslavo:
I've pushed the version using Ginkgo, PTAL. Also enabled ginkgolinter in `.golangci.yaml` to cover some common pitfalls (didn't enable `forbid-spec-pollution` yet, because it affects many of the existing E2E cases - we may want to fix all existing initialization happening outside of `BeforeEach` before we enable that, I'll create a tracker).

Update: tracker for `forbid-spec-pollution` https://github.com/scylladb/scylla-operator/issues/3190.

rzetelskik:
> Also enabled ginkgolinter in .golangci.yaml to cover some common pitfalls (didn't enable forbid-spec-pollution yet, because it affects many of the existing E2E cases - we may want to fix all existing initialization happening outside of BeforeEach before we enable that, I'll create a tracker).

I highly doubt this is going to happen anytime soon, it'd require a major refactoring of the E2E framework. At the same time it'd be good to prevent it in envtests. Could we configure the linter to only work with envtests for now and redefine #3190 to enable it for E2Es and fix the existing issues?

test/envtest/controllers/bootstrap_barrier_test.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+
+	type testCase struct {
+		objects               []client.Object
+		expectPreconditionMet bool
+	}
+
+	g.DescribeTable("Bootstrap precondition evaluation", func(ctx g.SpecContext, tc testCase) {
+		env := envtest.Setup(ctx)
```
rzetelskik: if this is common to all the tests, should we use BeforeEach to initialize?

test/envtest/controllers/bootstrap_barrier_test.go
```diff
@@ -0,0 +1,206 @@
...
+var _ = g.Describe("BootstrapBarrierController", func() {
+	const (
+		nodeServiceName                      = "dc1-rack1-1"
+		nodeOrdinal                          = 1
+		nodesStatusReportSelector            = "test-selector"
+		singleReportAllowNonReportingHostIDs = true
+		rackName                             = "rack1"
+		dcName                               = "dc1"
+		hostID                               = "host1"
+	)
```
rzetelskik, line 30: Shouldn't this be picked up by linter now? Or does that only target variables, not consts?

czeslavo:
> > Also enabled ginkgolinter in .golangci.yaml to cover some common pitfalls (didn't enable forbid-spec-pollution yet, because it affects many of the existing E2E cases - we may want to fix all existing initialization happening outside of BeforeEach before we enable that, I'll create a tracker).
> 
> I highly doubt this is going to happen anytime soon, it'd require a major refactoring of the E2E framework. At the same time it'd be good to prevent it in envtests. Could we configure the linter to only work with envtests for now and redefine #3190 to enable it for E2Es and fix the existing issues?

Good idea :) Added exclusion rule and updated the issue to specifically target E2Es.

docs/source/management/monitoring/external-prometheus-on-openshift.md
```diff
@@ -33,7 +33,7 @@ You can do this by following the [official OpenShift documentation][user-workloa
 
 To allow `ScyllaDBMonitoring`-managed Grafana to access the OpenShift User Workload Monitoring Prometheus instance,
 you need to create a `ServiceAccount` and a `ClusterRoleBinding` that grants the necessary permissions. We will use its
-`ServiceAccount` token for configuring Grafana datasource. See the OpenShift's [Accessing metrics as a developer](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/monitoring/accessing-metrics#accessing-metrics-as-a-developer) 
+`ServiceAccount` token for configuring Grafana datasource. See the OpenShift's [Accessing metrics as a developer](https://docs.redhat.com/en/documentation/monitoring_stack_for_red_hat_openshift/4.20/html/accessing_metrics/accessing-metrics-as-a-developer)
```
rzetelskik, line 36: seems completely unrelated?

test/envtest/controllers/bootstrap_barrier_test.go
czeslavo: I assume consts are safe to be defined here, as any test case cannot change them.

docs/source/management/monitoring/external-prometheus-on-openshift.md
czeslavo: Yeah, it was a dead link that's been fixed in a separate PR.

PR #3159, opened 2025-12-18
Include the `sha256:` prefix in Renovate's digest matcher

### LINKED ISSUE #3154: Renovate for ScyllaDB Manager and Manager Agent
Configure Renovate to update the following values in `assets/config/config.yaml`: `scyllaDBManagerVersion` and `scyllaDBManagerAgentVersion`, similarly to how it's currently done for `scyllaDBVersion`.

`scyllaDBManagerVersion` refers to the package `docker.io/scylladb/scylla-manager`.
`scyllaDBManagerVersion` refers to the package `docker.io/scylladb/scylla-manager-agent`.

The difference between `scyllaDBVersion` and the newly added update mechanism lies in digest pinning. Check if it's possible to use the Renovate digest pinning functionality to reference the exact digests in this YAML file. If yes, do it.

mflendrich:
Follow-up to #3159, resolves #3154

Renovate logs show that the `currentDigest` format should include the `sha256:` prefix:

```json
          {
            "currentDigest": "9718a9333f9337418be6d301521ccb49ef1449ea693bb178c4b5f5e8feedaf23",
            "updates": [
              {
                "updateType": "digest",
                "newDigest": "sha256:9718a9333f9337418be6d301521ccb49ef1449ea693bb178c4b5f5e8feedaf23",
              }
            ]
          },
```

This PR makes it happen.
Regex test: https://regex101.com/r/sJ4HmU/1

/kind machinery
/priority important-soon

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,7 +15,7 @@
         "/^assets/config/config\\.yaml$/"
       ],
       "matchStrings": [
-        "#\\s*renovate:\\s*datasource=(?<datasource>\\S+)\\s+depName=(?<depName>\\S+)(\\s+packageName=(?<packageName>\\S+))?(\\s+registryUrl=(?<registryUrl>\\S+))?(\\s+versioning=(?<versioning>\\S+))?\\s*\\n\\s*\\S+:\\s*\"?(?<currentValue>[^@\"\\s]+)(?:@sha256:(?<currentDigest>[a-f0-9]+))?\"?"
+        "#\\s*renovate:\\s*datasource=(?<datasource>\\S+)\\s+depName=(?<depName>\\S+)(\\s+packageName=(?<packageName>\\S+))?(\\s+registryUrl=(?<registryUrl>\\S+))?(\\s+versioning=(?<versioning>\\S+))?\\s*\\n\\s*\\S+:\\s*\"?(?<currentValue>[^@\"\\s]+)(?:@(?<currentDigest>sha256:[a-f0-9]+))?\"?"
       ]
     }
   ]


renovate.json
```diff
@@ -15,7 +15,7 @@
         "/^assets/config/config\\.yaml$/"
       ],
       "matchStrings": [
-        "#\\s*renovate:\\s*datasource=(?<datasource>\\S+)\\s+depName=(?<depName>\\S+)(\\s+packageName=(?<packageName>\\S+))?(\\s+registryUrl=(?<registryUrl>\\S+))?(\\s+versioning=(?<versioning>\\S+))?\\s*\\n\\s*\\S+:\\s*\"?(?<currentValue>[^@\"\\s]+)(?:@sha256:(?<currentDigest>[a-f0-9]+))?\"?"
+        "#\\s*renovate:\\s*datasource=(?<datasource>\\S+)\\s+depName=(?<depName>\\S+)(\\s+packageName=(?<packageName>\\S+))?(\\s+registryUrl=(?<registryUrl>\\S+))?(\\s+versioning=(?<versioning>\\S+))?\\s*\\n\\s*\\S+:\\s*\"?(?<currentValue>[^@\"\\s]+)(?:@(?<currentDigest>sha256:[a-f0-9]+))?\"?"
```
rzetelskik, line 18: nit: we could follow https://github.com/containers/container-libs/blob/b0f86df5a66598b80ddf7d4cbd88165981d3370c/image/docker/reference/reference.go#L20 and use a regex for the algorithm instead of hardcoding to sha256 (although I have personally never seen any other algorithm in use)

rzetelskik (approved):
/lgtm
/hold for optional comment

renovate.json
mflendrich: If we wanted to do that by the book, we'd need to use the entire `digest` term from the grammar you linked which is even more complex - turns out that the digest algorithm can have special characters used as separators; I think the regexp we have here is already unreadable enough and I see more risk in overcomplicating it than in missing updates here if Docker/OCI all of a sudden changes the digest algorithm.

Learned something about the digest structure from this exercise. Leaving as-is to tip the scale towards readability of this regexp.

PR #3162, opened 2025-12-19
Load versions in docs myst_substitutions from config.yaml

czeslavo:
**Description of your changes:** Avoids manual updates to the dependencies versions in docs by reading them from the single source of truth (`assets/config/config.yaml`).

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -2,8 +2,30 @@
 import os
 from datetime import date
 import sys
+from pathlib import Path
+import yaml
 
 from sphinx_scylladb_theme.utils import multiversion_regex_builder
+
+
+def load_myst_substitutions():
+    """
+    Load dynamic configuration values from config.yaml.
+    """
+    config_path = Path(__file__).parent.parent.parent / 'assets' / 'config' / 'config.yaml'
+
+    with open(config_path, 'r') as config_file:
+        config = yaml.safe_load(config_file)
+
+    scylla_db_version = config['operator']['scyllaDBVersion']
+    scylla_db_manager_version = config['operator']['scyllaDBManagerVersion'].split('@')[0]
+
+    return {
+        "imageTag": scylla_db_version,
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -53,16 +75,17 @@
...
     "productName": "Scylla Operator",
     "repository": "scylladb/scylla-operator",
     "revision": "master",
     "imageRepository": "docker.io/scylladb/scylla",
-    "imageTag": "2025.3.5",
     "enterpriseImageRepository": "docker.io/scylladb/scylla-enterprise",
-    "enterpriseImageTag": "2025.3.5",
-    "agentVersion": "3.7.0",
-    "scyllaDBRepositoryTag": "scylla-2025.3.5",
+    **dynamic_substitutions,
```
rzetelskik, line 88: Let's verify this doesn't break multiversion - it won't show until we have two branches with this code. Sent instructions internally.

czeslavo: It doesn't. :disappointed: All branches end up with the same versions (ones from the branch that's the "builder").

czeslavo: @dgarcia360, could you please take a look and tell us whether there's any reasonable way to make something like this work with multiversion builds? We'd like to dynamically load versions from a `config.yaml` file that's different for every branch.

czeslavo: I've found an answer to "why this doesn't work". 

> Since sphinx-multiversion copies the branch data into a temporary directory and builds them there while leaving the current working directory unchanged, relative paths in your conf.py will refer to the path of the version you’re building from, not the path of the version you are trying to build documentation for.

source: https://sphinx-contrib.github.io/multiversion/main/configuration.html#overriding-configuration-variables

czeslavo:
Closing in favor of https://github.com/scylladb/scylla-operator/pull/3163.

PR #3163, opened 2025-12-19
Update Scylla and Manager Agent versions injected to docs with Renovate

czeslavo:
**Description of your changes:** Renovate-driven alternative to https://github.com/scylladb/scylla-operator/pull/3162.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -53,16 +53,22 @@ todo_include_todos = True
 myst_enable_extensions = ["colon_fence", "attrs_inline", "substitution", "attrs_block"]
 myst_heading_anchors = 6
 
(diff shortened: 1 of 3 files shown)


czeslavo:
Sample PR generated in my fork: https://github.com/czeslavo/scylla-operator/pull/11

rzetelskik (approved):
I'd personally prefer that we generate the values in conf.py based on the config.yaml but I don't think this is incorrect, as this way the tool becomes the single source of truth. The only thing I'd be worried about is the renovate config getting out of sync.
To me it seems good enough either way.

@mflendrich wdyt?

czeslavo:
Had to resolve conflicts, PTAL again.

PR #3168, opened 2025-12-23
Set explicit PAT permissions for release GH Actions workflow

mykaul:
Potential fix for [https://github.com/scylladb/scylla-operator/security/code-scanning/1](https://github.com/scylladb/scylla-operator/security/code-scanning/1)

In general, the fix is to explicitly declare `permissions` for the `GITHUB_TOKEN` in the workflow, rather than relying on repository/organization defaults. This is done by adding a `permissions:` block either at the top workflow level (so it applies to all jobs without their own `permissions`) or within the specific job. The least-privilege baseline recommended by the alert is `contents: read`, which is sufficient for checking out code and many read-only operations.

For this workflow, the simplest and least intrusive change is to add a top-level `permissions:` block just after the `on:` section (before `env:`). This will apply to the single `release-notes` job and any future jobs unless they override it. We will set:
```yaml
permissions:
  contents: read
```
No imports or additional methods are needed—this is purely a YAML configuration change within `.github/workflows/releases.yaml`. Functionality should remain unchanged for operations that only need read access to repository contents; if the custom `release-notes` action requires more, those can be added later, but that is outside the scope of the current static-analysis fix.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/releases.yaml
+++ b/.github/workflows/releases.yaml
@@ -16,6 +16,9 @@ on:
         description: 'Tag for publishing a release, like `v1.3.0`. The old release must be deleted before this job.'
         required: true
 
+permissions:
+  contents: write
+
 env:
   go_version: '1.22'
 


mflendrich:
@copilot Please check what this workflow performs and what PAT permissions are necessary for successful execution.
In particular, it calls the GH Action defined in `.github/actions/release-notes/action.yaml`.

Update the PR accordingly to set the required permissions.

mflendrich:
I am confused by the keywords check - I am not seeing any keyword in the commit message.

rzetelskik:
handled in https://github.com/scylladb/scylla-operator/pull/3170

rzetelskik:
My bad, https://github.com/scylladb/scylla-operator/pull/3170 was only merged into this branch.

rzetelskik:
@copilot update the PR description to reflect the changes

rzetelskik:
> @rzetelskik I've opened a new pull request, #3189, to work on those changes. Once the pull request is ready, I'll request review from you.

Looks like it's incapable of updating the PR description.

rzetelskik:
We've switched to changelog-based release notes and removed the workload, so this is no longer applicable. Closing.

PR #3173, opened 2025-12-23
Remove default channel annotation from OLM bundle metadata

rzetelskik:
**Description of your changes:** The default channel annotation set in the OLM bundle metadata breaks the certification pipeline if the default channel is not among channels, as the temporary catalog created in the pipeline does not contain an entry for the default channel then. We circumvented it in our CI as we're building the scratch catalog ourselves, but it's not possible to override it in the pipeline.

I propose we remove this annotation as it's optional. 
**An alternative solution would be to only set it in the post-processing script iff the provided channel is `stable`.**

Resolves #

/kind machinery
/priority important-soon
/cc mflendrich

### DIFF
--- a/bundle/metadata/annotations.yaml
+++ b/bundle/metadata/annotations.yaml
@@ -5,7 +5,6 @@ annotations:
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: scylladb-operator
   operators.operatorframework.io.bundle.channels.v1: dev # Overwritten during publishing.
-  operators.operatorframework.io.bundle.channel.default.v1: stable
   operators.operatorframework.io.metrics.builder: operator-sdk-unknown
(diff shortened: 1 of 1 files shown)


rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[199b330](https://github.com/scylladb/scylla-operator/commit/199b33031d906bca029e01e095bf3658154290cb)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3173/pull-scylla-operator-master-e2e-gke-parallel/2003440118525857792)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3173). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).
> 
> Details

/retest

Issue #3144, opened 2025-12-16
Add lightweight kind E2E test suite


czeslavo:
### Context

After the spike completed in https://github.com/scylladb/scylla-operator/pull/3119, we've decided that we want to introduce a lightweight E2E suite that will be designed to be run on kind only. Its purpose will be to serve as a go-to test suite, feasible to be run locally, as well as in CI, as a quick feedback provider.

The goal of this issue is to add a new `kind-fast` (name to be discussed) E2E suite. Tests included in it should be enough to give a reasonable smoke-test quality response whether the operator is working as expected. They should run no longer than 20 minutes in total.

### Acceptance criteria

- [ ] A new `kind-fast` E2E test suite is added
- [ ] There's a documented way of running the suite locally

PR #3185, opened 2025-12-30
Allow configuring Scylla reactor backend in E2E tests

czeslavo:
**Description of your changes:**  Unlocks ability to set reactor backend to `io_uring` which, in contrast to `linux_aio`, doesn't require IO-related sysctls tweaking (which we wouldn't be able to tweak in KinD clusters). It also results in better performance in tests, reducing their total time to run.

Prerequisite for: https://github.com/scylladb/scylla-operator/issues/3144

### DIFF
--- a/pkg/cmd/tests/options.go
+++ b/pkg/cmd/tests/options.go
@@ -38,6 +38,7 @@ type ScyllaClusterOptions struct {
 	NodesBroadcastAddressType   string
 	ClientsBroadcastAddressType string
 	StorageClassName            string
+	ReactorBackend              string
 }
 
 var supportedNodeServiceTypes = []scyllav1.NodeServiceType{
@@ -48,6 +49,11 @@ var supportedNodeServiceTypes = []scyllav1.NodeServiceType{
 var supportedBroadcastAddressTypes = []scyllav1.BroadcastAddressType{
 	scyllav1.BroadcastAddressTypePodIP,
 	scyllav1.BroadcastAddressTypeServiceClusterIP,
+}
+
+var supportedReactorBackends = []string{
+	"io_uring",
+	"linux-aio",
 }
 
 type TestFrameworkOptions struct {
@@ -78,6 +84,7 @@ func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent stri
 			NodesBroadcastAddressType:   string(scyllav1.BroadcastAddressTypePodIP),
 			ClientsBroadcastAddressType: string(scyllav1.BroadcastAddressTypePodIP),
 			StorageClassName:            "",
+			ReactorBackend:              "", // Leaving empty to rely on ScyllaDB default.
 		},
 		ObjectStorageOptions:        NewObjectStorageOptions(),
 		ScyllaDBVersion:             configassets.Project.Operator.ScyllaDBVersion,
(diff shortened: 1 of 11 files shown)


mflendrich (commented):
If:
- using aio requires sysctls
- aio is the default the user will get

and this PR switches the default away from aio to avoid the need to configure sysctls

then we'll lose the e2e-on-GKE test coverage of verifying that our environment configuration is good enough to make scylla in its default configuration happy.

I am not yet convinced that loss of that coverage on GKE is justified, because it can lead to dead-on-arrival setups in the future. I would like to have some kind of defense against dead-on-arrival.

Some ideas:
- switch from aio to uring only for kind (and keep the default as is)
- go forward with the change of the default from aio to uring, but implement a smoke test for the default configuration

czeslavo:
> If:
> 
>     * using aio requires sysctls
> 
>     * aio is the default the user will get
> 
> 
> and this PR switches the default away from aio to avoid the need to configure sysctls
> 
> then we'll lose the e2e-on-GKE test coverage of verifying that our environment configuration is good enough to make scylla in its default configuration happy.
> 
> I am not yet convinced that loss of that coverage on GKE is justified, because it can lead to dead-on-arrival setups in the future. I would like to have some kind of defense against dead-on-arrival.

Fair enough. Let's stick to the default for now in all existing tests and use `uring` only with kind. Changed this PR to not change the default, but allow setting the reactor backend via a test runner arg.

test/e2e/fixture/scylla/scylladbcluster.yaml.tmpl
```diff
@@ -14,6 +14,10 @@ spec:
   scyllaDB:
     image: "{{ .scyllaDBRepository }}:{{ .scyllaDBVersion }}"
     enableDeveloperMode: true
+    {{- if .reactorBackend }}
+    additionalScyllaDBArguments:
+    - "--reactor-backend={{ .reactorBackend }}"
+    {{- end }}
```
mflendrich: Presence of `additionalScyllaDBArguments` in the template is tied to a custom `reactorBackend` being set. This means that if we add another additional argument (that may never happen in the future, though), we'll be stuck resolving debt that originated from rolling `additionalScyllaDBArguments` under the `reactorBackend` presence conditional.

I am wondering if there's an elegant way to move `additionalScyllaDBArguments` out of the conditional. One approach that comes to my mind is something along the lines of 

```yaml
    additionalScyllaDBArguments: [
        {{- if .reactorBackend }}
        "--reactor-backend={{ .reactorBackend }}",
        {{- end }}
    ]
```

By a quick glance at the yaml spec, it should work ([trailing comma is allowed in flow-style collections](https://yaml.org/spec/1.2.2/#74-flow-collection-styles)).

pkg/cmd/tests/options.go
```diff
@@ -78,6 +84,7 @@ func NewTestFrameworkOptions(streams genericclioptions.IOStreams, userAgent stri
 			NodesBroadcastAddressType:   string(scyllav1.BroadcastAddressTypePodIP),
 			ClientsBroadcastAddressType: string(scyllav1.BroadcastAddressTypePodIP),
 			StorageClassName:            "",
+			ReactorBackend:              "",
```
mflendrich: An empty value appears to be outside of `supportedReactorBackends`. What does it mean here?

I am asking to understand. Whatever the answer is - might be worth a comment.

test/e2e/fixture/scylla/scylladbcluster.yaml.tmpl
czeslavo: Changed it slightly to make the addition of potential new args less problematic (only one place - `ScyllaClusterOptions.ScyllaArgs()` method - has to be changed now).

pkg/cmd/tests/options.go
czeslavo: The intention is to leave the default empty to make Scylla use its own default. If it ever changes, we will consume it automatically. We want to test the default Scylla uses, not the one we think is the default.

czeslavo: Left a comment.

PR #3186, opened 2025-12-30
Add kind-fast suite

### LINKED ISSUE #3144: Add lightweight kind E2E test suite

czeslavo:
**Description of your changes:** It adds a `kind-fast` E2E suite that's intended to be run on containerized KinD clusters, along with helper scripts for setting up such clusters locally, running the tests on them, and tearing them down.

Note to reviewers: please try it out in your local environment if possible.

~Please also note that currently `kind` 0.31.0 with Kubernetes v1.35.0 node images doesn't work (kubelet doesn't boot up). Please use `kind` 0.30.0 until we figure it out.~ It's no longer the case - turned out I was relying on containerd config format that has been dropped in containerd >= 2.0, which kind v0.31.0 uses. Worked this around by setting the local registry host mapping manually in each kind cluster node with `podman exec`.

Resolves #3144

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,50 @@ it's up to the developer to ensure they are available in the environment:
   (**Note:** Until [#6978](https://github.com/operator-framework/operator-sdk/pull/6978) is merged and released, you
   need to use `operator-sdk` from the PR branch; `quay.io/scylladb/scylla-operator-images:golang-1.25` image has it preinstalled)
 - [ginkgo](https://onsi.github.io/ginkgo/#getting-started) - for running envtests.
+- [podman](https://podman.io/get-started) - for building and running container images.
+- [kind](https://kind.sigs.k8s.io/docs/user/quick-start/) - for running local Kubernetes clusters for E2E tests.
+
+### Running E2E tests locally
+
+We use [Kind](https://kind.sigs.k8s.io/) for basic local testing. There's a Makefile target to create a Kind cluster with a local registry configured
+so we can use locally built images. To set up the Kind cluster, run:
+
+```bash
+make kind-setup
+```
+
+You can then run the E2E tests (`kind-fast` suite) with:
+
+```bash
+make test-e2e-kind
+```
+
+The tests will run against the Kind cluster created in the previous step. A pod running the tests will be created in the `e2e` namespace.
+A ScyllaDB Operator image based on the local code will be built and loaded into the Kind cluster automatically.
+
+You can skip building if you choose the Scylla Operator image to be used in the tests by setting the `SO_IMAGE` environment variable. For example:
+
+```bash
+SO_IMAGE="quay.io/scylladb/scylla-operator:latest" make test-e2e-kind
+``` 
+
+Please note that the image must be accessible from within the Kind cluster. If you're using a locally built image, you should:
+
+1. Run `make kind-setup` first. 
+2. Tag your image with `localhost:5001/your-image-name:tag`. 
+3. Push it to the local registry: `podman push --tls-verify=false localhost:5001/your-image-name:tag`.
+
+To run specific test cases, use the `SO_FOCUS` environment variable (it accepts [a regexp filtering ginkgo test cases](https://onsi.github.io/ginkgo/#description-based-filtering)). For example: 
+
+```bash
+SO_FOCUS=".*TestCaseName.*" make test-e2e-kind
+```
+
+If you ever need to start from a clean state, you can delete the Kind cluster with:
+
+```bash
+make kind-teardown
+```
 
 ## Coding convention
 


test/e2e/framework/meta.go
```diff
@@ -14,6 +14,7 @@ const (
 	RequiresObjectStorageLabelName    = "RequiresObjectStorage"
 	NotSupportedOnOpenShiftLabelName  = "NotSupportedOnOpenShift"
 	SupportedOnlyOnOpenShiftLabelName = "SupportedOnlyOnOpenShift"
+	NotSupportedOnKindLabelName       = "NotSupportedOnKind"
```
rzetelskik, line 17: it could use a comment explaining why the tests are not supported as from what I see there are multiple reasons (either here or on particular tests)

Makefile
```diff
@@ -5,8 +5,12 @@ SHELL :=/bin/bash -euEo pipefail -O inherit_errexit
...
 
 IMAGE_TAG ?= latest
-IMAGE_REF ?= docker.io/scylladb/scylla-operator:$(IMAGE_TAG)
-BUNDLE_IMAGE_REF ?= docker.io/scylladb/scylla-operator-bundle:$(IMAGE_TAG)
+REGISTRY ?= docker.io
+IMAGE_REPO ?= scylladb/scylla-operator
+BUNDLE_REPO ?= scylladb/scylla-operator-bundle
+
+IMAGE_REF ?= $(REGISTRY)/$(IMAGE_REPO):$(IMAGE_TAG)
+BUNDLE_IMAGE_REF ?= $(REGISTRY)/$(BUNDLE_REPO):$(IMAGE_TAG)
```
rzetelskik: Why the change? I don't see the vars being overriden anywhere?

hack/kind/run-e2e-tests.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
rzetelskik, line 1: why not put this alongside the existing scripts for running e2es?

hack/kind/cluster-setup.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [ -z "${CLUSTER_NAME}" ]; then
+  echo "CLUSTER_NAME must be set" > /dev/stderr
+  exit 1
+fi
+
+# Ensure there's a `kind` IPv4 network.
+if ! podman network inspect kind >/dev/null 2>&1; then
+  echo "Creating kind IPv4-only network..."
+  podman network create kind
```
rzetelskik, line 27: shouldn't this also be deleted if RECREATE is set?

hack/kind/cluster-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+        kind: InitConfiguration
+        nodeRegistration:
+          kubeletExtraArgs:
+            feature-gates: KubeletInUserNamespace=true
+  - role: worker
+    labels:
+      scylla.scylladb.com/node-type: scylla
+  - role: worker
+    labels:
+      scylla.scylladb.com/node-type: scylla
```
rzetelskik, line 49: should we also taint these nodes? And have a separate node for other components?

Makefile
czeslavo: It's an artifact of my initial attempt, where I wanted to rely on these variables when building an image locally, but later I realized it won't be handy from the UX perspective, and made the makefile targets self-contained.

Reverted the changes to not introduce noise.

hack/kind/run-e2e-tests.sh
czeslavo: I was thinking about it, but decided to keep all kind-related stuff in one place. It also didn't really fit the real use case for this script, as it's used both in the Makefile and in CI, while the other scripts are placed under `hack/.ci/`, which suggests they're meant exclusively for CI.

test/e2e/framework/meta.go
czeslavo: Added comments and a separate `LongRunning` label, as that might be used more universally in the future.

hack/kind/cluster-setup.sh
czeslavo: Fixed.

hack/kind/cluster-config.yaml
czeslavo: Added taints to Scylla-dedicated nodes, and also added a worker for other components.

pkg/cmd/tests/tests.go
```diff
@@ -88,6 +88,23 @@ var Suites = ginkgotest.TestSuites{
...
+	{
+		Name: "kind-fast",
+		Description: templates.LongDesc(`
+		Relatively fast tests that can be run on kind clusters.
+		`),
+		LabelFilter: fmt.Sprintf(
+			"!%s && !%s && !%s && !%s && !%s && !%s && !%s",
+			framework.NotSupportedOnKindLabelName,
+			framework.LongRunningLabelName,
+			framework.RequiresObjectStorageLabelName,
```
rzetelskik, line 100: is this due to some restrictions in kind or just a missing local object storage setup? I suppose in the long run we'd like to be able to test it locally? It would actually be a good improvement as now it's somewhat annoying to set it up

czeslavo:
@rzetelskik I've just pushed a fix for kind v0.31.0 not working with our config. Turned out I was relying on the containerd config format that has been dropped in containerd >= 2.0, which kind v0.31.0 uses. Worked this around by setting the local registry host mapping manually in each kind cluster node with podman exec.

/hold cancel
I've verified the presubmit with mkpj, so I think it's fine to go ahead and merge this PR.

pkg/cmd/tests/tests.go
czeslavo: The latter. I didn't want to tackle everything at once. If you think this could be useful, I can create a tracking issue to set up some local object storage to unblock this.

hack/kind/cluster-setup.sh
```diff
@@ -0,0 +1,74 @@
...
+if [ "$(podman inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
+  podman run \
+    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --replace --network bridge --name "${reg_name}" \
+    registry:2
+fi
+
+# Add the registry config to the KinD nodes.
+REGISTRY_DIR="/etc/containerd/certs.d/localhost:${reg_port}"
+for node in $(kind get nodes --name="${CLUSTER_NAME}"); do
+  podman exec "${node}" mkdir -p "${REGISTRY_DIR}"
```
rzetelskik: isn't it possible to just configure a volume mount in the cluster config manifest?

czeslavo: It is possible - I've changed the cluster config to use mounts instead of this manipulation. Not sure it's better this way, but

pkg/cmd/tests/tests.go
rzetelskik: IIRC @mflendrich was very strongly against creating tracking issues

PR #3195, opened 2026-01-12
Fix renovate not calling go mod tidy/vendor and updating replaced deps

czeslavo:
**Description of your changes:** Adds `postUpdateOptions` for running `go mod tidy` and `go mod vendor` before submitting PRs with dependencies' updates. Also, adds a package rule to stop updating deps that are `replace`d in go.mod as these are intended to stay on a fixed version.

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -1942,6 +1942,6 @@ sigs.k8s.io/structured-merge-diff/v6/value
 ## explicit; go 1.22
 sigs.k8s.io/yaml
 sigs.k8s.io/yaml/kyaml
-# github.com/gocql/gocql => github.com/scylladb/gocql v1.14.5
 # github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring => github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.86.1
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -10,11 +10,23 @@
...
+  ],
   "packageRules": [
     {
       "description": "scylla-manager and scylla-manager-agent must be updated together",
       "matchPackageNames": ["scylla-manager", "scylla-manager-agent"],
       "groupName": "scylla-manager"
+    },
+    {
+      "description": "replaced go.mod dependencies should not be updated",
+      "matchPackageNames": [
```
mflendrich, line 25: 1. Why is `github.com/gocql/gocql` not included here? (Please help me understand, and persist that understanding in a comment)
2. This collection needs to be kept in sync with the list of `replace` directives in `go.mod`. (suggestion: make it happen somehow, e.g. by annotating the `replace` block in `go.mod` with a comment)

czeslavo: Re 1. Added a comment to the `go.mod` to explain this.
Re 2. Added a comment to the `go.mod` suggesting that. 

I've also changed the Renovate package rule name to `go.mod dependencies that should not be updated` so it better reflects reality and not suggest that _all_ replaced modules have to have fixed versions.

PR #3196, opened 2026-01-12
Update controller-gen to 0.20.0 along with its deps

czeslavo:
**Description of your changes:** Updates `sigs.k8s.io/controller-tools/cmd/controller-gen` from `0.18.0` to `0.20.0`. It required manual intervention by bumping `google.golang.org/genproto/googleapis/api` and `google.golang.org/genproto/googleapis/rpc` as otherwise we'd be getting the following error (similar to https://github.com/open-telemetry/opentelemetry-collector/issues/10476):

```console
 go: sigs.k8s.io/controller-tools/cmd/controller-gen imports
sigs.k8s.io/controller-tools/pkg/crd tested by
sigs.k8s.io/controller-tools/pkg/crd.test imports
k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation imports
github.com/google/cel-go/cel imports
google.golang.org/genproto/googleapis/api/expr/v1alpha1: ambiguous import: found package google.golang.org/genproto/googleapis/api/expr/v1alpha1 in multiple modules:
google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1 (/runner/cache/others/go/pkg/mod/google.golang.org/genproto@v0.0.0-20230410155749-daa745c078e1/googleapis/api/expr/v1alpha1)
google.golang.org/genproto/googleapis/api v0.0.0-20251029180050-ab9386a59fda (/runner/cache/others/go/pkg/mod/google.golang.org/genproto/googleapis/api@v0.0.0-20251029180050-ab9386a59fda/expr/v1alpha1)
go: sigs.k8s.io/controller-tools/cmd/controller-gen imports
sigs.k8s.io/controller-tools/pkg/crd tested by
sigs.k8s.io/controller-tools/pkg/crd.test imports
k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation imports
k8s.io/apiserver/pkg/util/webhook imports
k8s.io/component-base/tracing imports
go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc imports
go.opentelemetry.io/proto/otlp/collector/trace/v1 imports
github.com/grpc-ecosystem/grpc-gateway/v2/runtime imports
google.golang.org/genproto/googleapis/api/httpbody: ambiguous import: found package google.golang.org/genproto/googleapis/api/httpbody in multiple modules:
[...]

PR #3201, opened 2026-01-15
Update go to 1.25 and adapt to internal crypto changes

czeslavo:
**Description of your changes:** It solves an issue with the leaf certificates generation loop that manifests after updating Go to 1.25. 

TLDR; We were relying on [this](https://github.com/golang/go/blob/2bc4315d92a70d9a5e895d60defba4f799798806/src/crypto/x509/x509.go#L1755) behavior [here](https://github.com/scylladb/scylla-operator/blob/82bee228414498dead4626a7d114a7ed9df4ff03/pkg/kubecrypto/certs.go#L129-L128), while in go 1.25 it changed to [this](https://github.com/golang/go/blob/2bc4315d92a70d9a5e895d60defba4f799798806/src/crypto/x509/x509.go#L1762).

Go 1.25 changed the way `x509.CreateCertificate` populates `SubjectKeyId` field when it's not set explicitly in a template. Before 1.25, it was using the SHA-1 of a public key, while afterwards, it switched to the truncated SHA-256. See [Go 1.25 changelog](https://go.dev/doc/go1.25) and [issue](https://github.com/golang/go/issues/71746).

Our code was relying on this internal behavior in the `getAuthorityKeyIDFromSignerKey` function by assuming this key is computed as the SHA-1 of a public key.

This causes the Operator to constantly regenerate certificates because the actual (generated by `x509.CreateCertificate`) and expected (computed by the Operator based on the CA public key)  `AuthorityKeyId` never match.

Unblocks https://github.com/scylladb/scylla-operator/pull/3196 and consequently https://github.com/scylladb/scylla-operator/pull/3195.

### DIFF
--- a/pkg/crypto/signers.go
+++ b/pkg/crypto/signers.go
@@ -18,7 +18,7 @@ var (
 
 type Signer interface {
 	Now() time.Time
-	GetPublicKey() *rsa.PublicKey
+	GetSubjectKeyID() []byte
 	SignCertificate(template *x509.Certificate, requestKey *rsa.PublicKey) (*x509.Certificate, error)
 	VerifyCertificate(cert *x509.Certificate) error
 }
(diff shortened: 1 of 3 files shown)


rzetelskik (approved):
Good catch, well done! Thanks.
/lgtm

PR #3209, opened 2026-01-20
Generate renovate package rule from gomod replace directives

czeslavo:
**Description of your changes:** Adds a `hack/sync-renovate-config-with-go-mod-replace.sh` that enforces `go.mod dependencies that should not be updated (rule auto-generated by make update-renovate-config)` package rule to include LHSes of all replace directives in `go.mod` in the `matchPackageNames` array.

It wires the script up with `make update/verify` targets.

Addresses https://github.com/scylladb/scylla-operator/pull/3181#issuecomment-3772289503

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -230,16 +230,12 @@ tool (
 	sigs.k8s.io/controller-tools/cmd/controller-gen
 )
 
-// prometheus-operator modules are intentionally kept in sync with the version defined in `assets/metadata/metadata.yaml`.
-// Synchronization is performed automatically by the `make update-go-mod-replace` target, do not edit manually.
-//
-// If any other module version needs to be fixed to a specific version, please make sure it's also added to the `renovate.json`'s
-// `go.mod dependencies that should not be updated` package rule.
 replace (
+	// Use ScyllaDB's maintained fork of gocql.
+	github.com/gocql/gocql => github.com/scylladb/gocql v1.17.1
+
+	// prometheus-operator modules are intentionally kept in sync with the version defined in `assets/metadata/metadata.yaml`.
+	// Synchronization is performed automatically by the `make update-go-mod-replace` target, do not edit manually.
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring => github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.86.1
 	github.com/prometheus-operator/prometheus-operator/pkg/client => github.com/prometheus-operator/prometheus-operator/pkg/client v0.86.1
 )
-
-// Use ScyllaDB's maintained fork of gocql. It's not included in the `renovate.json`'s `go.mod dependencies that should not be updated` package rule
-// because we intend to keep track of `github.com/scylladb/gocql` updates.
-replace github.com/gocql/gocql => github.com/scylladb/gocql v1.17.1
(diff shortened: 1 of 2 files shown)


hack/sync-renovate-config-with-go-mod-replace.sh
```diff
@@ -0,0 +1,52 @@
...
+    mv "${renovate_config_file}.tmp" "${renovate_config_file}"
+}
+
+readarray -t modules < <(get_replace_lhs_modules)
+
+if [[ ${#modules[@]} -eq 0 ]]; then
+    echo "No replace directives found in ${go_mod_file}"
+    modules_json="[]"
+else
+    # Convert bash array to JSON array.
```
rzetelskik: Can't get_replace_lhs_modules just return a json array? We could avoid this back and forth

Makefile
```diff
@@ -726,10 +726,22 @@ update-go-mod-replace:
...
 
-verify: verify-codegen verify-crds verify-helm-schemas verify-helm-charts verify-deploy verify-lint verify-helm-lint verify-links verify-examples verify-docs-api verify-monitoring verify-bundle
+update-renovate-config:
+	./hack/sync-renovate-config-with-go-mod-replace.sh
+.PHONY: update-renovate-config
+
+verify-renovate-config: tmp_dir :=$(shell mktemp -d)
+verify-renovate-config:
+	cp renovate.json "$(tmp_dir)/renovate.json.original"
+	./hack/sync-renovate-config-with-go-mod-replace.sh
```
rzetelskik: I'd say that verify shouldn't modify the source file at all, you may unexpectedly end up in a broken state that way. Wouldn't it be better if the script took a path to a target file as an argument?

hack/sync-renovate-config-with-go-mod-replace.sh
czeslavo: Simplified the script a bit following this suggestion. :+1:

Makefile
czeslavo: Good point, refactored the script to accept a target config file.

hack/sync-renovate-config-with-go-mod-replace.sh
```diff
@@ -0,0 +1,39 @@
...
+        .packageRules | map(
+            if .description == "'"${packageRuleDescription}"'" then
+                .matchPackageNames = $modules
+            else
+                .
+            end
+        )
+    )
+' "${renovate_config_file}" > "${renovate_config_file}.tmp"
+mv "${renovate_config_file}.tmp" "${renovate_config_file}"
```
rzetelskik, line 37: is the mv necessary? Can't we edit inplace?

czeslavo: `jq` doesn't support in-place edits out of the box, thus using `mv` is the most obvious solution for me.

https://stackoverflow.com/questions/36565295/jq-to-replace-text-directly-on-file-like-sed-i

rzetelskik: TIL, thanks

rzetelskik:
/lgtm
Thanks for the updates!

czeslavo:
@rzetelskik I had to rebase onto master.

czeslavo:
@rzetelskik Had to rebase once again :sweat_smile:

PR #3212, opened 2026-01-20
Define internal resources in ClusterServiceVersion metadata

rzetelskik:
**Description of your changes:** Our internal API resources are now showing up in the "Provided APIs" in OpenShift console. We should mark them as internal to save the users the confusion.

image

Resolves #

/kind machinery
/priority important-soon


rzetelskik:
> Edit: // +kubebuilder:metadata:annotations/labels [can be used](https://book.kubebuilder.io/reference/markers/crd) on a type to add annotations/labels on a CustomResourceDefinition generated by controller-gen. We could add something like operator.scylladb.com/internal-api-type: true and rely on it in such cases.

Good idea, I'll try to implement this in this PR

PR #3223, opened 2026-01-26
Increase CPU limit to 50m for the scylladb-api-status-probe and scylladb-ignition containers

### LINKED ISSUE #2935: Configurable limits and requests for Memory and CPU, for scylladb-api-status-probe container via Helm Chart [CPU Throttling]

It should be possible to configure limits and requests for the scylladb-api-status-probe container via Helm chart

Unfortunately default settings ([those are not configurable](https://github.com/scylladb/scylla-operator/blob/master/pkg/controller/scylladbdatacenter/resource.go#L928-L937)) causes CPU Throttling for the scylladb-api-status-probe container in Kubernetes environments what causes alerts. The increased CPU limits may eliminate this problem. CPU Throttling can disrupt performance of ScyllaDB as a whole. It would be very useful to configure such limits, requests of containers within ScyllaDB pods to the condition of cluster that we have.

_No response_

sjoerdmulder:
Updated CPU limit from 10m to 50m in resource requirements for the containers
* scylladb-api-status-probe
* scylladb-ignition 

Resolves #2935

### DIFF
--- a/pkg/controller/scylladbdatacenter/resource.go
+++ b/pkg/controller/scylladbdatacenter/resource.go
@@ -950,11 +950,11 @@ wait
 							},
 							Resources: corev1.ResourceRequirements{
 								Limits: corev1.ResourceList{
-									corev1.ResourceCPU:    resource.MustParse("10m"),
+									corev1.ResourceCPU:    resource.MustParse("50m"),
 									corev1.ResourceMemory: resource.MustParse("40Mi"),
 								},
 								Requests: corev1.ResourceList{
-									corev1.ResourceCPU:    resource.MustParse("10m"),
+									corev1.ResourceCPU:    resource.MustParse("50m"),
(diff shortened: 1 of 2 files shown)


pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -950,7 +950,7 @@ wait
 							},
 							Resources: corev1.ResourceRequirements{
 								Limits: corev1.ResourceList{
-									corev1.ResourceCPU:    resource.MustParse("10m"),
+									corev1.ResourceCPU:    resource.MustParse("50m"),
```
rzetelskik, line 953: We're targeting a guaranteed QoS, so all of the containers in the pod need their limits to be equal to their requests, please update the requested value as well.

rzetelskik:
@sjoerdmulder thanks for the updates! Please squash the commits and update the commit subject line (the PR title is good now) and we're good to go

PR #3224, opened 2026-01-27
Add a ScyllaDB Operator installation guide in Red Hat OpenShift

rzetelskik:
**Description of your changes:** This PR introduces an installation guide for Red Hat OpenShift clusters.
Two installation methods are covered: OpenShift Container Platform web console and CLI.
The installation overview is modified to accommodate the new deployment method.

*Note*: this PR doesn't include upgrade steps for operator installed through OLM.

Resolves #

/priority important-soon
/kind documentation

### DIFF
--- a/docs/source/.internal/verify-operator-installation-gitops.md
+++ b/docs/source/.internal/verify-operator-installation-gitops.md
@@ -1,0 +1,24 @@
+Wait for CRDs to propagate to all apiservers:
+:::{code-block} shell
+kubectl wait --for='condition=established' crd/scyllaclusters.scylla.scylladb.com crd/nodeconfigs.scylla.scylladb.com crd/scyllaoperatorconfigs.scylla.scylladb.com crd/scylladbmonitorings.scylla.scylladb.com
+:::
+
+**Expected output**:
+:::{code-block} console
+customresourcedefinition.apiextensions.k8s.io/scyllaclusters.scylla.scylladb.com condition met
+customresourcedefinition.apiextensions.k8s.io/nodeconfigs.scylla.scylladb.com condition met
+customresourcedefinition.apiextensions.k8s.io/scyllaoperatorconfigs.scylla.scylladb.com condition met
+customresourcedefinition.apiextensions.k8s.io/scylladbmonitorings.scylla.scylladb.com condition met
+:::
+
+Wait for the components to roll out:
+:::{code-block} shell
+kubectl -n=scylla-operator rollout status --timeout=10m deployment.apps/{scylla-operator,webhook-server}
+:::
+
+**Expected output**:
+:::{code-block} console
+deployment "scylla-operator" successfully rolled out
+deployment "webhook-server" successfully rolled out
+:::
+
--- a/docs/source/installation/openshift.md
+++ b/docs/source/installation/openshift.md
@@ -1,0 +1,310 @@
+# Install ScyllaDB Operator in Red Hat OpenShift
+
+ScyllaDB Operator is a Red Hat OpenShift Certified Operator available for installation through the embedded software catalog, 
+a user interface for discovering Operators working in conjunction with Operator Lifecycle Manager (OLM), which installs and manages Operators in a cluster.
+This guide describes how to install ScyllaDB Operator from the software catalog in a Red Hat OpenShift cluster.
+
+## Prerequisites
+
+This guide requires you to have access to an OpenShift Container Platform cluster using an account with `cluster-admin` permissions.
+
+Ensure that your OpenShift cluster meets the [general prerequisites](./kubernetes-prerequisites.md) for ScyllaDB Operator installation.
+
+Additionally, ensure that you have `kubectl` or OpenShift CLI (`oc`) installed.
+
+:::{tip}
+Commands in this guide can be executed using the OpenShift CLI (`oc`) in place of `kubectl`.
+:::
+
+## Install ScyllaDB Operator
+
+ScyllaDB Operator can be installed in a Red Hat OpenShift cluster from the software catalog through one of the below methods: OpenShift Container Platform web console or CLI. 
+
+:::::{tabs}
+
+::::{group-tab} Web Console
+
+This procedure describes how to install and subscribe to ScyllaDB Operator from software catalog by using the OpenShift Container Platform web console.
+
+:::{note}
+This procedure aims to follow the generic Operator installation steps outlined in the upstream documentation: [Installing from the software catalog by using the web console](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/operators/administrator-tasks#olm-installing-from-software-catalog-using-web-console_olm-adding-operators-to-a-cluster).
+You can refer to it for more detailed instructions, or follow the below steps for a more contained, ScyllaDB Operator specific guidance.
+:::
+
+### Procedure
+
+1. Navigate to `Ecosystem` -> `Software Catalog`.
+2. Search for ScyllaDB Operator and select the `Certified` version by setting the `Source` filter to `Certified` or ensuring that the ScyllaDB Operator tile has the `Certified` tag next to it.
+3. In the description dialog, read the information about the operator and click `Install`.
+4. In the `Install Operator` dialog, configure your ScyllaDB Operator installation:
+   * For clusters on cloud providers with token-based authentication:
+     - AWS Security Token Service (STS Mode in the web console): enter the Amazon Resource Name (ARN) of the AWS IAM role of your service account in the role ARN field.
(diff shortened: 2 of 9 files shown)


docs/source/installation/openshift.md
```diff
@@ -0,0 +1,190 @@
+# Install {{productName}} in Red Hat OpenShift
```
mflendrich: This is how it looks in the sidebar:

Image

The first two options are simple terms, the third option is a full sentence.
The first two are brief and concise, the third one is elaborate.

To maintain consistency I think this should skip the "Install blablabla in" part and be  _OLM (OpenShift)_ or just _OpenShift_.

docs/source/installation/openshift.md
```diff
@@ -0,0 +1,190 @@
...
+
+:::{note}
+This procedure aims to follow the generic Operator installation steps outlined in the upstream documentation: [Installing from OperatorHub by using the web console](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/operators/administrator-tasks#olm-installing-from-operatorhub-using-web-console_olm-adding-operators-to-a-cluster).
+You can refer to it for more detailed instructions, or follow the below steps for a more contained, {{productName}} specific guidance.
+:::
+
+### Procedure
+
+1. Navigate to `Operators` -> `OperatorHub`.
+2. Search for {{productName}} and select the `Certified` version.
```
mflendrich:
```suggestion
2. Search for {{productName}} and select the `Certified` version (by setting the `Source` filter to `Certified` or ensuring that the ScyllaDB Operator tile has the `Certified` tag next to it).
```

docs/source/installation/openshift.md
```diff
@@ -0,0 +1,190 @@
...
+      startingCSV: scylladb-operator.{{latestStableRelease}}
+    EOF
+    :::
+
+::::
+:::::
+
+### Verify the {{productName}} installation
+
+:::{include} ./../.internal/verify-operator-installation-gitops.md
```
mflendrich, line 213: Why this and not wait for the `Succeeded` condition on the CSV?

And if we want to wait for the deployment, then why for its creation and not condition=Available?

docs/source/installation/openshift.md
```diff
@@ -0,0 +1,190 @@
...
+:::
+
+## Set up local storage on dedicated nodes and enable tuning
+
+{{productName}} enables local storage configuration and performance tuning through the [`NodeConfig`](../resources/nodeconfigs.md) resource.
+The below instructions provide a recommended configuration for a subset of Red Hat OpenShift deployment models and underlying platforms.
```
mflendrich: I think we should be precise as to what subset we're targetting. Saying "a subset" does not help the user if they don't know what subset it is.

We can solve this by saying something along the lines of
"for an example Red hat OpenShift deployment where xxxxx, yyyyy and zzzzz."

docs/source/installation/openshift.md
```diff
@@ -0,0 +1,190 @@
...
+
+:::{caution}
+Local storage configuration depends on the OpenShift deployment model and the underlying platform and infrastructure.
```
mflendrich, line 224: Similar to the above - this is too vague to be useful. we can chat to figure out how to make it more streamlined - that is, allow them to complete the guide without referring to unspecified decision-making or sources of knowledge

docs/source/installation/openshift.md
rzetelskik: > Why this and not wait for the Succeeded condition on the CSV?

I wanted to 1) share it between guides 2) do what we do in CI, but I suppose I can remove waiting for creation and make waiting for CSV specific to the openshift guide (and leave the remaining commands?).

> And if we want to wait for the deployment, then why for its creation and not condition=Available?

There's a wait for the condition earlier, without the wait for creation it's racy

docs/source/installation/openshift.md
rzetelskik: Isn't the subset obvious from the tabs that are present? Currently there is only one tab for ROSA.

docs/source/installation/openshift.md
rzetelskik: I'm happy to hear your ideas, I don't know how to cover a somewhat undefined set of environments

docs/source/installation/openshift.md
rzetelskik: I'd like to leave the actual header of the article as is, but I'll try to make the one displayed in the index brief

docs/source/installation/openshift.md
mflendrich: It was not obvious to me when I was reading it, it will be IMO even less obvious for the reader.

What I'm asking for is to avoid overly general terms like "a subset of deployment models and underlying platforms " and instead:
- be direct about what deployment model is covered
- be direct about what underlying platforms are covered

That is as simple as ROSA and AWS, no?

docs/source/installation/openshift.md
```diff
@@ -0,0 +1,190 @@
...
+
+{{productName}} is a Red Hat OpenShift Certified Operator available for installation through the embedded Red Hat OpenShift OperatorHub.
+This guide describes how to install {{productName}} in a Red Hat OpenShift cluster with OLM (Operator Lifecycle Manager).
+
+## Prerequisites
+
+This guide requires you to have access to an OpenShift Container Platform cluster using an account with `cluster-admin` permissions.
+
+The cluster must meet the following requirements for production-grade ScyllaDB deployments:
+- a dedicated node pool labeled with `scylla.scylladb.com/node-type: scylla`, with local NVMe storage devices attached.
```
czeslavo: I'm not sure if having NVMe drives should be communicated as essential to install the Operator on OpenShift. 

I imagine that if someone wants to try this out quickly without having NVMes, it would be handy for them to have an option here described with loop devices (similarly to the GitOps guide).

rzetelskik: Overall this is a good point, there's also already a dedicated document for "Kubernetes prerequisites" where this could probably live. Although I'm not sure if that's going to work that well with different OpenShift deployment models. Should we discuss how to structure it?

docs/source/installation/openshift.md
rzetelskik: done

docs/source/installation/openshift.md
rzetelskik: done

docs/source/installation/openshift.md
rzetelskik: done

docs/source/installation/openshift.md
rzetelskik: I removed this requirement and added a reference to the "Kubernetes prerequisites". I extended it with static cpu policy guide for openshift. I also added a tab for loop devices in the openshift guide.

docs/source/installation/openshift.md
rzetelskik: I reworded it to be clearer

docs/source/.internal/verify-operator-installation-gitops.md
```diff
@@ -0,0 +1,24 @@
+Wait for CRDs to propagate to all apiservers:
+:::{code-block} shell
+kubectl wait --for='condition=established' crd/scyllaclusters.scylla.scylladb.com crd/nodeconfigs.scylla.scylladb.com crd/scyllaoperatorconfigs.scylla.scylladb.com crd/scylladbmonitorings.scylla.scylladb.com
```
abramche, line 3: would it not be logical to use `oc` instead of `kubectl` throughout the docs, insofar OpenShift is concerned? (also adding it as a pre-requisite -- this is not strictly necessary, but perhaps is a logical forestalling for future deployments and surprise documentation changes between versions. Also, RedHat [suggests](https://docs.redhat.com/en/documentation/openshift_container_platform/3.11/epub/cli_reference/differences-oc-kubectl-using-oc#differences-oc-kubectl-using-kubectl) to use `oc` wherever possible)

rzetelskik: There is a mention at the beginning of the doc that it can be used interchangeably. I wouldn't go as far as using it by default as it requires duplicating all common paths

rzetelskik:
/hold
web console workflow changed in 4.20, I need to adjust it

rzetelskik:
- Adjusted the installation docs to 4.20.
- Changed Update Approval strategy to Manual.

There were some changes in the installation workflow, especially around providing the ARN for AWS with STS and the manual Update Approval strategy. PTAL.

rzetelskik:
@mflendrich ping

rzetelskik:
@ylebi rebased on top of current master, verify-docs passed now

PR #3229, opened 2026-01-29
Extend support matrix in docs with supported OpenShift version range

rzetelskik:
**Description of your changes:** This PR extends the support matrix in our docs with the supported OpenShift version range. The range is populated using the metadata assets file.

I also changed the syntax used for tables in this document as I personally find modifying the tables with existing syntax very annoying.

Resolves #

/kind documentation
/priority important-soon

### DIFF
--- a/docs/source/_ext/version_context_substitutions.py
+++ b/docs/source/_ext/version_context_substitutions.py
@@ -7,12 +7,58 @@ Sphinx extension to extend myst_substitutions with version context in build time
 import os
 import re
 import subprocess
-from typing import Any, Optional
+from functools import reduce
+from typing import Any, Optional, cast
 
 import yaml
 from sphinx.application import Sphinx
 from sphinx.config import Config
 from sphinx.errors import ExtensionError
+
+DEFAULT_BRANCH = 'master'
+
+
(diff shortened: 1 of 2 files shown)


rzetelskik:
I need to add handling for files missing in branches
/hold

docs/source/_ext/version_context_substitutions.py
```diff
@@ -7,57 +7,152 @@
...
+            pass
+
+        def string_constructor(loader, node):
+            return loader.construct_scalar(node)
+
+        # Override constructors for all scalar types to return strings.
+        # For reference, see https://yaml.org/type/.
+        StringLoader.add_constructor('tag:yaml.org,2002:int', string_constructor)
+        StringLoader.add_constructor('tag:yaml.org,2002:float', string_constructor)
+        StringLoader.add_constructor('tag:yaml.org,2002:bool', string_constructor)
```
czeslavo: Wouldn't it be simpler to correct the YAML side - i.e., do not specify versions as floats/integers, but strings? It seems to me like working around a mistake that we make, as clearly, versions are not numbers.

rzetelskik: Agreed, fixed

PR #3238, opened 2026-02-03
Set E2E suites' parallelism to match their test cases count

czeslavo:
**Description of your changes:** Sets the default parallelism for all E2E test suites to match their total test cases count (rounded up to 10s) so that all of them can run in a separate Ginkgo process.


rzetelskik (approved):
/lgtm

PR #3243, opened 2026-02-05
Automate updating grafana and prometheus versions in config on scylla-monitoring updates and verify that prometheus-operator supports the grafana version

rzetelskik:
This PR does a few of things which should allow us to configure renovate to manage the monitoring dependencies:
- adds automation populating our config with versions extracted from scylla-monitoring submodule,
- adds a unit test verifying that prometheus-operator version supports the specified prometheus version.

It also adds a unit test verifying that the default grafana platform dashboard is aligned with ScyllaDB version and that the dashboard exists. This could be automatically bumped together with ScyllaDB version update, but we don't have any automation there atm so I've just added a guard.

Resolves #

/kind machinery
/priority important-soon


assets/config/config_test.go
```diff
@@ -201,3 +205,69 @@ func TestProjectConfig(t *testing.T) {
...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(line, "PROMETHEUS_VERSION=") {
+			prometheusVersion = strings.TrimPrefix(line, "PROMETHEUS_VERSION=")
+		} else if strings.HasPrefix(line, "GRAFANA_VERSION=") {
+			grafanaVersion = strings.Trim(strings.TrimPrefix(line, "GRAFANA_VERSION="), `"`)
+		}
+	}
+
```
czeslavo: Nit: we could assert `prometheusVersion` and `grafanaVersion` were correctly read from the `version.sh` for cleaner errors if we ever hit that.

czeslavo (commented):
Nice, just one nit.

assets/config/config_test.go
rzetelskik: done

rzetelskik:
/hold
@czeslavo I wonder if we should also verify prometheus-operator that way? Not sure if a unit test would be a good place, perhaps we should move it to a verify target?

czeslavo:
> /hold @czeslavo I wonder if we should also verify prometheus-operator that way? Not sure if a unit test would be a good place, perhaps we should move it to a verify target?

I think it's not as trivial as with Grafana and Prometheus for which we can extract exact versions from the monitoring submodule. We'd have to have some kind of a more intelligent update target that would look at the Prometheus version used in config.yaml, and adjust Prometheus Operator version in the metadata.yaml, according to compatibility defined in https://prometheus-operator.dev/docs/getting-started/compatibility/. Doable, but we'd have to rely on scripting to extract the right version from this compatibility page. On the other hand, according to this page, latest Prometheus Operator supports Prometheus >= 2.0 so maybe we could just always rely on the latest Operator version (in such a case setting up Renovate for the version defined in metadata would be enough).

hack/third-party/generate-in-tree-prometheus-operator-exports.sh
```diff
@@ -0,0 +1,76 @@
...
+  exit 1
+fi
+
+# Generate the Go file matching source repo structure.
+output_file="${output_dir}/pkg/operator/defaults.go"
+mkdir -p "$(dirname "${output_file}")"
+
+cat > "${output_file}" << EOF
+// Code generated by hack/generate-in-tree-prometheus-operator-exports.sh. DO NOT EDIT.
+// Extracted from ${prometheus_operator_defaults_url}.
```
czeslavo: I think it will be more visible if we move that as a comment above the `PrometheusCompatibilityMatrix` variable. We can also point out this is the version extracted from config.yaml.

hack/lib/assets.sh
```diff
@@ -0,0 +1,30 @@
...
+# Usage: get-metadata <yaml-key>
+function get-metadata() {
+    get-yaml-value "${metadata_file}" "${1}"
+}
+
+# get-config retrieves a value from the config YAML file.
+# Usage: get-config <yaml-key>
+function get-config() {
+    get-yaml-value "${config_file}" "${1}"
+}
```
czeslavo: Nit: missing newline

rzetelskik: done

hack/third-party/generate-in-tree-prometheus-operator-exports.sh
rzetelskik: > I think it will be more visible if we move that as a comment above the PrometheusCompatibilityMatrix variable.

done

> We can also point out this is the version extracted from config.yaml.

It already says which script this file is generate with so I'd not go into that much detail as it may rot

czeslavo (approved):
/lgtm

PR #3254, opened 2026-02-10
Update min supported Kubernetes version to 1.32

rzetelskik:
**Description of your changes:** Update minKubernetesVersion metadata to 1.32 - the minimum from versions in GKE and EKS standard channels.

Resolves #

/kind machinery
/priority important-soon

### DIFF
--- a/bundle/manifests/scylladb-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/scylladb-operator.clusterserviceversion.yaml
@@ -562,4 +562,4 @@ spec:
       targetPort: 5000
       type: ValidatingAdmissionWebhook
       webhookPath: /validate
-  minKubeVersion: 1.31.0
+  minKubeVersion: 1.32.0
--- a/bundle/patches/manifests/versions.clusterserviceversion.yaml
+++ b/bundle/patches/manifests/versions.clusterserviceversion.yaml
@@ -3,4 +3,4 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 spec:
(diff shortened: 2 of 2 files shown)


assets/metadata/metadata.yaml
```diff
@@ -3,7 +3,7 @@
 operator:
   minOpenShiftVersion: "4.20"
   maxOpenShiftVersion: "4.20"
-  minKubernetesVersion: "1.31.0"
+  minKubernetesVersion: "1.32.9"
```
czeslavo: I'm wondering whether that makes sense to be patch-precise here. I don't think we'll be able to keep this up to date when GKE/EKS has another patch available, so maybe a safe move would be to just use a minor (`1.32`) release here? In the end, that's how we approach it in testing as well - we only ask for deploying the latest `1.3x` patch.

rzetelskik: This is propagated to the CSV and a patch version is required: https://github.com/operator-framework/api/blob/884ff9542f2dc7a12f8cbecc18f88f7e7bb17e44/pkg/validation/internal/csv.go#L248-L257.
We can either:
1) decouple it from the generally declared min kube version (it's not really coupled atm anyway, but I hope it will be propagated to our support matrix in the future and aligned with what we declare in the bundle)
2) just conventionally always set the patch to 0

I'd go with 2 in that case.

czeslavo: I wasn't aware of the CSV requirement. Yeah, in such a case, 2nd option looks good to me.

rzetelskik: done

PR #3257, opened 2026-02-11
Automate the propagation of supported Kubernetes version range to the compatibility matrix in docs

rzetelskik:
**Description of your changes:** This PR adds a max supported Kubrnetes version to our metadata assets and automates the propagation of supported Kubernetes version range to the compatibility matrix in docs.

Resolves #

/kind machinery
/priority important-soon

### DIFF
--- a/docs/source/_ext/version_context_substitutions.py
+++ b/docs/source/_ext/version_context_substitutions.py
@@ -113,6 +113,32 @@ def get_versions_from_config(repo_root: str) -> dict[str, str]:
 
     return result
 
+
+def strip_patch_version(version: str) -> str:
+    """
+    Parse a semantic X.Y.Z version string and reduce it to X.Y format.
+
+    Args:
+        version: Semantic version string (e.g., '1.28.0', '1.28.0', '1.28').
+
+    Returns:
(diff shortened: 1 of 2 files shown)


docs/source/_ext/version_context_substitutions.py
```diff
@@ -81,76 +101,151 @@ def strip_digest(version: str) -> str:
 
     result = {}
 
+    # Try to extract ScyllaDB version
     scylla_db_version = get_nested_value(config_data, 'operator.scyllaDBVersion')
-    if not scylla_db_version:
-        raise ExtensionError(f"operator.scyllaDBVersion not found in {config_path}")
```
czeslavo, line 86: Do I understand correctly that we're dropping this explicit error handling in favor of later build-time failures caused by usage of undefined variables?

rzetelskik: Yes, otherwise we'd have to maintain a map of key->version and verify that, which to me seems over the top as either way it results in a build-time error.

czeslavo: Yeah, as long as we get feedback at build time, it's fine with me.

PR #3259, opened 2026-02-11
Remove redundant checkout steps from release-notes composite action

rzetelskik:
**Description of your changes:** This PR fixes the "Duplicate header: 'Authorization'" error in the release-notes workflow by removing redundant checkout steps from the composite action that were causing authentication conflicts.

After the releases workflow was upgraded to use actions/checkout@v6, the release-notes job started failing with:
```
remote: Duplicate header: "Authorization"
fatal: unable to access 'https://github.com/scylladb/scylla-operator/': The requested URL returned error: 400
```
https://github.com/scylladb/scylla-operator/actions/runs/21901321999/job/63230718464

The composite action contained two redundant checkout steps that were re-checking out the repository that the workflow had already checked out. I assume the intent was that the gen-release-notes binary is built from the target branch, but I don't see any benefits here.
The version mismatch created conflicting authentication mechanisms that resulted in duplicate Authorization headers.

Resolves #

/kind machinery
/priority important-soon
/cc czeslavo

### DIFF
--- a/.github/actions/release-notes/action.yaml
+++ b/.github/actions/release-notes/action.yaml
@@ -13,9 +13,6 @@ inputs:
   goVersion:
     description: "Version of Go"
     required: true
-  genReleaseNotesVersionRef:
-    description: "Version ref of gen-release-notes"
-    required: true
   containerImageName:
     description: "Full name of the container image"
     required: true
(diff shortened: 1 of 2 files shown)


.github/actions/release-notes/action.yaml
```diff
@@ -26,31 +26,18 @@ runs:
...
+      cd ${{ github.workspace }}/go/src/github.com/${{ inputs.githubRepository }}
       make build GO_BUILD_PACKAGES=./cmd/gen-release-notes --warn-undefined-variables
       
       sudo mv ./gen-release-notes /usr/local/bin/gen-release-notes
-  
-  - uses: actions/checkout@v3
-    with:
-      repository: ${{ inputs.githubRepository }}
-      path: ${{ github.workspace }}/go/src/github.com/${{ inputs.githubRepository }}
-      fetch-depth: 0
```
czeslavo, line 52: Shouldn't we, instead of deleting, move this checkout step before the `Install release notes generator` step?

rzetelskik: there's already a checkout step in workflows/releases.yaml - won't this suffice?

czeslavo: Hm, I'm not sure, to be honest.

.github/actions/release-notes/action.yaml
czeslavo: It's no longer in use - let's drop it. We'll need to update the caller side as it was required before.

rzetelskik: done

PR #3261, opened 2026-02-11
Add utilities for debugging flaky tests

czeslavo:
**Description of your changes:** Adds a script that runs a suspect flaky E2E test in a loop until it fails N times and collects artifacts from all the runs (both success and failure scenarios). This data can later be used to feed an OpenCode custom agent added in this PR to analyze the artifacts and pinpoint the potential root cause of the flakiness.

Proven to work on one of our flakes: https://github.com/scylladb/scylla-operator/pull/3258. See the [logs](https://gist.github.com/czeslavo/13138d56861043347666c6c3a67571fb) of the agent run.

### DIFF
--- a/.opencode/agents/flaky-tests-debugger.md
+++ b/.opencode/agents/flaky-tests-debugger.md
@@ -1,0 +1,274 @@
+---
+description: Runs analysis of flaky E2E tests in the Scylla Operator project, comparing successful and failed runs to identify root causes and propose solutions.
+mode: primary 
+temperature: 0.3
+tools:
+  write: false
+  edit: false
+  bash: true
+---
+
+# Flaky E2E Tests Debugger Agent
+
+You are a specialized agent for debugging flaky end-to-end (E2E) tests in the Scylla Operator project. Your primary goal is to analyze test artifacts from multiple test runs, identify patterns in failures, compare them with successful runs, and provide actionable insights into why tests fail inconsistently.
+
+## Your Capabilities and Constraints
+
+**YOU CAN:**
+- Analyze test artifacts from multiple test runs
+- Compare successful and failed test executions
+- Examine logs, events, and Kubernetes resource states
+- Identify timing issues, race conditions, and inconsistent behavior
+- Explain root causes and propose potential solutions
+- Review test source code and operator logic to understand failure mechanisms
+
+**YOU CANNOT:**
+- Modify or write any code (test or operator)
+- Execute commands or make changes to the repository
+- Run tests or create new test runs
+
+## Your Analysis Process
+
+When given test artifacts from the `debug-flake.sh` script, you should follow this systematic approach:
+
+### Phase 1: Understand the Test Context
+
+1. **Parse the test summary:**
+   - Total runs, passed/failed count, success rate
+   - Test name and focus pattern
+   - Identify which runs failed (e.g., run-13, run-14)
+
+2. **Locate the test source code:**
+   - Find the test in `./test/e2e/`
+   - Understand what the test is verifying
+   - Identify key assertions and expectations
+   - Note any timing-sensitive operations or observers
+
+### Phase 2: Establish a Golden Reference (Successful Run)
+
+Select one successful test run as your baseline for comparison. For each successful run, locate:
+
+1. **Test execution details** in `run-N/e2e.json`:
+   - Find the specific test spec in the JSON (search for `"State": "passed"` or the test name)
+   - Note the `StartTime`, `EndTime`, and `RunTime`
+   - Review `CapturedGinkgoWriterOutput` for the test's log output
+   - Examine `SpecEvents` to understand the test flow
+
+2. **Test-specific logs** in `run-N/e2e/cluster/namespaces/<test-namespace>/`:
+   - Identify the test's namespace (e.g., `e2e-test-scyllacluster-6fcr2-x9flf`)
+   - Find pods created during the test
+   - For each pod, review container logs in `pods/<pod-name>/<container-name>.current`
+   - Review events in `events.events.k8s.io/`
+
+3. **Operator logs** in `run-N/must-gather/cluster/namespaces/scylla-operator/`:
+   - Find the active operator pod during the test timeframe
+   - Review `pods/<operator-pod-name>/scylla-operator.current`
+   - Look for events and decisions made by the operator
+
+4. **Kubernetes resources** created during the test:
+   - Jobs: `run-N/e2e/cluster/namespaces/<test-namespace>/jobs/`
+   - StatefulSets: `run-N/e2e/cluster/namespaces/<test-namespace>/statefulsets.apps/`
+   - ScyllaCluster/ScyllaDBCluster: Look in custom resource directories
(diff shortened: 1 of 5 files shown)


.opencode/agents/flaky-tests-debugger.md
```diff
@@ -0,0 +1,274 @@
...
+Based on your analysis, explain:
+
+1. **Root Cause:** Clearly state what you believe is causing the flake
+2. **Evidence:** Reference specific differences you found in logs, timings, or states
+3. **Proposed Fixes:** Describe (but don't code) potential solutions:
+   - Changes to test logic (longer waits, more robust assertions, etc.)
+   - Changes to operator behavior (if a bug is found)
+   - Changes to test infrastructure or setup
+
+## Artifact Structure Reference
```
rzetelskik, line 141: what's the plan for keeping it up to date?

.opencode/agents/flaky-tests-debugger.md
```diff
@@ -0,0 +1,274 @@
...
+- Compare successful and failed test executions
+- Examine logs, events, and Kubernetes resource states
+- Identify timing issues, race conditions, and inconsistent behavior
+- Explain root causes and propose potential solutions
+- Review test source code and operator logic to understand failure mechanisms
+
+**YOU CANNOT:**
+- Modify or write any code (test or operator)
+- Execute commands or make changes to the repository
+- Run tests or create new test runs
```
rzetelskik, line 28: I haven't had any experience with opencode, and I'm not opposing this constrained approach, but I'd like to understand why. Is it just a matter of making the goal more well-defined and constrained, or did you just find it to not work well?

hack/kind/debug-flake.sh
```diff
@@ -0,0 +1,229 @@
+#!/bin/bash
```
rzetelskik, line 1: If this is essentially just a wrapper for running tests in a loop and aggregating the reports, why make it kind-specific? Instead of just letting it take a kubeconfig or a script as an argument?

mflendrich: Do you have a concrete way of using it in mind, or just trying to future-proof it with no firm use case today?

rzetelskik: Kind covers a subset of tests. If you want to debug a manager-related flake with backup/restore for instance (which actually is very flaky, at least as of recently), this won't help - at least judging by the tests included in the suite.

mflendrich: > is very flaky

This tool serves no purpose in that case - it's supposed to be run on a reasonably healthy suite of tests to detect (mild) regression. Persistently flaky tests are a lost cause IMO and I would rather take a "two-speed approach" with good tooling for quality tests, while managing/maintaining/reducing the set of tests that don't work well.

rzetelskik: > This tool serves no purpose in that case - it's supposed to be run on a reasonably healthy suite of tests to detect (mild) regression. 

The script description says something different

> This script runs a suspect test case in a loop to verify if it's flaky and collect artifacts for debugging.

mflendrich: I am just trying to understand - if you know that the test case is very flaky (and I believe that with "very flaky" you can just have the logs/artifacts from a failed run just about immediately, in 1-2 iterations) - then what is the actual use case in your mind?

rzetelskik: This will also apply to any NodeConfig-related tests for instance. From my perspective it seems like the script could easily cover all tests in all environments, at least by design. If it's too difficult/unstable to be worth the effort now - sure, we can approach it iteratively. But to say that the tool would serve no purpose for tests from other suites because they are "too flaky" is quite baffling to me.

rzetelskik: > I am just trying to understand - if you know that the test case is very flaky (and I believe that with "very flaky" you can just have the logs/artifacts from a failed run just about immediately, in 1-2 iterations) - then what is the actual use case in your mind?

I see this script as a tool to iterate on a flaky test (or a suspected flaky test) - this is exactly what the PR description says. It helps you run it a couple times easily and collects artifacts to work with the AI agent. You try a fix, you run it again. Rinse and repeat. A test can be very flaky due to a bug in the operator or a component we integrate with, it doesn't discard the test in my opinion. So I don't see why this wouldn't be valid for all tests and potentially all platforms on principle.

mflendrich: Is this expensive to do? Will it significantly complicate the tool? Will it make the tool more painful to use in the kind case? @czeslavo

.opencode/agents/flaky-tests-debugger.md
czeslavo: I wanted to make it clear that it shouldn't do anything beyond analysing what it has at hand, provided by us explicitly in the form of test artifacts. I also don't want it to jump straight to implementing a solution to the problem, as I want to validate its suspicions and proposed paths to fixing them first, picking the solution that I find is the right one.

Once it figures out the plan, it should be trivial to implement it by hand or by feeding the plan to another agent that will focus on the implementation only. I noticed that after running the analysis, the limit of tokens for context was almost exhausted, so it seemed to me like the most sane approach to split the plan/implement stages (it's actually a pattern advised by OpenCode as well).

.opencode/agents/flaky-tests-debugger.md
czeslavo: That's a good question; I don't have a definite answer to it. But my intuition is that we should revisit it every time we actually use it. If the structure changes, it should be quite an easy task for an agent to look into the current structure and update this markdown for us.

I think the whole point about keeping it in repo is that we can share a common, loosely defined workflow, ready to be used (and saving tokens for exploratory work that an agent would have to do if we didn't provide this context). In the end, when we use it, it will always be a bit unique experience. And I guess that if the structure is off, the agent will realize the discrepancy when working with it (as something expected to be in a specific directory is not there, but somewhere else), so we'll end up using a bit more tokens for it to learn the structure on the go again.

I suggest we experiment and see how it will go. Maybe we will find this to be "dead code" at some point. Maybe OpenCode will implement a feature for "agent feedback loop" that will adjust the custom agent on every use. :shrug:

hack/kind/debug-flake.sh
czeslavo: I think that Kacper's point is correct, and it might make sense to make this script ready to be used both with GKE and kind. We can have flakes in any suite, so why not be able to debug all of them using the same workflow.

There are kind-specific steps; I must see what kind of effort it would take to restructure it a bit to fit both scenarios.

czeslavo: I spent some time trying to refactor this so it works nicely with both kind and GKE (or any other KUBECONFIG-provided cluster), but I failed to make it look as I would like (I always ended up with spaghetti). As we agreed that we will timebox it to the previous week, I suggest we merge this machinery as-is, since it's already value-added. Once we hit a limit (we need to debug a test that's not able to run on kind), we'll revisit it.

rzetelskik: Sounds good

PR #3276, opened 2026-02-20
Fix missing namespace for SA creation in Setting up monitoring on OpenShift guide

rzetelskik:
**Description of your changes:** A command creating the SA for OpenShift UWM was missing a namespace which was breaking the workflow. This PR fixes it.

Resolves #

/kind documentation
/priority important-longterm
/cc @czeslavo


rzetelskik:
Needs to be backported to 1.20 as well

PR #3283, opened 2026-02-24
Change scylla-monitoring renovate to rely on GH release instead of git-refs

czeslavo:
**Description of your changes:** As `scylla-monitoring` happens to move git tags for already published proper release tags (e.g., `1.14.0`), we need to adapt to take that into account so that we do not use git tags that are not considered by the team as properly published (i.e., having a production Github release created).

This PR changes the Renovate config so that we do not rely on the git-submodule manager, which can only use git-refs as its datasource. Instead, we use a custom regex manager that will detect a GH release and will update the `.gitmodules`' `branch` field. Because we use Mend.io workers, we cannot use a custom post upgrade action to update the git submodules for us. To work around this, there will be a `do-not-merge/hold` label assigned to PRs updating this dependency, and instructions for required manual actions will be added to the PR description.

Sample PR generated by Renove in my fork: https://github.com/czeslavo/scylla-operator/pull/28

Related to https://github.com/scylladb/scylla-operator/pull/3282.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,13 +5,9 @@
   ],
   "enabledManagers": [
     "custom.regex",
-    "git-submodules",
     "gomod",
     "poetry"
   ],
-  "git-submodules": {
-    "enabled": true
-  },
   "schedule": [
     "before 5am on monday"
   ],
@@ -88,7 +84,7 @@
     {
       "description": "Monitoring dependencies should be grouped and updated every weekday",
       "matchDepNames": [
-        "submodules/github.com/scylladb/scylla-monitoring",
+        "scylladb/scylla-monitoring",
         "prometheus-operator"
       ],
       "groupName": "monitoring",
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -107,6 +103,36 @@
...
+        "   cd submodules/github.com/scylladb/scylla-monitoring",
+        "   git fetch --tags",
+        "   git checkout tags/{{{newValue}}}",
+        "   cd ../../../..",
+        "   git add submodules/github.com/scylladb/scylla-monitoring",
+        "   make update",
+        "   git add .",
+        "   git commit --amend && git push --force",
+        "   ```",
+        "3. Remove the `do-not-merge` label."
```
rzetelskik: nit: do-not-merge/hold, there are other labels with this prefix

renovate.json
```diff
@@ -129,6 +155,15 @@
...
+    },
+    {
+      "customType": "regex",
+      "fileMatch": [
+        "^\\.gitmodules$"
+      ],
+      "matchStrings": [
+        "# renovate: datasource=(?<datasource>[a-z-]+) depName=(?<depName>[^\\s]+)\\s+branch = (?<currentValue>.*)"
+      ]
     }
```
rzetelskik, line 168: let's give it a description, e.g. `Custom manager for git submodules with Renovate annotations matching the regex.`

renovate.json
czeslavo: Addressed.

renovate.json
czeslavo: Addressed.

PR #3285, opened 2026-02-24
Add a default channel annotation to OLM bundle for stable releases

rzetelskik:
**Description of your changes:** Add default channel annotation to stable releases. It is not added to dev releases in case there is no stable channel release for a particular OpenShift version yet.

Resolves #

/kind machinery
/priority important-soon
/cc czeslavo


rzetelskik:
Needs to be backported to 1.20 as well.

PR #3286, opened 2026-02-25
Make sure podman is always used with kind

czeslavo:
**Description of your changes:** This ensures that all scripts that use `kind`, set `KIND_EXPERIMENTAL_PROVIDER=podman` for consistency and correctness. Without that, it was possible for a script to create a cluster using podman but list them with Docker, leading to unexpected behavior because the cluster is never found in such cases.


czeslavo:
@mflendrich ping :wave:

PR #3291, opened 2026-02-25
Drop no longer necessary workaround for xfsprogs on GKE

czeslavo:
**Description of your changes:** As GKE added xfsprogs back to their Ubuntu image, it's no longer needed to either apply or document the workaround for its absence. Let's drop it to save us the maintenance burden.


docs/source/installation/kubernetes-prerequisites.md
```diff
@@ -67,19 +67,3 @@ For the purposes of the installation guides, we assume that the nodes meant to r
 
 ScyllaDB Operator's [NodeConfig](../resources/nodeconfigs.md) controller requires `xfsprogs` to be installed on the Kubernetes
 nodes to format the local disks with XFS file system.
```
rzetelskik, line 69: Should we remove this note as well? It's somewhat random without that context

czeslavo: I believe the purpose of this document was to gather all dependencies, and `xfsprogs` remain as such. Maybe we should add a sentence to this like "Please ensure it's present in your Kubernetes worker hosts' PATH"?

rzetelskik: Let's leave it as is, my point was that it's just one of many and this becomes the only one we document out of context

PR #3292, opened 2026-02-26
Make shell scripts discover bash instead of hardcoding the interpreter

abramche:
This change replaces the hardcoded bash shebang with a OS-agnostic discovery lookup.
The benefits are that the shell scripts will be able to run on non-Linux OS (MacOS, BSD....).


abramche:
I ran tests thrice and there are different minor (1-2) amounts if failures -- all unrelated to anything resembling the environment variables, $PATH or interpreter itself. I am confident they have nothing to do with this change.

PR #3293, opened 2026-02-26
Bump Grafana minimum zoom in to 30s

### LINKED ISSUE #2990: monitoring - the prometheus source is mis-configured with a 5s scrape_interval - should be 30s (see the configmap)
### What happened?

the $__rate_interval is messed up given the wrong scrape_interval - which is typically 4xScrapeIntervel

some charts have no data - with a proper scrape_interval they do

easy to see in the configmap scylla-grafana-provisioning in the datasource.yaml key as shown fixed in this section:

  datasources.yaml: |
    apiVersion: 1
    datasources:
    - name: prometheus
      type: prometheus
      access: proxy
      url: "https://scylla-prometheus:9090"
      isDefault: true
      version: 1
      editable: false
      jsonData:
        **timeInterval: "30s"**
        tlsAuthWithCACert: true
      secureJsonData:
        tlsCACert: "$__file{/var/run/configmaps/prometheus-serving-ca/ca-bundle.crt}"
        tlsClientCert: "$__file{/var/run/secrets/prometheus-client-certs/tls.crt}"
        tlsClientKey: "$__file{/var/run/secrets/prometheus-client-certs/tls.key}"
  notifiers.yaml: ''

### Scylla Operator version

1.17.1 and 1.18.1

1.33

n/a

n/a

abramche:
This change:
- bumps minimum Grafana window 5s -> 30s (to be more than 4x of the scrape interval set to 5s)
- forces a 5s scrape interval on ServiceMonitor level, as a precaution for external Prometheus cases (OpenShift)
- adjusts the tests for the changes

Resolves #2990

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@
   a patch update to the Grafana image used (`docker.io/grafana/grafana` from `12.3.2` to `12.3.3`).
   [#3282](https://github.com/scylladb/scylla-operator/pull/3282), [#3284](https://github.com/scylladb/scylla-operator/pull/3284)
 
+### Bug fixes
+
+-  Fixed [#2990](https://github.com/scylladb/scylla-operator/issues/2990): set monitoring scrape intervals to: 5s in 
+  `ServiceMonitor`, and 30s in Grafana. This overrides the global scrape interval set in Prometheus.
+
 ## 1.20.0 and older
 
 For those versions, the changelog information can be found in two places:


abramche:
Verified locally - works even for smallest intervals of 10s (with two data points, scraped by 5s). Also verified that ServiceMonitor enforces 5s scrape interval.

abramche:
image

czeslavo:
/hold 
Ah, we've missed one thing. Let's add a CHANGELOG.md entry for this, as this is a user-facing change in the end.

rzetelskik:
@czeslavo should we verify this works on OpenShift or are we reasonably sure it doesn't break UWM and external prometheus mode?

rzetelskik:
> The priority can mirror the issue priority, or be set according to your opinion - it has no significant effect on anything in the process in practice, it's just Prow requiring it

Btw it does affect the ordering of PRs in queue to merge, but that's probably hardly ever making any difference. Using something out of the ordinary (like critical-urgent) would also be a signal for a reviewer to prioritise it.

rzetelskik:
Also @czeslavo @abramche should we backport it to 1.20?

abramche:
> Also @czeslavo @abramche should we backport it to 1.20?

Do you mean, as a 1.20.1 content? I think yes, it makes sense. @czeslavo WDYT?

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 - Updated ScyllaDB Monitoring (`github.com/scylladb/scylla-monitoring` git submodule) from `4.14.0` (`88dd086`) pre-release version
   to the final `4.14.0` (`9dcb579`). The update includes several Grafana dashboard improvements and bug fixes, as well as
   a patch update to the Grafana image used (`docker.io/grafana/grafana` from `12.3.2` to `12.3.3`).
-  [#3282](https://github.com/scylladb/scylla-operator/pull/3282), [#3284](https://github.com/scylladb/scylla-operator/pull/3284)
+  [#3282](https://github.com/scylladb/scylla-operator/pull/3282), [#3284](https://github.com/scylladb/scylla-operator/pull/3284), [3293](https://github.com/scylladb/scylla-operator/pull/3293)
```
czeslavo: I think this PR deserves a separate entry - this one is about updating our dependency on scylladb/scylla-monitoring git submodule.

Let's make it clear in the new entry that the scrape interval will now be hardcoded to 5s (due to ServiceMonitor settings), as that's the most important user-facing change in our PR.

czeslavo:
> @czeslavo should we verify this works on OpenShift or are we reasonably sure it doesn't break UWM and external prometheus mode?

UWM doesn't differ from the external mode much (and that's been tested by @abramche). In UWM, Prometheus Operator is also used underneath, and our ServiceMonitor will affect the scrape interval the same way as outside of OpenShift. We _could_ test that, but I don't think it's worth the effort.

Regarding 1.20 backport, I think that makes sense, but let's all agree on that during the team sync, as while it fixes a bug, it's still a default behavior change.

rzetelskik:
> We could test that, but I don't think it's worth the effort.

The effort is just triggering the test, no? Rember we don't have any coverage for OpenShift other than the presubmits now btw.

I'll trigger it but I'm leaving it to @czeslavo to decide when this is ready to merge.

/test e2e-openshift-aws-parallel

czeslavo:
> The effort is just triggering the test, no? Rember we don't have any coverage for OpenShift other than the presubmits now btw.

Ah, you meant that. Yes, that's what we should definitely do. I was thinking you were asking for manual tests.

rzetelskik:
OpenShift tests failed due to an unrelated upgrade test flake, other than that it's good.

abramche:
CHANGELOG.md entry added

CHANGELOG.md
```diff
@@ -4,20 +4,21 @@
...
-  [#3282](https://github.com/scylladb/scylla-operator/pull/3282), [#3284](https://github.com/scylladb/scylla-operator/pull/3284)
+  [#3282](https://github.com/scylladb/scylla-operator/pull/3282), [#3284](https://github.com/scylladb/scylla-operator/pull/3284
+
+### Bug fixes
+
+- Made sure that ScyllaDBMonitoring-deployed ServiceMonitors have an explicit scrape interval of 5s, which is the recommended
+  default for ScyllaDB monitoring. Also, adjusted the Grafana dashboard data source configuration to use 30s as the default
+  scrape interval for queries to ensure there are no issues with missing data points in the graphs.
+  Please note this change may affect your setup if you have an external Prometheus instance configured with a global scrape
+  interval - from now on it will be overridden by the ServiceMonitor configuration generated by the operator.
```
mflendrich: IMO this can be made much more information-dense:

```suggestion
-  Fixed [#2990](https://github.com/scylladb/scylla-operator/issues/2990): set monitoring scrape intervals to: 5s in `ServiceMonitor`, and 30s in Grafana. This overrides the global scrape interval set in Prometheus.
```

abramche: This is in similar vein that I originally submitted. If this is alright with you @czeslavo I'll change to Michal's suggestion

mflendrich: What I am asking for is:
- encode the same information in fewer words
- point at the bug being fixed

I am not particularly attached to any specific wording. I just want these 2 qualities above somehow met.

abramche: suggestion applied verbatim; if you guys feel OK with the change, let's merge

Issue #3294, opened 2026-02-27
Operator fails silently when ScyllaCluster's task names do not adhere to RFC 1123 subdomain requirements


rzetelskik:
When a ScyllaCluster is created with a manager task (either baackup or repair) with a name that doesn't adhere to RFC 1123 subdomain requirements, the underlying ScyllaDBManagerTask creation fails and propagates to the controller condition, but 1) the admission webhook does not forbid the creation of ScyllaCluster, 2) the error does not propagate to task errors in ScyllaCluster's status, and 3) the controller condition is not aggregated to the "global" ScyllaCluster conditions (Available, Progressing, Degraded). What this means is that the manager task is never created and the error is silent - it can only be observed in ScyllaCluster's controller conditions.

```
  - lastTransitionTime: "2026-02-26T13:15:11Z"
    message: 'can''t apply ScyllaDBManagerTask: ScyllaDBManagerTask.scylla.scylladb.com
      "scylla-repair-invalid_name-2n2u3" is invalid: metadata.name: Invalid value:
      "scylla-repair-invalid_name-2n2u3": a lowercase RFC 1123 subdomain must consist
      of lower case alphanumeric characters, ''-'' or ''.'', and must start and end
      with an alphanumeric character (e.g. ''example.com'', regex used for validation
      is ''[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'')'
    observedGeneration: 1
    reason: Error
    status: "True"
    type: ScyllaDBManagerTaskControllerDegraded
```

```
  - lastTransitionTime: "2026-02-26T10:55:33Z"
    message: ""
    observedGeneration: 1
    reason: AsExpected
    status: "False"
    type: Degraded
```

This is a regression introduced when adding the under-the-hood translation from ScyllaCluster's inlined tasks to ScyllaDBManagerTask objects.

## Acceptance criteria
- [ ] `ScyllaDBManagerTaskControllerDegraded` condition is taken into account to calculate the aggregate `Degraded` condition
- [ ] Task names not adhering to RFC 1123 are forbidden by the admission webhook.

Proposed solution:
- In a next patch release, extend the admission webhook to log a warning when a ScyllaCluster with task names not adhering to RFC1123 subdomain requirements is created or updated.
- In a next minor release:
  - Extend the admission webhook to forbid task names not adhering to RFC1123 subdomain requirements in ScyllaCluster spec.
[...]

PR #3295, opened 2026-02-27
Add warnings for ScyllaCluster's manager task names not adhering to RFC 1123 subdomain requirements

rzetelskik:
**Description of your changes:** As per https://github.com/scylladb/scylla-operator/issues/3294, operator fails silently when ScyllaCluster's task names do not adhere to RFC 1123 subdomain requirements. As part of the proposed solution, this PR extends the admission webhook to log warnings and inform about backwards-incompatible validation tightening in a future minor release.

It's an intermediate step towards https://github.com/scylladb/scylla-operator/issues/3294

/kind api-change
/priority important-soon
/cc @czeslavo

### DIFF
--- a/pkg/api/scylla/validation/cluster_validation.go
+++ b/pkg/api/scylla/validation/cluster_validation.go
@@ -510,16 +510,50 @@ func isRackStatusUpToDate(sc *scyllav1.ScyllaCluster, rack string) bool {
 func GetWarningsOnScyllaClusterCreate(sc *scyllav1.ScyllaCluster) []string {
 	var warnings []string
 
-	warnings = append(warnings, warningsForScyllaClusterSpec(&sc.Spec, field.NewPath("spec"))...)
+	warnings = append(warnings, getWarningsForScyllaClusterSpec(&sc.Spec, field.NewPath("spec"))...)
 
 	return warnings
 }
 
-func warningsForScyllaClusterSpec(spec *scyllav1.ScyllaClusterSpec, fldPath *field.Path) []string {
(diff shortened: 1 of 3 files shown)


pkg/api/scylla/validation/cluster_validation.go
```diff
@@ -510,25 +510,59 @@ func isRackStatusUpToDate(sc *scyllav1.ScyllaCluster, rack string) bool {
...
+	warnings = append(warnings, getWarningsForTaskSpec(&repairTaskSpec.TaskSpec, fldPath)...)
+
+	return warnings
+}
+
+func getWarningsForTaskSpec(taskSpec *scyllav1.TaskSpec, fldPath *field.Path) []string {
+	var warnings []string
+
+	for _, msg := range apimachineryvalidation.NameIsDNSSubdomain(taskSpec.Name, false) {
+		warnings = append(warnings, fmt.Sprintf("%s. An invalid task name will result in a validation error in the next minor release.", field.Invalid(fldPath.Child("name"), taskSpec.Name, msg)))
```
czeslavo: I'm wondering whether we could make it a bit clearer what the consequences are - "will result in a validation error" doesn't explain directly what that will mean in practice (I suppose not everyone will know the difference between warnings/errors).

Maybe we could rephrase that to "ScyllaDB Operator will reject ScyllaCluster objects' creation/update with this invalid task name in the next minor release."?

rzetelskik: Good pointed, changed

czeslavo:
Let's remember about backporting to v1.20.

PR #3306, opened 2026-03-04
Refactor e2e tests to delint from spec pollution

### LINKED ISSUE #3190: Do not assign variables within ginkgo container nodes in E2E tests
### Problem

Ginkgo suggests not assigning variables within container nodes as it may lead to bugs (source: https://onsi.github.io/ginkgo/#avoid-spec-pollution-dont-initialize-variables-in-container-nodes). We should fix the existing issues in our E2E suite. Those can be detected by `ginkgolinter` with the `forbid-spec-pollution` flag enabled. Currently, we have an exclusion that mutes their failures.

### Acceptance criteria

- [ ] Existing occurrences of assignments within ginkgo container nodes are fixed in E2E tests
- [ ] Exclusion rule for "use BeforeEach..." inside of `test/e2e` is removed from `.golangci.yaml`

abramche:
This change:
- enables the previously disabled [ginkgo spec pollution linting rule](https://onsi.github.io/ginkgo/#avoid-spec-pollution-dont-initialize-variables-in-container-nodes)
- adjusts all the e2e tests so they pass the linting check
- as a side effect, limited the scope of tests using k8s AdminClient in setting up ginkgo framework to a more limited (user) Client (where applicable)

Resolves #3190 

Tested the linting rule with running `$ ginkgolinter -forbid-spec-pollution ./...` (should also auto-check now since the rule is not muted any more)

image

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -46,7 +46,13 @@ type Framework struct {
 var _ FullClientInterface = &Framework{}
 var _ ClusterInterface = &Framework{}
 
-func NewFramework(namePrefix string) *Framework {
+// NewFramework creates a new test framework. It creates a unique namespace for a test to run in and provides a client
+// with permissions limited to that namespace. It also creates clients for worker clusters (for multi-DC) if configured.
+// It should be called in a ginkgo BeforeEach closure.
+//
+// At the end of the test, artifacts are going to be collected, and the namespace and all objects in it will be
+// automatically cleaned up, unless the cleanup policy is set to skip.
+func NewFramework(ctx context.Context, namePrefix string) *Framework {
 	var err error
 
 	e2eArtifactsDir := ""
@@ -105,11 +111,19 @@ func NewFramework(namePrefix string) *Framework {
 		)
 	}
 
-	g.BeforeEach(f.beforeEach)
-	g.JustAfterEach(f.justAfterEach)
-	g.AfterEach(f.afterEach)
+	ns, nsClient := f.CreateUserNamespace(ctx)
+	f.cluster.defaultNamespace = ns
+	f.cluster.defaultClient = nsClient
+	f.FullClient.Client = nsClient
+
+	g.DeferCleanup(f.cleanup)
 
 	return f
+}
+
+func (f *Framework) cleanup(ctx context.Context) {
+	f.collectArtifacts(ctx)
+	f.cleanupObjects(ctx)
 }
 
 func (f *Framework) Cluster() ClusterInterface {
@@ -258,14 +272,7 @@ func (f *Framework) GetObjectStorageSettingsForWorkerCluster(workerName string) 
 	return settings
 }
 
-func (f *Framework) beforeEach(ctx context.Context) {
-	ns, nsClient := f.CreateUserNamespace(ctx)
-	f.cluster.defaultNamespace = ns
-	f.cluster.defaultClient = nsClient
-	f.FullClient.Client = nsClient
-}
-
-func (f *Framework) justAfterEach(ctx context.Context) {
+func (f *Framework) collectArtifacts(ctx context.Context) {
 	ginkgoNamespace := f.cluster.defaultNamespace.Name
 	f.cluster.Collect(ctx, ginkgoNamespace)
 	for _, c := range f.workerClusters {
@@ -273,15 +280,7 @@ func (f *Framework) justAfterEach(ctx context.Context) {
 	}
 }
 
-func (f *Framework) afterEach(ctx context.Context) {
-	nilClient := Client{
-		Config: nil,
-	}
-
-	f.cluster.defaultNamespace = nil
-	f.cluster.defaultClient = nilClient
-	f.FullClient.Client = nilClient
-
+func (f *Framework) cleanupObjects(ctx context.Context) {
 	shouldCleanup := true
 	switch TestContext.CleanupPolicy {
 	case CleanupPolicyNever:
--- a/test/e2e/set/nodeconfig/nodeconfig_optimizations.go
+++ b/test/e2e/set/nodeconfig/nodeconfig_optimizations.go
@@ -14,6 +14,7 @@ import (
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	configassests "github.com/scylladb/scylla-operator/assets/config"
+	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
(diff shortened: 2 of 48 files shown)


test/e2e/framework/framework.go
```diff
@@ -258,22 +267,22 @@ func (f *Framework) GetObjectStorageSettingsForWorkerCluster(workerName string)
 	return settings
 }
 
-func (f *Framework) beforeEach(ctx context.Context) {
+func (f *Framework) BeforeEach(ctx context.Context) {
```
rzetelskik: looks like this is never being called now

test/e2e/framework/framework.go
```diff
@@ -258,22 +267,22 @@ func (f *Framework) GetObjectStorageSettingsForWorkerCluster(workerName string)
...
+func (f *Framework) collectArtifacts(ctx context.Context) {
 	ginkgoNamespace := f.cluster.defaultNamespace.Name
 	f.cluster.Collect(ctx, ginkgoNamespace)
 	for _, c := range f.workerClusters {
 		c.Collect(ctx, ginkgoNamespace)
 	}
 }
 
-func (f *Framework) afterEach(ctx context.Context) {
+func (f *Framework) cleanupObjects(ctx context.Context) {
```
rzetelskik, line 283: I think the func name doesn't reflect the semantics now - it doesn't only cleanup objects, but also "resets" the framework:
```
	f.cluster.defaultNamespace = nil
	f.cluster.defaultClient = nilClient
	f.FullClient.Client = nilClient
```

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
```diff
@@ -36,15 +37,20 @@ const (
...
+		f             *framework.Framework
+		ncTemplate    *scyllav1alpha1.NodeConfig
+		matchingNodes []*corev1.Node
+	)
 
 	g.JustBeforeEach(func() {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		f = framework.NewFramework(ctx, "nodesetup")
```
rzetelskik, line 50: why is this sometimes called in `BeforeEach` nodes and other times in `JustBeforeEach` nodes?

test/e2e/framework/framework.go
czeslavo: I think we could just drop these reset actions after the changes we've made, as the framework is recreated for every test case anyway.

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
czeslavo: As long as we have only `BeforeEach` or `JustBeforeEach` in a given test, I guess it doesn't really change anything. I suggest we unify the approach and just use `BeforeEach` everywhere in such cases.

test/e2e/framework/framework.go
```diff
@@ -46,8 +46,11 @@ type Framework struct {
 var _ FullClientInterface = &Framework{}
 var _ ClusterInterface = &Framework{}
 
-func NewFramework(namePrefix string) *Framework {
-	var err error
+func NewFramework(ctx context.Context, namePrefix string) *Framework {
+	var (
+		err error
+		f   *Framework
```
czeslavo: There's no reason to split the `f` declaration and initialization here; let's keep it as it was before.

czeslavo:
Looks like we've got a unit test flake: https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3306/pull-scylla-operator-master-unit/2029492306859528192. :disappointed: Not related to the changes in this PR though.

test/e2e/set/nodeconfig/nodeconfig_optimizations.go
abramche: I was conservative and kept `BeforeEach` and `JustBeforeEach` as they were used in the tests. i.e. if the test had the `JustBeforeEach` block already, I just moved it there. I'll check if we can unify these

abramche: Changed every affected test to `BeforeEach` in 4a1d4f889117d5f1fc536786231d65fee1431aac and the kind local suite passed

test/e2e/framework/framework.go
abramche: Good spot, fixed in 4a1d4f889117d5f1fc536786231d65fee1431aac

test/e2e/framework/framework.go
abramche: Dropped in 4a1d4f889117d5f1fc536786231d65fee1431aac

test/e2e/framework/framework.go
abramche: removed in 4a1d4f889117d5f1fc536786231d65fee1431aac

test/e2e/framework/framework.go
```diff
@@ -46,7 +46,7 @@ type Framework struct {
 var _ FullClientInterface = &Framework{}
 var _ ClusterInterface = &Framework{}
 
-func NewFramework(namePrefix string) *Framework {
+func NewFramework(ctx context.Context, namePrefix string) *Framework {
```
rzetelskik, line 55: Since the constructor now implicitly requires to be called in BeforeEach, should we document it in a comment (on the function or at the top of the file)?

abramche: Ah, that explains the comment that @czeslavo submitted before (and I did not understand why) - re-introduced in b604a8775db1f03fd1b5d865be134aed324475ef

test/e2e/framework/framework.go
```diff
@@ -46,7 +46,13 @@ type Framework struct {
 var _ FullClientInterface = &Framework{}
 var _ ClusterInterface = &Framework{}
 
-func NewFramework(namePrefix string) *Framework {
+// NewFramework creates a new test framework. It creates a unique namespace for a test to run in and provides a client
+// with permissions limited to that namespace. It also creates clients for worker clusters (for multi-DC) if configured.
+// It should be called in a ginkgo BeforeEach closure.
+//
+// At the end of the test, artifacts are going to be collected, and the namespace and all objects in it will be
+// automatically, unless the cleanup policy is set to skip cleanup.
```
rzetelskik:
```suggestion
// At the end of the test, artifacts are going to be collected, and the namespace and all objects in it will be
// automatically cleaned up, unless the cleanup policy is set to skip cleanup.
```

abramche: done, also reduced tautology in the following sentence `, unless the cleanup policy is set to skip.`

rzetelskik:
lgtm
/assign @czeslavo

Issue #2778, opened 2025-07-15
Updates to `.spec.rackTemplate.scyllaDBManagerAgent.CustomConfigSecretRef` cause a config drift


rzetelskik:
### What happened?

Updates to `.spec.rackTemplate.scyllaDBManagerAgent.CustomConfigSecretRef` of v1alpha1.ScyllaDBDatacenter are allowed, but they result in an error from STS controller:
```
E0715 10:27:13.969355       1 scylladbdatacenter/controller.go:268] "Unhandled Error" err="syncing key 'scylla/basic' failed: can't sync stateful sets: can't apply statefulset update: can't update apps/v1, Kind=StatefulSet \"scylla/basic-us-east-1-us-east-1a\": StatefulSet.apps \"basic-us-east-1-us-east-1a\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden" logger="UnhandledError"
```

This can also break integration with ScyllaDB Manager. If auth_token is specified in the custom config, it is propagated to ScyllaDB Manager server, but not to ScyllaDB Manager agents, resulting in a drift.

### What did you expect to happen?

Updates to SDC which should but can't be propagated to the underlying StatefulSets through a regular apply, should either be forbidden or propagate correctly. 

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a ScyllaDBDatacenter with no `.spec.rackTemplate.scyllaDBManagerAgent.CustomConfigSecretRef` defined (and no non-templated fields overriding it).
2. Wait for readiness.
3. Update ScyllaDBDatacenter to specify `.spec.rackTemplate.scyllaDBManagerAgent.CustomConfigSecretRef`.

### Scylla Operator version

master

### Kubernetes platform name and version

n/a

### Please attach the must-gather archive.

n/a

mflendrich:
Proposed solution: implement a validating webhook banning the secret ref change.

mflendrich:
Now tracked as [OPERATOR-31](https://scylladb.atlassian.net/browse/OPERATOR-31) in the internal Jira.

[...]

PR #3309, opened 2026-03-06
Avoid disallowed StatefulSet volumeClaimTemplates mutations when rackTemplate is set

czeslavo:

Setting `.spec.rackTemplate` on an existing `ScyllaDBDatacenter` (e.g., to configure `scyllaDBManagerAgent.CustomConfigSecretRef`) triggers `applyRackTemplateOnRackSpec`, which unconditionally returns a non-nil `Storage.Metadata` pointer with empty `Labels`/`Annotations` maps. `StatefulSetForRack` function used `!= nil` to decide which PVC template labels to use, so that the check always runs, causing it to compute different `volumeClaimTemplates` labels/annotations than the ones initially written to `StatefulSet`. This caused the controller to update `volumeClaimTemplates`, which are immutable, putting the controller in a permanent error loop and blocking the intended change from taking effect.

This PR fixes the logic so that for an existing `StatefulSet`, `volumeClaimTemplates` labels and annotations are always copied from the existing object rather than recomputed. It also tightens the "explicitly configured metadata" check from `!= nil` to `len(...) > 0` to correctly distinguish an empty map produced by template merging from a map that was actually set by the user.

Resolves https://github.com/scylladb/scylla-operator/issues/2778

### DIFF
--- a/pkg/api/scylla/validation/scylladbdatacenter_validation_test.go
+++ b/pkg/api/scylla/validation/scylladbdatacenter_validation_test.go
@@ -771,6 +771,193 @@ func TestValidateScyllaDBDatacenterUpdate(t *testing.T) {
 			expectedErrorString: "spec.racks[0].scyllaDB.storage: Forbidden: changes in storage are currently not supported",
 		},
 		{
+			name: "rack storage storageClassName changed",
+			old: func() *scyllav1alpha1.ScyllaDBDatacenter {
+				sdc := newValidScyllaDBDatacenter()
+				sdc.Spec.Racks[0].RackTemplate.ScyllaDB.Storage.StorageClassName = pointer.Ptr("old-class")
+				return sdc
+			}(),
+			new: func() *scyllav1alpha1.ScyllaDBDatacenter {
+				sdc := newValidScyllaDBDatacenter()
+				sdc.Spec.Racks[0].RackTemplate.ScyllaDB.Storage.StorageClassName = pointer.Ptr("new-class")
+				return sdc
+			}(),
+			expectedErrorList: field.ErrorList{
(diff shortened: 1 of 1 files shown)


pkg/controller/scylladbdatacenter/resource_test.go
```diff
@@ -1902,6 +1902,136 @@ exec scylla-manager-agent \
...
+					"scylla/datacenter":            "dc",
+					"scylla/rack":                  "rack",
+					"custom-pvc-label":             "custom-pvc-label-value",
+				}
+				return sts
+			}(),
+			expectedError: nil,
+		},
+		{
+			name: "existing StatefulSet PVC template labels are preserved and not overwritten by current rack storage metadata labels",
```
rzetelskik, line 1968: Q: Is preserving the annotations/labels actually the desired behaviour if they differ (as compared to e.g. throwing an error)? Looking at the validation, isn't this an anomaly? It seems like the labels/annotations in the volume claim template should be immutable.

rzetelskik: On the other hand, it's probably better to err on the side of doing something reasonable here than throwing an error

czeslavo: I believe that we should leave this approach in the controller as a safety net, leaving the system working instead of blocking the controller with an error.

The other aspect is that, indeed, we could validate in the webhook that anything that directly maps to the PVC template is immutable.

czeslavo: The validation is already there. I added explicit unit test cases that cover every field separately.

Issue #2985, opened 2025-09-29
v1.ScyllaCluster/v1alpha1.ScyllaDBDatacenter translation does not aggregate status conditions


rzetelskik:
### What happened?

The v1.ScyllaCluster/v1alpha1.ScyllaDBDatacenter translation controller does not aggregate status conditions. The aggregated status conditions, which are part of our publicly documented API (https://operator.docs.scylladb.com/v1.18/resources/scyllaclusters/basics.html), are copied over verbatim from the underlying ScyllaDBDatacenter, instead of being aggregated from all conditions. This may result in some partial conditions not being accounted for.

### What did you expect to happen?

The aggregated conditions should be aggregated from all conditions, instead of copied over verbatim from the underlying resource.

### How can we reproduce it (as minimally and precisely as possible)?

n/a

### Scylla Operator version

master

### Kubernetes platform name and version

n/a

### Please attach the must-gather archive.

n/a

mflendrich:
What conditions are there  for ScyllaCluster that aren't there on the ScyllaDBDatacenter?

Why are they important?
If they are important - why aren't they present on ScyllaDBDatacenter?

What I'm trying to get at -- maybe it is good to keep the ScyllaCluster-ScyllaDBDatacenter translator trivial and keep its logic to a minimum.

rzetelskik:
> What conditions are there for ScyllaCluster that aren't there on the ScyllaDBDatacenter?

Anything related to ScyllaDBManager which is a part of the legacy ScylaCluster and not a part of ScyllaDBDatacenter - for instance anything ScyllaDB Manager related (moved to ScyllaDBManagerClusterRegistration, ScyllaDBManagerTask).

> Why are they important?
> If they are important - why aren't they present on ScyllaDBDatacenter?

They are still a functionality of a ScyllaCluster controller, but no longer of a ScyllaDBDatacenter controller.

> What I'm trying to get at -- maybe it is good to keep the ScyllaCluster-ScyllaDBDatacenter translator trivial and keep its logic to a minimum.

It's already extended to translate ScyllaCluster fields into ScyllaDBManagerTasks etc. So it goes one way, but some of the status conditions of the translated resources are not taken into account in status calculation. The top-level condition aggregates are copied over from ScyllaDBDatacenter instead of being calculated by the translation controller.

mflendrich:
Step at clarifying the definition of done: (@rzetelskik 's words)

[...]

Issue #2984, opened 2025-09-29
v1.ScyllaCluster/v1alpha1.ScyllaDBDatacenter translation does not account for generation drift


rzetelskik:
### What happened?

ScyllaCluster/ScyllaDBDatacenter translation controller does not account for potential generation drift between the two resources. 
Some mechanism can cause the generation of the two resources to drift. An example of such mechanism is the translation of `ScyllaCluster/v1.spec.sysctls` to a annotation in `ScyllaDBDatacenter/v1alpha1`. Adjusting the spec increases a resource's generation, while an update of annotations does not. 
Meanwhile, the controller copies some of the status and the status conditions over verbatim from ScyllaDBDatacenter to ScyllaCluster. This results in a drift between ScyllaCluster's generation and its observed generation. This in in turn breaks our mechanism of broadcasting the resource's readiness.

### What did you expect to happen?

The controller should account for the potential generation drift.

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a ScyllaCluster without sysctls defined
```
curl -L https://raw.githubusercontent.com/scylladb/scylla-operator/v1.18/examples/scylladb/scylla.scyllacluster.yaml | yq 'del(.spec.sysctls)' - | kubectl apply -n=scylla --server-side -f=-
```
2. Wait for it to roll out
```
kubectl -n scylla wait --timeout=5m --for='condition=Progressing=False' scyllacluster.scylla.scylladb.com/scylla
kubectl -n scylla wait --timeout=5m --for='condition=Degraded=False' scyllacluster.scylla.scylladb.com/scylla
kubectl -n scylla wait --timeout=5m --for='condition=Available=True' scyllacluster.scylla.scylladb.com/scylla
```
3. Update the ScyllaCluster with sysctls' spec
```
kubectl apply -n=scylla --server-side -f=https://raw.githubusercontent.com/scylladb/scylla-operator/v1.18/examples/scylladb/scylla.scyllacluster.yaml
```
4. Observe the ScyllaCluster never reaching the desired status conditions (as in step 2)

### Scylla Operator version

master

### Kubernetes platform name and version

n/a

### Please attach the must-gather archive.

n/a

mflendrich:
> The controller should account for the potential generation drift.

I am sensing some XY problem here. Generation drift is a consequence of the API design choice - that we translate from spec to an annotation. Maybe this was a wrong design choice that we should backtrack from (and consider retiring the offending annotations in favor of spec fields).

PR #3311, opened 2026-03-09
Fix condition aggregation in ScyllaCluster controller

rzetelskik:
ScyllaCluster translation controller currently propagates status conditions directly from the underlying ScyllaDBDatacenter.
This is incorrect as it doesn't take it's own partial conditions into account, resulting in e.g. ScyllaDBManagerTask application errors not being included in the aggregates. See https://github.com/scylladb/scylla-operator/issues/2985 and https://github.com/scylladb/scylla-operator/issues/3294.
Translation also doesn't account for the generation drift between SC and SDC resources, leading to issues such as described in https://github.com/scylladb/scylla-operator/issues/2984.

This change fixes the aggregation logic and covers it with envtest tests.

The controller now:
- Propagates SDC's controller conditions into SC's status, offset by the generation skew between SC and SDC so that conditions' observed generation matches SC's generation space.
- Combines SDC-sourced conditions with its own controller partial conditions (ScyllaDBDatacenterControllerProgressing/Degraded, ScyllaDBManagerTaskControllerProgressing/Degraded) to produce the final aggregate. If the conditions are missing from the underlying SDC, the aggregates are either set to unknown or an "offending" condition coming from SC's own controller partial conditions.

Resolves https://github.com/scylladb/scylla-operator/issues/2985
Resolves https://github.com/scylladb/scylla-operator/issues/2984
Partially addresses https://github.com/scylladb/scylla-operator/issues/3294.

/kind bug
/priority important-soon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@
 
 -  Fixed [#2990](https://github.com/scylladb/scylla-operator/issues/2990): set monitoring scrape intervals to: 5s in 
   `ServiceMonitor`, and 30s in Grafana. This overrides the global scrape interval set in Prometheus.
+- `ScyllaCluster`'s translation controller now combines `ScyllaDBDatacenter` status conditions with its own controller partial conditions when aggregating `ScyllaCluster`'s status conditions,
+  and correctly offsets their observed generations by the generation skew between the two resources.
+  [#3311](https://github.com/scylladb/scylla-operator/pull/3311)
 
 ## 1.20.0 and older
 


pkg/controller/scyllacluster/sync.go
```diff
@@ -271,10 +274,27 @@ func (scmc *Controller) sync(ctx context.Context, key string) error {
...
+		// and users can also directly bump SDC's generation without touching SC.
+		// We compensate by shifting all SDC-reported ObservedGenerations into SC's generation space.
+		// The result is clamped to 0 to satisfy the CRD's minimum: 0 constraint: a condition whose
+		// shifted ObservedGeneration would be negative is from a generation too old to be relevant,
+		// and clamping it to 0 ensures it won't match sc.Generation.
+		generationOffset := sc.Generation - sdc.Generation
+		for i := range sdcConditions {
+			sdcConditions[i].ObservedGeneration = max(0, sdcConditions[i].ObservedGeneration+generationOffset)
+		}
+	}
```
czeslavo, line 283: It's quite an important domain logic that I believe we should extract to a helper and have it unit tested for easier understanding of cases that we're expected to handle with this. I had to spend some time wrapping my head around this, and such a unit test with explicit test cases would be helpful, I believe.

test/envtest/setup.go
```diff
@@ -106,3 +116,94 @@ func (e *Environment) Namespace() string {
 func (e *Environment) Config() *rest.Config {
 	return e.config
 }
+
+// SetupValidatingWebhook installs a ValidatingWebhookConfiguration that intercepts all operations
+// on all resources under the scylla.scylladb.com API group (v1 and v1alpha1) and dispatches them
+// to handleFunc. The webhook server is started automatically and cleaned up when the test ends.
```
czeslavo: Let's make it clear in the comment that this is not running the webhook server Operator implements. We could also improve the name to emphasize this aspect, e.g., by changing it to `SetupMockValidatingWebhook`.

test/envtest/controllers/scyllacluster_controller_test.go
czeslavo, line 1: As discussed offline, let's see whether there's space for deduplication by extracting helper functions (e.g., for creating SC and waiting for SDC).

test/envtest/setup.go
rzetelskik: done

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@
 
 -  Fixed [#2990](https://github.com/scylladb/scylla-operator/issues/2990): set monitoring scrape intervals to: 5s in 
   `ServiceMonitor`, and 30s in Grafana. This overrides the global scrape interval set in Prometheus.
+- ScyllaCluster's translation controller now combines ScyllaDBDatacenter status conditions with its own controller partial conditions when aggregating ScyllaCluster's status conditions,
+  and correctly offsets their observed generations by the generation skew between the two resources.
+  [#3311](https://github.com/scylladb/scylla-operator/pull/3311)
```
czeslavo: Not strictly related to this entry, as we already have the inconsistency in this file, but let's agree whether we put CR kinds in ticks or not. I think that using ticks might help with readability a bit?

pkg/controller/scyllacluster/sync.go
rzetelskik: done, ptal

czeslavo: Thanks, that helps.

CHANGELOG.md
rzetelskik: It might be hard to maintain a unified approach but I don't mind, I'll amend it

mflendrich: > It might be hard to maintain a unified approach

I'll post a PR shortly

mflendrich: https://github.com/scylladb/scylla-operator/pull/3318

test/envtest/controllers/scyllacluster_controller_test.go
rzetelskik: gave it a shot, ptal

CHANGELOG.md
rzetelskik: amended

test/envtest/controllers/scyllacluster_controller_test.go
czeslavo: Looks more readable to me now, thanks.

PR #3313, opened 2026-03-09
Improve ScyllaCluster orphaned PV tests

czeslavo:

Tries stabilizing the `ScyllaCluster should replace a node with orphaned PV` E2E test by making its mock PVC/PV cloning ready for transient Kubernetes API errors: `AlreadyExists` and `NotFound` errors during creation and cleanup are ignored. PVC rebinding retries on update conflicts, skipping work if the original PVC is already bound to prevent update conflicts from causing the test to fail. The test has been run 10 times locally after the fix, with 100% success rate.

Also:
- improves the controller's coverage by adding an envtest suite.
- adds a timeout to cql `DataInserter`'s `Close` method as it managed to get into a deadlock (see [logs](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3313/pull-scylla-operator-master-e2e-gke-parallel/2031090024505675776))

Closes https://github.com/scylladb/scylla-operator/issues/2444

### DIFF
--- a/test/e2e/utils/datainserter.go
+++ b/test/e2e/utils/datainserter.go
@@ -18,7 +18,13 @@ import (
 	apimachineryutilrand "k8s.io/apimachinery/pkg/util/rand"
 )
 
-const nRows = 10
+const (
+	nRows = 10
+
+	// cqlSessionCloseTimeout is the maximum time to wait for gocql Session.Close() to complete. If Close() does not finish
+	// within this window we abandon it and log a warning to avoid hanging the entire test suite.
+	cqlSessionCloseTimeout = 30 * time.Second
+)
 
 type DataInserter struct {
 	session           *gocqlx.Session
@@ -85,8 +91,18 @@ func NewMultiDCDataInserter(dcHosts map[string][]string, options ...DataInserter
 }
 
 func (di *DataInserter) Close() {
-	if di.session != nil {
-		di.session.Close()
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		if di.session != nil {
+			di.session.Close()
+		}
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(cqlSessionCloseTimeout):
+		framework.Infof("WARNING: gocql Session.Close() did not return within %v; abandoning to avoid suite hang", cqlSessionCloseTimeout)
 	}
 }
 


czeslavo:
The E2E test was stable locally and in isolation (no other tests were run alongside), but it looks like after changes it managed to get stuck in a brand new way. Will look into that.

test/e2e/set/scyllacluster/scyllacluster_pv.go
```diff
@@ -154,7 +156,7 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", func() {
 
 					framework.Infof("Creating clone PVC for %q", pvc.Name)
 					pvcClone, err := f.KubeClient().CoreV1().PersistentVolumeClaims(f.Namespace()).Create(ctx, pvcClone, metav1.CreateOptions{})
-					o.Expect(err).NotTo(o.HaveOccurred())
+					o.Expect(err).To(framework.NotHaveOccurredExceptAlreadyExists())
```
rzetelskik: why not use Apply if we expect it to already exist?

test/envtest/controllers/orphanedpv_test.go
```diff
@@ -0,0 +1,259 @@
...
+			kubeInformers.Core().V1().PersistentVolumeClaims(),
+			kubeInformers.Core().V1().Nodes(),
+			scyllaInformers.Scylla().V1alpha1().ScyllaDBDatacenters(),
+		)
+		o.Expect(err).NotTo(o.HaveOccurred(), "Failed to create orphaned PV controller")
+
+		kubeInformers.Start(ctx.Done())
+		scyllaInformers.Start(ctx.Done())
+
+		go controller.Run(ctx, 1)
```
rzetelskik:
```suggestion
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		controller.Run(ctx, 1)
	}()

	g.DeferCleanup(func() {
		kubeInformers.Shutdown()
		scyllaInformers.Shutdown()
		wg.Wait()
	})
```

test/e2e/set/scyllacluster/scyllacluster_pv.go
czeslavo: Good point. Switched all `NotHaveOccurredExceptAlreadyExists` sites to use `Apply*` pattern from `resourceapply` package.

test/envtest/controllers/orphanedpv_test.go
czeslavo: Applied.

PR #3318, opened 2026-03-10
Ensure that CRD names in CHANGELOG.md are quoted with `backticks`

mflendrich:
A crapshoot attempt at addressing https://github.com/scylladb/scylla-operator/pull/3311#discussion_r2911474911.

Open questions
- Is the unit test suite the right place for this test? Should it be a separate suite from `make test.unit`? Maybe it should live under `verify` or somewhere else?
- Should we expand this to Kubernetes resource names `Service`, `Pod` etc? and Prometheus/cert-manager CRDs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@
 
 ### API Changes
 
-- Extended the admission webhook to emit warnings when ScyllaCluster's backup or repair task names do not adhere to RFC 1123
-  subdomain requirements. Invalid task names currently cause silent failures where the underlying ScyllaDBManagerTask objects fail to be created.
-  **In the next minor release, these warnings will become validation errors that prevent ScyllaCluster creation or updates.**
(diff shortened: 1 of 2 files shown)


rzetelskik:
> Is the unit test suite the right place for this test? Should it be a separate suite from make test.unit? Maybe it should live under verify or somewhere else?

A verify target would be more aligned with current practices. While I don't see anything wrong with this approach in principle, it could use a separate "check", i.e. it seems confusing that a misaligned changelog entry would result in a unit test suite failure.

> Should we expand this to Kubernetes resource names Service, Pod etc? and Prometheus/cert-manager CRDs?

I'd say so. A good enough rule of thumb could be anything that's in our schema: https://github.com/scylladb/scylla-operator/blob/f37e0386d690fec61e3883ccd1f55dc28a5a7f9e/pkg/scheme/scheme.go. Although that would miss some dependencies like cert-manager CRDs or anything OpenShift-related.

PR #3319, opened 2026-03-10
Introduce cert manager version update (semi-)automation

### LINKED ISSUE #2914: Pin cert-manager version and automate its update
## Acceptance criteria
- [ ] Cert-manager version is pinned in a single source of truth and manifests update process is automated.

/kind machinery
/priority important-longterm

abramche:
Introduce cert-manager update infra and renovate config (based off prometheus operator)

Resolves #2914

### DIFF
--- a/assets/config/config.yaml
+++ b/assets/config/config.yaml
@@ -23,6 +23,9 @@ operatorTests:
   # renovate: datasource=custom.envtest depName=envtest versioning=semver
   envTestKubernetesVersion: "v1.35.0"
 thirdParty:
+  certManager:
+    # renovate: datasource=github-releases depName=cert-manager packageName=cert-manager/cert-manager versioning=semver
+    version: "1.17.4"
   prometheusOperator:
     # renovate: datasource=github-releases depName=prometheus-operator packageName=prometheus-operator/prometheus-operator versioning=semver
     version: "0.89.0" # Latest compatible with .operator.prometheusVersion (https://prometheus-operator.dev/docs/getting-started/compatibility/)
--- a/hack/third-party/build-cert-manager-manifest.sh
+++ b/hack/third-party/build-cert-manager-manifest.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2026 ScyllaDB
+#
+# This script builds the cert-manager manifest with a specified version following the upstream instructions.
+# https://cert-manager.io/docs/installation/kubectl/
+
+set -euxEo pipefail
+shopt -s inherit_errexit
+
+source "$( dirname "${BASH_SOURCE[0]}" )/../lib/assets.sh"
+
+if [[ -n "${1+x}" ]]; then
+    target="${1}"
+else
+    printf 'Missing target file.\nUsage: %s <target>\n' "$0" >&2
+    exit 1
+fi
+
+version=$( get-config ".thirdParty.certManager.version" )
+
+url="https://github.com/cert-manager/cert-manager/releases/download/v${version}/cert-manager.yaml"
+curl --fail --retry 5 --retry-all-errors -L "${url}" -o "${target}"
--- a/renovate.json
+++ b/renovate.json
@@ -82,6 +82,32 @@
       "enabled": false
     },
     {
+      "description": "cert-manager dependency should be updated every weekday",
+      "matchDepNames": [
+        "cert-manager"
+      ],
+      "schedule": [
+        "before 5am every weekday"
+      ],
+      "addLabels": [
+        "do-not-merge/hold"
+      ],
+      "prBodyNotes": [
+        "### ⚠️ Manual Action Required",
+        "Renovate has bumped the cert-manager dependency, but cannot automatically regenerate the manifests.",
+        "",
+        "**Please complete the following steps before merging:**",
+        "1. Check out this branch locally.",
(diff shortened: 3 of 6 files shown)


renovate.json
```diff
@@ -81,6 +81,32 @@
...
+        "**Please complete the following steps before merging:**",
+        "1. Check out this branch locally.",
+        "2. Propagate changes by running:",
+        "   ```bash",
+        "   make update",
+        "   git add .",
+        "   git commit --amend && git push --force",
+        "   ```",
+        "3. Remove the `do-not-merge/hold` label."
+      ]
```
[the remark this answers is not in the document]
abramche: it's working in the same way as prometheus operator and is intended -> if we change it to full auto, it will be a separate change with a separate discussion

hack/third-party/build-cert-manager-manifest.sh
```diff
@@ -0,0 +1,23 @@
...
+    target="${1}"
+else
+    echo "Missing target file.\nUsage: ${0} <target>" >&2 >/dev/null
+    exit 1
+fi
+
+version=$( get-config ".thirdParty.certManager.version" )
+
+url="https://github.com/cert-manager/cert-manager/releases/download/v${version}/cert-manager.yaml"
+curl --fail --retry 5 --retry-all-errors -L "${url}" -o "${target}"
```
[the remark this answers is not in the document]
abramche: likewise, same approach as prometheus operator

renovate.json
```diff
@@ -81,6 +81,32 @@
...
+      "prBodyNotes": [
+        "### ⚠️ Manual Action Required",
+        "Renovate has bumped the cert-manager dependency, but cannot automatically regenerate the manifests.",
+        "",
+        "**Please complete the following steps before merging:**",
+        "1. Check out this branch locally.",
+        "2. Propagate changes by running:",
+        "   ```bash",
+        "   make update",
+        "   git add .",
```
rzetelskik: Can we make it more specific? I know this is what other notes are using but it's actually quite annoying if you have any local uncommitted files.

assets/config/config.yaml
```diff
@@ -23,6 +23,9 @@ operatorTests:
   # renovate: datasource=custom.envtest depName=envtest versioning=semver
   envTestKubernetesVersion: "v1.35.0"
 thirdParty:
+  certManager:
+    # renovate: datasource=github-releases depName=cert-manager packageName=cert-manager/cert-manager versioning=semver
+    version: "1.17.4"
```
rzetelskik, line 28: let's extend `TestProjectConfig` table with an entry verifying that it's present and it's a valid semantic version

rzetelskik, line 28: Should we also propagate it to the support matrix in docs? It's a bit out of scope for the issue this PR is addressing but we discussed including the dependencies there - perhaps as a separate "tested dependency versions" matrix. cc @czeslavo @mflendrich

hack/third-party/build-cert-manager-manifest.sh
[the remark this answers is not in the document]
rzetelskik: It'd be a valid point if this wasn't a one time action. We'd learn about a change by verify makefile target failing.

assets/config/config.yaml
abramche: It would be a drag for us to maintain a third-party version matrix imo, but maybe we can add that we ship the release with the latest version of cert-manager (at the time of the release) - that should tick the checkbox?

abramche: good point, addressed in 494358288ff26435a7d3619271aef612b048e30d

rzetelskik: That's what I mean - documenting a single version we test with and know it works.

czeslavo: I'd be happy to see that added. I imagine it as a separate "tested with" table with just "name" and "version" columns, where we put the "runtime dependencies", i.e., cert-manager, prometheus-operator, prometheus, and grafana. At some point, we might also add the Kubernetes/GKE/Kind version we test with, if we shift where the version is defined to the operator repo.

czeslavo: But I think it would be perfectly fine to tackle this separately from this PR, adding the whole table at once for all dependencies.

abramche: I am in favour of a separate PR because it makes sense to do it for all dependencies

renovate.json
abramche: addressed in 744b7502dc024f93e9eb7c3075b43e57691e2867

renovate.json
```diff
@@ -81,6 +81,32 @@
...
+      "prBodyNotes": [
+        "### ⚠️ Manual Action Required",
+        "Renovate has bumped the cert-manager dependency, but cannot automatically regenerate the manifests.",
+        "",
+        "**Please complete the following steps before merging:**",
+        "1. Check out this branch locally.",
+        "2. Propagate changes by running:",
+        "   ```bash",
+        "   make update",
+        "   git add assets/config/config.yaml examples/third-party/cert-manager.yaml",
```
rzetelskik: nit: `git add examples/third-party/cert-manager.yaml` should suffice, assets should already be covered by the renovate action

abramche: done in 3cc5abd7284c6d8b3eb05c35f873b07240aee04e

PR #3320, opened 2026-03-10
Fix relative paths in debug-flake.sh

mflendrich:
The script fails early because everything relative to `$parent_dir` is incorrectly referenced.

This PR fixes the paths.

/kind machinery
/priority important-soon


czeslavo (approved):
/lgtm

PR #3322, opened 2026-03-10
Do not use legacy cert-manager locations in the docs and infra code

abramche:
Replace the obsolete usages of symlinked `examples/common/cert-manager.yaml` with the current, new location. The old symlink is still in place for redundancy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -809,7 +809,7 @@ help:
 .PHONY: help
 
 cert-manager:
-	kubectl apply -f examples/common/cert-manager.yaml
+	kubectl apply -f examples/third-party/cert-manager.yaml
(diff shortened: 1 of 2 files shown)


abramche:
> There seems to be a spurious submodule update that shouldn't be part of this PR I think?
> 
> otherwise lgtm

yes, I am unsure what happened there. Dropped that change

mflendrich (approved):
> yes, I am unsure what happened there. Dropped that change

Most likely, some other branch you had checked out previously has a different version of the monitoring submodule, and two things happened:
1. you haven't run `git submodule update` after checking out the other branch,
2. you staged the unwanted submodule change when committing (`git commit -a` or similar).

/lgtm

PR #3332, opened 2026-03-12
docs(CONTRIBUTING.md, copilot-instructions.md): introduce conventional commits

abramche:
- docs(CONTRIBUTING.md): introduce conventional commits
   also update copilot-instructions.md to reflect this change

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,13 +1,9 @@
 ## Copilot Instructions for ScyllaDB Operator
 
-The commit message should always (no exceptions) be an imperative sentence without a trailing dot. Some examples:
-* "Document automatic data cleanups"
-* "Add IPv6 support for ScyllaClusters"
-* "Reword platforms to environments"
-* "Update maintainers in the Helm charts"
+Follow the project's [CONTRIBUTING.md](../CONTRIBUTING.md) and [API_CONVENTIONS.md](../API_CONVENTIONS.md).
 
-Prefer commit message length under 72 characters.
+### Commits and PRs
 
-The commit message should never include "@mentions" or hashtag ("#") references. The commit message should never include [keywords that close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).
-
-The PR title should follow the same rules as the commit message.
+- Format commit messages and PR titles as [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/): `<type>(<scope>): <subject>` or `<type>: <scopeless subject>`, where subject is imperative, lowercase, ≤72 chars, and has no trailing period. 
+- Never include `@mentions`, [keywords that close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) or hashtag ("#") references in the commit messages. 
+- If the PR fixes an issue, reference it with `Resolves #<number>` only in the PR description.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,22 +109,7 @@ To maintain a clean commit history, aim for a minimal number of logical commits.
 
 ### Commits and PRs
 
-The **subject line** of your commit message **and the PR title** should summarize the change in one clear sentence that would be meaningful to a user of the Operator. The sentence should be **written in the imperative**, i.e. written as if giving a command or instruction, e.g. "Add support for XYZ". A properly formed Git commit subject line should always be able to complete the sentence "If applied, this commit will...", e.g.  "If applied, this commit will **Add support for XYZ**".
-Changelog entries are verbatim PR titles, so make them concise yet informative.
-
-Further details should be added after a blank line. Explain why the change was necessary, not just what was changed. In the general case, extensive descriptions are well-received. Comparing the behavior before and after the change is especially helpful. Write the message with the mindset that you'll need to revisit the code in the future.
-If your PR fixes an issue, include "Resolves #1234" in the description, replacing "1234" with the issue number.
-
-```
-**Description of your changes:**
-Add new XYZ field to ScyllaCluster CRD.
-The new field allows for configuration of ZYX feature of ScyllaDB.
-<more details>
-API change was discussed in the following enhancement: <link>.
-
-**Which issue is resolved by this Pull Request:**
-Resolves #1234 
-```
+We follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for all commit messages and PR titles.
 
 ### Changelog entries
 


.github/copilot-instructions.md
```diff
@@ -1,13 +1,14 @@
 ## Copilot Instructions for ScyllaDB Operator
 
-The commit message should always (no exceptions) be an imperative sentence without a trailing dot. Some examples:
-* "Document automatic data cleanups"
-* "Add IPv6 support for ScyllaClusters"
-* "Reword platforms to environments"
-* "Update maintainers in the Helm charts"
+Follow the project's [CONTRIBUTING.md](../CONTRIBUTING.md) and [API_CONVENTIONS.md](../API_CONVENTIONS.md) for all coding conventions, commit message formatting, changelog entries, and PR guidelines.
```
mflendrich: I'm worried that the rest of this sentence might narrow down the extent to which agents will follow the instructions. Suggested improvement.

```suggestion
Follow the project's [CONTRIBUTING.md](../CONTRIBUTING.md) and [API_CONVENTIONS.md](../API_CONVENTIONS.md).
```

CONTRIBUTING.md
```diff
@@ -109,19 +109,34 @@ To maintain a clean commit history, aim for a minimal number of logical commits.
...
-Changelog entries are verbatim PR titles, so make them concise yet informative.
+We follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for all commit messages and PR merge messages.
 
-Further details should be added after a blank line. Explain why the change was necessary, not just what was changed. In the general case, extensive descriptions are well-received. Comparing the behavior before and after the change is especially helpful. Write the message with the mindset that you'll need to revisit the code in the future.
-If your PR fixes an issue, include "Resolves #1234" in the description, replacing "1234" with the issue number.
+#### Format of commit messages and PR titles
```
mflendrich: This section looks completely redundant - no?
suggestion: remove

I feel that, with respect to rules, "less is more". It might be better to have a narrow set of strictly enforced rules than an elaborate description that will blur the picture.

.github/copilot-instructions.md
```diff
@@ -1,13 +1,14 @@
...
-The commit message should never include "@mentions" or hashtag ("#") references. The commit message should never include [keywords that close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).
+Format commit messages and PR titles as [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/): `<type>(<scope>): <subject>` or `<type>: <scopeless subject>`, where subject is imperative, lowercase, ≤72 chars, and has no trailing period. Never include `@mentions` or issue-closing keywords in the subject line. Reference issues with `#<number>` only in the commit footer or PR body.
 
-The PR title should follow the same rules as the commit message.
+### Useful commands
+
+* `make build` - build project binaries
+* `make test-unit` - run unit tests
+* `make update` - run all code generators
+* `make verify` - run all linters and verify autogenerated files are up to date
```
rzetelskik: seems unrelated to the PR and not very practical for AI instructions

.github/copilot-instructions.md
```diff
@@ -1,13 +1,14 @@
...
-* "Document automatic data cleanups"
-* "Add IPv6 support for ScyllaClusters"
-* "Reword platforms to environments"
-* "Update maintainers in the Helm charts"
+Follow the project's [CONTRIBUTING.md](../CONTRIBUTING.md) and [API_CONVENTIONS.md](../API_CONVENTIONS.md) for all coding conventions, commit message formatting, changelog entries, and PR guidelines.
 
-Prefer commit message length under 72 characters.
+### Commits and PRs
 
-The commit message should never include "@mentions" or hashtag ("#") references. The commit message should never include [keywords that close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).
```
rzetelskik, line 11: why remove the reference?

CONTRIBUTING.md
```diff
@@ -109,19 +109,34 @@ To maintain a clean commit history, aim for a minimal number of logical commits.
 
 ### Commits and PRs
 
-The **subject line** of your commit message **and the PR title** should summarize the change in one clear sentence that would be meaningful to a user of the Operator. The sentence should be **written in the imperative**, i.e. written as if giving a command or instruction, e.g. "Add support for XYZ". A properly formed Git commit subject line should always be able to complete the sentence "If applied, this commit will...", e.g.  "If applied, this commit will **Add support for XYZ**".
-Changelog entries are verbatim PR titles, so make them concise yet informative.
+We follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for all commit messages and PR merge messages.
```
rzetelskik: `and PR merge messages`  - doesn't seem to be true as of now?

CONTRIBUTING.md
```diff
@@ -109,19 +109,34 @@ To maintain a clean commit history, aim for a minimal number of logical commits.
...
 
+type: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`, `build`, `ci`
+
+scope: optional, short module/component
+
+subject: imperative, lowercase, ≤72 chars, no trailing period
+
+Keep in mind:
+- Use **BREAKING CHANGE: description** in the body for breaking changes
+- Reference issues with `#<number>` in the commit footer or PR body (if applicable)
```
rzetelskik: nit: this is less precise than it was before, you need a keyword for it to close an issue

CONTRIBUTING.md
```diff
@@ -109,19 +109,34 @@ To maintain a clean commit history, aim for a minimal number of logical commits.
...
+- Keep messages concise and meaningful
+- Squash/merge commits must preserve a Conventional Commit as the final message
```
rzetelskik: a contributor has no control over these, no?

CONTRIBUTING.md
rzetelskik: +1, I think the syntax part is redundant

.github/copilot-instructions.md
abramche: Ah, I see, I supposed an agent may run build commands - this does not belong here

.github/copilot-instructions.md
abramche: By the conventional commits, these do not belong to the commit messages anyway. But I don't mind leaving it for explicit instruction.

CONTRIBUTING.md
abramche: Hmm I just noticed there's a conflicting suggestion allowing the use of `#<number>` in the commit footer. I'll constrict to the PR body (if it fixes an issue)

CONTRIBUTING.md
abramche: Yes, good catch. I meant `PR titles` - adjusted for that

CONTRIBUTING.md
abramche: A contributor is fully in control of a squash commit, including the message. Do you think this sentence is confusing?

mflendrich: I still think that these instructions are excessively specific. There is no way that we will be enforcing them, and they will turn ineffective. I stand by https://github.com/scylladb/scylla-operator/pull/3332#discussion_r2945559415 with respect to the sentiment that we should not be over-regulating; just state the preferred format and leave it at that.

CONTRIBUTING.md
abramche: OK with me, dropped the section

CONTRIBUTING.md
abramche: I dropped the section

.github/copilot-instructions.md
```diff
@@ -1,13 +1,9 @@
...
-
-The PR title should follow the same rules as the commit message.
+- Format commit messages and PR titles as [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/): `<type>(<scope>): <subject>` or `<type>: <scopeless subject>`, where subject is imperative, lowercase, ≤72 chars, and has no trailing period. 
+- Never include `@mentions`, [keywords that close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) or hashtag ("#") references in the commit messages. 
+- If the PR fixes an issue, reference it with `#<number>` only in the PR description.
```
rzetelskik: In that case it also needs to be prefixed with Fixes/Resolves - otherwise it won't close the issue. Not sure if we need to mention this explicitly as it's a part of the template.

abramche: It's an agentic instruction, better to err on the explicit side, no?

```
- If the PR fixes an issue, reference it with `Resolves #<number>` only in the PR description.
```

abramche: 16154909143c1f8b801524aa7ad9d92b49c8957d

rzetelskik:
lgtm
/assign @mflendrich

Issue #3063, opened 2025-11-10
Unify dependency requirement syntax in docs


rzetelskik:
The [support matrix in Releases document](https://operator.docs.scylladb.com/stable/support/releases.html#support-matrix) uses multiple ways of defining versions requirements, some of which not following any widely recognised syntax. We should unify these and use a known syntax.

### Acceptence criteria:
- [x] Unify version requirements syntax in compatibility matrix.
- [x] Clarify the "CRD" version requirement for ScyllaDB Monitoring.

/kind documentation
/priority important-longterm
/triage accepted

mflendrich:
Likely can be done in one pass with https://github.com/scylladb/scylla-operator/issues/3296

mflendrich:
Now tracked as [OPERATOR-52](https://scylladb.atlassian.net/browse/OPERATOR-52) in the internal Jira.

[OPERATOR-52]: https://scylladb.atlassian.net/browse/OPERATOR-52?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3333, opened 2026-03-12
Clarify supported ScyllaDB Monitoring version and extend the support matrix with tested versions of third-party dependencies

rzetelskik:
**Description of your changes:** Currently, the support matrix uses "(CRD)" as a supported version of ScyllaDB Monitoring. The meaning behind this was supposed to be that the version is embedded in a given operator version. This PR clarifies it by explicitly stating the supported version and adding a footnote explaining that the version is embedded and what it means from the user perspective.

image

image

Additionally, this PR extends the support matrix with a table specifying the supported and tested versions of third-party dependencies.

image

All of the values are dynamically sourced from config files.

Resolves https://github.com/scylladb/scylla-operator/issues/3063

/kind documentation
/priority important-longterm

### DIFF
--- a/docs/source/_ext/version_context_substitutions.py
+++ b/docs/source/_ext/version_context_substitutions.py
@@ -112,6 +112,30 @@ def get_versions_from_config(repo_root: str) -> dict[str, str]:
     if scylla_db_manager_agent_version is not None:
         result['agentVersion'] = strip_digest(scylla_db_manager_agent_version)
 
+    # Try to extract cert-manager version
+    cert_manager_version = get_nested_value(config_data, 'thirdParty.certManager.version')
+    if cert_manager_version is not None:
+        result['certManagerVersion'] = cert_manager_version
+
+    # Try to extract Grafana version from the full image string (e.g. "docker.io/grafana/grafana:12.3.3" -> "12.3.3")
+    grafana_image = get_nested_value(config_data, 'operator.grafanaImage')
+    if grafana_image is not None:
+        # Strip digest if present, then extract the tag after the last ':'
+        grafana_image_no_digest = strip_digest(grafana_image)
+        colon_idx = grafana_image_no_digest.rfind(':')
+        if colon_idx != -1:
+            result['grafanaVersion'] = grafana_image_no_digest[colon_idx + 1:]
+
+    # Try to extract Prometheus version, stripping the leading 'v' prefix (e.g. "v3.9.1" -> "3.9.1")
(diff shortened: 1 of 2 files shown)


docs/source/_ext/version_context_substitutions.py
```diff
@@ -111,6 +112,30 @@ def strip_digest(version: str) -> str:
     if scylla_db_manager_agent_version is not None:
         result['agentVersion'] = strip_digest(scylla_db_manager_agent_version)
 
+    # Try to extract cert-manager version
```
mflendrich, line 115: 1. For maintainability, I suggest that we extract the "try to extract"s into functions.

2. Do we want to silently allow the situation when the value is not found? (this appears to be the case now). I feel as if it should be a loud failure.

Assuming 1 and 2:
This way the `result` assignment could take a more elegant form of (pardon my python - this is just so that you get the idea)

```python
result = {
    'certManagerVersion`: get_nested_value(...),
    ...
}
```

docs/source/support/releases.md
```diff
@@ -150,7 +150,29 @@ The support matrix table shows the version requirements for ScyllaDB Operator. M
...
 * - ScyllaDB Monitoring
-  - (CRD)
+  - {{scyllaDBMonitoringVersion}} [^scylladb-monitoring-version]
+:::
+
+[^scylladb-monitoring-version]: ScyllaDB Operator embeds the specified version of ScyllaDB Monitoring, which includes a set of Grafana dashboards and Prometheus rules.
+
+### Third-party dependencies
+
+The following table lists the versions of third-party dependencies that ScyllaDB Operator supports and is tested against.
```
mflendrich: I would rather not make a statement about "supported" versions of 3rd party deps, unless necessary. It will be painful to maintain, for no benefit that is clear to me.

```suggestion
The following table lists the versions of third-party dependencies that ScyllaDB Operator is tested against.
```

mflendrich:
Great to have this change btw!

docs/source/_ext/version_context_substitutions.py
rzetelskik: It's handled later when docs are being built - it errors on an undefined substitution. Otherwise you'd need a map of which vars are expected in each branch or backport every source file change 4 versions back. That's also the reason for gracefully handling missing files and keys.

docs/source/support/releases.md
rzetelskik: Agreed, I was actually considering putting it this way initially. Changed.

PR #3385, opened 2026-04-01
 fix(validation): reject ScyllaCluster with empty spec.version at admission time

czeslavo:

Makes `ScyllaCluster` `spec.version` a required field. Previously, an empty value was silently accepted but caused the migration controller to fail, leaving the cluster degraded. This is shifting the feedback from runtime to object creation phase.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-70.

/cc @mflendrich

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,8 @@
 ### Highlights
 
 ### Upgrade requirements
-
-- Rename `ScyllaCluster` backup and repair tasks to comply with RFC 1123 (e.g., if they contain underscores `_`).
-- Ensure all `ScyllaCluster`'s `spec.version` is not empty.
   
-For details, refer to the [1.20 to 1.21 upgrade guide](https://operator.docs.scylladb.com/stable/management/upgrading/upgrade/#to-1-21).
+Please refer to the [1.20 to 1.21 upgrade guide](https://operator.docs.scylladb.com/stable/management/upgrading/upgrade/#to-1-21).
 
 ### Deprecations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@
 ### Upgrade requirements
 
 - Rename `ScyllaCluster` backup and repair tasks to comply with RFC 1123 (e.g., if they contain underscores `_`).
-  For details, refer to the [1.20 to 1.21 upgrade guide](https://operator.docs.scylladb.com/stable/management/upgrading/upgrade/#to-1-21).
+- Ensure all `ScyllaCluster`'s `spec.version` is not empty.
+  
+For details, refer to the [1.20 to 1.21 upgrade guide](https://operator.docs.scylladb.com/stable/management/upgrading/upgrade/#to-1-21).
 
 ### Deprecations
 
@@ -38,6 +40,7 @@
   [#3363](https://github.com/scylladb/scylla-operator/pull/3363)
 - `must-gather` resource collection now tolerates partial API discovery failures (e.g., when aggregated API servers like `metrics.k8s.io` are transiently unavailable) instead of failing entirely.
   [#3396](https://github.com/scylladb/scylla-operator/pull/3396)
+- Fixed [#3302](https://github.com/scylladb/scylla-operator/issues/3302): `ScyllaCluster` `spec.version` is now a required field. Previously, an empty value was accepted but caused a silent failure in the migration controller.
 
 ### Dependencies
 
--- a/docs/source/management/upgrading/upgrade.md
+++ b/docs/source/management/upgrading/upgrade.md
@@ -62,6 +62,8 @@ Make sure to familiarize yourself with both the standard upgrade procedure and t
 
 ### 1.20 to 1.21
 
+#### Ensure ScyllaCluster repair and backup task names are RFC 1123 compliant
+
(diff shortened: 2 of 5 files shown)


docs/source/management/upgrading/upgrade.md
```diff
@@ -62,6 +62,8 @@ Make sure to familiarize yourself with both the standard upgrade procedure and t
 
 ### 1.20 to 1.21
 
+#### Ensure ScyllaCluster repair and backup task names are RFC 1123 compliant
```
mflendrich, line 65: nit: simpler language in the session description

```suggestion
#### Check if your repair and backup task names include newly forbidden characters
```

+ this description appears (if I'm reading it right) to go in depth about what's allowed, but glosses over the fundamental fact that we're effectively disallowing the underscore (`_`)

suggestion: make an example of what is no longer allowed under the new rules (like in the changelog)

docs/source/management/upgrading/upgrade.md
```diff
@@ -105,6 +107,25 @@ Starting with v1.20.1, ScyllaDB Operator emitted warnings for `ScyllaCluster` re
 In v1.21, these warnings have been replaced with hard validation errors, and the operator will refuse to start if any existing `ScyllaClusters` have non-conforming task names.
 :::
 
+#### Ensure ScyllaCluster spec.version is not empty
+
+`ScyllaCluster` `spec.version` is now a required field. Any create or update request with an empty value will be rejected by the admission webhook.
```
mflendrich, line 112: I think that the user can read "create or update request" as "USER's create or update request" while the truth is that these operations happen under the hood all the time.

I think what it should say is that "having it unset will stall operator completely"

suggestion: use stronger language about what the consequences of non-conformance are (a complete stall)

CHANGELOG.md
```diff
@@ -12,7 +12,9 @@
 ### Upgrade requirements
 
 - Rename `ScyllaCluster` backup and repair tasks to comply with RFC 1123 (e.g., if they contain underscores `_`).
-  For details, refer to the [1.20 to 1.21 upgrade guide](https://operator.docs.scylladb.com/stable/management/upgrading/upgrade/#to-1-21).
+- Ensure all `ScyllaCluster`'s `spec.version` is not empty.
```
mflendrich: The instructions under `Upgrade requirements` appear to be a more concise duplicate of the `upgrade.md` page. It means that:

- we're repeating ourselves
- we're contradicting ourselves (there are more steps in the markdown page than in the changelogw)

suggestion: find a way to fix these two problems. Some options that come to my mind:
- move these concise bullets to a new TLDR on the markdown page,
- drop them altogether and just refer to the markdown page.

docs/source/management/upgrading/upgrade.md
czeslavo: For the context: I added this header because before this PR, we only had one "action" to take in the 1.20 -> 1.21 upgrade path. Now we have two, thus the headers.

I think that using precise "RFC 1123" in the header is fine; the section below explains what it means in practice. This way it's also shorter.

docs/source/management/upgrading/upgrade.md
czeslavo: > I think that the user can read "create or update request" as "USER's create or update request" while the truth is that these operations happen under the hood all the time.

That was the intention, as that's what will happen - the admission webhook will reject creation and updates if the object doesn't have a version specified.

CHANGELOG.md
czeslavo: I felt there was some value in summarizing the steps that may be required in the changelog, but linking to docs for detailed instructions. If you think that's not helpful, I'm ok with "drop them altogether and just refer to the markdown page." option.

mflendrich: What worries me is that we effectively have two sources of truth (of instructions to execute) that don't say exactly the same thing. I will be happy with anything that addresses this problem.

docs/source/management/upgrading/upgrade.md
mflendrich: The admission webhook will reject creation and updates - true.

But it will also reject updates that are part of Operator's reconciliation loop - like status updates, right? Which makes it a complete stall of operator, even without any user interaction. Am I missing something?

docs/source/management/upgrading/upgrade.md
mflendrich: Ok, fair enough.

To me however, the "what it means in practice" part is completely silent about the most glaring violation the user can make - and the one which we saw in real life - which is putting an underscore `_` in the name. But that's expressed in my other comments, so resolving this one.

docs/source/management/upgrading/upgrade.md
czeslavo: > But it will also reject updates that are part of Operator's reconciliation loop - like status updates, right? Which makes it a complete stall of operator, even without any user interaction. Am I missing something?

No, it's not this way. We're adding webhook validation to `ScyllaCluster`. Before this PR, `ScyllaCluster` with an invalid version was accepted on the API level, and the migration controller couldn't create a `ScyllaDBDatacenter` for it. I think that's the behavior you're describing - if we leave the version empty, `ScyllaCluster` will remain as broken as it was before this change. It's pointed out in the `Why is this not a breaking change?` note below.

CHANGELOG.md
czeslavo: :+1: Let's drop the steps in the changelog then and link to the upgrade guide as the source of truth.

czeslavo:
Rebased onto master.

czeslavo:
@mflendrich, is there anything else you'd like me to address?

mflendrich (approved):
/lgtm

czeslavo:
For some reason Prow didn't `lgtm`-labeled the PR, added the label manually.

PR #3305, opened 2026-03-04
[WIP] Proposed approach to docs restructuring

mflendrich:
I recommend viewing using `make preview`.

For details / agent context: see
- `DOCS_HINTS.md` (mostly a result of prompts listed in `DOCS_WORKLOG.md` + some manual changes to my taste)
- `NEW_DOCS_STRUCTURE.md`.

PR #3397, opened 2026-04-10
docs: add docs-staging skeleton with new documentation structure

mflendrich:
First part of [OPERATOR-32](https://scylladb.atlassian.net/browse/OPERATOR-32).

This PR is based on the draft of new documentation that is present in #3305 (that is in turn heavily inspired by lessons learned by review of #3305, #3303 and @rzetelskik's draft that was a private GH gist).

With the (bold) assumption that #3305 implements a structure that is generally acceptable (but details may be off or hypothetically hallucinated within the individual documents - but the top-level TOC, structure and scope is uncontroversial), this PR:
- creates a `/docs-staging` directory which is a shallow copy of `/docs` (that is: symlinks instead of copies) - this is so that `make preview` works (see the `conf.py` caveat)
- takes the TOCs and index pages from #3305 and applies them verbatim
- takes the titles of individual markdown pages from #3305 and applies **their stubs** (only a title + a TODO note box) verbatim.

The plan is that the `/docs-staging` directory will exist for the duration of OPERATOR-32 being under implementation, and (many) subsequent iterations to take small chunks of #3305 (individual documents or small groups thereof) and post PRs for them after self-review.

Once all content is migrated from #3305 to `/docs-staging`, the plan is to replace `/docs/source` with `/docs-staging/source` and delete `/docs-staging`.

`conf.py` caveat:

The reason is that conf.py can't be a symlink because it needs two staging-specific modifications:
1. Warning suppression for the theme's grid directive: The root index.md (copied from the draft docs) uses {grid} with a :type: option and {topic-box} children. The sphinx_scylladb_theme registers its own grid directive that doesn't fully match sphinx_design's option schema, producing warnings on myst.directive_option and design.grid. These are cosmetic (the page renders correctly) but fail the build under -W (warnings-as-errors). Adding suppress_warnings resolves this without touching the source index.md or the Makefile.
2. Removal of the version-conditional sphinx_design loading: The upstream conf.py only loads sphinx_design for v1.18 multiversion builds. In staging we don't use multiversion, and the theme already provides the directives the landing page needs (hero-box, topic-box, grid). Loading sphinx_design unconditionally would conflict with the theme's registrations (duplicate grid directive warning), so the conditional block is simply dropped.
Both changes are minimal and clearly commented in the file. If the upstream conf.py evolves, the staging copy will need to be manually synced -- but since the changes are isolated to one small block at the end, this should be low-friction.

/kind documentation
/cc czeslavo

[OPERATOR-32]: https://scylladb.atlassian.net/browse/OPERATOR-32?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


czeslavo:
The high-level structure looks good to me in general. :+1:  Signalling a few points that we might want to address by adding more pages/nesting. Not all of them have to be addressed in this issue, but I suggest that we keep track of them somehow at least (e.g., by adding notes to the articles' stubs):

1. Setting up monitoring - in the old docs, we have a dedicated guide for OpenShift - I believe we should fit it somewhere (maybe as a subpage of "Set up monitoring"?
2. I couldn't find a page mirroring [Support overview](https://operator.docs.scylladb.com/stable/support/overview.html) from the old docs. I think we should have a page that will cover the "formal" explanation of what support channels we provide (addressing community/enterprise).
3. We're dropping the `/resources` section (and that's good), but I think we should make sure that the content that belonged there (at least its relevant parts) is migrated to the API reference (as inline Go comments that will be included in the autogenerated ref).
4. I'm on a fence on whether "Set up monitoring" should belong to "Deploy ScyllaDB" section or "Operate (Day 2)" as monitoring itself is a day 2 activity, but setting it up is day 1 for sure.
5. I feel we might be missing a clear path for the " day 0/"I just want to try this out" scenario. I imagine "Get started" could have a doc that would have a clear path for such a scenario, mostly linking to other ones as its dependencies: create a kind cluster, install operator with gitops, create an example scyllacluster, connect a client to it.
6. (nit) "(Day 2)" in "Operate" makes me look for a complementary "Day 1" (and maybe "Day 0") section, and it's not there. I suggest that we "annotate" other sections this way for symmetry, or drop this altogether.

mflendrich:
> The high-level structure looks good to me in general. 👍 Signalling a few points that we might want to address by adding more pages/nesting. Not all of them have to be addressed in this issue, but I suggest that we keep track of them somehow at least (e.g., by adding notes to the articles' stubs):
> 
>     1. Setting up monitoring - in the old docs, we have a dedicated guide for OpenShift - I believe we should fit it somewhere (maybe as a subpage of "Set up monitoring"?

It's exactly there - as a section within the "Set up monitoring" document. I suggest that we iron out the exact details (like whether this should be a separate page or a section on the page is good enough) as we review this specific file. But I believe the content is there (unless I'm missing your point).

https://github.com/mflendrich/scylla-operator/blob/docs-vibecode/docs/source/deploy-scylladb/set-up-monitoring.md#openshift-user-workload-monitoring

> 
>     2. I couldn't find a page mirroring [Support overview](https://operator.docs.scylladb.com/stable/support/overview.html) from the old docs. I think we should have a page that will cover the "formal" explanation of what support channels we provide (addressing community/enterprise).

Valid point. How would you like this covered? I suggest a "Get Support" top-level category (added in a later PR). It would be a verbatim copy of "Support overview" and "Releases" pages.

>     3. We're dropping the `/resources` section (and that's good), but I think we should make sure that the content that belonged there (at least its relevant parts) is migrated to the API reference (as inline Go comments that will be included in the autogenerated ref).

Is there any specific bit of information you see missing? I believe that the mapping in #3305 should be reasonably complete.

> 
>     4. I'm on a fence on whether "Set up monitoring" should belong to "Deploy ScyllaDB" section or "Operate (Day 2)" as monitoring itself is a day 2 activity, but setting it up is day 1 for sure.

Set up monitoring is a requirement for a complete production-ready setup, so it's clearly an installation step for a production grade setup - hence "Deploy ScyllaDB". What exact journeys would you like covered under day2?

>     5. I feel we might be missing a clear path for the " day 0/"I just want to try this out" scenario. I imagine "Get started" could have a doc that would have a clear path for such a scenario, mostly linking to other ones as its dependencies: create a kind cluster, install operator with gitops, create an example scyllacluster, connect a client to it.

I like this idea. Do you agree that we handle this in the PR introducing the "deploy scylladb" guide? Is there any action you'd like taken beforehand?

>     6. (nit) "(Day 2)" in "Operate" makes me look for a complementary "Day 1" (and maybe "Day 0") section, and it's not there. I suggest that we "annotate" other sections this way for symmetry, or drop this altogether.

Dropped the "day 2" language.

mflendrich:
@czeslavo PTAL. Let me know if you want any reminders written to the stub pages, or we can revisit them as we go chapter by chapter with final content (in the former case, I'd be grateful for well-defined suggestions of entries in specific files)

czeslavo (commented):
> It's exactly there - as a section within the "Set up monitoring" document. I suggest that we iron out the exact details (like whether this should be a separate page or a section on the page is good enough) as we review this specific file. But I believe the content is there (unless I'm missing your point).

Alright, I was reviewing the structure alone and did not look into the draft PR. It's fine, let's focus on that when we review this specific file, then.

> Valid point. How would you like this covered? I suggest a "Get Support" top-level category (added in a later PR). It would be a verbatim copy of "Support overview" and "Releases" pages.

I like the "Get Support" category as it nicely aligns with the verb-based pattern we established here. :+1: 

> Is there any specific bit of information you see missing? I believe that the mapping in https://github.com/scylladb/scylla-operator/pull/3305 should be reasonably complete.

Nothing specific - just a call from my side to make sure that we do not drop any "reference" information that might be embedded in the old "resources" docs. I believe we discussed that during the review call with Kacper, and we agreed that our API reference has virtually no high-level overview of resources that it documents (e.g., [NodeConfig](https://operator.docs.scylladb.com/stable/reference/api/groups/scylla.scylladb.com/nodeconfigs.html) - there's no comment on the CRD type level explaining its use). I'd like us to cover this, and I believe that most of the content that should be there is already somewhere between the lines in "resources" articles.

I suggest we have a separate PR that "enriches" k8s API top-level types with comments describing their meaning/use.

> Set up monitoring is a requirement for a complete production-ready setup, so it's clearly an installation step for a production grade setup - hence "Deploy ScyllaDB". What exact journeys would you like covered under day2?

Yeah, I think that makes sense. Installation is a day 1 task. Regarding day 2, I was thinking about a guide that would help with understanding the dashboards we ship, but that may be a stretch, given that the dashboards can change often, and we do not track that anyhow. Let's drop it for now.

> I like this idea. Do you agree that we handle this in the PR introducing the "deploy scylladb" guide? Is there any action you'd like taken beforehand?

Makes total sense to me. Let's tackle it in the "deploy scylladb" guide PR.

> PTAL. Let me know if you want any reminders written to the stub pages, or we can revisit them as we go chapter by chapter with final content (in the former case, I'd be grateful for well-defined suggestions of entries in specific files)

I think it's good that we have the points surfaced here in the form of a written text. Let's focus on the content now.

/lgtm

PR #3401, opened 2026-04-14
feat: add ECDSA key support as an alternative to RSA for certificate generation

czeslavo:

Adds ECDSA key support as an alternative to RSA for TLS certificate generation. The operator now accepts a `--crypto-key-type=ECDSA` flag, with algorithm-aware `KeyUsage` handling, automatic `SignatureAlgorithm` selection, and all key operations unified behind Go's `crypto.Signer` interface.

Note: RSA remains the default. A migration plan for switching the default can be discussed separately to avoid disruptive changes without notice between minor releases.

Closes https://scylladb.atlassian.net/browse/OPERATOR-51.


pkg/cmd/operator/operator.go
```diff
@@ -63,9 +65,25 @@ const (
 	rsaKeySizeMin = 2048
 	rsaKeySizeMax = 4096
 
+	// ECDSA key size of 384 is a security strength equivalent of RSA 7680 bits. We use 4096 as default for RSA, so 384 is a good default for ECDSA.
+	// See https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate-characteristics.html.
```
mflendrich:
```suggestion
	// Recommendations (like the below) claim that a 384-bit ECDSA key is comparable with 7680 bits for RSA.
	// See https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate-characteristics.html.
```

I suggest more future-proof language here - "security strength equivalent to" is a risky thing to say in cryptography IMO 🙂

pkg/cmd/operator/operator.go
```diff
@@ -170,12 +191,23 @@ func (o *OperatorOptions) Validate() error {
 		errs = append(errs, errors.New("operator image can't be empty"))
 	}
 
-	if o.CryptoKeySize < rsaKeySizeMin {
-		errs = append(errs, fmt.Errorf("crypto-key-size must not be less than %d", rsaKeySizeMin))
-	}
+	switch o.CryptoKeyType {
```
mflendrich: I don't think that this degree of crypto logic sprawl in `pkg/cmd/operator` is justified (key manipulation living in `pkg/cmd/operator` as opposed to a specialized package like `pkg/crypto` feels hard to maintain to me).

If refactoring has been tried but hasn't resulted in a more modular structure (that is: logic in `pkg/cmd/operator` limited to wiring flag values to modular dependencies), I'd be happy to sit down for a pair programming session to try figure it out.

suggestion: get logic out of `pkg/cmd/operator`

pkg/crypto/certcreators.go
```diff
@@ -54,16 +53,33 @@ func (c *X509CertCreator) MakeCertificate(ctx context.Context, keyGetter RSAKeyG
...
 		return nil, nil, err
 	}
 
 	return cert, privateKey, nil
 }
 
+// AdjustKeyUsageForKeyType removes KeyUsageKeyEncipherment for non-RSA key types.
+// ECDSA keys cannot perform key encipherment (RSA key transport), so this
+// usage bit must be stripped to produce spec-compliant certificates.
+func AdjustKeyUsageForKeyType(keyUsage x509.KeyUsage, keyType KeyType) x509.KeyUsage {
```
mflendrich, line 82: Is there a reason why we'd want different key usage based on cipher type?
We'd need the same usage regardless of key type, no? Why was key encipherment there for RSA in the first place? Do we need it at all? Can it be removed for all ciphers?

pkg/crypto/certcreators_test.go
```diff
@@ -74,29 +111,186 @@ func TestX509CertCreator_MakeCertificate(t *testing.T) {
...
+		},
+		{
+			name: "ECDSA client cert with P-521",
+			certCreator: (&ClientCertCreatorConfig{
+				Subject: pkix.Name{
+					CommonName: "test-ecdsa-client",
+				},
+				DNSNames: []string{"client.local"},
+			}).ToCreator(),
+			makeKeyGetter: func(t *testing.T) KeyGetter {
```
mflendrich: These `makeKeyGetter` lambdas look unmaintainable (with their sync use, calls to the test harness, and repetition) and I'm looking for a way to make it less so. I can see two directions to explore:

1. declare and initialize the key getters outside of the table-based test with a less convoluted cleanup flow
2. refactor setup/teardown/MakeCertificate into a synchronous operation?

mflendrich (commented):
First round -- with an impression that the solution that was in place for RSA no longer works with multiple supported ciphers.

Comments are largely exploratory - instead of a discussion in comments, I think it would be more productive to have a pairing session if there's no immediate low hanging fruit found.

pkg/cmd/operator/operator.go
czeslavo: Good point, applied your suggestion verbatim. :+1:

pkg/cmd/operator/operator.go
czeslavo: Agreed. The CLI flags validation/completion logic has been moved out of `pkg/cmd/operator/operator.go` and moved to a dedicated `pkg/cmd/operator/cryptokeyoptions.go` (unit tests added). Also, a factory function has been added to the `pkg/crypto/keygenerator.go` so that the dispatch happens there, not in the CLI package.

pkg/crypto/certcreators.go
czeslavo: I've been considering dropping it as well, but I dug into this and concluded it could break some clients that negotiate `TLS_RSA_*` cipher suites.

Scylla uses the priority string [`SECURE128`](https://github.com/scylladb/scylladb/blob/8e7ba7efe25cdb18923f0e350b022bef88c62194/db/config.cc#L1755), which keeps `TLS_RSA_*` cipher suites active. Those suites require the client to encrypt the pre-master secret with the server's RSA public key, so `keyEncipherment` is required to be set in the certificate in such a case.

When the server certificate is ECDSA, the TLS stack filters those suites out entirely before the handshake begins (ECDSA cannot be used to encrypt data, only to sign - for auth). The `keyEncipherment` bit would therefore be meaningless even if present, and including it produces a non-compliant certificate.

I've extended the comment to explain it better.

pkg/crypto/certcreators_test.go
czeslavo: Simplified it by using the new factory function and a common runner test helper. Only its parameters are now part of the test case spec.

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,127 @@
...
+			errs = append(errs, fmt.Errorf("crypto-key-size must not be more than %d for RSA keys", rsaKeySizeMax))
+		}
+
+	case crypto.KeyTypeECDSA:
+		if !slices.Contains(crypto.SupportedECDSACurveBitSizes(), o.KeySize) {
+			errs = append(errs, fmt.Errorf("crypto-key-size must be one of %s, got: %d", supportedECDSACurveBitSizesAsString(), o.KeySize))
+		}
+
+	default:
+		errs = append(errs, fmt.Errorf("unsupported crypto-key-type %q, supported values are: %s, %s", o.KeyType, crypto.KeyTypeRSA, crypto.KeyTypeECDSA))
```
mflendrich: Inconsistency with flag description. This line uses `crypto.KeyTypeXxx` constants while the flag desc inlines the type literals.

suggestion: unify

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if o.BufferSizeMin < 1 {
```
mflendrich: nit: `1` is a magic number here; this appears to violate the pattern established above that has min and max values as constants

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (o *CryptoKeyOptions) Complete(cmd *cobra.Command) error {
+	maxChanged := cmd.Flags().Lookup(cryptoKeyBufferSizeMaxFlagKey).Changed
+	if !maxChanged && o.BufferSizeMin > o.BufferSizeMax {
+		o.BufferSizeMax = o.BufferSizeMin
+	}
+
+	// When ECDSA is selected but key size wasn't explicitly provided, default to P-384.
+	keySizeSetByUser := cmd.Flags().Lookup(cryptoKeySizeFlagKey).Changed
```
mflendrich, line 105: This appears to be a workaround for the fact that you can't have a default flag value for RSA and ECDSA at the same time, if reusing the same flag.

I think it would be much cleaner to split the key size flag into two flags: one for ecdsa, one for rsa. Each one having a sensible default (4096, 384), and not requiring magic to tell whether the default has changed.

pkg/cmd/operator/cryptokeyoptions_test.go
```diff
@@ -0,0 +1,256 @@
...
+		expectedErr string
+	}{
+		{
+			name: "RSA with minimum valid size passes",
+			opts: CryptoKeyOptions{
+				KeyType:       crypto.KeyTypeRSA,
+				KeySize:       rsaKeySizeMin,
+				BufferSizeMin: 1,
+				BufferSizeMax: 1,
+				BufferDelay:   42 * time.Hour,
```
mflendrich: suggestion: `const defaultBufferDelay` for maintainability

pkg/crypto/certcreators_test.go
```diff
@@ -15,6 +15,25 @@ import (
 	"github.com/scylladb/scylla-operator/pkg/helpers"
 )
 
+// startKeyGenerator starts the generator's background goroutine and registers
+// cleanup that cancels the context, waits for the goroutine, and closes the generator.
+func startKeyGenerator(t *testing.T, kg RunnableKeyGenerator) KeyGenerator {
+	t.Helper()
+	ctx, cancel := context.WithCancel(context.Background())
```
mflendrich: why not propagate context from caller?

pkg/crypto/ecdsa_test.go
```diff
@@ -0,0 +1,203 @@
...
+		{
+			name:  "P-521",
+			curve: elliptic.P521(),
+		},
+	}
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			gen, err := NewECDSAKeyGenerator(1, 1, tc.curve, 42*time.Hour)
```
mflendrich, line 37: Why not using `startKeyGenerator` here to avoid the waitgroup boilerplate? (applies throughout the PR)

pkg/crypto/helpers.go
```diff
@@ -85,20 +107,31 @@ func DecodeCertificates(certBytes []byte) ([]*x509.Certificate, error) {
...
-		return nil, fmt.Errorf("can't parse private key from block type %q: %w", block.Type, err)
+	var key crypto.Signer
+	switch block.Type {
+	case "EC PRIVATE KEY":
+		var err error
+		key, err = x509.ParseECPrivateKey(block.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("can't parse EC private key from block type %q: %w", block.Type, err)
+		}
+	default:
```
mflendrich, line 130: this logic (that the default case means that the key is RSA) looks not defensive enough to me:

Suppose that the input is the following

BEGIN BLABLABLA KEY
bytes bytes bytes
END BLABLABLA KEY

where BLABLABLA != "RSA"

I think we'd want to reject such input.

suggestion: put RSA in a separate `case` with `RSA PRIVATE KEY`, and handle the `default` case as an error

pkg/kubecrypto/certs_test.go
```diff
@@ -927,3 +939,153 @@ func Test_makeCertificate_certStability(t *testing.T) {
...
+	}
+	caKey, err := caResult.GetKey()
+	if err != nil {
+		t.Fatalf("failed to get CA key: %v", err)
+	}
+
+	// Verify the ECDSA CA cert does not have KeyEncipherment.
+	if caCert.KeyUsage&x509.KeyUsageKeyEncipherment != 0 {
+		t.Errorf("ECDSA CA cert should not have KeyUsageKeyEncipherment, got KeyUsage=%v", caCert.KeyUsage)
+	}
```
mflendrich: Trying to understand why there is a need to check it in this test?

pkg/kubecrypto/certs_test.go
```diff
@@ -927,3 +939,153 @@ func Test_makeCertificate_certStability(t *testing.T) {
...
+		t.Errorf("ECDSA serving cert should not have KeyUsageKeyEncipherment, got KeyUsage=%v", servingCert1.KeyUsage)
+	}
+
+	// Verify the serving cert's AuthorityKeyId matches the CA's SubjectKeyId.
+	if !reflect.DeepEqual(servingCert1.AuthorityKeyId, caCert.SubjectKeyId) {
+		t.Errorf("serving cert AuthorityKeyId doesn't match CA SubjectKeyId: got %x, want %x",
+			servingCert1.AuthorityKeyId, caCert.SubjectKeyId)
+	}
+
+	// Simulate a reconciliation cycle - call makeCertificate again with the existing secret.
```
mflendrich: This is at odds with the promise in the test docstring.

_is not regenerated on subsequent reconciliation cycles_

is not the same thing as

_Simulate a reconciliation cycle - call makeCertificate again with the existing secret_

I believe what this test is testing is that `makeCertificate` is idempotent - having nothing to do with reconciliation.

pkg/kubecrypto/tlssecret.go
```diff
@@ -24,9 +26,32 @@ const (
...
+	case *rsa.PrivateKey:
+		return "RSA"
+	case *ecdsa.PrivateKey:
+		return "ECDSA"
+	default:
+		return "Unknown"
+	}
+}
+
+func keySizeString(key crypto.Signer) string {
```
mflendrich: `size` is an overloaded concept in this function. It uses the term `size` for bits and for bytes interchangeably.

suggestion: use something along the lines of "Hungarian notation" to denote whether functions return (variables hold) a bits size of a bytes size. Or maybe use different types?

And it feels odd to cast the resulting integer into a string before returning.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: > This appears to be a workaround for the fact that you can't have a default flag value for RSA and ECDSA at the same time, if reusing the same flag.

Kind of, yes. But I wouldn't call this "magic", it's just using a public interface of `pflag`.

The motivation for keeping it as a single flag is that it already exists, and we cannot rename it (e.g., to `crypto-rsa-key-size`) without breaking changes. If we'd like to split, we'd have to add two new flags (e.g., `crypto-rsa-key-size`, `crypto-ecdsa-key-size`), keep the old one for backwards compatibility, and plan its deprecation. This way, we make it more complex than it's worth in my opinion. From the UX perspective, I believe that having it as a single flag also makes sense. Happy to discuss alternatives.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: Done. Flag descriptions and validation now use `crypto.RSAKeyType` / `crypto.ECDSAKeyType` consistently via `fmt.Sprintf`.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: Extracted `cryptoKeyBufferSizeMinValue` constant, used in both validation and flag description.

pkg/cmd/operator/cryptokeyoptions_test.go
czeslavo: Done.

pkg/crypto/certcreators_test.go
czeslavo: Done with `t.Context()`.

pkg/crypto/ecdsa_test.go
czeslavo: Extracted a `StartKeyGenerator` test helper and used across the board.

pkg/crypto/helpers.go
czeslavo: Good point. Made `RSA PRIVATE KEY` an explicit case + covered this function with unit tests.

pkg/kubecrypto/certs_test.go
czeslavo: You're right to call it. Removed it from here, as this test should focus on idempotency.

It made me realize, though, that I haven't extended the general `Test_makeCertificate` with ECDSA cases where we should verify this aspect.

Extended this test by adding a `keyConfig` dimension to the matrix (there's one for RSA and one for ECDSA) to avoid unnecessary duplication of test cases. All pre-existing test cases will run for both key configs. I generated ECDSA test "existing" certificates similarly to RSA ones and persisted them in `pkg/crypto/testfiles`. PTAL https://github.com/scylladb/scylla-operator/pull/3401/changes/e0349a9a988a5b4fe975c5910426110a5131fa88

pkg/kubecrypto/certs_test.go
czeslavo: You're right. I would call it a domain leak as `makeCertificate` is indeed used in the reconciliation loop, thus this lingua. Changed wording to call out the intention directly.

pkg/kubecrypto/tlssecret.go
czeslavo: This function is a narrow helper used only to generate a string representation of the key bit size for annotation purposes. It doesn't mix bits and bytes; it returns a bit size always (RSA key `.Size()` returns bytes, thus 8x multiplier). 

I renamed it to `keyBitSizeStringForAnnotation` for clarity.

pkg/cmd/operator/cryptokeyoptions.go
mflendrich: > If we'd like to split, we'd have to add two new flags (e.g., crypto-rsa-key-size, crypto-ecdsa-key-size), keep the old one for backwards compatibility, and plan its deprecation

Yup - this was exactly the intention behind my comment (see below for rationale)

> This way, we make it more complex than it's worth in my opinion. From the UX perspective, I believe that having it as a single flag also makes sense.

I understand the tradeoff. I have two arguments that might have enough weight to tip the tradeoff balance, though:

1. This is security code. And in security code tradeoffs, "simple/straightforward/clear" beats "clever/abstract/hidden" more than elsewhere. This implementation adds a functional dependency of the result value on three very unrelated inputs: (1) what cipher has been selected, (2) what the default value per cipher is, (3) whether the user has specified an override - resolved into a single value. Which leads me to a conclusion that this mechanism belongs to the "clever/abstract/hidden" class.
2. I have limited trust for the `Changed` mechanism because there is more than way to set the flag value (by cmdline argument, by env var ([`ReadFlagsFromEnv`](https://github.com/scylladb/scylla-operator/blob/b1f1e80a3349fabfeb42e0fa1b22a5ec0cdf0e9f/pkg/cmdutil/helpers.go#L36)). After a quick search I haven't found any unit test that would verify that. And it's not part of the obvious contract of the parts. Even if it works today (I haven't checked), what guarantee will we have that a future refactoring does not break it?

On the other hand, the rename can be carried out in a non-breaking fashion (that is - make the `crypto-key-size` flag an alias of the new `crypto-rsa-key-size` flag; possibly with a deprecation warning and a validation error if set when using ECDSA; and removal in a future version can be purely optional)

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,129 @@
...
+// CryptoKeyOptions holds CLI options for cryptographic key generation.
+type CryptoKeyOptions struct {
+	KeyType       string
+	KeySize       int
+	BufferSizeMin int
+	BufferSizeMax int
+	BufferDelay   time.Duration
+}
+
+func NewCryptoKeyOptions() CryptoKeyOptions {
```
mflendrich: nit: `DefaultCryptoKeyOptions`?

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,129 @@
...
+		errs = append(errs, fmt.Errorf("unsupported crypto-key-type %q, supported values are: %s, %s", o.KeyType, crypto.RSAKeyType, crypto.ECDSAKeyType))
+	}
+
+	if o.BufferSizeMin < cryptoKeyBufferSizeMinValue {
+		errs = append(errs, fmt.Errorf("crypto-key-buffer-size-min (%d) has to be at least %d", o.BufferSizeMin, cryptoKeyBufferSizeMinValue))
+	}
+
+	if o.BufferSizeMax < cryptoKeyBufferSizeMinValue {
+		errs = append(errs, fmt.Errorf("crypto-key-buffer-size-max (%d) has to be at least %d", o.BufferSizeMax, cryptoKeyBufferSizeMinValue))
+	}
```
mflendrich: this check looks redundant given the one immediately below

suggestion: remove

pkg/crypto/keygenerator.go
```diff
@@ -12,15 +15,74 @@ import (
...
+func ecdsaCurveForKeySize(keySize int) (elliptic.Curve, error) {
+	switch keySize {
+	case 256:
+		return elliptic.P256(), nil
+	case 384:
+		return elliptic.P384(), nil
+	case 521:
+		return elliptic.P521(), nil
+	default:
+		return nil, fmt.Errorf("unsupported ECDSA key size %d, supported values are: 256, 384, 521", keySize)
```
mflendrich: code smell: the [256, 384, 521] collection is scattered across the codebase; this line could make use of `func supportedECDSACurveBitSizesAsString`

suggestion: have a single authoritative definition of the `[256, 384, 521]` collection in one place, and reference it wherever needed; it looks like the only necessary exception from this rule would be where implementing the strategy pattern to discern behavior based on curve

pkg/cmd/operator/cryptokeyoptions_test.go
mflendrich: looks not done -- `42 * time.Hour` literal repetitions still present in the PR

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,129 @@
...
+		BufferDelay:   o.BufferDelay,
+	}
+}
+
+func supportedECDSACurveBitSizesAsString() string {
+	var sizes []string
+	for _, size := range crypto.SupportedECDSACurveBitSizes() {
+		sizes = append(sizes, fmt.Sprintf("%d", size))
+	}
+	return strings.Join(sizes, ", ")
```
mflendrich: nit - less code for a very similar result

```suggestion
	return fmt.Sprintf("%v", crypto.SupportedECDSACurveBitSizes())
```

mflendrich (commented):
Generally looking good - see comments.

One major feeling I have is that the addition of ECDSA has caused a cardinality explosion in almost every place that touches encyption or keys (now every function needs a fork on the road for both RSA and ECDSA). I'm wondering if a completely separate (but swappable) implementation of crypto having the same interface but ECDSA instead of RSA would have been a more maintainable alternative - but I assume that such an alternative was rejected in development for good reasons.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: Added both new `--crypto-rsa-key-size` and `--crypto-ecdsa-key-size` flags. Marked `--crypto-key-size` as deprecated, mapping only to RSA key size if used.

pkg/cmd/operator/cryptokeyoptions_test.go
czeslavo: Covered other files.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: Renamed.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: Removed.

pkg/crypto/keygenerator.go
czeslavo: Switched to using `SupportedECDSACurveBitSizes()` everywhere.

pkg/cmd/operator/cryptokeyoptions.go
czeslavo: Switched to using `fmt.Sprintf("%v", crypto.SupportedECDSACurveBitSizes())` inline, dropped this helper.

pkg/cmd/operator/cryptokeyoptions.go
```diff
@@ -0,0 +1,128 @@
...
+	cmd.Flags().IntVarP(&o.ECDSAKeySize, cryptoECDSAKeySizeFlagKey, "", o.ECDSAKeySize, fmt.Sprintf("The ECDSA curve bit size, used when --crypto-key-type=%s. Must be one of %v.", crypto.ECDSAKeyType, crypto.SupportedECDSACurveBitSizes()))
+	cmd.Flags().IntVarP(&o.BufferSizeMin, "crypto-key-buffer-size-min", "", o.BufferSizeMin, fmt.Sprintf("Minimal number of pre-generated crypto keys that are used for quick certificate issuance. The minimum size is %d.", cryptoKeyBufferSizeMinValue))
+	cmd.Flags().IntVarP(&o.BufferSizeMax, cryptoKeyBufferSizeMaxFlagKey, "", o.BufferSizeMax, fmt.Sprintf("Maximum number of pre-generated crypto keys that are used for quick certificate issuance. The minimum size is %d. If not set, it will adjust to be at least the size of crypto-key-buffer-size-min.", cryptoKeyBufferSizeMinValue))
+	cmd.Flags().DurationVarP(&o.BufferDelay, "crypto-key-buffer-delay", "", o.BufferDelay, "Delay is the time to wait when generating next certificate in the (min, max) range. Certificate generation bellow the min threshold is not affected.")
```
mflendrich:
```suggestion
	cmd.Flags().DurationVarP(&o.BufferDelay, "crypto-key-buffer-delay", "", o.BufferDelay, "Delay is the time to wait when generating next certificate in the (min, max) range. Certificate generation below the min threshold is not affected.")
```

mflendrich (approved):
/lgtm
/hold for typo

czeslavo:
@mflendrich Fixed the typo. Also, changed the number of nodes in the E2E test to 1 - we do not support node - node communication over TLS, so there's no need to have more than 1 node in the cluster.

mflendrich (approved):
Thanks!
/lgtm

mykaul:
New deployments should actually move to ECDSA P-384. See https://scylladb.atlassian.net/browse/CLOUD-4122 and others.

mflendrich:
> New deployments should actually move to ECDSA P-384. See https://scylladb.atlassian.net/browse/CLOUD-4122 and others.

@mykaul I believe this is the case already:

https://github.com/scylladb/scylla-operator/blob/5525e605bdacc6311b21b307c204fc8db16c5634/pkg/cmd/operator/cryptokeyoptions.go#L20-L22

https://github.com/scylladb/scylla-operator/blob/5525e605bdacc6311b21b307c204fc8db16c5634/pkg/crypto/keygenerator.go#L72-L73

I reckon there's no action necessary in that case - please confirm.

mykaul:
> Note: RSA remains the default. A migration plan for switching the default can be discussed separately to avoid disruptive changes without notice between minor releases.

This from the PR description. So it confused me.

mflendrich:
Ah, pardon - I misread the expectation. You are right.

If ECDSA is chosen, then 384 is the default curve. But RSA remains the default. We'll make a switch to ECDSA P-384 by default happen.

mflendrich:
Switch of the default tracked in the epic OPERATOR-450.
I made a crapshoot attempt with AI in #3684 but that's not good enough (details in the epic).

PR #3403, opened 2026-04-15
fix(test): wait for ghost node cleanup before quorum check in re-bootstrap test

czeslavo:

This PR makes the most noisy `ScyllaCluster should re-bootstrap from old PVCs` E2E test case stable. It adds a dedicated `WaitForScyllaClusterGhostNodeCleanup` step (with 10 min timeout)  that waits for the ghost nodes to disappear before we proceed to the actual `WaitForFullQuorum` check.

This is needed as after re-bootstrapping a ScyllaCluster on existing PVCs, ghost host IDs from the previous cluster persist in `system.topology`. ScyllaDB's topology coordinator cleans them up in the background, which takes 1-5 minutes (observed in CI). The existing `WaitForFullQuorum` (5-min timeout) could time out before cleanup finishes, causing flaky failures. To make the test stable, I didn't just extend the `WaitForFullQuorum` timeout, but added a deterministic check, focused on waiting for a precondition that's required for full quorum as we expect it.

Resolves major instability reason of the kind-fast suite (based on the past 2 month results analysis):

- Total failure rate: 43/214 = 20.1%
- Infra failure rate: 16/214 = 7.5%
- Test failure rate: 27/214 = 12.6%

| Rank | Test Name | Failures | % of test failures |
|------|-----------|----------|-------------------|
| 1 | ScyllaCluster should re-bootstrap from old PVCs | 22 | 81.5% (22/27) |
| 2 | ScyllaCluster should connect to cluster via Ingresses (already addressed in https://github.com/scylladb/scylla-operator/pull/3392) | 5 | 18.5% (5/27) |
| - | No JUnit artifact (failed before tests) | 3 | - |

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_rebootstrap.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_rebootstrap.go
@@ -110,6 +110,11 @@ var _ = g.Describe("ScyllaCluster", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		scyllaclusterverification.Verify(ctx, f.KubeClient(), f.ScyllaClient(), sc)
+
+		framework.By("Waiting for ghost nodes from the old cluster to be cleaned up")
+		err = utils.WaitForScyllaClusterGhostNodeCleanup(ctx, f.KubeClient().CoreV1(), sc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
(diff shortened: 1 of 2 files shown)


mflendrich (approved):
Good catch 👏 
/lgtm

czeslavo:
Same case as [here](https://github.com/scylladb/scylla-operator/pull/3385#issuecomment-4279059295). Adding `lgtm` label manually.

PR #3404, opened 2026-04-15
chore(jira): add a calling action to the GH-Jira sync automation

dani-tweig:
Add call_jira_sync.yml to scylladb/scylla-operator repository so that GH PRs will be synced with their linked Jira issues

Fixes PM-283


mflendrich (approved):
/lgtm

Thanks @dani-tweig.
Let's see how it works and adjust as we go.

PR #3408, opened 2026-04-17
fix: explicitly create a user for alternator API tests

abramche:
https://github.com/scylladb/scylladb/pull/27215 dropped the implicit creation of the `cassandra/cassandra` user. This change explicitly creates one for the purpose of the tests. It is backward compatible (ScyllaDB [supports](https://github.com/scylladb/scylladb/blob/f6532aca96cfc6140d522e46e54eddd8319430a4/db/config.cc#L1347) this config syntax since [ancient](https://github.com/scylladb/scylladb/blob/next-6.2/db/config.cc) open source versions). 

Essentially this fix enables further testing of Alternator API as is.

NB: This only addresses the newly failing tests in 2026.2 onwards. There is a related Operator production usage of these hardcoded credentials which will need to be addressed separately - see https://scylladb.atlassian.net/browse/OPERATOR-88

Resolves OPERATOR-86

**Testing**
The fix can be verified locally by running a 2026.2 build, i.e.
```
SCYLLADB_IMAGE_REF="scylladb/scylla-nightly:2026.2.0-dev-0.20260420.6011cb8a4c25" SO_FOCUS=".*should set up Alternator API when enabled.*" ./hack/kind/run-e2e-tests.sh
```

/kind failing-test
/priority important-soon

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_alternator.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_alternator.go
@@ -37,8 +37,10 @@ import (
 )
 
 const (
-	cqlDefaultUser     = "cassandra"
-	cqlDefaultPassword = "cassandra"
+	cqlAlternatorUser     = "nondefaultcassandra"
+	cqlAlternatorPassword = "nondefaultcassandra"
+	// the SHA-512 crypt hash of cqlAlternatorPassword, for ScyllaDB's auth_superuser_salted_password config
+	cqlAlternatorPasswordSaltedHash = "$6$VbAwLOEIu18EYeOF$zb0/MC1/KmI3OZg8im1XjC4RmpqkZ/Xh0LDYBun/mKRjUFIh16t3.K.OO7y17XMR4ZiYTsYX3FcPxeoL9oiiD."
 )
 
 type Movie struct {
@@ -72,10 +74,12 @@ var _ = g.Describe("ScyllaCluster", func() {
 		defer cancel()
 
 		cm, _, err := scyllafixture.ScyllaDBConfigTemplate.RenderObject(map[string]any{
-			"config": strings.TrimPrefix(`
+			"config": strings.TrimPrefix(fmt.Sprintf(`
 authenticator: PasswordAuthenticator
 authorizer: CassandraAuthorizer
-`, "\n"),
+auth_superuser_name: %s
+auth_superuser_salted_password: '%s'
+`, cqlAlternatorUser, cqlAlternatorPasswordSaltedHash), "\n"),
 		})
 		o.Expect(err).NotTo(o.HaveOccurred())
 
@@ -117,8 +121,8 @@ authorizer: CassandraAuthorizer
 
 		cqlConfig := gocql.NewCluster(svcIP)
 		cqlConfig.Authenticator = gocql.PasswordAuthenticator{
-			Username: cqlDefaultUser,
-			Password: cqlDefaultPassword,
+			Username: cqlAlternatorUser,
+			Password: cqlAlternatorPassword,
 		}
 		cqlConfig.Consistency = gocql.All
 		cqlSession, err := cqlConfig.CreateSession()
@@ -126,7 +130,7 @@ authorizer: CassandraAuthorizer
 		defer cqlSession.Close()
 
 		awsCredentials := aws.Credentials{
-			AccessKeyID:     cqlDefaultUser,
+			AccessKeyID:     cqlAlternatorUser,
 			SecretAccessKey: "",
 		}
 


test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -39,6 +39,8 @@ import (
 const (
 	cqlDefaultUser     = "cassandra"
 	cqlDefaultPassword = "cassandra"
+	// cqlDefaultPasswordSaltedHash is the SHA-512 crypt hash of cqlDefaultPassword.
+	cqlDefaultPasswordSaltedHash = "$6$cassandra$9q9MZlACAy/AUNIoeAEYN0D5UQQ8bU/Wnjl7sX7sm90qKJh3SUxzLAaQ73/5MTk8PWhWpGMNjYB8Yd/QLOIt1."
 )
```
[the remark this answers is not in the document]
abramche: renamed for clarity

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -72,10 +74,12 @@ var _ = g.Describe("ScyllaCluster", func() {
 		defer cancel()
 
 		cm, _, err := scyllafixture.ScyllaDBConfigTemplate.RenderObject(map[string]any{
-			"config": strings.TrimPrefix(`
+			"config": strings.TrimPrefix(fmt.Sprintf(`
 authenticator: PasswordAuthenticator
 authorizer: CassandraAuthorizer
-`, "\n"),
+auth_superuser_name: %s
+auth_superuser_salted_password: '%s'
```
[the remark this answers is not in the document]
abramche: not a concern - the username is not going to change. the password is only quoted because the hash contains special symbols.

test/e2e/set/scyllacluster/scyllacluster_alternator.go
```diff
@@ -37,8 +37,10 @@ import (
 )
 
 const (
-	cqlDefaultUser     = "cassandra"
-	cqlDefaultPassword = "cassandra"
+	cqlTestUser     = "cassandra"
+	cqlTestPassword = "cassandra"
```
mflendrich: If the test is creating the user explicitly, then I assume that giving it exactly the same values as the [previously existing by default](https://github.com/scylladb/scylladb/pull/27215) is a choice that will leave ambiguity as to whether "the new path" (user has been successfully created) or "the old path" (the user has been created by something out of our control) was taken.

suggestion: do not "shadow" the default user, use a non-default username and password

abramche: I do not mind that we change it this way for 2026.2.

Note that *if* the test was initially designed this way, we would not have discovered the original problem. Something to keep in mind for the new tests

abramche: addressed in 630c9f4c0bee7275a9aa8094ba5fd70b658c3d40

mflendrich: I think it boils down to the question - what semantics we expect from certain versions.

In my understanding, the test could implement these 2 possible assertions:

1. Alternator API tests pass calling scylla with `cassandra/cassandra` requiring no extra configuration
2. Alternator API tests pass calling scylla with a custom user (that is configured by the test as part of SUT initialization).

It is my understanding that the status quo before this PR was that assertion 1 was tested and assertion 2 was not, across all versions (and you learned that assertion 1 will no longer hold with 2026.2).

This PR in its current shape effectively implements `assert(1 OR 2)` if I understand it correctly

What we're doing here is trying to answer the two questions
- which assertions we want to verify for scylla <2026.2
- which assertions we want to verify for scylla >=2026.2

I see three ways forward that this PR could implement
The "cheap way": drop assertion 1 altogether, check assertion 2 across both <2026.2 and >=2026.2
The "middle way": assertion 1 for <2026.2, assertion 2 for >2026.2
The "complete way": assertion 1 for <2026.2; assertion 2 for both <2026.2 and >=2026.2

Do you have a preference/recommendation which way to take? Or do you propose a completely different framing of the problem?

abramche: I have a strong preference for the "cheap way" - at least for now, primarily because this is still testing exactly what the test is supposed to (alternator connectivity) - and no more than that - it does not test connectivity with *default* credentials, it is not in the name or in the test logic. So this choice is consistent with the test name and logic.

The test assumes that the supplied credentials are valid. The fixture change ensures they are - for all the versions.

Now, if the test was testing the connectivity with the default credentials (= `g.Describe("ScyllaCluster" ... g.It("should set up Alternator API with default credentials when enabled"`), we would have more options on the table. But, at any rate, introducing the version check logic would have unnecessarily overcomplicate the test in my view. In this hypothetical situation, I would have proposed to keep "assertion 1" in the branches where we do not support 2026.2, and replace with "assertion 2" for any later branches and master.

mflendrich (approved):
/lgtm

PR #3409, opened 2026-04-21
fix: narrow identity service selector down to scylla nodes only

### LINKED ISSUE #3407: Identity service selector also matched cleanup job pods
We were adding node to a scylladb cluster, once the new node is up, the operator is scheduling cleanup job pods. Around the same time we were triggering deployment on the db client, when the new client pod starts, gocql client reported this error  `gocql: unable to create session: unable to connect to the cluster, last error: unable to discover protocol version: dial tcp connect: connection refused`.

The db client is using identity service to connect the scylladb cluster. We noticed that the identity service selector is like this
 ```
  selector:
    app: scylla
    app.kubernetes.io/managed-by: scylla-operator
    scylla/cluster: <cluster-name>
```

Above selector end up matching both actual scylladb pods and the cleanup job pods. The ideal setup is to have the service selector only match with the actual scylladb pods

Operator version: 1.16.2

### PR DESCRIPTION
**Pull request by czeslavo**

Adds `scylla-operator.scylladb.com/pod-type: scylladb-node` to the identity service selector to prevent it from matching other cluster pods (e.g., cleanup job's ones) and causing connection issues.

Resolves https://github.com/scylladb/scylla-operator/issues/3407

### DIFF
--- a/pkg/controller/scylladbdatacenter/resource_test.go
+++ b/pkg/controller/scylladbdatacenter/resource_test.go
@@ -2751,6 +2751,93 @@ func TestMakeIngresses(t *testing.T) {
 	}
 }
 
+func TestIdentityService(t *testing.T) {
+	basicSDC := &scyllav1alpha1.ScyllaDBDatacenter{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "basic",
+			UID:  "the-uid",
+			Labels: map[string]string{
+				"default-sc-label": "foo",
+			},
+			Annotations: map[string]string{
+				"default-sc-annotation": "bar",
+			},
+		},
+		Spec: scyllav1alpha1.ScyllaDBDatacenterSpec{
+			ClusterName:    "basic",
+			DatacenterName: pointer.Ptr("dc"),
+			ScyllaDB: scyllav1alpha1.ScyllaDB{
+				Image: "scylladb/scylla:latest",
+			},
+			RackTemplate: &scyllav1alpha1.RackTemplate{},
+			Racks: []scyllav1alpha1.RackSpec{
+				{
+					Name: "rack",
+				},
+			},
+		},
+	}
+
+	got, err := IdentityService(basicSDC)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	expected := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "basic-client",
+			Labels: map[string]string{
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: pkg/naming/labels.go [hunk 1]
```diff
@@ -34,6 +34,15 @@ func ScyllaDBNodesPodsLabelsForScyllaCluster(sc *scyllav1.ScyllaCluster) map[str
 	return labels
 }
 
+// IdentityServiceSelectorLabels returns label keys and values used as the pod selector
+// for the identity service of the given ScyllaDBDatacenter.
+// It restricts the selector to only match ScyllaDB node pods, excluding other (e.g., cleanup job pods).
+func IdentityServiceSelectorLabels(sdc *scyllav1alpha1.ScyllaDBDatacenter) map[string]string {
```
rzetelskik: nit: perhaps we could reuse `ScyllaDBNodePodLabels` and optionally clean up the existing functions with similar purpose (e.g. `ScyllaDBNodesPodsLabelsForScyllaCluster`) by making the naming more uniform?

pkg/controller/scylladbdatacenter/resource_test.go
```diff
@@ -2751,6 +2751,93 @@ func TestMakeIngresses(t *testing.T) {
...
+				{Name: "agent-api", Port: 10001},
+				{Name: "prometheus", Port: 9180},
+				{Name: "agent-prometheus", Port: 5090},
+				{Name: "node-exporter", Port: 9100},
+				{Name: "thrift", Port: 9160},
+			},
+		},
+	}
+	if !reflect.DeepEqual(got, expected) {
+		t.Errorf("identity service selector differs from expected:\n%s", cmp.Diff(expected, got))
```
rzetelskik: it seems that the equality test checks more than just the selector so the message is confusing

pkg/naming/labels.go
czeslavo: Reused `ScyllaDBNodePodLabels` and renamed `ScyllaDBNodesPodsLabelsForScyllaCluster` to `ScyllaDBNodePodsSelectorLabelsForScyllaCluster` to match the pattern (`ScyllaCluster` variants get the explicit `ForScyllaCluster` suffix).

pkg/controller/scylladbdatacenter/resource_test.go
czeslavo: Fixed.

rzetelskik (approved):
/lgtm

PR #3410, opened 2026-04-22
feat(api): deprecate serverless CQL-SNI APIs and default ScyllaDBMonitoring to Platform

rzetelskik:
Deprecates `ScyllaCluster`/`ScyllaDBDatacenter` `spec.exposeOptions.cql` and `ScyllaDBMonitoring.spec.type = SaaS`, and flips the `ScyllaDBMonitoring.spec.type` default from `SaaS` to `Platform`. Existing `ScyllaDBMonitoring` objects that omit `spec.type` will render `Platform` dashboards after the upgrade; `exposeOptions.cql` continues to work with an admission warning.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-92

/kind api-change
/priority important-longterm

### DIFF
--- a/docs/source/management/upgrading/upgrade.md
+++ b/docs/source/management/upgrading/upgrade.md
@@ -126,6 +126,35 @@ If the command returns no output, all your `ScyllaClusters` are unaffected. If a
 A `ScyllaCluster` with an empty `spec.version` was never functional. The migration controller would fail to reconcile it, leaving the cluster in a permanently degraded state. Making the field required only prevents new broken clusters from being created.
 :::
 
+#### Review ScyllaDBMonitoring spec.type default change
+
+The default value of `ScyllaDBMonitoring` `spec.type` has changed from `SaaS` to `Platform`. Any existing `ScyllaDBMonitoring`
+object that omits `spec.type` will render `Platform` dashboards after the upgrade instead of `SaaS`. The `SaaS` value is
+also now deprecated and will be removed in a future release; the admission webhook will emit a warning when it is set explicitly.
+
+You can run the following snippet to list `ScyllaDBMonitoring` objects that will be affected by the default change or that
+still use the deprecated `SaaS` value:
+
+```bash
+output=$(kubectl get scylladbmonitorings --all-namespaces -o json | jq -r '
+  .items[] |
+  select((.spec.type // "SaaS") == "SaaS") |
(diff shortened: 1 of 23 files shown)


docs/source/management/upgrading/upgrade.md
```diff
@@ -126,6 +126,35 @@ If the command returns no output, all your `ScyllaClusters` are unaffected. If a
...
+If the command returns no output, none of your `ScyllaDBMonitoring` objects are affected. Otherwise, for each listed object,
+decide whether you want to:
+- keep the previous behavior by setting `spec.type: SaaS` explicitly before upgrading (the admission webhook will emit a
+  deprecation warning, and you will need to migrate to `Platform` before a future release removes `SaaS`), or
+- adopt the new default by either setting `spec.type: Platform` explicitly or leaving `spec.type` unset.
+
+:::{note}
+**Why is this not a breaking change?**
+The `Platform` dashboards are a superset of the `SaaS` dashboards, so switching from `SaaS` to `Platform` does not remove
+functionality. Users who still need the `SaaS` behavior can opt in explicitly until the value is removed in a future release.
```
czeslavo: > Users who still need the `SaaS` behavior can opt in explicitly until the value is removed in a future release.

I'd drop this sentence altogether - it's better not to suggest sticking to the `SaaS` type, given we're going to remove it soon.

czeslavo (commented):
Should we also document how to migrate from having `spec.exposeOptions.cql` set, or leave it until the actual removal of support for it? In the end, the sooner people migrate, the better. If we give them warnings, but no migration path, that's not the most comfortable situation.

Or do we ignore it, given that the field was always an experimental, undocumented feature?

rzetelskik:
> Should we also document how to migrate from having spec.exposeOptions.cql set, or leave it until the actual removal of support for it? In the end, the sooner people migrate, the better. If we give them warnings, but no migration path, that's not the most comfortable situation.
> 
> Or do we ignore it, given that the field was always an experimental, undocumented feature?
> 

Given the functionality is being dropped from the drivers, I think we can safely assume there are no real life users of CQL over SNI and we can drop it without providing any alternatives.

docs/source/management/upgrading/upgrade.md
rzetelskik: good point, done

czeslavo (approved):
/lgtm

PR #3414, opened 2026-04-29
docs(staging): add OKE reference deployment, gitops install, set up node pools, cpu pinning, nodes guides

czeslavo:

- Adds an end-to-end OKE cluster setup guide and reference deployment for ScyllaDB on Oracle Cloud, along with  NodeConfig manifest and executable setup script for OKE infrastructure provisioning
- Restructures install-operator/ and deploy-scylladb/ sections with reusable snippets, platform tabs, and proper cross-linking
- Ports setup dedicated node pools, configure cpu pinning, configure nodes guides

Structural changes (compared to the PoC/staging structure):
- prerequisites.md → provision-infrastructure/ directory with per-platform cluster setup stubs
- reference-deployment-*.md → reference-deployments/ subdirectory
- Software prereqs (cert-manager, Prometheus) moved from provision-infrastructure to install-operator index

The network diagram is an SVG file with an embedded XML that can be imported into draw.io directly and edited there. I tried Mermaid, but it doesn't give that much flexibility that's required for this kind of architecture diagrams.

The OKE cluster setup script includes markers for `literalinclude` directives, so we can maintain a single source of truth for the script content while still displaying fragments of it in the guide. 

Resolves https://scylladb.atlassian.net/browse/OPERATOR-82
Part of https://scylladb.atlassian.net/browse/OPERATOR-32.


czeslavo:
Converting to draft - we need to figure out business requirements regarding documenting the whole infra setup on different platforms (should we maintain it or not).

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
+# Reference deployment: OKE
+
+This guide walks you through a complete end-to-end ScyllaDB deployment on Oracle Container Engine for Kubernetes (OKE), from creating the network and the Kubernetes cluster all the way to running CQL queries.
+The deployment is suitable as a baseline for production workloads.
+
+## Prerequisites
+
+- An Oracle Cloud Infrastructure account with permissions to create networking, Container Engine, and Compute resources in a target compartment.
```
mykaul: need a link to which permissions are needed / some script or whatnot to verify them / set them up. Not high priority to fix right now.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+- The [`oci` CLI](https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/cliinstall.htm) installed and configured (`oci setup config`).
+- `kubectl` installed.
```
mykaul: link to https://kubernetes.io/docs/tasks/tools/  ?

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+- A Dense I/O instance shape available in your target region with capacity for 3 worker nodes.
```
mykaul: link to https://docs.scylladb.com/manual/stable/getting-started/cloud-instance-recommendations.html#oracle-cloud-infrastructure-oci

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+
+## Set environment variables
+
+The rest of the guide refers to the variables defined here. Adjust them to match your environment, then export them in your shell:
+
+:::{code-block} bash
```
mykaul: If someone is copy-pasting this (and they should), it needs to be one block at a time. This is also applicable later on.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+:::
+
+:::{note}
+In single-AD regions (such as `us-sanjose-1`), high availability for ScyllaDB is achieved by spreading nodes across the three fault domains within the AD (`FAULT-DOMAIN-1`, `FAULT-DOMAIN-2`, `FAULT-DOMAIN-3`).
+The ScyllaCluster manifest below uses one rack per fault domain.
+:::
+
+## Create the network
+
+OKE requires a VCN with appropriate subnets, gateways, route tables, and security rules.
```
mykaul: First time you use an acronym (vcn here), expand it.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+The topology created in this section is the minimum needed for an OKE cluster with VCN-Native Pod Networking and a public Kubernetes API endpoint.
```
mykaul: Is the default to have a public K8S API endpoint?

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+
+The resources created are:
+
+- a VCN (`10.0.0.0/16`),
+- an Internet Gateway and a NAT Gateway,
+- a public route table (default route → IGW) and a private route table (default route → NAT GW),
+- a public security list (ingress TCP `6443` and `443` from anywhere, plus all intra-VCN traffic) and a private security list (ingress all from inside the VCN),
+- three regional subnets — control plane, workers, and Kubernetes Service load balancers.
+
+```{figure} oke-network.drawio.svg
```
mykaul: scary stuff. Should be in a different page. Linked from here.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+)"
+:::
+
+:::{seealso}
+[Creating a VCN](https://docs.oracle.com/en-us/iaas/Content/Network/Tasks/create_vcn.htm) in the OCI documentation.
+:::
+
+### Create gateways
+
+An Internet Gateway for the public subnets and a NAT Gateway so private worker nodes can reach the public internet for image pulls and OS updates:
```
mykaul:
```suggestion
An Internet Gateway for the public subnets and a NAT Gateway so private worker nodes can reach the public Internet for image pulls and OS updates:
```

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
```diff
@@ -0,0 +1,821 @@
...
+
+:::{seealso}
+[Internet Gateway](https://docs.oracle.com/en-us/iaas/Content/Network/Tasks/managingIGs.htm) and [NAT Gateway](https://docs.oracle.com/en-us/iaas/Content/Network/Tasks/NATgateway.htm) in the OCI documentation.
+:::
+
+### Create route tables
+
+A public route table sending `0.0.0.0/0` to the Internet Gateway, and a private route table sending `0.0.0.0/0` to the NAT Gateway:
+
+:::{code-block} bash
```
mykaul: same comment as above elsewhere - two commands - two separate code blocks

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Linked to permissions guide in upstream.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Added.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Linked.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: All steps/commands are copy-pastable and work this way - verified manually.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Addressed in the `Set up an OKE cluster for ScyllaDB` guide intro.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Yes, and it's the simplest approach; otherwise, we'd have to deploy a bastion host and SSH into it - I'm not sure we should go that far in our guide.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Networking/cluster setup has been extracted to a separate page.

docs-staging/source/deploy-scylladb/reference-deployment-oke.md
czeslavo: Addressed.

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -0,0 +1,279 @@
...
+At the end, you will have:
+
+- A Virtual Cloud Network (VCN) with public and private subnets.
+- An OKE Enhanced Cluster with VCN-Native Pod Networking.
+- A general-purpose node pool for system workloads.
+- A dedicated Dense I/O node pool for ScyllaDB with the static CPU manager policy enabled.
+
+:::{tip}
+All the steps in this guide are also available as a single executable script:
+[`examples/oke/setup-oke-cluster.sh`](https://github.com/scylladb/scylla-operator/blob/master/examples/oke/setup-oke-cluster.sh).
```
rzetelskik: link should be versioned

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -0,0 +1,279 @@
...
+Set `OCI_REGION` and `OCI_COMPARTMENT_OCID`, then run the script to provision everything in one go.
+:::
+
+## Prerequisites
+
+- An Oracle Cloud Infrastructure account with [permissions to create networking, Container Engine, and Compute resources](https://docs.oracle.com/en-us/iaas/Content/ContEng/Concepts/contengpolicyconfig.htm) in a target compartment.
+- The [`oci` CLI](https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/cliinstall.htm) installed and configured (`oci setup config`).
+- [`kubectl`](https://kubernetes.io/docs/tasks/tools/) installed.
```
rzetelskik: nit: add anchor (https://kubernetes.io/docs/tasks/tools/#kubectl)

examples/oke/setup-oke-cluster.sh
```diff
@@ -0,0 +1,296 @@
...
+#   ./setup-oke-cluster.sh
+
+set -euo pipefail
+
+# [START env_vars]
+# Target region and compartment.
+# Replace the compartment OCID below with one of your own; you can list
+# compartments with: oci iam compartment list --all
+: "${OCI_REGION:?'Set OCI_REGION (e.g. us-sanjose-1)'}"
+: "${OCI_COMPARTMENT_OCID:?'Set OCI_COMPARTMENT_OCID'}"
```
rzetelskik, line 26: this works in a script but not as a docs snippet, assuming the intent is to copy and paste?

Image

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -0,0 +1,279 @@
...
+
+The resources created are:
+
+- a VCN (`10.0.0.0/16`),
+- an Internet Gateway and a NAT Gateway,
+- a public route table (default route → IGW) and a private route table (default route → NAT GW),
+- a public security list (ingress TCP `6443` and `443` from anywhere, plus all intra-VCN traffic) and a private security list (ingress all from inside the VCN),
+- three regional subnets — control plane, workers, and Kubernetes Service load balancers.
+
+```{figure} ../../deploy-scylladb/oke-network.drawio.svg
```
rzetelskik: The diagram shows a CSI driver deployed in the system pool. If the CSI driver is supposed to represent our own, then it's incorrect - both the controller and node RPC services are served by pods running on scylla-dedicated nodes, there is no separate controller service deployment that could run in the other pool

rzetelskik: nit: I'd try to be consistent wrt capitalization (Services, pods)

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -0,0 +1,279 @@
...
+
+```{literalinclude} ../../../../examples/oke/setup-oke-cluster.sh
+:language: bash
+:start-after: "# [START create_system_pool]"
+:end-before: "# [END create_system_pool]"
+```
+
+### Create the dedicated ScyllaDB node pool
+
+This pool runs ScyllaDB pods exclusively.
```
rzetelskik: I get the message but it's technically not true as it also runs node config and CSI driver pods at the very least. It's an early stage of the user journey so we shouldn't confuse the user here - we could say that its dedicated to ScyllaDB workloads.

rzetelskik: s/pods/Pods

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -0,0 +1,279 @@
...
+
+Expected output (node names and ages will differ):
+
+```
+NAME          STATUS   ROLES   AGE   VERSION   NODE-TYPE   FAULT-DOMAIN
+10.0.1.10    Ready    node    10m   v1.32.1                FAULT-DOMAIN-1
+10.0.1.20    Ready    node    5m    v1.32.1   scylla      FAULT-DOMAIN-1
+10.0.1.21    Ready    node    5m    v1.32.1   scylla      FAULT-DOMAIN-2
+10.0.1.22    Ready    node    5m    v1.32.1   scylla      FAULT-DOMAIN-3
+```
```
rzetelskik: nit: let's be consistent about code-block usage, you could also use 'console' decorator with `:::{code-block} console`

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -0,0 +1,279 @@
...
+:end-before: "# [END get_credentials]"
+```
+
+Verify connectivity and node readiness:
+
+:::{code-block} bash
+kubectl get nodes -L scylla.scylladb.com/node-type -L oci.oraclecloud.com/fault-domain
+:::
+
+Expected output (node names and ages will differ):
```
rzetelskik: nit: I'd just say "Example expected output" instead of listing the differences, the version and fault domains can differ too, no?

docs-staging/source/deploy-scylladb/before-you-deploy/configure-cpu-pinning.md
```diff
@@ -1,5 +1,151 @@
...
+If any of these is missing, CPU pinning silently does not apply.
+
+## Enable the static CPU manager policy
+
+The kubelet on each dedicated ScyllaDB node must be started with `cpuManagerPolicy: static`.
+This tells the kubelet to assign exclusive CPUs to containers that request integer CPU amounts in pods with Guaranteed QoS class.
+
+How you configure this depends on your platform:
+
+::::{tabs}
```
rzetelskik, line 34: we've lost OpenShift when compared to current docs: https://operator.docs.scylladb.com/stable/installation/kubernetes-prerequisites.html#static-cpu-policy

docs-staging/source/deploy-scylladb/before-you-deploy/configure-cpu-pinning.md
```diff
@@ -1,5 +1,151 @@
...
+```
+
+Check that the `ContainerPerftune` Job ran for each ScyllaDB pod:
+
+```shell
+kubectl -n scylla-operator-node-tuning get jobs \
+  -l scylla-operator.scylladb.com/node-config-job-type=ContainerPerftune
+```
+
+## Common pitfalls
```
rzetelskik: I don't feel strongly about this, but isn't this breaking the diataxis split a bit? Should we consider just referencing the troubleshooting section?

examples/oke/setup-oke-cluster.sh
```diff
@@ -0,0 +1,296 @@
...
+  --placement-configs "[{\"availabilityDomain\": \"${OCI_AD}\", \"subnetId\": \"${OKE_WORKERS_SUBNET_OCID}\"}]" \
+  --pod-subnet-ids "[\"${OKE_WORKERS_SUBNET_OCID}\"]" \
+  --size 1 \
+  --max-wait-seconds 1800 --wait-interval-seconds 30 \
+  --wait-for-state SUCCEEDED \
+  --wait-for-state FAILED
+# [END create_system_pool]
+
+# [START create_scylla_pool]
+CLOUD_INIT_BASE64="$(base64 -w0 <<'CLOUD_INIT'
```
rzetelskik: nit: personal preference only but I find these custom delimiters less readable than `EOF`

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
```diff
@@ -0,0 +1,227 @@
+# Reference deployment: OKE
```
rzetelskik, line 1: what's going to be the difference between these reference deployments and the "Deploy your first cluster" article? Is the latter supposed to be a generic deployment guide? It sounds like a development/local cluster creation guide, so it's hard to say

rzetelskik, line 1: I'm not sure this "Reference deployments" section fits its place that well, e.g. the "Before you deploy" subsection kind of assumes you already have the cluster with operator installed, and this section comes after it. Is the main role of these articles to be platform-specific deployment guides or serve as quickstarts?

docs-staging/source/deploy-scylladb/index.md
rzetelskik, line 13: is there no generic gitops guide for deploying a scylla cluster? Only the platform-specific references?

rzetelskik (commented):
first batch, more to come tomorrow

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
czeslavo: We discussed this with @mflendrich, and we both agreed that the "Deploy your first cluster" should be a generic, non-prod cluster deployment guide (as simple as possible). It should serve a use case of someone who would like to spin up a ScyllaDB cluster in any compatible Kubernetes cluster for PoC/development/demonstration purposes.

czeslavo: > Is the main role of these articles to be platform-specific deployment guides or serve as quickstarts?

Following business requirements, we want to make these guides production baselines on a specific platform. I propose we drop the operator and cert-manager installation steps from this article and make the operator installation a prerequisite, as it's generic anyway.

docs-staging/source/deploy-scylladb/index.md
czeslavo: `Deploy your first cluster` will serve this purpose.

examples/oke/setup-oke-cluster.sh
czeslavo: Changed it so we clearly ask for exporting `OCI_REGION` and `OCI_COMPARTMENT_OCID` (with the article's inline code). The rest remains a `literalinclude` from the script content.

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
czeslavo: Dropped the examples of workloads from the picture as it's not important in this context.

czeslavo: Done.

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
czeslavo: Reworded.

czeslavo: No longer relevant, rephrased to `This pool is dedicated to ScyllaDB workloads.`

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
czeslavo: Done.

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
czeslavo: If the user sticks to the default region, the fault domains should stay the same. But anyway, it makes sense to make this as you suggested - let's not get into details of the potential differences.

docs-staging/source/deploy-scylladb/before-you-deploy/configure-cpu-pinning.md
czeslavo: Brought back, thanks for catching this.

docs-staging/source/deploy-scylladb/before-you-deploy/configure-cpu-pinning.md
czeslavo: Extracted to the separate "Troubleshoot performance" guide and linked to it.

examples/oke/setup-oke-cluster.sh
czeslavo: Changed to EOF.

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
rzetelskik: something doesn't render correctly here, at least in the preview it redirects to a broken `https://github.com/%7B%7Brepository%7D%7D/blob/%7B%7Brevision%7D%7D/examples/oke/setup-oke-cluster.sh`

docs-staging/source/deploy-scylladb/before-you-deploy/configure-cpu-pinning.md
```diff
@@ -1,5 +1,126 @@
...
+
+:::{note}
+Use **integer CPU values** (e.g., `4`, not `4000m`) for the ScyllaDB container.
+The kubelet only assigns exclusive cores for integer CPU requests.
+Millicpu values result in a shared CPU pool.
+:::
+
+## Troubleshooting
+
+If CPU pinning is not working as expected, see [Troubleshoot performance](../../troubleshoot/troubleshoot-performance.md) for verification steps and common issues.
```
rzetelskik: nit: it's probably worth to go through the verification steps either way, so maybe we could skip the "If CPU pinning is not working as expected" part

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
czeslavo: Thanks, fixed. I remember checking this, perhaps changing the block from ``` to `:::` broke the rendering.

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
```diff
@@ -0,0 +1,214 @@
...
+:::
+
+To tear down the OKE infrastructure, follow the [Clean up](../../install-operator/provision-infrastructure/set-up-oke-cluster.md#clean-up) section of the OKE cluster setup guide.
+
+## Next steps
+
+- [Set up monitoring](../set-up-monitoring.md) with Prometheus and Grafana dashboards.
+- Review the [production checklist](../production-checklist.md) before going live.
+- Learn how to [connect your application](../../connect-your-app/index.md) to ScyllaDB.
+- Read the [Tuning](../../understand/tuning.md) and [Configure nodes](../before-you-deploy/configure-nodes.md) references to adjust the deployment for your workload.
```
rzetelskik: shouldn't this go before the deployment?

docs-staging/source/deploy-scylladb/index.md
```diff
@@ -6,20 +6,14 @@ Deploy and configure ScyllaDB clusters on Kubernetes. Start with a quick single-
 
 **Already have the Operator installed?** Go straight to [Deploy your first cluster](deploy-your-first-cluster.md).
 
-**Starting from scratch on a specific platform?** Follow a reference deployment:
-
-- [Reference deployment: GKE](reference-deployment-gke.md) — end-to-end on Google Kubernetes Engine.
-- [Reference deployment: EKS](reference-deployment-eks.md) — end-to-end on Amazon EKS.
-- [Reference deployment: OpenShift](reference-deployment-openshift.md) — end-to-end on Red Hat OpenShift.
+**Starting from scratch on a specific platform?** Follow a [reference deployment](reference-deployments/index.md).
```
rzetelskik: I think we need to align the messaging here with the intent. Assuming I'm following the user journey for one of the supported platforms, let's say OKE, it's not clear to me which one I should follow. I already have the platform set up and operator installed - so I should follow `Before you deploy` and `Deploy your first cluster`. But my understanding from https://github.com/scylladb/scylla-operator/pull/3414#discussion_r3224832979 is that I should be guided to a reference deployment, as  `Deploy your first cluster` will be a generic guide to setting up a dev cluster.

Also, the index is hard to bump into as, in the sidebar, it just shows up as the section heading. Perhaps we should have an "Overview" document here explaining this? It's more of a UI issue though so I don't feel particularly strong about this

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
rzetelskik: let's continue in https://github.com/scylladb/scylla-operator/pull/3414/changes/528173053bf0b377679b4a9f72f0a27efc12bf3c#r3233250798

rzetelskik: let's continue in https://github.com/scylladb/scylla-operator/pull/3414/changes/528173053bf0b377679b4a9f72f0a27efc12bf3c#r3233250798

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
```diff
@@ -0,0 +1,214 @@
...
+
+```
+NAME       READY   MEMBERS   RACKS   AVAILABLE   PROGRESSING   DEGRADED   AGE
+scylladb   3       3         3       True        False         False      5m
+```
+
+## Connect with cqlsh
+
+:::{include} /_snippets/connect-with-cqlsh.md
+:::
```
rzetelskik: I'm not sure I understand the purpose of this - wouldn't it be better to just link to the "Connect your app" section?

docs-staging/source/_snippets/connect-with-cqlsh.md
```diff
@@ -0,0 +1,40 @@
...
+:::{code-block} shell
+kubectl exec "pod/${SCYLLA_POD}" -c scylla -- nodetool status
+:::
+
+Run a few CQL commands:
+
+:::{code-block} shell
+SCYLLA_DC="$(kubectl get scyllaclusters.scylla.scylladb.com -o jsonpath='{.items[0].spec.datacenter.name}')"
+
+kubectl exec -i "pod/${SCYLLA_POD}" -c scylla -- cqlsh localhost <<EOF
```
rzetelskik: this won't work if we have a PasswordAuthenticator configured, and we should recommend it by default (as part of production readiness checklist, I suppose)

docs-staging/source/install-operator/install-with-gitops.md
```diff
@@ -1,5 +1,71 @@
...
+
+Wait for the driver to roll out:
+
+:::{code-block} shell
+kubectl -n=local-csi-driver rollout status --timeout=10m daemonset.apps/local-csi-driver
+:::
+
+:::{note}
+The Local CSI Driver requires that a [`NodeConfig`](../deploy-scylladb/before-you-deploy/configure-nodes.md) resource has already prepared the local disks on the target nodes.
+When following a platform-specific reference deployment, `NodeConfig` setup is covered there.
```
rzetelskik: isn't this out of order then? The user will do this before they go through the platform-specific deployment, no? Will the CSI driver reach readiness without it?

docs-staging/source/install-operator/install-with-gitops.md
```diff
@@ -1,5 +1,71 @@
...
+:::
+
+## Next steps
+
+- [Deploy your first ScyllaDB cluster](../deploy-scylladb/deploy-your-first-cluster.md).
+- Follow a [platform-specific reference deployment](../deploy-scylladb/reference-deployments/index.md).
```
rzetelskik: as per the comment on "Deploy ScyllaDB" index - it's not that obvious which one to choose. The user will also skip the "Before you deploy" section that way. Perhaps we should just link the the index/overview?

docs-staging/source/install-operator/install-with-gitops.md
```diff
@@ -1,5 +1,71 @@
...
-:::{todo}
-This page is not yet written.
+Install ScyllaDB Operator and its dependencies by applying raw manifests from the project repository.
+This method works with any GitOps tool (Argo CD, Flux, etc.) or plain `kubectl apply`.
+
+## Prerequisites
+
+- A Kubernetes cluster meeting the [infrastructure requirements](provision-infrastructure/index.md).
+- [`kubectl`](https://kubernetes.io/docs/tasks/tools/) configured to communicate with the cluster.
+
```
rzetelskik, line 19: we've lost some stuff compared to the existing document here:
- a caution block about replacing the operator image reference with a specific version tag/digest: https://github.com/scylladb/scylla-operator/blob/72b848db686b5e338fc472344b4a90a258986f47/docs/source/installation/gitops.md?plain=1#L7-L10
- a note about reserved namespaces: https://github.com/scylladb/scylla-operator/blob/72b848db686b5e338fc472344b4a90a258986f47/docs/source/installation/gitops.md?plain=1#L12-L13
- a caution block about keeping the SCYLLA_OPERATOR_IMAGE env var in sync with the image reference: https://github.com/scylladb/scylla-operator/blob/72b848db686b5e338fc472344b4a90a258986f47/docs/source/installation/gitops.md?plain=1#L61-L90

Prometheus Operator deployment is also missing - operator will throw reconciliation errors about missing CRDs (correct me if I'm wrong here).

docs-staging/source/deploy-scylladb/before-you-deploy/set-up-dedicated-node-pools.md
```diff
@@ -1,5 +1,123 @@
...
+
+Apply the ScyllaDB node label so that pod affinity rules can target these nodes:
+
+:::{code-block} bash
+kubectl label nodes <node-name> scylla.scylladb.com/node-type=scylla
+:::
+
+When using managed Kubernetes services, apply the label at node pool creation time to avoid manual labeling:
+
+::::{tabs}
```
rzetelskik, line 63: We're missing OpenShift here but I also know it'll be hard to cover as it's dependent on the platform and deployment model. Maybe we should just revert to a generic message saying you need the nodes labeled? Or just provide references to upstream?

docs-staging/source/deploy-scylladb/before-you-deploy/set-up-dedicated-node-pools.md
```diff
@@ -1,5 +1,123 @@
...
+Dedicated ScyllaDB nodes must have:
+
+- **Local NVMe storage** — ScyllaDB stores data on local SSDs for maximum I/O performance.
+  Use instance types with directly attached NVMe drives.
+- **Sufficient CPU and memory** — see the [ScyllaDB system requirements](https://docs.scylladb.com/manual/stable/getting-started/system-requirements.html) for minimum and recommended specifications.
+  Plan for at least 2 CPUs reserved for the OS, kubelet, and `DaemonSets`.
+
+### Recommended instance types
+
+::::{tabs}
```
rzetelskik, line 28: missing ROSA

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
```diff
@@ -1,5 +1,235 @@
...
+These must match the same nodes where ScyllaDB pods will run — if `NodeConfig` and `ScyllaCluster` target different node sets, performance tuning and disk setup will not apply to ScyllaDB pods.
+
+The recommended convention is label `scylla.scylladb.com/node-type: scylla` and taint `scylla-operator.scylladb.com/dedicated=scyllaclusters:NoSchedule`.
+Both `NodeConfig` placement and `ScyllaCluster` rack placement must reference the same label and tolerate the same taint.
+
+## Apply a NodeConfig
+
+`NodeConfig` manifests are platform-specific because disk device paths and naming conventions differ.
+
+::::{tabs}
```
rzetelskik, line 26: missing ROSA

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
```diff
@@ -1,5 +1,235 @@
...
+## Disk setup
+
+`NodeConfig` creates RAID arrays from local NVMe instance storage, formats them with XFS, and mounts them for the Local CSI Driver.
+
+The setup pipeline runs in order: loop devices (if configured) → RAID arrays → filesystems → mounts.
+After this, the Local CSI Driver can provision `PersistentVolumes` from directories on the mount point.
+
+### Platform differences
+
+::::{tabs}
```
rzetelskik, line 83: ditto, missing ROSA

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
```diff
@@ -1,5 +1,235 @@
...
+
+Wait for `NodeConfig` to finish reconciling:
+
+:::{code-block} bash
+kubectl wait --timeout=10m --for='condition=Progressing=False' nodeconfigs.scylla.scylladb.com/scylladb-nodepool-1
+kubectl wait --timeout=10m --for='condition=Degraded=False' nodeconfigs.scylla.scylladb.com/scylladb-nodepool-1
+kubectl wait --timeout=10m --for='condition=Available=True' nodeconfigs.scylla.scylladb.com/scylladb-nodepool-1
+:::
+
+## Disk setup
```
rzetelskik, line 74: nit: not sure if this is particularly important still but we've lost a note about xfsprogs requirement: https://operator.docs.scylladb.com/master/installation/kubernetes-prerequisites.html#xfsprogs

docs-staging/source/deploy-scylladb/before-you-deploy/configure-cpu-pinning.md
czeslavo: Addressed.

docs-staging/source/deploy-scylladb/index.md
czeslavo: Created a `deploy-scylladb/overview.md` that clearly separates two paths: 

1. Reference deployment for production on a supported platform (links to the platform-specific guides)
2. Generic/development cluster for quick experimentation (links to "Deploy your first cluster"). 

The overview also links to the supported platforms table in the releases reference. Index pages are now toctree-only, with the overview listed first.

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
czeslavo: As discussed offline - dropped the inline cqlsh snippet entirely. The "Next steps" section now links to "Connect your application" instead.

docs-staging/source/_snippets/connect-with-cqlsh.md
czeslavo: As discussed offline - auth configuration will be covered in the production checklist. The "Connect your app" section is the right place for connection instructions.

docs-staging/source/install-operator/install-with-gitops.md
czeslavo: Moved the Local CSI Driver installation to `configure-nodes.md` (the "Before you deploy" section), where it runs after `NodeConfig` is reconciled. The `install-with-gitops.md` guide no longer includes the CSI driver step.

docs-staging/source/install-operator/install-with-gitops.md
czeslavo: The "Next steps" in `install-with-gitops.md` now links to the deploy overview (`deploy-scylladb/overview.md`), which explains both paths clearly.

docs-staging/source/install-operator/install-with-gitops.md
czeslavo: Restored all three: (1) caution block about pinning operator image to a specific version tag/digest, (2) caution block about keeping `SCYLLA_OPERATOR_IMAGE` in sync, (3) note about scylla-operator and scylla-manager reserved namespaces. Also added Prometheus Operator CRD installation as a prerequisite step (optional, though, as the mentioned behavior was fixed recently).

docs-staging/source/deploy-scylladb/before-you-deploy/set-up-dedicated-node-pools.md
czeslavo: Added OpenShift/ROSA tabs to both tab sections in `set-up-dedicated-node-pools.md`. For labeling, the OpenShift tab provides a generic kubectl label command with a note about using MachineSets/MachinePools for managed clusters. For tainting, same approach.

docs-staging/source/deploy-scylladb/before-you-deploy/set-up-dedicated-node-pools.md
czeslavo: Added OpenShift/ROSA tabs.

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
czeslavo: Added OpenShift/ROSA tabs.

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
czeslavo: Added OpenShift/ROSA tabs.

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
czeslavo: Added it as a "Requirements" bullet point to "Set up dedicated node pools" guide.

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
rzetelskik: this one is still unaddressed

docs-staging/source/install-operator/install-with-gitops.md
rzetelskik: ad 3. there is no mention of scylla-manager namespace - will this be covered by manager section?
ad prometheus operator. if the behaviour is fixed, perhaps this should just be covered in `Set up monitoring`?

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
```diff
@@ -1,5 +1,276 @@
...
+# Verify mount
+kubectl -n scylla-operator-node-tuning exec -it <node-tuning-pod> -- df -h /mnt/hostfs/mnt/raid-disks/
+
+# Verify sysctls
+kubectl -n scylla-operator-node-tuning exec -it <node-tuning-pod> -- sysctl fs.nr_open
+:::
+
+## Install Local CSI Driver
+
+The Local CSI Driver is needed for provisioning `PersistentVolumes` for `ScyllaClusters` using the mounted NVMe storage:
```
rzetelskik: nit: not just NVMe, we also cover loop devices for dev environments in this doc

docs-staging/source/install-operator/install-with-gitops.md
```diff
@@ -1,5 +1,96 @@
...
+This method works with any GitOps tool (Argo CD, Flux, etc.) or plain `kubectl apply`.
+
+:::{caution}
+The manifests in this guide use a rolling tag for the ScyllaDB Operator image.
+For production deployments, replace the image reference in all manifests with a pinned version tag or SHA digest.
+:::
+
+:::{note}
+ScyllaDB Operator must run in the `scylla-operator` namespace.
+You can run your `ScyllaCluster` in any namespace of your choice.
```
rzetelskik: should we have a note about scylla namespaces in the deployment docs instead?

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
```diff
@@ -0,0 +1,196 @@
...
+## Deploy a ScyllaDB cluster
+
+Create a ScyllaDB cluster with one rack per fault domain.
+The resource requests are sized for `VM.DenseIO2.8` (8 OCPUs, 120 GB RAM), leaving headroom for the OS, kubelet, and `DaemonSets`.
+See the [ScyllaDB system requirements](https://docs.scylladb.com/manual/stable/getting-started/system-requirements.html) for general sizing guidance.
+Adjust if you use a different shape.
+
+:::{code-block} console
+:substitutions:
+kubectl apply --server-side -f=- <<EOF
```
rzetelskik, line 34: the example does not configure a dedicated namespace, I assume this expects users to use something like kubens, but we should at least have a note/info block informing about best practices (as per https://github.com/scylladb/scylla-operator/issues/2718)?

docs-staging/source/deploy-scylladb/overview.md
```diff
@@ -0,0 +1,27 @@
...
+This guide deploys a minimal ScyllaDB cluster and is not intended for production use.
+For production, complete the [Before you deploy](before-you-deploy/overview.md) steps first.
+
+## Further configuration
+
+After your cluster is running, see these guides for additional setup:
+
+- [Set up networking](set-up-networking/index.md) — expose ScyllaDB outside the Kubernetes cluster.
+- [Set up monitoring](set-up-monitoring.md) — integrate with Prometheus and Grafana.
+- [Deploy a multi-datacenter cluster](deploy-multi-dc-cluster.md) — span ScyllaDB across multiple Kubernetes clusters.
```
rzetelskik, line 26: not actionable: not sure how modular this can be - let's keep this in mind when adding the multi-dc docs

rzetelskik (commented):
Thanks, I think the user journey is much better now. Left a few minor comments.
One thing that still bothers me is that manager installation is completely gone from the install/deploy path. Maybe it can be covered separately from the operator installation (although the manager agent is going to throw errors, not sure about readiness - I assume you verified it), but I don't see a placeholder in the staging docs that could cover it. Is there a plan in place for covering it?

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
```diff
@@ -22,7 +22,7 @@ Follow [Configure nodes](../before-you-deploy/configure-nodes.md), selecting the
 ## Deploy a ScyllaDB cluster
 
 Create a ScyllaDB cluster with one rack per fault domain.
-The resource requests are sized for `VM.DenseIO2.8` (8 OCPUs, 120 GB RAM), leaving headroom for the OS, kubelet, and `DaemonSets`.
+The resource requests are sized for `VM.DenseIO.E4.Flex` with 8 OCPUs (128 GB RAM), leaving headroom for the OS, kubelet, and `DaemonSets`.
```
rzetelskik: as per internal thread - we need to revert to VM.DenseIO2.8 for now unfortunately

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
```diff
@@ -19,7 +19,7 @@ Set `OCI_REGION` and `OCI_COMPARTMENT_OCID`, then run the script to provision ev
 - An Oracle Cloud Infrastructure account with [permissions to create networking, Container Engine, and Compute resources](https://docs.oracle.com/en-us/iaas/Content/ContEng/Concepts/contengpolicyconfig.htm) in a target compartment.
 - The [`oci` CLI](https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/cliinstall.htm) installed and configured (`oci setup config`).
 - [`kubectl`](https://kubernetes.io/docs/tasks/tools/#kubectl) installed.
-- A Dense I/O instance shape (e.g., `VM.DenseIO2.8`) available in your target region with capacity for 3 worker nodes.
+- A Dense I/O Flex instance shape (e.g., `VM.DenseIO.E4.Flex`) available in your target region with capacity for 3 worker nodes.
```
rzetelskik: ditto

docs-staging/source/install-operator/install-with-gitops.md
czeslavo: I'd leave Prometheus Operator installation here, as optional, just because it has no other dependencies (unlike manager, which requires NodeConfig and local CSI driver installed).

docs-staging/source/deploy-scylladb/before-you-deploy/configure-nodes.md
czeslavo: Dropped NVMe part.

docs-staging/source/install-operator/install-with-gitops.md
czeslavo: Added a snippet and included it in all deployment guides.

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
czeslavo: Covered with a snippet.

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
czeslavo: Reverted.

docs-staging/source/install-operator/provision-infrastructure/set-up-oke-cluster.md
czeslavo: Reverted.

docs-staging/source/deploy-scylladb/reference-deployments/reference-deployment-oke.md
czeslavo: Dropped this point.

rzetelskik (approved):
Thanks for all the updates, it's a great improvement.
/lgtm

PR #3419, opened 2026-05-05
Potential fix for 1 code quality finding - typo

mykaul:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/scylladb/scylla-operator/security/quality/ai-findings)._

### DIFF
--- a/pkg/controller/scylladbdatacenter/resource.go
+++ b/pkg/controller/scylladbdatacenter/resource.go
@@ -2436,7 +2436,7 @@ func makeNodeStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec *scyl
 
 	if len(hostID) == 0 {
 		// Host ID hasn't been propagated yet, report an empty status without a hostID.
-		klog.V(4).InfoS("HostID of an expected node has nod been propagated yet, reporting an empty status", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KRef(sdc.Namespace, svcName))
+		klog.V(4).InfoS("HostID of an expected node has not been propagated yet, reporting an empty status", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KRef(sdc.Namespace, svcName))
 		return nodeStatusReport, true, nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/scylladbdatacenter/resource.go
mflendrich, line 1: Please fix the commit message & PR title as they violate CONTRIBUTING.md in their current form:

- this will fail on CI: commit message must be Conventional Commits, e.g. `fix(controller): typo in log print`. The rule is [here](https://github.com/scylladb/scylla-operator/blob/master/CONTRIBUTING.md#commits-and-prs), the validator is [here](https://github.com/scylladb/scylla-operator/blob/master/hack/.ci/verify-commits.sh).
- PR message must be Conventional Commits, e.g. `fix(controller): typo in log print`. The rule is [here](https://github.com/scylladb/scylla-operator/blob/master/CONTRIBUTING.md#commits-and-prs).
- must have a changelog entry: [rule](https://github.com/scylladb/scylla-operator/blob/master/CONTRIBUTING.md#changelog-entries)

It's probably a good idea to make a pointer to CONTRIBUTING.md inside an AGENTS.md for this repo.

mykaul: I'm fixing a single letter typo here... Do we need to spend so much time (yours and mine) on it? Does it really add any value?
(moreover, it was auto-created by CoPilot, I guess there's not a lot we can expect from it. In the future I'll probably won't use it).

mflendrich: Probably not. But from a management perspective, it means one of three things
1. The autogenerated PRs today are not good enough (and we should invest in autogeneration to make it better)
2. We should lower the bar for PRs
3. We don't need those autogenerated PRs

I am just stating a fact that CI will not allow this PR through (with some other bits not checked by CI, but still relevant).

I would like to hear your opinion which way (1-3) forward you see as the most productive, and we'll implement it.

mykaul: Given up on this.

PR #3421, opened 2026-05-05
docs(staging): reference: releases, feature gates and glossary

mflendrich:
Ports more reference articles from various sources into `docs-staging`:

- `reference/feature-gates.md`: from #3305 
- `reference/releases.md`: a verbatim copy of the existing docs `support/releases.md`
- `reference/glossary.md`: once I started reviewing I realized that it's mostly an amorphous slop and decided to drop it altogether.

/kind documentation
/priority important-longterm
/cc czeslavo


docs-staging/source/reference/feature-gates.md
```diff
@@ -1,5 +1,96 @@
...
+::::
+
+::::{group-tab} Helm
+Add the `--feature-gates` argument through the `additionalArgs` value:
+
+:::{code-block} yaml
+:emphasize-lines: 2
+additionalArgs:
+- --feature-gates=AutomaticTLSCertificates=true,BootstrapSynchronisation=true
+:::
```
czeslavo, line 48: Let's make it clear it's in `values.yaml` file. We could also add an alternative way to set it via `--set` CLI flag (not required though).

docs-staging/source/reference/feature-gates.md
```diff
@@ -1,5 +1,96 @@
...
+:::{list-table}
+:widths: 50 15 20
+:header-rows: 1
+
+* - Feature gate
+  - Default
+  - Since
+* - `AutomaticTLSCertificates`
+  - `true`
+  - alpha: v1.8, beta: v1.11
```
czeslavo: How about dropping the alpha/beta lingua? I remember us considering it for feature gates.

docs-staging/source/reference/releases.md
```diff
@@ -1,5 +1,219 @@
...
+  - v1
+* - ScyllaDB
+  - 2025.1, 2025.3 - 2026.1
+* - ScyllaDB Manager
+  - 3.7 - 3.9
+* - ScyllaDB Monitoring
+  - {{scyllaDBMonitoringVersion}} [^scylladb-monitoring-version]
+:::
+
+[^scylladb-monitoring-version]: ScyllaDB Operator embeds the specified version of ScyllaDB Monitoring, which includes a set of Grafana dashboards and Prometheus rules.
```
czeslavo, line 156: It renders a bit weird in my browser. If we cannot improve that, maybe let's switch to "manual" mode like suggested? 

Image

````suggestion
* - ScyllaDB Monitoring [1]
  - {{scyllaDBMonitoringVersion}}
:::

:::{note}
[1] ScyllaDB Operator embeds the specified version of ScyllaDB Monitoring, which includes a set of Grafana dashboards and Prometheus rules.
:::
:::
````

czeslavo (commented):
A few nits, nothing critical.

docs-staging/source/reference/feature-gates.md
rzetelskik: We not only considered it, it has been deliberately avoided in the existing docs: https://operator.docs.scylladb.com/master/reference/feature-gates.html

mflendrich: Of course - this is an oversight. Fixed. Also renamed the column to "Last changed".

docs-staging/source/reference/releases.md
mflendrich: This is already so on master: https://operator.docs.scylladb.com/master/support/releases.html#id1
I don't disagree with this suggestion, but I'd rather treat it as a master bug and backport here than start from here. With your permission let's handle it in a separate PR.

Created a tracker [OPERATOR-105](https://scylladb.atlassian.net/browse/OPERATOR-105).

docs-staging/source/reference/feature-gates.md
mflendrich: Done.

czeslavo (approved):
/lgtm

PR #3433, opened 2026-05-14
docs(deploy-scylladb): add Install ScyllaDB Manager page

czeslavo:
## Summary

Add a standalone "Install ScyllaDB Manager" page under the Deploy ScyllaDB section.

## Why a separate page?

The existing install guides (Install with Helm, Install with GitOps) cover Manager installation as one of their steps. However, Manager depends on node configuration (specifically the local CSI driver) because it deploys a small internal ScyllaCluster for its own state. Node configuration is covered in the Deploy ScyllaDB section as a deployment prerequisite, not in the install guides. A dedicated page under Deploy ScyllaDB makes this dependency explicit and places Manager installation in the correct order within the deployment flow — after nodes are configured and a cluster is deployed.

## Notes

- The licensing note was intentionally omitted pending clarification from the Manager team.

/cc @mflendrich

### DIFF
--- a/docs-staging/source/deploy-scylladb/index.md
+++ b/docs-staging/source/deploy-scylladb/index.md
@@ -7,6 +7,7 @@ before-you-deploy/index
 deploy-your-first-cluster
 reference-deployments/index
 deploy-multi-dc-cluster
+install-scylladb-manager
 set-up-networking/index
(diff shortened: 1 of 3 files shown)


mflendrich (approved):
/lgtm

PR #3436, opened 2026-05-14
fix(test): avoid nil pointer dereference in coredump test cleanup

rzetelskik:
The coredump E2E test panicked in OKE serial CI because the `DeferCleanup` closure
captured the `ds` variable by reference, which was later reassigned to `nil` when
`WaitForDaemonSetState` timed out. The cleanup then dereferenced `nil` when accessing
`ds.Name`.

The underlying timeout was caused by the coredump setup DaemonSet (from
`examples/gke/coredumps/`) running `apt-get` via `nsenter` on the host. OKE nodes run
Oracle Linux, which doesn't have `apt-get`, so the container crash-looped and the
DaemonSet never rolled out.

This PR fixes the panic by recording `ds.Name` and `ds.Namespace` in local variables
before registering `DeferCleanup`, so the closure no longer depends on the mutable `ds`
variable. The OKE platform incompatibility of the coredump setup DaemonSet is a separate
issue.

Linked issue: https://scylladb.atlassian.net/browse/OPERATOR-80

/kind bug
/priority important-soon


czeslavo (approved):
/lgtm

PR #3437, opened 2026-05-15
fix: address default SA existence race conditions in e2e.sh

abramche:
Kubernetes creates a default SA for new namespaces, but this process is async. e2e tests rely on the default SA in e2e and must-gather namespaces, and there are no guard rails.

This leads to situations where the e2e tests can [occasionally](https://jenkins.scylladb.com/job/scylla-master/job/docker/1309/) [fail](https://jenkins.scylladb.com/job/scylla-master/job/docker/1310/)

```
 Error from server (Forbidden): pods "e2e" is forbidden: error looking up service account e2e/default: serviceaccount "default" not found
 Error from server (Forbidden): pods "must-gather" is forbidden: error looking up service account gather-artifacts/default: serviceaccount "default" not found
```

Normally SA is created near-instantly, however on Jenkins runners it can occasionally jitter, introducing a race condition for e2e and must-gather pods. This PR ensures that the SA exists.

Added an acknowledgement to a potential further RBAC race (if RBAC is not yet propagated to API servers) - this did not float up so far; will be handled in a follow-up ticket if necessary.

Fixes https://scylladb.atlassian.net/browse/OPERATOR-115

**Testing**
See ticket

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -96,6 +96,9 @@ function gather-artifacts {
   kubectl create namespace gather-artifacts --dry-run=client -o=yaml | kubectl_apply -f=-
   kubectl create clusterrolebinding gather-artifacts --clusterrole=cluster-admin --serviceaccount=gather-artifacts:default --dry-run=client -o=yaml | kubectl_apply -f=-
   kubectl create -n=gather-artifacts pdb must-gather --selector='app=must-gather' --max-unavailable=0 --dry-run=client -o=yaml | kubectl_apply -f=-
+  # Wait until SA is created - eliminates an existing race condition; there is a potential further race if
+  # RBAC is not yet propagated to k8s API servers - in which case more hardening is due.
+  kubectl -n=gather-artifacts wait --for=jsonpath='{.metadata.name}'=default serviceaccount/default --timeout=60s
 
   kubectl_create -n=gather-artifacts -f=- <<EOF
 apiVersion: v1
@@ -401,6 +404,10 @@ function run-e2e {
   if [[ -n "${SCYLLADB_IMAGE_REF}" ]]; then
     e2e_command_args+=( "--scylladb-image-ref=${SCYLLADB_IMAGE_REF}" )
   fi
+
+  # Wait until SA is created - eliminates an existing race condition; there is a potential further race if
+  # RBAC is not yet propagated to k8s API servers - in which case more hardening is due.
+  kubectl -n=e2e wait --for=jsonpath='{.metadata.name}'=default serviceaccount/default --timeout=60s
 
   kubectl_create -n=e2e -f=- <<EOF
 apiVersion: v1


hack/.ci/lib/e2e.sh
```diff
@@ -96,6 +96,7 @@ function gather-artifacts {
   kubectl create namespace gather-artifacts --dry-run=client -o=yaml | kubectl_apply -f=-
   kubectl create clusterrolebinding gather-artifacts --clusterrole=cluster-admin --serviceaccount=gather-artifacts:default --dry-run=client -o=yaml | kubectl_apply -f=-
   kubectl create -n=gather-artifacts pdb must-gather --selector='app=must-gather' --max-unavailable=0 --dry-run=client -o=yaml | kubectl_apply -f=-
+  kubectl -n=gather-artifacts wait --for=jsonpath='{.metadata.name}'=default serviceaccount/default --timeout=60s
```
mflendrich, line 101: I am not sure that this is a correct check (i.e. that it indeed fixes the race condition) - reasoning below:

IIRC, propagation of RBAC to apiservers is eventually consistent. Which means that the instant that you create a SA isn't necessarily the moment that all apiservers in the cluster immediately start permitting requests configured in RBAC.
Therefore I believe that a more correct (still not completely airtight though) check would be `kubectl auth can-i ...`. But it's more awkward - not suppoting `wait` semantics, and it still doesn't involve a barrier for all apiservers to agree.

I don't want to block the change on it (does not hurt to add a wait for an intermediate step), but let's add a comment warning that this wait is for partial state, and edit the PR title. Example comment on my mind:

```
// This check waits until the default SA for the namespace is created, but it is not guaranteed that apiservers pick up the new RBAC rules by that time, so the race condition is still present.
```

and matching reword of PR title.

(Or - if I'm wrong and it indeed is guaranteed by semantics that SA name being present in apiserver means that up-to-date RBAC is served already - please prove me wrong and go forward with this change)

mflendrich (dismissed):

mflendrich:
/unapprove
(misclicked approve when commenting)

hack/.ci/lib/e2e.sh
abramche: It's a valid observation - and potentially addressing further, currently-unseen problems.

However at this stage, there is nothing to point fingers to RBAC about: the errors in the jenkins pipelines stem from absence of SA, not by missing (or not-yet-configured/propagated) RBAC. As such, I do not think that the comment is really on point here. We may add a custom loop with a `can-i` logic after the waiter, but imo that's overengineering just at this stage - it is uncertain the problems would float up at all. I [ran](https://jenkins.scylladb.com/job/releng-testing/job/docker/) the Jenkins pipeline with this change (plus some added throttling executions) and kind-cluster creation and configuration did not fail any more with the original issue - or any other RBAC error.

Wait addresses the current known problem (the one I raised the ticket about) - if we face further issues, I'm happy to take them on in a follow-up change (and raise a separate ticket for that of course 🙂)  

WDYT?

mflendrich: I am not disagreeing. I am however pointing out that a statement "this fixes the race condition" is unfounded unless proven. All I am asking for is that we relax the PR title and add this remark in a comment. I hope that's uncontroversial?

abramche: addressed in c3d0ecd3b6c183c22aa8977d622b80e281a24031

PR title and description adjusted to match

mflendrich (approved):
Thanks!
/lgtm

PR #3441, opened 2026-05-18
[WIP] chore: dummy commit to open PR

abramche:
Empty PR to run the OpenShift tests

/kind machinery
/priority important-soon

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_tls.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_tls.go
@@ -7,7 +7,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/x509"
 	"encoding/base64"
-	"encoding/json"
 	"fmt"
 	"time"
 
@@ -38,19 +37,20 @@ import (
 )
 
 var (
-	deploymentResourceInfo = collect.ResourceInfo{
+	subscriptionResourceInfo = collect.ResourceInfo{
 		Resource: schema.GroupVersionResource{
-			Group:    "apps",
-			Version:  "v1",
-			Resource: "deployments",
+			Group:    "operators.coreos.com",
+			Version:  "v1alpha1",
+			Resource: "subscriptions",
 		},
 		Scope: meta.RESTScopeNamespace,
 	}
 )
 
 const (
-	operatorNamespace      = "scylla-operator"
-	operatorDeploymentName = "scylla-operator"
+	operatorNamespace        = "scylla-operator"
+	operatorDeploymentName   = "scylla-operator"
+	operatorSubscriptionName = "scylladb-operator-subscription"
 )
 
 var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuiteParallelOpenShift, framework.SuiteKindFast, func() {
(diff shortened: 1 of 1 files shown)


czeslavo:
> @abramche: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/e2e-openshift-aws-serial 	[b70ac1d](https://github.com/scylladb/scylla-operator/commit/b70ac1da301ce018d8f4cacd997a034f39106302) 	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3441/pull-scylla-operator-master-e2e-openshift-aws-serial/2056338751247880192) 	false 	`/test e2e-openshift-aws-serial`
> 
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3441). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:abramche).
> Details

Expected failures:
- `[It] NodeConfig Optimizations should correctly project state for each scylla pod [Serial, SuiteSerial]` (due to perftune.py failing with missing systemctl in the Scylla image: https://scylladb.atlassian.net/browse/OPERATOR-113)
- `[It] ScyllaCluster coredumps should persist a coredump when ScyllaDB crashes with SIGABRT [Serial, SuiteSerial]` (the test needs to be adapted to handle hosts other than GKE's Ubuntu - OKE and EKS - needs a tracker to be created; @rzetelskik correct me if I'm wrong)

Unexpected failure: `[It] ScyllaCluster ECDSA should create TLS certificates using ECDSA keys when the operator is configured with --crypto-key-type=ECDSA [Serial, SuiteSerial]`.

abramche:
@czeslavo @rzetelskik there appears to be operator deployment spec drift, the [resulting](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3441/pull-scylla-operator-master-e2e-openshift-aws-serial/2056338751247880192/artifacts/must-gather/cluster/namespaces/scylla-operator/deployments.apps/scylla-operator.yaml) spec does not contain ECDSA flag at all, while the test appears to have tried patching it - and [this spec includes the flag](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3441/pull-scylla-operator-master-e2e-openshift-aws-serial/2056338751247880192/artifacts/must-gather/cluster/namespaces/scylla-operator/replicasets.apps/scylla-operator-75f76d4c98.yaml)

It seems to have caused a situation that the test progressed past the spec patch, but ended up with a replica set deployment with different configuration. The correct deployment with ECDSA flags had 0 pods as per must-gather.

It looks like there is an outside factor in causing the scylladb operator spec to drift, could it be something from the OpenShift itself?

czeslavo:
> @czeslavo @rzetelskik there appears to be operator deployment spec drift, the [resulting](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3441/pull-scylla-operator-master-e2e-openshift-aws-serial/2056338751247880192/artifacts/must-gather/cluster/namespaces/scylla-operator/deployments.apps/scylla-operator.yaml) spec does not contain ECDSA flag at all, while the test appears to have tried patching it - and [this spec includes the flag](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3441/pull-scylla-operator-master-e2e-openshift-aws-serial/2056338751247880192/artifacts/must-gather/cluster/namespaces/scylla-operator/replicasets.apps/scylla-operator-75f76d4c98.yaml)
> 
> It seems to have caused a situation that the test progressed past the spec patch, but ended up with a replica set deployment with different configuration. The correct deployment with ECDSA flags had 0 pods as per must-gather.
> 
> It looks like there is an outside factor in causing the scylladb operator spec to drift, could it be something from the OpenShift itself?

Ah, in OpenShift tests, we're using OLM to install the Operator, and this patch is ineffective. @rzetelskik do you know how we could patch the Operator Deployment to set a flag for this test and then revert it in cleanup? As I understand, we should patch `Subscription` by setting `spec.config.env[].SCYLLA_OPERATOR_CRYPTO_KEY_TYPE` to `ECDSA`.

czeslavo:
Anyway, I think that to unblock the release, we should create a tracker for it and tackle it offline, as it's not a bug, but a problem with a gap in OpenShift testing that wasn't taken into account. @mflendrich do you agree?

mflendrich:
> Anyway, I think that to unblock the release, we should create a tracker for it and tackle it offline, as it's not a bug, but a problem with a gap in OpenShift testing that wasn't taken into account. @mflendrich do you agree?

If the test result doesn't mean that we cannot enable ECDSA for OpenShift (and we are convinced that it works) - yes.

rzetelskik:
> [It] ScyllaCluster coredumps should persist a coredump when ScyllaDB crashes with SIGABRT [Serial, SuiteSerial] (the test needs to be adapted to handle hosts other than GKE's Ubuntu - OKE and EKS - needs a tracker to be created; @rzetelskik correct me if I'm wrong)

We have an existing tracker for OKE: https://scylladb.atlassian.net/browse/OPERATOR-116. It should either be extended for OpenShift (and conversely EKS), or separate trackers should be created. We should also extend our documentation as it only covers GKE.

> @rzetelskik do you know how we could patch the Operator Deployment to set a flag for this test and then revert it in cleanup? As I understand, we should patch Subscription by setting spec.config.env[].SCYLLA_OPERATOR_CRYPTO_KEY_TYPE to ECDSA.

Sounds correct.

mflendrich:
> > @rzetelskik do you know how we could patch the Operator Deployment to set a flag for this test and then revert it in cleanup? As I understand, we should patch Subscription by setting spec.config.env[].SCYLLA_OPERATOR_CRYPTO_KEY_TYPE to ECDSA.
> 
> Sounds correct.

In that case my hunch is: to unblock 1.21, verify manually that this works.
If yes -> create a tracker to write a test and cover openshift by testing as well
If not -> create a tracker to implement ECDSA for OpenShift

PR #3443, opened 2026-05-18
docs(staging): promote staging docs to the mainline

mflendrich:
Part of OPERATOR-32.

Removes the `docs-staging/` dir and effectively overwrites the old docs `docs/` with `docs-staging/`.

Also:
- moves overviews into index files (fixes OPERATOR-114)
- deletes the leftover stub todo files for docs that were not created before this promotion
- files some build errors caused by invalid references


docs/source/conf.py
```diff
@@ -173,11 +173,16 @@
     "https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds",
     # Ignore anchor that works as expected in the browser (probably a matter of the JS page rendering).
     "https://www.scylladb.com/product/support/#enterprise-support",
+    # docs.openshift.com redirects to docs.redhat.com which blocks automated link checkers with 403.
+    "https://docs.openshift.com/container-platform/.*",
+    "https://docs.openshift.com/rosa/.*",
 ]
```
czeslavo, line 179: Is that still the case with `linkcheck_request_headers` added? It shouldn't be 🤔

mflendrich: This was the error without it

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3443/pull-scylla-operator-master-verify-docs/2056396244346998784#1:build-log.txt%3A186

czeslavo: Ah. Alright, let's figure this out separately, then (perhaps remove the custom headers as they're no longer effective anyway).

rzetelskik: @mflendrich did you try adding the header for the other urls? It's kind of surprising given we didn't get the linkcheck errors in periodics

rzetelskik: https://github.com/scylladb/scylla-operator/pull/3444

PR #3444, opened 2026-05-18
docs: override User-Agent for docs.openshift.com linkcheck instead of ignoring

rzetelskik:
docs.openshift.com redirects to docs.redhat.com, which blocks automated link checkers with 403. bc4f503e2 worked around this by adding docs.openshift.com URLs to linkcheck_ignore, but 4954628e4 introduced a User-Agent override for docs.redhat.com that addresses the root cause.
Apply the same User-Agent override for docs.openshift.com and remove it from linkcheck_ignore.

Resolves #

/kind machinery
/priority important-soon
/cc


mflendrich (approved):
Ah. It didn't cross my mind that redhat and openshift were different domains. Good call. Thanks for this PR.

/lgtm

PR #3449, opened 2026-05-20
chore(changelog,docs): add 1.21.0, promote docs to stable

mflendrich:
Linked issue: OPERATOR-109

Adds a changelog for 1.21.0.
Also promotes docs to stable.
Setting a release date for 1.22 will come in a separate PR.

/kind cleanup
/priority important-soon

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -84,12 +84,12 @@ smv_tag_whitelist = multiversion_regex_builder(TAGS)
 # Whitelist pattern for branches (set to None to ignore all branches)
 BRANCHES = ["master", "v1.18", "v1.19", "v1.20", "v1.21"]
 # Set which versions are not released yet.
-UNSTABLE_VERSIONS = ["master", "v1.21"]
+UNSTABLE_VERSIONS = ["master"]
 DEPRECATED_VERSIONS = ["v1.18", "v1.19"]
 smv_branch_whitelist = multiversion_regex_builder(BRANCHES)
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -84,7 +84,7 @@
 # Whitelist pattern for branches (set to None to ignore all branches)
 BRANCHES = ["master", "v1.18", "v1.19", "v1.20", "v1.21"]
 # Set which versions are not released yet.
-UNSTABLE_VERSIONS = ["master", "v1.21"]
+UNSTABLE_VERSIONS = ["master"]
```
rzetelskik, line 87: out of order or missing all other steps from `Adjust built branches and configure versions` section - is this on purpose?

mflendrich: Yes - I intend to edit the releases page based on conversations tomorrow, so that we accurately represent our plan for 1.22 (as the release date will not follow the usual pattern).

Also some of the changes were made already by the PR promoting the new docs from staging. I think the current state is what the conf.py file shoud look like; the release procedure is a bit bitrotten with regard to it from what I can see.

I missed `smv_latest_version` though - fixed it in my most recent push. PTAL.

rzetelskik: Ok, as long as it goes together with smv_latest_version it's fine with me (although I'd like to hear why X.Y-3 is not being deprecated like before).

rzetelskik (approved):
/lgtm

PR #3468, opened 2026-06-05
chore(kind): enable nftables and tune etcd

abramche:
- Switch kind suite to nftables routing (much faster than iptables). The [Jenkins](https://jenkins.scylladb.com/job/releng-testing/job/operator/9/) [results](https://jenkins.scylladb.com/job/releng-testing/job/operator/10/) hint at much [improved](https://jenkins.scylladb.com/job/releng-testing/job/operator/13/) [stability](https://jenkins.scylladb.com/job/releng-testing/job/operator/14/).
- Tune etcd to use 500m CPU (and wait for the new instances to spin up before proceeding with the cluster deployment

With these changes we should see fewer flakes in Prow and Jenkins alike

Resolves https://scylladb.atlassian.net/browse/OPERATOR-137

### DIFF
--- a/hack/kind/cluster-config.yaml
+++ b/hack/kind/cluster-config.yaml
@@ -6,6 +6,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 networking:
   ipFamily: ipv4
+  kubeProxyMode: "nftables"
 nodes:
   - role: control-plane
     kubeadmConfigPatches:
--- a/hack/kind/cluster-setup.sh
+++ b/hack/kind/cluster-setup.sh
@@ -52,6 +52,17 @@ else
     echo "Reusing existing KinD cluster: ${CLUSTER_NAME}"
 fi
 
+# Bump etcd CPU request from the kubeadm default of 100m to avoid compaction stalls under high test parallelism.
+podman exec "${CLUSTER_NAME}-control-plane" sed -i 's/cpu: 100m/cpu: 500m/' /etc/kubernetes/manifests/etcd.yaml
+echo "Waiting for API server to become ready after etcd CPU bump..."
+wait_kubeconfig="$(mktemp)"
+kind get kubeconfig --name="${CLUSTER_NAME}" > "${wait_kubeconfig}"
+until KUBECONFIG="${wait_kubeconfig}" kubectl get --raw /readyz >/dev/null 2>&1; do
+    echo "API server not ready yet, retrying..."
+    sleep 2
+done
+rm "${wait_kubeconfig}"
+
 # Set up a local registry for the KinD cluster following https://kind.sigs.k8s.io/docs/user/local-registry/.
 reg_name='kind-registry'
 reg_port='5001'


mflendrich:
> The [Jenkins](https://jenkins.scylladb.com/job/releng-testing/job/operator/9/) [results](https://jenkins.scylladb.com/job/releng-testing/job/operator/10/) hint at much [improved](https://jenkins.scylladb.com/job/releng-testing/job/operator/13/) [stability](https://jenkins.scylladb.com/job/releng-testing/job/operator/14/).

All four runs are failed. What is the exact stability claim you're making here @abramche ?

abramche:
> > The [Jenkins](https://jenkins.scylladb.com/job/releng-testing/job/operator/9/) [results](https://jenkins.scylladb.com/job/releng-testing/job/operator/10/) hint at much [improved](https://jenkins.scylladb.com/job/releng-testing/job/operator/13/) [stability](https://jenkins.scylladb.com/job/releng-testing/job/operator/14/).
> 
> All four runs are failed. What is the exact stability claim you're making here @abramche ?

1. We expect suite failures"because we have unresolved situation with node-exporter. The tests that failed in these runs also fail on Jenkins.
2. Before this fix, there were more flaky failures (the ones that are not normally flaky in Prow) and the suite was frequently killed on timeout with less than 58 specs run. With the performance gain from nftables I was able to increase parallelism from 3 to 20 tests, cutting down execution time from 2h to 30min (comparable to Prow now)

I would suppose there is more tuning to be done, but this brings in much needed performance improvement. I am seeing significantly fewer etcd fails and no tests failing on network connectivity, compared to iptables-powered suite.

-
As an alternative, I can inline these fixes into the Jenkins pipeline, if you think they need more maturity in the operator repo

mflendrich (commented):
The etcd change looks incorrect.

etcd is a static pod, so setting a container **request** for it does not impact the amount of ~memory~ **cpu** available to the container (the cgroup ceiling is controlled by the limit).
Therefore - I do not think how setting the request on a static pod's container would achieve etcd's capability to "use 500m CPU" as the PR claims.

As for nftables - this change feels very arbitrary. I can be convinced that our network configuration and traffic patterns are such that there's a significant difference. But I'm not convinced yet. I'd expect higher quality evidence - e.g. "50 runs with iptables with N% flakiness" vs "50 runs with nftables with M% flakiness" and M less than N, but deferring this to @czeslavo .

abramche:
> etcd is a static pod, so setting a container **request** for it does not impact the amount of memory available to the container (the cgroup ceiling is controlled by the limit)

I think there's some misunderstanding going on. Firstly, I am changing CPU resource allocation, not memory; Secondly, I am changing the CPU **floor**, not the ceiling (which is undefined, and, as such, is indeed controlled by cgroups)

> Therefore - I do not think how setting the request on a static pod's container would achieve etcd's capability to "use 500m CPU" as the PR claims.

It does; you can see it in the build's [etcd must-gather artifacts](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3468/pull-scylla-operator-master-e2e-kind-fast/2062915652544368640/artifacts/must-gather/cluster/namespaces/kube-system/pods/etcd-scylla-operator-e2e-control-plane.yaml).

Though, I am still seeing big etcd [pressure](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3468/pull-scylla-operator-master-e2e-kind-fast/2062915652544368640/artifacts/must-gather/cluster/namespaces/kube-system/pods/etcd-scylla-operator-e2e-control-plane/etcd.current) even with the current setting, so probably `500m` is still too low.

> As for nftables - this change feels very arbitrary. I can be convinced that our network configuration and traffic patterns are such that there's a significant difference. But I'm not convinced yet. I'd expect higher quality evidence - e.g. "50 runs with iptables with N% flakiness" vs "50 runs with nftables with M% flakiness" and M less than N, but deferring this to @czeslavo .

I am happy to spend a day to generate this evidence (it would require some cooperation with the infra team - I'll need to utilize one of the runners.). I know it is still a small sample - but I have seen strong correlation with these changes and much more stable Jenkins results ("much" - as in, neither of four runs I scheduled on a Jenkins runner had tests failing due to this - all of them failed due to node-exporter situation. Before the change, the situation had been catastrophic for the past 10 days with incredibly slow performance and many more flaky tests as evidenced by the CI channel). I attribute this to a different Jenkins runner distro/setup (kind is sensitive to this).

mflendrich:
> I think there's some misunderstanding going on. Firstly, I am changing CPU resource allocation, not memory; Secondly, I am changing the CPU **floor**, not the ceiling (which is undefined, and, as such, is indeed controlled by cgroups)

No misunderstanding, just a typo on my end - replace "memory" with CPU. Fixing it in the comment right now.
I am quite positive that this PR does not have any impact on CPU being available to etcd, contrary to the claim in the PR title, because it only controls the node's schedulable (not runtime) capacity for other pods.

> > Therefore - I do not think how setting the request on a static pod's container would achieve etcd's capability to "use 500m CPU" as the PR claims.
> 
> It does; you can see it in the build's [etcd must-gather artifacts](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3468/pull-scylla-operator-master-e2e-kind-fast/2062915652544368640/artifacts/must-gather/cluster/namespaces/kube-system/pods/etcd-scylla-operator-e2e-control-plane.yaml).

You seem to be pointing at the fact that this PR has successfully updated `requests` to `500m` - let me know if I understand you wrong.
My point in response to that is that setting `requests` on a static pod  does not affect the number of CPU cycles available to the container.

> Though, I am still seeing big etcd [pressure](https://gcsweb.scylla-operator.scylladb.com/gcs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3468/pull-scylla-operator-master-e2e-kind-fast/2062915652544368640/artifacts/must-gather/cluster/namespaces/kube-system/pods/etcd-scylla-operator-e2e-control-plane/etcd.current) even with the current setting, so probably `500m` is still too low.

Ditto. I think you're seeing this result not because this value is too low, but because this knob does not have the effect of giving etcd more cpu time - at most it might make other pods unschedulable on the node.

We should chat to explore possible ways to guarantee more CPU for etcd (if that's indeed necessary).

As for nftables - deferring the decision here to @czeslavo

abramche:
> You seem to be pointing at the fact that this PR has successfully updated `requests` to `500m` - let me know if I understand you wrong.
> My point in response to that is that setting `requests` on a static pod does not affect the number of CPU cycles available to the container.

Updating the CPU floor definition still causes Kubernetes to re-deploy container with these settings (so yes, it applies to scheduling) -- but as the pod is redeployed (hence the wait), it is my understanding that runtime CPU quota is still increased (CPU requests map to CFS shares).

There are no limits, so under node CPU pressure (a lot of scylla pods running) and no etcd requests, kubernetes is less likely to allow etcd CPU burst.

It's easy to find out though, I'll raise the requests and see if it makes a difference.

If it's more straightforward, I can split the PR into two changes to unblock nftables promotion. @czeslavo WDYT?

czeslavo:
> > You seem to be pointing at the fact that this PR has successfully updated `requests` to `500m` - let me know if I understand you wrong.
> > My point in response to that is that setting `requests` on a static pod does not affect the number of CPU cycles available to the container.
> 
> Updating the CPU floor definition still causes Kubernetes to re-deploy container with these settings (so yes, it applies to scheduling) -- but as the pod is redeployed (hence the wait), it is my understanding that runtime CPU quota is still increased (CPU requests map to CFS shares).
> 
> There are no limits, so under node CPU pressure (a lot of scylla pods running) and no etcd requests, kubernetes is less likely to allow etcd CPU burst.
> 
> It's easy to find out though, I'll raise the requests and see if it makes a difference.
> 
> If it's more straightforward, I can split the PR into two changes to unblock nftables promotion. @czeslavo WDYT?

Yes, let's extract the nftables change to a separate PR. I find it uncontroversial - it's an upgrade to a recommended modern approach.

abramche:
superseded with https://github.com/scylladb/scylla-operator/pull/3472

etcd tuning will be a topic for a future discussion

PR #3473, opened 2026-06-09
chore(deps): bump Python deps in docs/uv.lock for security fixes

mykaul:
## What

Bumps transitive Python dependencies in `docs/uv.lock` to resolve open [Dependabot security alerts](https://github.com/scylladb/scylla-operator/security/dependabot).

| Package | Old | New | First patched | Alert(s) |
|---------|-----|-----|---------------|----------|
| urllib3 | 2.6.3 | 2.7.0 | 2.7.0 | GHSA-qccp-gfcp-xxvc (high), GHSA-mf9v-mfxr-j63j (high) |
| idna | 3.11 | 3.18 | 3.15 | GHSA-65pc-fj4g-8rjx (medium) |
| starlette | 0.52.1 | 1.2.1 | 1.0.1 | GHSA-86qp-5c8j-p5mr (medium) |
| requests | 2.32.5 | 2.34.2 | 2.33.0 | GHSA-gc5v-m9x4-r6x2 (medium) |
| Pygments | 2.19.2 | 2.20.0 | 2.20.0 | GHSA-5239-wwwm-4pmq (low) |

## How

All five are transitive deps, so they were bumped via `uv lock --upgrade-package <pkg>=<min>` rather than hand-editing the lock, keeping version + sdist/wheel hashes consistent. `pyproject.toml` is **not** changed.

`uv 0.10.12` was used to match the project's pinned `required-version` range (`>=0.10.9, <0.11.0`), which tracks the upstream `scylladb/scylla-operator-images` `uv/0.10` CI image — so the lock stays compatible with CI and the pin was left untouched.

## Verification

- `uv lock --check` passes (lock is internally consistent).
- `uv sync --locked` succeeds (resolution is installable, no drift).
- Diff is limited to `docs/uv.lock`: 15 insertions / 15 deletions, only the five intended packages.

> Note: the remaining open Dependabot alerts are all Go (`github.com/docker/docker`) and out of scope for this change.

### DIFF
--- a/docs/uv.lock
+++ b/docs/uv.lock
@@ -139,11 +139,11 @@ wheels = [
 
 [[package]]
 name = "idna"
-version = "3.11"
+version = "3.18"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/6f/6d/0703ccc57f3a7233505399edb88de3cbd678da106337b9fcde432b65ed60/idna-3.11.tar.gz", hash = "sha256:795dafcc9c04ed0c1fb032c2aa73654d8e8c5023a7df64a53f39190ada629902", size = 194582, upload-time = "2025-10-12T14:55:20.501Z" }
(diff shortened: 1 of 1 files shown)


rzetelskik (approved):
/lgtm

PR #3478, opened 2026-06-19
test(serial): add new iotune/non-devmode cluster test

abramche:
- create a new test deploying production mode cluster (we do not have any) - this automatically kicks off IO tuning - restart the cluster and assert that iotune is not performed after restart

Resolves https://scylladb.atlassian.net/browse/OPERATOR-155

/kind new-test
/priority important-soon

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryutilwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/storage/names"
@@ -227,6 +228,16 @@ func (f *Framework) GetDefaultScyllaCluster() *scyllav1.ScyllaCluster {
 
 	sc, _, err := scyllafixture.ScyllaClusterTemplate.RenderObject(renderArgs)
 	o.Expect(err).NotTo(o.HaveOccurred())
+
+	return sc
+}
+
+func (f *Framework) GetNonDevModeScyllaCluster() *scyllav1.ScyllaCluster {
+	sc := f.GetDefaultScyllaCluster()
+	sc.Spec.DeveloperMode = false
+	// ScyllaDB production mode requires at least 10Gi to run iotune and 4Gi of memory (system requirements)
+	sc.Spec.Datacenter.Racks[0].Storage.Capacity = "10Gi"
+	sc.Spec.Datacenter.Racks[0].Resources.Limits[corev1.ResourceMemory] = resource.MustParse("4Gi")
 
 	return sc
 }
--- a/test/e2e/set/scyllacluster/scyllacluster_iotune.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_iotune.go
@@ -1,0 +1,173 @@
+// Copyright (C) 2021 ScyllaDB
+
+package scyllacluster
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/pkg/pointer"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
+	scyllaclusterverification "github.com/scylladb/scylla-operator/test/e2e/utils/verification/scyllacluster"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/exec"
+)
+
+// scyllaIOPropertiesPath is where the sidecar expects the IO properties file. It's a symlink to the
+// cached results on the data volume, dangling until iotune actually produces output.
+const scyllaIOPropertiesPath = "/etc/scylla.d/" + naming.ScyllaIOPropertiesName
+
+func IsIOTuneRequested(ctx context.Context, f *framework.Framework, podName string) (bool, error) {
+	// IsIOTuneRequested reports whether iotune was requested by the entrypoint for the given Pod.
+	// It inspects the arguments the sidecar rendered onto the long-lived /docker-entrypoint.py
+	// process (it blocks on supervisord.wait(), so its args are readable for the whole container
+	// lifetime by execing into the container). --io-setup=0 and --io-properties-file are passed
+	// only when the cached IO properties file already exists, i.e. when iotune was skipped.
+	stdout, stderr, err := utils.ExecWithOptions(ctx, f.ClientConfig(), f.KubeClient().CoreV1(), utils.ExecOptions{
+		Command:       []string{"/usr/bin/pgrep", "-af", "docker-entrypoint.py"},
+		Namespace:     f.Namespace(),
+		PodName:       podName,
+		ContainerName: naming.ScyllaContainerName,
+		CaptureStdout: true,
+		CaptureStderr: true,
+	})
+	if err != nil {
+		return false, fmt.Errorf("can't read entrypoint process args from pod %q: %w (stderr: %q)", podName, err, stderr)
+	}
+
+	entrypointCommand := stdout
+	if !strings.Contains(entrypointCommand, "docker-entrypoint.py") {
+		return false, fmt.Errorf("can't find docker-entrypoint.py process in pod %q", podName)
+	}
+
+	// In developer mode iotune is never run, so there's nothing to perform or skip.
+	if strings.Contains(entrypointCommand, "--developer-mode=1") {
+		return false, nil
+	}
+
+	hasIOSetupDisabled := strings.Contains(entrypointCommand, "--io-setup=0")
+	hasIOPropertiesFile := strings.Contains(entrypointCommand, "--io-properties-file")
+	isIOTuneSkipped := hasIOSetupDisabled && hasIOPropertiesFile
+
+	return !isIOTuneSkipped, nil
+}
+
+// IOPropertiesFileExists reports whether the resolved io_properties file exists in the ScyllaDB
+// container. The path is a symlink that's always present, but its target only exists once iotune has
+// actually produced output, so `test -f` (which follows the symlink) confirms iotune was performed
+// rather than merely requested.
+func IOPropertiesFileExists(ctx context.Context, f *framework.Framework, podName string) (bool, error) {
+	_, stderr, err := utils.ExecWithOptions(ctx, f.ClientConfig(), f.KubeClient().CoreV1(), utils.ExecOptions{
+		Command:       []string{"test", "-f", scyllaIOPropertiesPath},
+		Namespace:     f.Namespace(),
+		PodName:       podName,
+		ContainerName: naming.ScyllaContainerName,
+		CaptureStdout: true,
+		CaptureStderr: true,
+	})
+	if err != nil {
+		var codeExitErr exec.CodeExitError
+		if errors.As(err, &codeExitErr) && codeExitErr.Code == 1 {
+			// `test -f` exits with 1 when the file doesn't exist (or the symlink is dangling).
+			return false, nil
+		}
+		return false, fmt.Errorf("can't check for io_properties file in pod %q: %w (stderr: %q)", podName, err, stderr)
+	}
+
+	return true, nil
+}
+
+var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuiteParallelOpenShift, func() {
+	var f *framework.Framework
+
+	g.BeforeEach(func(ctx context.Context) {
+		f = framework.NewFramework(ctx, "scyllacluster")
+	})
+
+	g.It("should skip iotune on restart when io properties are cached", func(testCtx context.Context) {
+		framework.By("Creating a ScyllaCluster with developer mode disabled")
+		sc := f.GetNonDevModeScyllaCluster()
+		sc.Spec.Datacenter.Racks[0].Members = 1
+
+		sc, err := f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(testCtx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
+		initialRolloutCtx, initialRolloutCtxCancel := utils.ContextForRollout(testCtx, sc)
+		defer initialRolloutCtxCancel()
+		sc, err = controllerhelpers.WaitForScyllaClusterState(initialRolloutCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		scyllaclusterverification.Verify(testCtx, f.KubeClient(), f.ScyllaClient(), sc)
+
+		pod, err := f.KubeClient().CoreV1().Pods(f.Namespace()).Get(
+			testCtx,
+			utils.GetNodeName(sc, 0),
+			metav1.GetOptions{},
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Verifying that iotune was requested on initial rollout")
+		isIOTuneRequested, err := IsIOTuneRequested(testCtx, f, pod.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(isIOTuneRequested).To(o.BeTrue())
+
+		framework.By("Verifying that iotune produced the io_properties file")
+		ioPropertiesFileExists, err := IOPropertiesFileExists(testCtx, f, pod.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(ioPropertiesFileExists).To(o.BeTrue())
+
+		initialPodUID := pod.UID
+		framework.Infof("Initial pod %q UID is %q", pod.Name, initialPodUID)
+
+		framework.By("Restarting the ScyllaDB Pod")
+		err = f.KubeClient().CoreV1().Pods(f.Namespace()).Delete(testCtx, pod.Name, metav1.DeleteOptions{
+			Preconditions: &metav1.Preconditions{
+				UID: &initialPodUID,
+			},
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the old ScyllaDB Pod to be deleted")
+		deletionCtx, deletionCtxCancel := context.WithTimeoutCause(
+			testCtx,
+			utils.ScyllaDBTerminationTimeout,
+			fmt.Errorf("pod %q has not finished termination in time", naming.ObjRef(pod)),
+		)
+		defer deletionCtxCancel()
+		err = framework.WaitForObjectDeletion(
+			deletionCtx,
+			f.DynamicClient(),
+			corev1.SchemeGroupVersion.WithResource("pods"),
+			pod.Namespace,
+			pod.Name,
+			pointer.Ptr(initialPodUID),
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
+		restartRolloutCtx, restartRolloutCtxCancel := utils.ContextForRollout(testCtx, sc)
+		defer restartRolloutCtxCancel()
+		sc, err = controllerhelpers.WaitForScyllaClusterState(restartRolloutCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Verifying that iotune was skipped on restart")
+		isIOTuneRequested, err = IsIOTuneRequested(testCtx, f, pod.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(isIOTuneRequested).To(o.BeFalse())
+
+		framework.By("Verifying that io_properties file was present on restart")
+		ioPropertiesFileExists, err = IOPropertiesFileExists(testCtx, f, pod.Name)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(ioPropertiesFileExists).To(o.BeTrue())
+
+	})
+})


abramche:
/test images
/test e2e-gke-serial

the test will not be suited for Kind
```
2026-07-07T17:14:57.382250515Z ERROR:root:Command '['/usr/bin/iotune', '--format', 'envfile', '--options-file', '/etc/scylla.d/io.conf', '--properties-file', '/etc/scylla.d/io_properties.yaml', '--evaluation-directory', '/var/lib/scylla/data', '--cpuset', '0,1,2,3,4,5,6,7']' returned non-zero exit status 1.
2026-07-07T17:14:57.382260576Z ERROR:root:['/var/lib/scylla/data'] did not pass validation tests, it may not be on XFS and/or has limited disk space.
2026-07-07T17:14:57.382262046Z This is a non-supported setup, and performance is expected to be very bad.
2026-07-07T17:14:57.382263737Z For better performance, placing your data on XFS-formatted directories is required.
2026-07-07T17:14:57.382265179Z To override this error, enable developer mode as follow:
2026-07-07T17:14:57.382266913Z sudo /opt/scylladb/scripts/scylla_dev_mode_setup --developer-mode 1
2026-07-07T17:14:57.425312361Z failed!
2026-07-07T17:14:57.494247390Z Traceback (most recent call last):
2026-07-07T17:14:57.494255411Z   File "/docker-entrypoint.py", line 26, in <module>
2026-07-07T17:14:57.494257539Z     setup.io()
2026-07-07T17:14:57.494259160Z     ~~~~~~~~^^
2026-07-07T17:14:57.494261162Z   File "/scyllasetup.py", line 83, in io
2026-07-07T17:14:57.494262820Z     self._run(['/opt/scylladb/scripts/scylla_io_setup'])
2026-07-07T17:14:57.494264931Z     ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-07-07T17:14:57.494267102Z   File "/scyllasetup.py", line 53, in _run
2026-07-07T17:14:57.494269000Z     subprocess.check_call(*args, **kwargs)
2026-07-07T17:14:57.494270878Z     ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
2026-07-07T17:14:57.494272968Z   File "/opt/scylladb/python3/lib64/python3.14/subprocess.py", line 419, in check_call
2026-07-07T17:14:57.494274729Z     raise CalledProcessError(retcode, cmd)
2026-07-07T17:14:57.494276778Z subprocess.CalledProcessError: Command '['/opt/scylladb/scripts/scylla_io_setup']' returned non-zero exit status 1.
```

test/e2e/fixture/scylla/nondevmode.scyllacluster.yaml.tmpl
```diff
@@ -0,0 +1,55 @@
...
+        type: {{ .nodesBroadcastAddressType }}
+      clients:
+        type: {{ .clientsBroadcastAddressType }}
+  datacenter:
+    name: us-east-1
+    racks:
+    - name: us-east-1a
+      members: 1
+      storage:
+        capacity: 11Gi
```
abramche: info: ScyllaDB will not run IO tune on <=10 GB storage

test/e2e/fixture/scylla/nondevmode.scyllacluster.yaml.tmpl
czeslavo, line 1: The new template differs from `scyllacluster.yaml.tmpl` in only three fields (developerMode, storage.capacity, resources.limits.memory). 

I suggest that we use the default template as a base, and we modify it "in-flight" in the `GetNonDevModeScyllaCluster` function. This way, we keep a single base without duplication.

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +213,83 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+		scyllaclusterverification.Verify(ctx, f.KubeClient(), f.ScyllaClient(), sc)
+
+		pod, err := f.KubeClient().CoreV1().Pods(f.Namespace()).Get(
+			ctx,
+			utils.GetNodeName(sc, 0),
+			metav1.GetOptions{},
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		initialPodUID := pod.UID
+		framework.Infof("Initial pod %q UID is %q", pod.Name, initialPodUID)
```
czeslavo: Let's add a precondition check that will ensure that right after the first rollout, there's no `--io-properties-file` passed to the `scylla` process. Without that, we can't be sure it isn't always passed.

To not introduce duplication, we can add a helper `IOPropertiesPassedToScyllaProcess(...) (bool, error)` and use it in both cases.

abramche: Addressed both in cdfa762caf22c29b794a415e03cc0570df2a4543

test/e2e/fixture/scylla/nondevmode.scyllacluster.yaml.tmpl
abramche: addressed in cdfa762caf22c29b794a415e03cc0570df2a4543

czeslavo (approved):
/lgtm

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +231,83 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuiteParallelOpenShift, func() {
+	var f *framework.Framework
+
+	g.BeforeEach(func(ctx context.Context) {
+		f = framework.NewFramework(ctx, "scyllacluster")
+	})
+
+	g.It("should skip iotune on restart when io properties are cached", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+		defer cancel()
```
rzetelskik: nit: use specCtx instead

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +231,83 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+		framework.By("Creating a ScyllaCluster with developer mode disabled")
+		sc := f.GetNonDevModeScyllaCluster()
+		sc.Spec.Datacenter.Racks[0].Members = 1
+		var ioPropertiesPassed bool
+
+		sc, err := f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
+		waitCtx1, waitCtx1Cancel := utils.ContextForRollout(ctx, sc)
```
rzetelskik: nit: I know this is a prevalent pattern in the e2e tests but let's not further cement this code smell - please give ctx and ctxCancels meaningful names, the suffixed vars are hard to maintain

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +231,83 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+			f.DynamicClient(),
+			corev1.SchemeGroupVersion.WithResource("pods"),
+			pod.Namespace,
+			pod.Name,
+			pointer.Ptr(initialPodUID),
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
+		waitCtx2, waitCtx2Cancel := utils.ContextForRollout(ctx, sc)
```
rzetelskik: ditto

rzetelskik (commented):
a few nits - not required but an opportunity to improve code health

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: Addressed all of the suggestions while I was hardening the test logic against IO properties file being always present in the binary arguments

abramche:
The test was incorrectly reading from the ignition container (the logs are in the scylla container)

test/e2e/framework/framework.go
```diff
@@ -231,6 +232,15 @@ func (f *Framework) GetDefaultScyllaCluster() *scyllav1.ScyllaCluster {
 	return sc
 }
 
+func (f *Framework) GetNonDevModeScyllaCluster() *scyllav1.ScyllaCluster {
+	sc := f.GetDefaultScyllaCluster()
+	sc.Spec.DeveloperMode = false
+	sc.Spec.Datacenter.Racks[0].Storage.Capacity = "11Gi"
+	sc.Spec.Datacenter.Racks[0].Resources.Limits[corev1.ResourceMemory] = resource.MustParse("3Gi")
```
rzetelskik: magic values - please add comments (also iirc 10Gi is the threshold)

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -30,6 +36,29 @@ func addQuantity(lhs resource.Quantity, rhs resource.Quantity) *resource.Quantit
...
+	}
+
+	entrypointCommand := logs.String()
+	if !strings.Contains(entrypointCommand, "Scylla entrypoint") {
+		return false, fmt.Errorf("can't find Scylla entrypoint log in pod %q", podName)
+	}
+
+	hasIOSetupDisabled := strings.Contains(entrypointCommand, "--io-setup=0")
+	hasIOPropertiesFile := strings.Contains(entrypointCommand, "--io-properties-file")
+	isIOTuneSkipped := hasIOSetupDisabled && hasIOPropertiesFile
```
rzetelskik: why don't we verify the flags passed to the process instead of what's printed in the logs?

czeslavo: It would be more robust not to rely on logs, but (as before) check arguments/flags passed to a process. Please consider reworking this.

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +241,83 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+
+var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuiteParallelOpenShift, func() {
+	var f *framework.Framework
+
+	g.BeforeEach(func(ctx context.Context) {
+		f = framework.NewFramework(ctx, "scyllacluster")
+	})
+
+	g.It("should skip iotune on restart when io properties are cached", func(specCtx context.Context) {
+		testCtx, testCtxCancel := context.WithTimeout(specCtx, testTimeout)
```
rzetelskik: do we need the test timeout? Either way that'll fail the suite, so we might as well get limited by the suite timeout

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +241,83 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+		scyllaclusterverification.Verify(testCtx, f.KubeClient(), f.ScyllaClient(), sc)
+
+		pod, err := f.KubeClient().CoreV1().Pods(f.Namespace()).Get(
+			testCtx,
+			utils.GetNodeName(sc, 0),
+			metav1.GetOptions{},
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Verifying that iotune was performed on initial rollout")
```
rzetelskik: this doesn't really check if it was performed, but if it was requested. To check if it was actually performed, we could add an additional check for the existence of io_properties file

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -30,6 +36,29 @@ func addQuantity(lhs resource.Quantity, rhs resource.Quantity) *resource.Quantit
...
+	entrypointCommand := logs.String()
+	if !strings.Contains(entrypointCommand, "Scylla entrypoint") {
+		return false, fmt.Errorf("can't find Scylla entrypoint log in pod %q", podName)
+	}
+
+	hasIOSetupDisabled := strings.Contains(entrypointCommand, "--io-setup=0")
+	hasIOPropertiesFile := strings.Contains(entrypointCommand, "--io-properties-file")
+	isIOTuneSkipped := hasIOSetupDisabled && hasIOPropertiesFile
+
+	return !sc.Spec.DeveloperMode && !isIOTuneSkipped, nil
```
czeslavo: Let's consider dropping `!sc.Spec.DeveloperMode` (and passing `sc` at all to this helper) - it's set by us in the test, doesn't really matter IMO.

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: @rzetelskik @czeslavo it is complicated. None of the flags in use are passed explicitly or visible from the spec, both `--io-setup` and `--io-properties-file` are internals in `pkg/sidecar/config/config.go:190`. 

`docker-entrypoint.py` only accepts the `--developer-mode=0` argument from the spec (which implies it should be eligible for IO tuning). The script spawns the scylla process with the flags -- and scylla binary effectively always has the flag `--io-properties-file`, regardless if IO tune ran (cold start - tuning runs - new file) or not (restart - tuning skipped - cached file); and `--io-tune=0` is not passed to the binary at all.

 This makes it difficult to understand if IO tuning was done or not without reading the logs; and there is a race element trying to intercept the entrypoint flags.

Or is there a simpler idea I do not see?

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: How do you mean? As far as I can see, all test packages define a per-test timeout. `SO_E2E_TIMEOUT` is rich `24h` (yikes), it's best to fail fast.

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: I don't mind (it's a small check), but one could argue it's area of Scylla's responsibility

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: I added it there as the function is exported and could be potentially used elsewhere - and the name does not make it explicit it's only supposed to be done against Prod mode clusters.

By the old logic, running the function against a devmode cluster would falsely confirm iocheck being run (as neither `--io-tune` or `--io-properties-file` were present in a devmode command)

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: I agree it's better not to read logs and rely on arguments, but it's more complicated than I thought - see https://github.com/scylladb/scylla-operator/pull/3478#discussion_r3552609878

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: > How do you mean? As far as I can see, all test packages define a per-test timeout.

Some tests define them, some don't. Some specify this through this context, some set it as a decorator. We should fail on particular assertions or on granular contexts. Arbitrary, per-test timeouts just seem like noise to me - after all why is it 45m? Suite timeouts are overriden in the actual jobs. But I won't die on that hill since it's very prevalent in the entire test suite, so leaving this up to you.

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: We set the flags for docker-entrypoint.py so you should be able to read them by execing to the container. It's not pretty, but it verifies what the sidecar actually set.

> there is a race element trying to intercept the entrypoint flags

Assuming the container is not restarted, how is there a race?

abramche: > Assuming the container is not restarted, how is there a race?

`docker-entrypoint.py` process is short-lived, once it spawns ScyllaDB binary, it exits. Container exec may be too late. I'll investigate

abramche: `docker-entrypoint.py` lives for a window of 2s in the container

```
2026-07-09T14:27:12.277637798Z I0709 14:27:12.277553       1 config/config.go:278] "Scylla entrypoint" Command="/docker-entrypoint.py --seeds=10.57.1.159 --smp=1 --prometheus-address=0.0.0.0 --cpuset=0-7 --listen-address=0.0.0.0 --overprovisioned=1 --broadcast-address=10.57.1.159 --broadcast-rpc-address=10.57.1.159 --developer-mode=0 --io-setup=0 --io-properties-file=/etc/scylla.d/io_properties.yaml"

2026-07-09T14:27:14.494190199Z command used: "/usr/bin/scylla --log-to-syslog 0 --log-to-stdout 1 --network-stack posix --cpuset 0-7 --smp 1 --overprovisioned --listen-address 0.0.0.0 --rpc-address 0.0.0.0 --seed-provider-parameters seeds=10.57.1.159 --broadcast-address 10.57.1.159 --broadcast-rpc-address 10.57.1.159 --alternator-address 0.0.0.0 --blocked-reactor-notify-ms 25 --prometheus-address=0.0.0.0 --io-properties-file=/etc/scylla.d/io_properties.yaml"
```

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: removed test timeout in bffc964dba4d57385567130398dedfb0651ab637

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: io properties file is always present, regardless if IOtune ran or not, so presence of the file is not an indication.

The current implementation (with log parsing) already checks whether it was run on the initial rollout.

rzetelskik: > io properties file is always present, regardless if IOtune ran or not, so presence of the file is not an indication.

If that's true, then your IsIOTunePerformed helper would always return false?
https://github.com/scylladb/scylla-operator/blob/f8f93c62da496a58f2c24945b28402fa002c054f/pkg/sidecar/config/config.go#L260-L264

The file is the output of iotune. Well, it's a symlink to the output technically.

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: > docker-entrypoint.py process is short-lived, once it spawns ScyllaDB binary, it exits

The content of the script doesn't match this claim:
```
$ cat docker-entrypoint.py | grep -B 1 wait
    supervisord = subprocess.Popen(["/usr/bin/supervisord", "-c",  "/etc/supervisord.conf"])
    supervisord.wait()
```

The container would also complete in that case.
And I don't understand how these logs prove it.

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -211,3 +241,80 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
 		verification.VerifyCQLData(ctx, di)
 	})
 })
+
+var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuiteParallelOpenShift, func() {
```
rzetelskik: one thing I didn't notice before - why is this being added to scyllacluster_updates.go file? It doesn't have much to do with the existing test. How about scyllacluster_iotune.go or something along these lines?

test/e2e/set/scyllacluster/scyllacluster_updates.go
```diff
@@ -30,6 +36,29 @@ func addQuantity(lhs resource.Quantity, rhs resource.Quantity) *resource.Quantit
 	return &res
 }
 
+func IsIOTunePerformed(ctx context.Context, f *framework.Framework, podName string, sc *scyllav1.ScyllaCluster) (bool, error) {
+	logOptions := &corev1.PodLogOptions{
```
rzetelskik: if you want to keep the guard, you can return early
```
if sc.Spec.DeveloperMode {
  return false, nil
}
```

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: https://github.com/rzetelskik/scylla-operator/commit/ebb95d2f6e01ce83cb492103df3fd976b2049752 (ai generated poc)

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: https://github.com/scylladb/scylla-operator/commit/8291172053b5960a974db507fec9d279dacf3640 (ai generated poc)

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: I stand corrected - I must have inspected the wrong container initially and relied on false evidence since. I incorporated your changes into this PR

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: Great point, extracted the test

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: I meant to say that the IO properties check is performed early, namely in https://github.com/scylladb/scylla-operator/blob/f8f93c62da496a58f2c24945b28402fa002c054f/pkg/sidecar/config/config.go#L56
and therefore the argument would always be passed to the binary, whether IO tune just ran or was cached. But it's double safeguarded with presence or absence of `--io-setup=0` so no uncovered logic branch exists.

Anyway, I believe this thread is addressed now (as your POC is incorporated into this PR)

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: I believe this got lost (assuming that thumbs up is an indication you wanted to apply this)

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: > and therefore the argument would always be passed to the binary, whether IO tune just ran or was cached

Yes. But not before the first IO tune run. So by checking the file's presence you're verifying that it actually ran, and not only that it was requested. But one could argue it's a bit void since we're not really able to verify the file wasn't there before the entrypoint was ran.

> Anyway, I believe this thread is addressed now (as your POC is incorporated into this PR)

If that was the intention, I think you lost it on the way, because I don't see it

test/e2e/set/scyllacluster/scyllacluster_updates.go
abramche: Yes - that is, before the PR update; however in the current iteration I think it will unnecessarily clutter the method - the rest of the logic is devmode-safe, does not take time, and the single point of return is clean. Adding the statement will remove the devmode check from return and add 3 more lines. Do you think it's worth it?

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: To reiterate - checking for existence of the file actually changes the semantics of the check to "performed". Now, despite the name, the check just checks if it's requested.

test/e2e/set/scyllacluster/scyllacluster_updates.go
rzetelskik: Actually, after reconsideration - I think there's a third option that would align the logic better to check if io tune is requested or not. If developer-mode short circuits to false, why don't we verify the flag, instead of the spec field? In the current shape that only functions as an assertion.

```
// IsIOTuneRequested reports whether iotune was requested by the entrypoint for the given Pod.
// It inspects the arguments the sidecar rendered onto the long-lived /docker-entrypoint.py
// process (it blocks on supervisord.wait(), so its args are readable for the whole container
// lifetime by execing into the container). --io-setup=0 and --io-properties-file are passed
// only when the cached IO properties file already exists, i.e. when iotune was skipped.
// In developer mode iotune never runs, so --developer-mode=1 short-circuits to false.
func IsIOTuneRequested(ctx context.Context, f *framework.Framework, podName string) (bool, error) {
	stdout, stderr, err := utils.ExecWithOptions(ctx, f.ClientConfig(), f.KubeClient().CoreV1(), utils.ExecOptions{
		Command:       []string{"/usr/bin/pgrep", "-af", "docker-entrypoint.py"},
		Namespace:     f.Namespace(),
		PodName:       podName,
		ContainerName: naming.ScyllaContainerName,
		CaptureStdout: true,
		CaptureStderr: true,
	})
	if err != nil {
		return false, fmt.Errorf("can't read entrypoint process args from pod %q: %w (stderr: %q)", podName, err, stderr)
	}

	entrypointCommand := stdout
	if !strings.Contains(entrypointCommand, "docker-entrypoint.py") {
		return false, fmt.Errorf("can't find docker-entrypoint.py process in pod %q", podName)
	}

	// In developer mode iotune is never run, so there's nothing to perform or skip.
	if strings.Contains(entrypointCommand, "--developer-mode=1") {
		return false, nil
	}

	hasIOSetupDisabled := strings.Contains(entrypointCommand, "--io-setup=0")
	hasIOPropertiesFile := strings.Contains(entrypointCommand, "--io-properties-file")
	isIOTuneSkipped := hasIOSetupDisabled && hasIOPropertiesFile

	return !isIOTuneSkipped, nil
}
```

As for the early return discussion  - if the `io-setup` and `io-properties-file` have no effect when `developer-mode=1` is set, then to me the early return makes the logic easier to understand (if `developer-mode=1`, always return false as the other flags have no effect).

abramche: Ah, this makes sense - no need to pass along the spec any more.

Do you think we need an extra assignment?
```
	entrypointCommand := stdout
```

abramche: Simplified the helper as advised above

rzetelskik (approved):
thanks, let's squash the fixup commits and we can land this

abramche:
> thanks, let's squash the fixup commits and we can land this

done

rzetelskik (approved):
Thanks for the updates
/lgtm

PR #3482, opened 2026-07-01
feat: switch to scylladb-node-exporter sidecar for ScyllaDB >=2026.3 

rzetelskik:

Starting with ScyllaDB 2026.3, supervisord no longer launches the internal scylladb-node-exporter process within the ScyllaDB container. It silently breaks cluster monitoring and alert routing in the operator, as Prometheus can no longer scrape node metrics from port 9100 on the ScyllaDB container.

To restore monitoring parity for ScyllaDB >=2026.3, this PR adds an operator-managed scylladb-node-exporter  sidecar container.
The changes follow the design proposed in the RFC: https://scylladb.atlassian.net/browse/OPERATOR-200.

Sidecar injection is behind `ScyllaDBVersionWithoutNodeExporter` semver gate. If the target cluster version is lower, the operator exposes port 9100 on the main container as before. For versions 2026.3 and newer, the operator appends a separate scylladb-node-exporter container.
The sidecar mounts the `/var/lib/scylla` persistent volume as read-only. This grants the exporter access to the mount path so it can track local disk usage and capacity metrics.

`scyllaDBNodeExporterImage` field is added to `ScyllaOperatorConfig`. This allows operators to override the exporter image cluster-wide while falling back to a default defined in the operator assets.
This PR configures it with the upstream image in place of the scylladb-node-exporter as the container image is not being built yet. The switch is tracked in https://scylladb.atlassian.net/browse/OPERATOR-219.

This PR only extends the existing documentation with information about the new sidecar and config options.
Documenting the inconsistencies in node-exporter metric collection is tracked in https://scylladb.atlassian.net/browse/OPERATOR-218.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-130.

/cc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@
 
 ## Unreleased
 
+### Other changes
+
+- ScyllaDB Operator now runs `scylladb-node-exporter` as a dedicated sidecar container for ScyllaDB clusters running
+  version 2026.3 or later, since these versions no longer bundle node-exporter. The node-exporter container image is configurable via
+  `ScyllaOperatorConfig.spec.scyllaDBNodeExporterImage`.
+  [#3482](https://github.com/scylladb/scylla-operator/pull/3482)
+
 ## [1.21.0](https://github.com/scylladb/scylla-operator/releases/tag/v1.21.0)
 
 Release date: 2026-05-20
--- a/assets/config/config.yaml
+++ b/assets/config/config.yaml
@@ -2,6 +2,8 @@ operator:
   # renovate: datasource=docker depName=scylladb packageName=docker.io/scylladb/scylla versioning=semver
   scyllaDBVersion: "2026.2.1"
   scyllaDBUtilsImage: "docker.io/scylladb/scylla:2026.1.0@sha256:d450d2d8636ab7b511b68180b4c535bf2294d0d6427adf11fc7f4bddfdbff35a"
+  # FIXME: switch to scylladb-node-exporter image once https://scylladb.atlassian.net/browse/RELENG-711 is completed.
+  scyllaDBNodeExporterImage: "quay.io/prometheus/node-exporter:v1.11.1"
   # renovate: datasource=docker depName=scylla-manager packageName=docker.io/scylladb/scylla-manager versioning=semver
   scyllaDBManagerVersion: "3.11.2@sha256:61ec1396d438945abf3db46db893d7e4d2324acb5523d143077f8fdfea05779d"
   # renovate: datasource=docker depName=scylla-manager-agent packageName=docker.io/scylladb/scylla-manager-agent versioning=semver
--- a/docs/source/deploy-scylladb/before-you-deploy/configure-operator.md
+++ b/docs/source/deploy-scylladb/before-you-deploy/configure-operator.md
@@ -27,6 +27,7 @@ The `status` section shows the resolved values that are actually in use, includi
 ```yaml
 status:
   scyllaDBUtilsImage: docker.io/scylladb/scylla:2025.1.9@sha256:...
+  scyllaDBNodeExporterImage: quay.io/prometheus/node-exporter:v1.11.1
   bashToolsImage: registry.access.redhat.com/ubi9/ubi:9.5-...@sha256:...
   grafanaImage: docker.io/grafana/grafana:12.2.0@sha256:...
   prometheusVersion: v3.6.0
@@ -35,7 +36,7 @@ status:
 
 ## Configurable fields
 
-```{list-table}
+:::{list-table}
 :header-rows: 1
 
 * - Spec field
@@ -44,6 +45,9 @@ status:
 * - `scyllaUtilsImage`
   - ScyllaDB image used for running utility scripts (perftune, sysctl). Determines which tuning scripts are used for performance optimization.
   - Latest ScyllaDB image.
+* - `scyllaDBNodeExporterImage`
+  - `scylladb-node-exporter` image that ScyllaDB Operator runs as a sidecar to expose node-level OS metrics for ScyllaDB clusters. For versions before 2026.3.0, this field is ineffective because node-exporter is bundled and run directly within `scyllaDBImage`.
+  - Latest scylladb-node-exporter image.
 * - `configuredClusterDomain`
   - Kubernetes cluster domain. Must be a fully qualified domain name.
   - Auto-discovered via DNS lookup of `kubernetes.default.svc`.
@@ -56,7 +60,7 @@ status:
 * - `unsupportedPrometheusVersionOverride`
   - Override the Prometheus version. **Unsupported** — for advanced use only.
   - Latest tested Prometheus version.
-```
+:::
 
 :::{caution}
 Fields prefixed with `unsupported` are not covered by the regular support policy. Use them only if you have a specific reason and understand the implications.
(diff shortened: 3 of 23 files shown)


rzetelskik:
@mflendrich since @czeslavo is pretty busy with IDP, would you be able to take the review of this one?

rzetelskik:
@mflendrich I'm wondering if the regression test with a hardcoded version is the right approach given we'd have to handle https://scylladb.atlassian.net/browse/OPERATOR-226, or if creating dedicated CI (potentially with a dedicated, smaller suite) job would be a cleaner solution, given there are no specific steps required? WDYT?

mflendrich:
> @mflendrich I'm wondering if the regression test with a hardcoded version is the right approach given we'd have to handle https://scylladb.atlassian.net/browse/OPERATOR-226, or if creating dedicated CI (potentially with a dedicated, smaller suite) job would be a cleaner solution, given there are no specific steps required? WDYT?

We should discuss on Monday how we want to approach testing multiple versions,  given the fact that GHA will open up the possibility of creating parameterized workflows (where we can provide the scylladb version as a workflow parameter) + matrix workflows (where we can fan out the same workflow across a number of ScyllaDB versions, for example). I do not have a simple answer.

rzetelskik:
> > @mflendrich I'm wondering if the regression test with a hardcoded version is the right approach given we'd have to handle https://scylladb.atlassian.net/browse/OPERATOR-226, or if creating dedicated CI (potentially with a dedicated, smaller suite) job would be a cleaner solution, given there are no specific steps required? WDYT?
> 
> We should discuss on Monday how we want to approach testing multiple versions, given the fact that GHA will open up the possibility of creating parameterized workflows (where we can provide the scylladb version as a workflow parameter) + matrix workflows (where we can fan out the same workflow across a number of ScyllaDB versions, for example). I do not have a simple answer.

Until we switch to GHA, I'd leave it as is rather than copy-paste a Prow job definition for a particular version. https://scylladb.atlassian.net/browse/OPERATOR-226 will most likely be able to exclude this test using a label or an env var.

assets/config/config.yaml
```diff
@@ -14,6 +16,8 @@ operatorTests:
   scyllaDBVersions:
     updateFrom: "2026.2.0" # One patch lower than .operator.scyllaDBVersion (if no patches yet for this minor, use a release candidate)
     upgradeFrom: "2026.1.8" # One minor lower than .operator.scyllaDBVersion
+    # TODO: replace with a version with supervisord-managed node-exporter when the default is updated to >=2026.3.
+    imageWithoutNodeExporter: "docker.io/scylladb/scylla-nightly:2026.3.0-dev-0.20260701.812cace29c42"
```
mflendrich: I think this field deserves a sentence of two of explanation of its semantics. Something like "Invariant: scyllaDBVersion runs node-exporter internally under supervisord. imageWithoutNodeExporter is a pre-release ScyllaDB image that has the new behavior (node-exporter removed from supervisord's governance)."

The TODO doesn't look correct- it should say "invert the logic of this check (and have a special case for older versions running supervisord under the hood) when the node-exporter invariant changes".


I'm sad that we need to resort to such acrobatics to test against versions of scylladb before and after a breaking change. I hope that the migration to a workflow engine based CI (+ introduction of testing against a matrix of ScyllaDB versions) will straighten it out.

docs/source/deploy-scylladb/before-you-deploy/configure-operator.md
```diff
@@ -44,6 +45,9 @@ status:
 * - `scyllaUtilsImage`
   - ScyllaDB image used for running utility scripts (perftune, sysctl). Determines which tuning scripts are used for performance optimization.
   - Latest ScyllaDB image.
+* - `scyllaDBNodeExporterImage`
+  - scylladb-node-exporter image used to expose node-level OS metrics for ScyllaDB clusters running version 2026.3 or later.
```
mflendrich:
```suggestion
  - `scylladb-node-exporter` image that ScyllaDB Operator runs as a sidecar to expose node-level OS metrics for ScyllaDB clusters. For versions before 2026.3.0 this field is ineffective (as node-exporter is bundled and run directly within `scyllaDBImage`)
```

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@
 
 ## Unreleased
 
+### Features & Enhancements
```
mflendrich: This is not a feature or enhancement imo - it's a breaking change / maintenance task. I'd move it to `Breaking Changes`.

rzetelskik: It's not really a breaking change from the user perspective though? Also our release notes generator skill only lists [these categories](https://github.com/scylladb/scylla-operator/blob/876554a6ae7e062c1a561f52a2ebde9d7a5c94ed/.opencode/skills/release-notes-generator/SKILL.md?plain=1#L26-L33) - do you think it's missing a category, or does it fit into one of the existing ones?

docs/source/deploy-scylladb/before-you-deploy/configure-operator.md
rzetelskik: done

assets/config/config.yaml
rzetelskik: done

> I'm sad that we need to resort to such acrobatics to test against versions of scylladb before and after a breaking change. I hope that the migration to a workflow engine based CI (+ introduction of testing against a matrix of ScyllaDB versions) will straighten it out.

Not by design imo. There's nothing prohibiting us from adding a job that will run a selected test suite against a specific version (although we don't have a nice mechanism for testing a version matrix atm). But I'm not sure I'd like to run an entire test suite against <2026.3 for two years just to test this specific code path. It's a dilemma we'll face regardless of the CI engine we use.

CHANGELOG.md
mflendrich: In that case, we're missing a category for stuff that is "none of the above". I am fairly positive that this isn't a feature or enhancement.

> It's not really a breaking change from the user perspective though?

By design - it's not. But the tech doctrine says that the reality doesn't care: https://www.hyrumslaw.com/

Maybe we just need "Other changes"?
suggestion: create a category that matches your taste in the changelog and expand the skill to allow it.

assets/config/config.yaml
mflendrich: > But I'm not sure I'd like to run an entire test suite against <2026.3 for two years just to test this specific code path.

The test suite relevant to node-exporter is just a few test cases, no?

> It's a dilemma we'll face regardless of the CI engine we use.

Yes - but a better CI engine makes new suites much cheaper than they're today.
My personal taste is to have multiple small suites over one that is bloated with custom exceptions, but it's just an engineering taste and I don't intend to spend our precious time pushing it in this PR.

(no further action required)

rzetelskik: > The test suite relevant to node-exporter is just a few test cases, no?

Yes. I can add a separate, small suite that we'll run with <2026.3 (and pre-release 2026.3.0 for now). There's a bit of bloat with defining new jobs, yes, but it's not terrible. If you think it's a cleaner approach, we can take it.

CHANGELOG.md
rzetelskik: moved to "Other changes", amended the skill

assets/config/config.yaml
mflendrich: I'm inclined to at least try it out.

And I'm rather convinced that this problem will not exist with parameterized builds. So the cost of having an intermediate phase of new jobs copy-pasted feels low to me.

mflendrich (approved):
/lgtm

mflendrich (approved):
/lgtm

rzetelskik:
> 
> ci/prow/e2e-oke-parallel	[9a11ba6](https://github.com/scylladb/scylla-operator/commit/9a11ba6ff19ca9b2691970890aa51a760b0ffe6a)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3482/pull-scylla-operator-master-e2e-oke-parallel/2074428222669328384)	false	`/test e2e-oke-parallel`
> ci/prow/e2e-gke-parallel-clusterip	[40116dc](https://github.com/scylladb/scylla-operator/commit/40116dc45f3d04664642c162ee86f74c434abf40)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3482/pull-scylla-operator-master-e2e-gke-parallel-clusterip/2078112013707382784)	unknown	`/test e2e-gke-parallel-clusterip`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3482). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).
> 
> Details

unrelated
/retest

PR #3484, opened 2026-07-02
feat(manager): start agent after scylla API is ready

abramche:
- make SM agent wait until after ScyllaDB API to become available
- adjust the test accordingly

Resolves https://scylladb.atlassian.net/browse/OPERATOR-199

### DIFF
--- a/pkg/controller/scylladbdatacenter/resource.go
+++ b/pkg/controller/scylladbdatacenter/resource.go
@@ -1371,7 +1371,7 @@ func getScyllaDBManagerAgentContainer(r scyllav1alpha1.RackSpec, sdc *scyllav1al
 		Name:            naming.ScyllaManagerAgentContainerName,
 		Image:           *sdc.Spec.ScyllaDBManagerAgent.Image,
 		ImagePullPolicy: corev1.PullIfNotPresent,
-		// There is no point in starting scylla-manager before ScyllaDB is tuned and ignited. The manager agent fails after 60 attempts and hits backoff unnecessarily.
+		// There is no point in starting ScyllaDB Manager Agent before ScyllaDB is ignited and tuned. The manager agent fails after 60 attempts and hits backoff unnecessarily.
 		Command: []string{
 			"/usr/bin/bash",
 			"-euEo",
@@ -1387,7 +1387,13 @@ until [[ -f "/mnt/shared/ignition.done" ]]; do
   sleep 1 &
   wait
 done
-printf '{"L":"INFO","T":"%s","M":"Ignited. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+printf '{"L":"INFO","T":"%s","M":"Ignited. Waiting for ScyllaDB API to become available"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+
+until curl --connect-timeout 5 --max-time 10 -s -o /dev/null -w "" "http://localhost:` + strconv.Itoa(naming.ScyllaAPIPort) + `/"; do
+  sleep 1 &
+  wait
+done
+printf '{"L":"INFO","T":"%s","M":"ScyllaDB API is available. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
 
 exec scylla-manager-agent \
 -c ` + fmt.Sprintf("%q ", naming.ScyllaAgentConfigDefaultFile) + `\
--- a/pkg/controller/scylladbdatacenter/resource_test.go
+++ b/pkg/controller/scylladbdatacenter/resource_test.go
@@ -5,6 +5,7 @@ import (
 	"maps"
 	"reflect"
 	"slices"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -1284,7 +1285,13 @@ until [[ -f "/mnt/shared/ignition.done" ]]; do
   sleep 1 &
   wait
 done
-printf '{"L":"INFO","T":"%s","M":"Ignited. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+printf '{"L":"INFO","T":"%s","M":"Ignited. Waiting for ScyllaDB API to become available"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+
+until curl --connect-timeout 5 --max-time 10 -s -o /dev/null -w "" "http://localhost:` + strconv.Itoa(naming.ScyllaAPIPort) + `/"; do
+  sleep 1 &
+  wait
+done
+printf '{"L":"INFO","T":"%s","M":"ScyllaDB API is available. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
 
 exec scylla-manager-agent \
 -c "/etc/scylla-manager-agent/scylla-manager-agent.yaml" \
--- a/docs/source/understand/ignition.md
+++ b/docs/source/understand/ignition.md
@@ -19,7 +19,7 @@ Ignition uses a simple file-based signal on the shared emptyDir volume mounted a
 1. The `scylladb-ignition` sidecar container runs the ignition controller, which continuously evaluates prerequisites.
 2. When **all** prerequisites are met, the controller creates the file `/mnt/shared/ignition.done`.
 3. The `scylla` container's entrypoint polls for this file in a shell loop. Once the file appears, it execs into the sidecar binary that configures and starts ScyllaDB.
-4. The ScyllaDB Manager Agent container (when present) uses the same wait loop, ensuring the agent does not start before ScyllaDB is configured.
+4. The ScyllaDB Manager Agent container (when present) uses the same wait loop for ignition, and additionally polls the ScyllaDB REST API (port 10000) to wait for ScyllaDB to finish IO tuning and become available before starting.
 
 ## Prerequisites evaluated
 
--- a/docs/source/understand/manager.md
+++ b/docs/source/understand/manager.md
@@ -24,7 +24,7 @@ The Agent communicates with Manager to execute operations on the local node (str
 
 The Agent:
 - Listens on port 10001.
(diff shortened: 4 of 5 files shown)


mflendrich (commented):
copilot comments make sense to me. otherwise lgtm

Low-key it might make sense to check if there's opportunity to replace 127.0.0.1 with localhost in other places as well, but only if doable with a 10 minute timebox.

abramche:
> copilot comments make sense to me. otherwise lgtm
> 
> Low-key it might make sense to check if there's opportunity to replace 127.0.0.1 with localhost in other places as well, but only if doable with a 10 minute timebox.

I agree with the documentation comments, adding to the batch.

Regarding the curl timeouts - it is already the same (unhandled) risk with ignition - no timeouts there either. What I propose instead, is going for a kubernetes native way and add a startup probe to the container (10s poll / 10min timeout -- `test/e2e/utils/config.go` defines "memberRolloutTimeout = 30s + imagePullTimeout(4m) + joinClusterTimeout(3m) ≈ 7.5 minutes", 10min seem reasonable buffer).

As for `localhost`, it is unconventional to our repo (the literal is only used in the machinery scripts), if you agree, I'll replace with an address resolution closure (= `sdc.Spec.GetIPFamily()` -> `::1/127.0.0.1`) similar to what is already widely used elsewhere (i.e. `pkg/controller/scylladbdatacenter/resource.go:733`)

WDYT?

mflendrich (commented):
> > copilot comments make sense to me. otherwise lgtm
> > Low-key it might make sense to check if there's opportunity to replace 127.0.0.1 with localhost in other places as well, but only if doable with a 10 minute timebox.
> 
> I agree with the documentation comments, adding to the batch.

I do not see any new commits.

> Regarding the curl timeouts - it is already the same (unhandled) risk with ignition - no timeouts there either. 

Ignition synchronization is based on a file. The phenomenon that _stalled local connection could hang the loop indefinitely_ does not apply there.

> What I propose instead, is going for a kubernetes native way and add a startup probe to the container (10s poll / 10min timeout -- `test/e2e/utils/config.go` defines "memberRolloutTimeout = 30s + imagePullTimeout(4m) + joinClusterTimeout(3m) ≈ 7.5 minutes", 10min seem reasonable buffer).

This might be an interesting idea but cutting too deep for such a small change (would require a design doc because it might have second-order implications and may be breaking) and will distract us again from the E2E work we're trying to complete. Let's stick to curl to unblock the tests, and maybe file an idea issue (possibly under the [Operator 2.0 epic](https://scylladb.atlassian.net/browse/OPERATOR-153)).

> As for `localhost`, it is unconventional to our repo (the literal is only used in the machinery scripts), if you agree, I'll replace with an address resolution closure (= `sdc.Spec.GetIPFamily()` -> `::1/127.0.0.1`) similar to what is already widely used elsewhere (i.e. `pkg/controller/scylladbdatacenter/resource.go:733`)

Using a DNS name for `localhost` is unconventional? Why would a convention matter here? What's wrong with localhost?

abramche:
> Ignition synchronization is based on a file. The phenomenon that _stalled local connection could hang the loop indefinitely_ does not apply there.

If ignition stalls for whatever reason (ScyllaDB bug or anything), the file is not created -> we wait indefinitely in the same fashion.

> This might be an interesting idea but cutting too deep for such a small change (would require a design doc because it might have second-order implications and may be breaking) and will distract us again from the E2E work we're trying to complete. Let's stick to curl to unblock the tests, and maybe file an idea issue (possibly under the [Operator 2.0 epic](https://scylladb.atlassian.net/browse/OPERATOR-153)).

in my view, a startup probe is a better, elegant long-term solution addressing the existing obscure problem - but I'm ok with the low hanging fruit. I'll go with curl.

> > As for `localhost`, it is unconventional to our repo (the literal is only used in the machinery scripts), if you agree, I'll replace with an address resolution closure (= `sdc.Spec.GetIPFamily()` -> `::1/127.0.0.1`) similar to what is already widely used elsewhere (i.e. `pkg/controller/scylladbdatacenter/resource.go:733`)
> Using a DNS name for `localhost` is unconventional? Why would a convention matter here? What's wrong with localhost?

I was writing a comment on IP v4/v6 [addresses](https://github.com/scylladb/scylla-operator/blob/9da4bb76dbce5f2e979dd97c2583bbff6e342ec6/pkg/cmd/operator/probeserver/scylladbapistatus.go#L53) [being](https://github.com/scylladb/scylla-operator/blob/9da4bb76dbce5f2e979dd97c2583bbff6e342ec6/pkg/cmd/operator/sidecar/sidecar.go#L174) more commonly used, but then I also spotted a test [usage](https://github.com/scylladb/scylla-operator/blob/7d56acaf855d858f368bccac2d54cd37f590952a/test/e2e/set/scyllacluster/scyllacluster_restarts.go#L85) of `localhost` as well. I'll leave the unification for another day...

mflendrich:
> > Ignition synchronization is based on a file. The phenomenon that _stalled local connection could hang the loop indefinitely_ does not apply there.
> 
> If ignition stalls for whatever reason (ScyllaDB bug or anything), the file is not created -> we wait indefinitely in the same fashion.

There's an important difference - "stalled connection" means that a single connection attempt stalls indefinitely (and a possible future agent change from unavailable to available goes unnoticed). This is not the case with checking if a file is present.

Anyway - looks like you have a way to proceed.

rzetelskik:
> > What I propose instead, is going for a kubernetes native way and add a startup probe to the container (10s poll / 10min timeout -- test/e2e/utils/config.go defines "memberRolloutTimeout = 30s + imagePullTimeout(4m) + joinClusterTimeout(3m) ≈ 7.5 minutes", 10min seem reasonable buffer).
> 
> This might be an interesting idea but cutting too deep for such a small change (would require a design doc because it might have second-order implications and may be breaking) and will distract us again from the E2E work we're trying to complete. Let's stick to curl to unblock the tests, and maybe file an idea issue (possibly under the [Operator 2.0 epic](https://scylladb.atlassian.net/browse/OPERATOR-153)).

~Startup probe is not there to enforce timeouts, but to give slow-starting apps some wiggle room and not bloat the liveness probe at the same time. Btw the node does not need to be a part of the cluster to be considered healthy - we just verify the API server is up. The image pull also happens before the container is created. The only reason to consider a startup probe imo would be to account for things like the ignition (and the whole sidecar wrapper), but I'd much rather we made it an actual sidecar and moved it to initContainers.~

Edit: I just noticed you're referring to the manager agent container, not scylla, discard the above. Imo the root of the issue here is that manager agent doesn't fit the sidecar model. The startup probe is still just going to be a workaround. Maybe we should think about if we really need it to restart when it can't reach scylla? Are there any scenarios where that's going to happen and scylla container itself won't propagate any issues?

abramche:
> Maybe we should think about if we really need it to restart when it can't reach scylla? Are there any scenarios where that's going to happen and scylla container itself won't propagate any issues?

I do not know, potentially yes (defect being a notable candidate)? The current approach certainly allows for this. But then we need to have ScyllaDB SLAs and it's a tough topic (with repairs, restores etc... it can wildly differ)

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -1387,7 +1387,12 @@ until [[ -f "/mnt/shared/ignition.done" ]]; do
   sleep 1 &
   wait
 done
-printf '{"L":"INFO","T":"%s","M":"Ignited. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+printf '{"L":"INFO","T":"%s","M":"Ignited. Waiting for ScyllaDB API to become ready"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+until curl --silent --fail --output /dev/null http://127.0.0.1:8080/readyz; do
```
rzetelskik: /readyz is stronger than what the comment says (ScyllaDB serving api), is that deliberate? This makes it closer to the hallucinated documentation (`it does not run until ScyllaDB itself is ready`), but I'm not sure that's correct (i.e. I'm not sure the agent should necessarily wait for a node to be UN and have native transport enabled), and if so, it should be explicitly stated in the comment.

abramche: @rzetelskik oh, good catch! No, that's not needed - a lot of overconstraining - I believe we could do with `/healthz` which should also check API, but not the follow-up steps like CQL interface coming up.

rzetelskik: Bear in mind it has side effects related to the maintenance mode. Why not make a direct API call?
Let's also adjust the docs

rzetelskik:
/test all
??

abramche:
Copilot suggestions contradict earlier agreements (plus curl is available in the image - we are not changing it and if we do, we need to test anyway, there are other dependencies in play as well).

Any further feedback?

pkg/controller/scylladbdatacenter/resource_test.go
```diff
@@ -1284,7 +1284,13 @@ until [[ -f "/mnt/shared/ignition.done" ]]; do
   sleep 1 &
   wait
 done
-printf '{"L":"INFO","T":"%s","M":"Ignited. Starting ScyllaDB Manager Agent"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+printf '{"L":"INFO","T":"%s","M":"Ignited. Waiting for ScyllaDB API to become available"}\n' "$( date -u '+%Y-%m-%dT%H:%M:%S,%3NZ' )" > /dev/stderr
+
+until curl --connect-timeout 5 --max-time 10 -s -o /dev/null -w "" "http://localhost:10000/"; do
```
mflendrich: why 10000 hardcoded here as opposed to `naming.ScyllaAPIPort`?

this makes this test a change detector

mflendrich (commented):
last comment from my end, otherwise lgtm

pkg/controller/scylladbdatacenter/resource_test.go
abramche: oops, I already did it in `resource.go` but somehow missed here, good catch

addressed in 85438f0eabfe6aeba2c52d235fa661af6216e045

I am also seeing in the iotune test results that this change makes agent not restart prematurely, so it achieves its purpose

rzetelskik (approved):
lgtm
/assign mflendrich

abramche:
/test images
/retest

```
+ kubectl -n ci-clusters create -f -
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.ci.scylladb.com": failed to call webhook: Post "[https://webhook-server.ci-controller-manager.svc:443/validate?timeout=10s](https://webhook-server.ci-controller-manager.svc/validate?timeout=10s)": no endpoints available for service "webhook-server"
```

mflendrich (approved):
/lgtm

PR #3488, opened 2026-07-06
test: introduce the helm install operator test

abramche:
- add the new Helm install test GH Actions

Resolves https://scylladb.atlassian.net/browse/OPERATOR-158

/kind new-test
/priority important-soon

### DIFF
--- a/.github/actions/setup-kind/action.yml
+++ b/.github/actions/setup-kind/action.yml
@@ -10,24 +10,11 @@ outputs:
 runs:
   using: composite
   steps:
-    - name: Write KinD cluster config
-      shell: bash
-      run: |
-        cat > /tmp/kind-config.yaml <<'EOF'
-        kind: Cluster
-        apiVersion: kind.x-k8s.io/v1alpha4
-        networking:
-          ipFamily: ipv4
-        nodes:
-          - role: control-plane
-          - role: worker
-        EOF
-
     - name: Create KinD cluster
       uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc
       with:
         cluster_name: e2e
-        config: /tmp/kind-config.yaml
+        config: hack/kind/cluster-config.yaml
 
     - name: Deploy cert-manager
       shell: bash
--- a/.github/actions/setup-kind/action.yml
+++ b/.github/actions/setup-kind/action.yml
@@ -7,6 +7,9 @@ outputs:
   image-ref:
     description: The operator image reference loaded into the cluster.
     value: ${{ steps.meta.outputs.image-ref }}
+  cluster-name:
+    description: The KinD cluster name used for this run.
+    value: ${{ steps.meta.outputs.cluster-name }}
 
 runs:
   using: composite
@@ -14,7 +17,12 @@ runs:
     - name: Set image metadata
       id: meta
       shell: bash
-      run: echo "image-ref=localhost/scylladb/scylla-operator:e2e" >> "${GITHUB_OUTPUT}"
+      env:
+        PR_NUMBER: ${{ github.event.pull_request.number || 'manual' }}
+      run: |
+        suffix="pr${PR_NUMBER}-run${GITHUB_RUN_ID}-a${GITHUB_RUN_ATTEMPT}"
+        echo "cluster-name=e2e-${suffix}" >> "${GITHUB_OUTPUT}"
+        echo "image-ref=localhost/scylladb/scylla-operator:${suffix}" >> "${GITHUB_OUTPUT}"
 
     - name: Write KinD cluster config
       shell: bash
@@ -30,9 +38,9 @@ runs:
         EOF
 
     - name: Create KinD cluster
-      uses: helm/kind-action@v1
+      uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc
       with:
-        cluster_name: e2e
+        cluster_name: ${{ steps.meta.outputs.cluster-name }}
         config: /tmp/kind-config.yaml
 
     - name: Build operator image
@@ -41,7 +49,7 @@ runs:
 
     - name: Load operator image into KinD
       shell: bash
-      run: kind load docker-image ${{ steps.meta.outputs.image-ref }} --name e2e
+      run: kind load docker-image ${{ steps.meta.outputs.image-ref }} --name ${{ steps.meta.outputs.cluster-name }}
 
     - name: Deploy cert-manager
       shell: bash
--- a/.github/actions/setup-kind/action.yml
+++ b/.github/actions/setup-kind/action.yml
@@ -1,29 +1,15 @@
-name: Setup KinD with ScyllaDB Operator image
+name: Setup KinD cluster
 description: >
-  Creates a KinD cluster, builds the operator image from source,
-  loads it into the cluster, and deploys cert-manager.
+  Creates a KinD cluster and deploys cert-manager.
 
 outputs:
-  image-ref:
-    description: The operator image reference loaded into the cluster.
-    value: ${{ steps.meta.outputs.image-ref }}
   cluster-name:
-    description: The KinD cluster name used for this run.
-    value: ${{ steps.meta.outputs.cluster-name }}
+    description: The KinD cluster name.
+    value: e2e
 
 runs:
   using: composite
   steps:
-    - name: Set image metadata
-      id: meta
-      shell: bash
-      env:
-        PR_NUMBER: ${{ github.event.pull_request.number || 'manual' }}
-      run: |
-        suffix="pr${PR_NUMBER}-run${GITHUB_RUN_ID}-a${GITHUB_RUN_ATTEMPT}"
-        echo "cluster-name=e2e-${suffix}" >> "${GITHUB_OUTPUT}"
-        echo "image-ref=localhost/scylladb/scylla-operator:${suffix}" >> "${GITHUB_OUTPUT}"
-
     - name: Write KinD cluster config
       shell: bash
       run: |
@@ -40,16 +26,8 @@ runs:
     - name: Create KinD cluster
       uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc
       with:
-        cluster_name: ${{ steps.meta.outputs.cluster-name }}
+        cluster_name: e2e
         config: /tmp/kind-config.yaml
-
-    - name: Build operator image
-      shell: bash
-      run: docker build . -t ${{ steps.meta.outputs.image-ref }}
-
-    - name: Load operator image into KinD
-      shell: bash
-      run: kind load docker-image ${{ steps.meta.outputs.image-ref }} --name ${{ steps.meta.outputs.cluster-name }}
 
     - name: Deploy cert-manager
       shell: bash
--- a/.github/actions/setup-kind/action.yml
+++ b/.github/actions/setup-kind/action.yml
@@ -1,0 +1,61 @@
+name: Setup KinD with ScyllaDB Operator image
+description: >
+  Creates a KinD cluster, builds the operator image from source,
+  loads it into the cluster, and deploys cert-manager.
+
+outputs:
+  image-ref:
+    description: The operator image reference loaded into the cluster.
+    value: ${{ steps.meta.outputs.image-ref }}
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-kind-helm.yaml
```diff
@@ -0,0 +1,138 @@
...
+        run: |
+          pod="$(kubectl -n scylla get pods -l "scylla/cluster=scylla" -o jsonpath='{.items[0].metadata.name}')"
+          echo "Running CQL query against pod ${pod}..."
+
+          kubectl -n scylla exec "${pod}" -c scylla -- \
+            cqlsh localhost -e "SELECT key, cluster_name, data_center FROM system.local"
+
+          echo "CQL connectivity verified."
+
+      - name: Collect debug info
```
mflendrich: let's make this a must-gather uploaded as an artifact

mflendrich: (only in the event of failure)

abramche: addressed in e9eeaecc6173f9f0393b4d89a44e67708a5fb44c
you can see debug-less result in this PR and the simulated failure result in https://github.com/scylladb/scylla-operator/actions/runs/29096738077/job/86375052843?pr=3495 (same change, but using `always()` as a condition instead of `failure()`)

.github/actions/setup-kind/action.yml
```diff
@@ -0,0 +1,69 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Set image metadata
+      id: meta
+      shell: bash
+      env:
+        PR_NUMBER: ${{ github.event.pull_request.number || 'manual' }}
+      run: |
```
mflendrich: 1. Is creating a custom image tag really necessary? Since the image is not being uploaded anywhere, why not keep it simpler by having a constant image tag - the same across runs?
2. If the workflow runs outside of a PR, the result is going to be `manual` while it's not always the case that it will be a manual run.

.github/actions/setup-kind/action.yml
```diff
@@ -0,0 +1,69 @@
...
+      uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc
+      with:
+        cluster_name: ${{ steps.meta.outputs.cluster-name }}
+        config: /tmp/kind-config.yaml
+
+    - name: Build operator image
+      shell: bash
+      run: docker build . -t ${{ steps.meta.outputs.image-ref }}
+
+    - name: Load operator image into KinD
```
mflendrich: For maintainability - I suggest that the image build becomes a separate workflow (with a [possibly output] parameter of what the name of the image is), so that the test workflow becomes:
- build the image
- (in parallel) provision kind
- when both complete: upload the image to kind
- run the tests

.github/workflows/e2e-kind-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+        uses: ./.github/actions/setup-kind
+
+      - name: Deploy operator (Helm)
+        env:
+          IMAGE_REF: ${{ steps.setup.outputs.image-ref }}
+        run: |
+          # Extract repository and tag from the full image reference.
+          # "localhost/scylladb/scylla-operator:pr123-run456-a1" -> repo="localhost/scylladb", tag="pr123-run456-a1"
+          repo="${IMAGE_REF%/*}"
+          tag="${IMAGE_REF##*:}"
```
mflendrich: code quality hint: these perhaps could be output parameters of the image build  (or image push to kind) step

.github/workflows/e2e-kind-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+
+          helm install scylla-operator helm/scylla-operator \
+            --create-namespace \
+            --namespace scylla-operator \
+            --set "image.repository=${repo}" \
+            --set "image.tag=${tag}" \
+            --set replicas=1 \
+            --set webhookServerReplicas=1
```
mflendrich: why need to be inconsistent with the helm install for the scylla chart below? which uses a heredoc values.yaml instead of a multitude of `--set`s?

.github/workflows/e2e-kind-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+            crd/scyllaclusters.scylla.scylladb.com
+
+          # Wait for rollout.
+          kubectl -n scylla-operator rollout status --timeout=5m deployment/scylla-operator
+          kubectl -n scylla-operator rollout status --timeout=5m deployment/webhook-server
+
+      - name: Deploy ScyllaDB cluster (Helm)
+        run: |
+          SCYLLADB_VERSION="$(grep -m1 '^  scyllaDBVersion:' assets/config/config.yaml | awk -F'"' '{print $2}')"
+          SCYLLADB_MANAGER_AGENT_VERSION="$(grep -m1 '^  scyllaDBManagerAgentVersion:' assets/config/config.yaml | awk -F'"' '{print $2}')"
```
mflendrich: code quality hint: use yaml-specific tooling to read yaml (e.g. `yq`) instead of mangling raw text

.github/workflows/e2e-kind-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+        run: |
+          pod="$(kubectl -n scylla get pods -l "scylla/cluster=scylla" -o jsonpath='{.items[0].metadata.name}')"
+          echo "Running CQL query against pod ${pod}..."
+
+          kubectl -n scylla exec "${pod}" -c scylla -- \
+            cqlsh localhost -e "SELECT key, cluster_name, data_center FROM system.local"
+
+          echo "CQL connectivity verified."
+
+      - name: Collect must-gather evidence
```
mflendrich, line 91: please make this a reusable sub-workflow - like kind setup and build

.github/workflows/e2e-kind-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+            exit 0
+          fi
+
+          ARTIFACT_DIR="${RUNNER_TEMP}/must-gather"
+          mkdir -p "${ARTIFACT_DIR}"
+
+          gather_failed=0
+
+          {
+            kubectl create namespace gather-artifacts --dry-run=client -o yaml | kubectl apply -f -
```
mflendrich: running must-gather as a pod in kind looks like overkill. you should be able to directly run it locally, just providing a kind kubeconfig, right?

.github/actions/setup-kind/action.yml
rzetelskik: W already have this wired as a makefile target. I don't really understand the need to reinvent the wheel here.

.github/actions/setup-kind/action.yml
abramche: 1. My original intent was to safeguard against potential run-ins if the action runs concurrently on the same runner. Also, would make it a bit easier to identify in (locally downloaded) must-gather, if we kept the PR number in the image name (and must-gather archive name, by extension). That said, if you feel it just adds unnecessary complexity, I'm not feeling strongly about it.

2. `manual` could have been `custom`

.github/actions/setup-kind/action.yml
abramche: > * build the image
> * (in parallel) provision kind
> * when both complete: upload the image to kind
> * run the tests

that's possible to do in several different ways:
1. Use multiple stages in the same workflow, make 3/"load image" dependent on 1/build and 2/provision. It would introduce extra overhead for uploading/downloading image as artifact.
2. Rely on Linux parallelism. Send `make image` (what Kacper suggested) into background with `&`, record PID, run kind setup, wait for make image process to finish, load image. No cross-stage/artifact overhead, but will be slower and more memory-hungry.
3. Separate workflows or external registry. Unreasonably difficult for this simple action.

.github/workflows/e2e-kind-helm.yaml
abramche: imo it reads (and edits) better than heredoc, also less prone to indentation breakage

.github/workflows/e2e-kind-helm.yaml
abramche: my original intent was to make it independent on `yq` being available on the runner, but I just checked, it's installed on `ubuntu-latest` by default, so no objections

.github/actions/setup-kind/action.yml
mflendrich: > potential run-ins if the action runs concurrently on the same runner

This does not happen - GHA runners are isolated in this fashion. And even if it were the case, then this mechanism would be incorrect because it wouldn't provide non-colliding names in the manual aka custom case.

I am strongly convinced that a stable image tag makes sense because it avoids introducing accidental complexity (of a need to deal with nondeterministic image tags). Please make it so. We can discuss tags that are guaranteed to be unique across runs if you have a good reason to do so, but I am near certain that it's not necessary.

.github/actions/setup-kind/action.yml
mflendrich: > W already have this wired as a makefile target. I don't really understand the need to reinvent the wheel here.

discussed offline - we'll make it so in a separate PR.

> that's possible to do in several different ways

Yes - please make a decision.

.github/workflows/e2e-kind-helm.yaml
mflendrich: I don't really mind which option you choose - I care that you are consistent about the choice.

mflendrich:
@abramche from what I can see - the comments have not been addressed? Was a push accidentally omitted?

Please clearly say which comments you have applied (with a simple statement like "Fixed" / "Done") or say so clearly if pushing back.

.github/actions/setup-kind/action.yml
abramche: > discussed offline - we'll make it so in a separate PR.

Kacper refers to `make image` which does the same - it's simple enough to be included in this PR.

> Use multiple stages in the same workflow, make 3/"load image" dependent on 1/build and 2/provision. It would introduce extra overhead for uploading/downloading image as artifact.

I'll pick this approach, as the other one runs on the same runner and performance gain will be questionable (and then why even bother).

rzetelskik: > Kacper refers to make image which does the same - it's simple enough to be included in this PR.

I wasn't referring to `make image` but to duplicating the entire `make kind-setup`, which handles kind cluster creation, image building and loading into the registry, and imo we should reuse these steps here - not sure if and how `make image` fits into it really. I'm staying out of this PR as deduplication is meant to be handled separately. :grin:

.github/workflows/e2e-kind-helm.yaml
abramche: this step is not actually needed (unless we already plan to deploy from custom repositories) - the values in helm/scylla/values.yaml are resolved to the same as in assets/config/config.yaml

dropped in 052aab2b2ffbfaa235cda8e5e5d0e6784fbcb56e altogether

.github/workflows/e2e-kind-helm.yaml
abramche: dropped heredoc in 052aab2b2ffbfaa235cda8e5e5d0e6784fbcb56e, read defaults from Helm, overridden only where needed with `--set`

.github/workflows/e2e-kind-helm.yaml
abramche: image build, kind-setup run concurrently (052aab2b2ffbfaa235cda8e5e5d0e6784fbcb56e) and load image waits on both to complete

.github/workflows/e2e-kind-helm.yaml
abramche: running with `go run ...` in 052aab2b2ffbfaa235cda8e5e5d0e6784fbcb56e

.github/actions/setup-kind/action.yml
abramche: the image ref is uniform now

.github/actions/setup-kind/action.yml
abramche: ~~image build is now concurrently run with kind-setup~~

serial again, after a discussion in Slack

mflendrich (approved):
/lgtm

abramche:
rebased on master (fixed documentation)

mflendrich (approved):
/lgtm

PR #3497, opened 2026-07-10
test: update spec during operator outage

abramche:
- introduce a new test to ensure Kubernetes state remains source of truth after operator outage

Resolves https://scylladb.atlassian.net/browse/OPERATOR-160

/kind new-test
/priority important-soon

### DIFF
--- a/pkg/controllerhelpers/apps.go
+++ b/pkg/controllerhelpers/apps.go
@@ -3,7 +3,9 @@ package controllerhelpers
 import (
 	"fmt"
 
+	oslices "github.com/scylladb/scylla-operator/pkg/helpers/slices"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 )
 
@@ -38,6 +40,44 @@ func IsStatefulSetRolledOut(sts *appsv1.StatefulSet) (bool, error) {
 	} else {
 		return sts.Status.UpdateRevision == sts.Status.CurrentRevision, nil
 	}
+}
+
+func IsDeploymentRolledOut(deploy *appsv1.Deployment) (bool, error) {
+	if deploy.Spec.Replicas == nil {
+		return false, fmt.Errorf("deployment.spec.replicas can't be nil")
+	}
+
+	if deploy.Status.ObservedGeneration == 0 || deploy.Generation > deploy.Status.ObservedGeneration {
+		klog.V(4).InfoS("Observed generation not caught up yet", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
+	if deploy.Status.UpdatedReplicas < *deploy.Spec.Replicas {
+		klog.V(4).InfoS("Not all replicas are updated yet", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
+	if deploy.Status.ReadyReplicas < *deploy.Spec.Replicas {
+		klog.V(4).InfoS("Not all replicas are ready yet", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
+	if deploy.Status.AvailableReplicas < *deploy.Spec.Replicas {
+		klog.V(4).InfoS("Not all replicas are available yet", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
+	_, _, availableConditionTrue := oslices.Find(deploy.Status.Conditions, func(condition appsv1.DeploymentCondition) bool {
+		return condition.Type == appsv1.DeploymentAvailable && condition.Status == corev1.ConditionTrue
+	})
+	if !availableConditionTrue {
+		klog.V(4).InfoS("Available condition is not True", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
+	klog.V(4).InfoS("Fully rolled out", "Deployment", klog.KObj(deploy))
+
+	return true, nil
 }
 
 func IsDaemonSetRolledOut(ds *appsv1.DaemonSet) (bool, error) {
--- a/test/e2e/set/scyllacluster/scyllacluster_outage.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_outage.go
@@ -1,0 +1,132 @@
+// Copyright (C) 2026 ScyllaDB
+
+package scyllacluster
+
+import (
+	"context"
+	"fmt"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	scyllav1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1"
+	scyllav1client "github.com/scylladb/scylla-operator/pkg/client/scylla/clientset/versioned/typed/scylla/v1"
+	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
+	"github.com/scylladb/scylla-operator/pkg/naming"
+	"github.com/scylladb/scylla-operator/test/e2e/framework"
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
+	scyllaclusterverification "github.com/scylladb/scylla-operator/test/e2e/utils/verification/scyllacluster"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+)
+
+const outageAnnotationKey = "e2e.scylla-operator.scylladb.com/outage-test"
+
+func setOperatorDeploymentReplicasAndWaitForRollout(ctx context.Context, deploymentsClient appsv1client.DeploymentInterface, replicas int32) {
+	framework.By("Setting %s deployment replicas to %d", operatorDeploymentName, replicas)
+	_, err := deploymentsClient.Patch(
+		ctx,
+		operatorDeploymentName,
+		types.MergePatchType,
+		[]byte(fmt.Sprintf(`{"spec":{"replicas":%d}}`, replicas)),
+		metav1.PatchOptions{},
+	)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	framework.By("Waiting for %s deployment to roll out", operatorDeploymentName)
+	_, err = controllerhelpers.WaitForDeploymentState(ctx, deploymentsClient, operatorDeploymentName, controllerhelpers.WaitForStateOptions{}, controllerhelpers.IsDeploymentRolledOut)
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+func updateClusterPodAnnotations(ctx context.Context, scyllaClient scyllav1client.ScyllaClusterInterface, sc *scyllav1.ScyllaCluster, key, value string) *scyllav1.ScyllaCluster {
+	sc, err := scyllaClient.Get(ctx, sc.Name, metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	scCopy := sc.DeepCopy()
+	if scCopy.Spec.PodMetadata == nil {
+		scCopy.Spec.PodMetadata = &scyllav1.ObjectTemplateMetadata{}
+	}
+	if scCopy.Spec.PodMetadata.Annotations == nil {
+		scCopy.Spec.PodMetadata.Annotations = map[string]string{}
+	}
+	scCopy.Spec.PodMetadata.Annotations[key] = value
+
+	patch, err := controllerhelpers.GenerateMergePatch(sc, scCopy)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	sc, err = scyllaClient.Patch(ctx, sc.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(sc.Spec.PodMetadata).NotTo(o.BeNil())
+	o.Expect(sc.Spec.PodMetadata.Annotations).To(o.HaveKeyWithValue(key, value))
+
+	return sc
+}
+
+var _ = g.Describe("ScyllaCluster", framework.SuiteSerial, func() {
+	var f *framework.Framework
+
+	g.BeforeEach(func(ctx context.Context) {
+		f = framework.NewFramework(ctx, "scyllacluster")
+	})
+
+	g.It("should reconcile spec changes after operator outage", func(ctx g.SpecContext) {
+		sc := f.GetDefaultScyllaCluster()
+		o.Expect(sc.Spec.Datacenter.Racks).To(o.HaveLen(1))
+		sc.Spec.Datacenter.Racks[0].Members = 1
+
+		framework.By("Creating a ScyllaCluster")
+		var err error
+		sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
+		initialRolloutCtx, initialRolloutCtxCancel := utils.ContextForRollout(ctx, sc)
+		defer initialRolloutCtxCancel()
+		sc, err = controllerhelpers.WaitForScyllaClusterState(initialRolloutCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		scyllaclusterverification.Verify(ctx, f.KubeClient(), f.ScyllaClient(), sc)
+		scyllaclusterverification.WaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		framework.By("Saving the original operator replica count")
+		operatorDeploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, operatorDeploymentName, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		originalOperatorReplicas := *operatorDeploy.Spec.Replicas
+		o.Expect(originalOperatorReplicas).To(o.BeNumerically(">", 0))
+
+		operatorDeployments := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace)
+
+		g.DeferCleanup(func(ctx context.Context) {
+			setOperatorDeploymentReplicasAndWaitForRollout(ctx, operatorDeployments, originalOperatorReplicas)
+		})
+
+		setOperatorDeploymentReplicasAndWaitForRollout(ctx, operatorDeployments, 0)
+
+		framework.By("Updating ScyllaCluster spec.podMetadata.annotations while operator is down")
+		outageAnnotationValue := sc.ResourceVersion
+		sc = updateClusterPodAnnotations(ctx, f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()), sc, outageAnnotationKey, outageAnnotationValue)
+
+		framework.By("Restoring the original operator replica count")
+		setOperatorDeploymentReplicasAndWaitForRollout(ctx, operatorDeployments, originalOperatorReplicas)
+
+		framework.By("Waiting for the ScyllaCluster to reconcile the spec change")
+		reconcileCtx, reconcileCtxCancel := utils.ContextForRollout(ctx, sc)
+		defer reconcileCtxCancel()
+		sc, err = controllerhelpers.WaitForScyllaClusterState(reconcileCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Verifying the annotation was propagated to ScyllaCluster pods")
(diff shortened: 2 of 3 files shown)


test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+		sc.Spec.Datacenter.Racks[0].Members = 1
+
+		framework.By("Creating a ScyllaCluster")
+		var err error
+		sc, err = f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()).Create(ctx, sc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for the ScyllaCluster to roll out (RV=%s)", sc.ResourceVersion)
+		waitCtx1, waitCtx1Cancel := utils.ContextForRollout(ctx, sc)
+		defer waitCtx1Cancel()
```
rzetelskik: please use meaningful variable names and avoid this ordering - if you wanted to add a context before this one and maintain this ordering, you'd have to rename all the existing ones

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+			)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			framework.By("Waiting for operator pods to become Ready")
+			o.Eventually(func(eg o.Gomega) {
+				deploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, operatorDeploymentName, metav1.GetOptions{})
+				eg.Expect(err).NotTo(o.HaveOccurred())
+				eg.Expect(deploy.Status.ReadyReplicas).To(o.Equal(originalOperatorReplicas))
+				eg.Expect(deploy.Status.AvailableReplicas).To(o.Equal(originalOperatorReplicas))
+			}).WithContext(ctx).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(o.Succeed())
```
rzetelskik: Use WaitForObjectState to avoid polling - see e.g. WaitForScyllaOperatorConfigState. Polling is not terrible here, but this is also introducing a lot of code repetition.

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for scylla-operator deployment to scale down")
+		o.Eventually(func(eg o.Gomega) {
+			deploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, operatorDeploymentName, metav1.GetOptions{})
+			eg.Expect(err).NotTo(o.HaveOccurred())
+			eg.Expect(deploy.Status.ObservedGeneration).To(o.BeNumerically(">=", deploy.Generation))
+			eg.Expect(deploy.Status.Replicas).To(o.BeZero())
+			eg.Expect(deploy.Status.ReadyReplicas).To(o.BeZero())
+			eg.Expect(deploy.Status.AvailableReplicas).To(o.BeZero())
```
rzetelskik: ditto, you could use a helper as well since there is a lot of code duplication in this test

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for webhook-server deployment rollout to complete")
+		o.Eventually(func(eg o.Gomega) {
+			deploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, webhookServerDeploymentName, metav1.GetOptions{})
+			eg.Expect(err).NotTo(o.HaveOccurred())
+			eg.Expect(deploy.Status.ObservedGeneration).To(o.BeNumerically(">=", deploy.Generation))
+			eg.Expect(deploy.Status.UpdatedReplicas).To(o.Equal(*deploy.Spec.Replicas))
+			eg.Expect(deploy.Status.ReadyReplicas).To(o.Equal(*deploy.Spec.Replicas))
+			eg.Expect(deploy.Status.AvailableReplicas).To(o.Equal(*deploy.Spec.Replicas))
```
rzetelskik: ditto

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+		// Trigger a rollout restart by patching the pod template annotation.
+		restartAnnotation := fmt.Sprintf(`{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/restartedAt":"%s"}}}}}`, time.Now().Format(time.RFC3339))
+		_, err = f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Patch(
+			ctx,
+			webhookServerDeploymentName,
+			types.MergePatchType,
+			[]byte(restartAnnotation),
+			metav1.PatchOptions{},
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
```
rzetelskik: can you explain what that's supposed to test?

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+
+			for _, cond := range deploy.Status.Conditions {
+				if cond.Type == appsv1.DeploymentAvailable {
+					eg.Expect(cond.Status).To(o.Equal(corev1.ConditionTrue))
+				}
+			}
+		}).WithContext(ctx).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(o.Succeed())
+
+		framework.By("Waiting for the ScyllaCluster to reconcile after operator restart")
+		waitCtx2, waitCtx2Cancel := utils.ContextForRollout(ctx, sc)
```
rzetelskik: ditto

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Waiting for scylla-operator pods to become Ready")
+		o.Eventually(func(eg o.Gomega) {
+			deploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, operatorDeploymentName, metav1.GetOptions{})
+			eg.Expect(err).NotTo(o.HaveOccurred())
+			eg.Expect(deploy.Status.ObservedGeneration).To(o.BeNumerically(">=", deploy.Generation))
+			eg.Expect(deploy.Status.UpdatedReplicas).To(o.Equal(*deploy.Spec.Replicas))
+			eg.Expect(deploy.Status.ReadyReplicas).To(o.Equal(*deploy.Spec.Replicas))
+			eg.Expect(deploy.Status.AvailableReplicas).To(o.Equal(*deploy.Spec.Replicas))
```
rzetelskik: ditto

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+				}
+			}
+		}).WithContext(ctx).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(o.Succeed())
+
+		// Ensure the webhook-server pods were actually replaced during the rollout.
+		updatedWebhookDeploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, webhookServerDeploymentName, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(updatedWebhookDeploy.Status.ObservedGeneration).To(o.BeNumerically(">", webhookDeploy.Status.ObservedGeneration))
+
+		framework.By("Updating ScyllaCluster spec.backups again after webhook-server rollout")
```
rzetelskik: again, what's the idea here? You've already updated the spec after the operator was scaled down to zero, I don't see what the webhook-server pod replacement changes

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
```diff
@@ -0,0 +1,224 @@
...
+		defer waitCtx2Cancel()
+		sc, err = controllerhelpers.WaitForScyllaClusterState(waitCtx2, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Verifying the operator does not revert the ScyllaCluster spec to a stale version")
+		o.Expect(sc.Spec.Backups).To(o.HaveLen(1), "spec.backups should not be cleared by the operator")
+		o.Expect(sc.Spec.Backups[0].Name).To(o.Equal("e2e-ha-backup"))
+		o.Expect(*sc.Spec.Backups[0].Cron).To(o.Equal("@daily"), "spec.backups[0].cron should remain @daily")
+		o.Expect(sc.Spec.Backups[0].Retention).To(o.Equal(int64(7)), "spec.backups[0].retention should remain 7")
+	})
```
rzetelskik: imo the more interesting part functionally is to test that the spec change still causes the desired effect after the outage, but for this you should probably use a less convoluted spec change than the manager backup, e.g. some config change that can be verified easily

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: point taken (I'll re-check all other PRs for this as well)

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: the original SCTest restarts webhook while updating the spec to make sure it's not reverted

```
db_cluster.k8s_cluster.kubectl("rollout restart deployment webhook-server", namespace=SCYLLA_OPERATOR_NAMESPACE)
db_cluster.k8s_cluster.kubectl("rollout status deployment webhook-server --watch=true", namespace=SCYLLA_OPERATOR_NAMESPACE)
```

I started going the same way and then changed my mind to update the backup spec instead -- indeed, it does not make sense standalone. So the choices are:

1. Leave this logic. Update the spec while both operator and one of webhook replicas are out (this is what SCT does).
2. Remove this logic, only update the spec while operator is out.

in my view, 1 is more complete, but also bulkier. Which makes more sense to you?

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: The idea is to ensure that no matter what happens, operator/webhook restarts do not revert kubernetes state, I suppose we do not need to carry the original intent from SCT and combine both (update once operator AND either of webhook replicas are down)

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: The original SCT test bumped `fs.aio-max-nr`, I thought the backup spec makes more sense from "real life" POV - though technically either ultimately proves the point.

>  spec change still causes the desired effect after the outage

What do you mean? I see it borderline responsibility between ScyllaDB and Operator; if we assert the config is propagated to the cluster shard successfully, imo we can stop there.

rzetelskik: > The original SCT test bumped fs.aio-max-nr

That's deprecated.

> What do you mean? I see it borderline responsibility between ScyllaDB and Operator

I understand the intent of the test as "ensure that, after the outage has finished, the operator reconciles the desired state introduced during an outage". At the end of the day, SC is just an API. The fact that something propagated to status is desired and we need to verify this, but what really matters is that operator actually reconciled the state - in this case, that it created/updated the task in ScyllaDB Manager state. Which is way too complicated for this sort of change, but you get my point. 

Maybe it doesn't even need to be tested with SC - e.g. propagating to status is the actual reconciliation logic in SOC controller. But I'm sure there are examples in our tests of a simple enough change to SC.

> if we assert the config is propagated to the cluster shard successfully, imo we can stop there.

What do you mean by propagated to the cluster shard?

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: There is no helper for Deployment though (`WaitForObjectState[*appsv1.Deployment...`), other tests did not need this functionality yet - I'll introduce it.

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
rzetelskik: > the original SCTest restarts webhook while updating the spec to make sure it's not reverted

I feel like this is coming from some misconceptions. I don't see any value in this. Even assuming this is supposed to test if our webhook server is HA, it's not only configurable by the user, but it's also borderline testing k8s API.

> in my view, 1 is more complete, but also bulkier. Which makes more sense to you?

In either case, these two feel like separate tests. But I personally don't see any value in the webhook test.

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
rzetelskik: I don't see how any of those could revert kubernetes state, especially how one of two webhook replicas being down could contribute to this. Let's continue in https://github.com/scylladb/scylla-operator/pull/3497#discussion_r3569088316.

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: > What do you mean by propagated to the cluster shard?

That's my mistake, these are "control plane" changes. You said what I actually meant with

> it doesn't even need to be tested with SC - e.g. propagating to status is the actual reconciliation logic in SOC controller

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: webhook behaviour dependency removed (renamed test accordingly) in 86d94bfaa0fb84b7c901925f0037b4cc5e1f2c9a

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: addressed in 86d94bfaa0fb84b7c901925f0037b4cc5e1f2c9a

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: added in f0140dce90fe386b4377a8dcb43d7c07e3d03154

test/e2e/set/scyllacluster/scyllacluster_webhook_ha.go
abramche: changed backups to simply new annotation (a spec change is a spec change ¯\_(ツ)_/¯) in 86d94bfaa0fb84b7c901925f0037b4cc5e1f2c9a

pkg/controllerhelpers/apps.go
```diff
@@ -40,6 +40,36 @@ func IsStatefulSetRolledOut(sts *appsv1.StatefulSet) (bool, error) {
...
+	if deploy.Status.ReadyReplicas < *deploy.Spec.Replicas {
+		klog.V(4).InfoS("Not all replicas are ready yet", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
+	if deploy.Status.AvailableReplicas < *deploy.Spec.Replicas {
+		klog.V(4).InfoS("Not all replicas are available yet", "Deployment", klog.KObj(deploy))
+		return false, nil
+	}
+
```
rzetelskik, line 77: I'd also add
```
_, _, availableConditionTrue := slices.Find(deploy.Status.Conditions, func(condition appsv1.DeploymentCondition) bool {
	return condition.Type == appsv1.DeploymentAvailable && condition.Status == corev1.ConditionTrue
})
if !availableConditionTrue {
	...
}
```

test/e2e/set/scyllacluster/scyllacluster_outage.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/scylladb/scylla-operator/test/e2e/utils"
+	scyllaclusterverification "github.com/scylladb/scylla-operator/test/e2e/utils/verification/scyllacluster"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+)
+
+const outageAnnotationKey = "e2e.scylla-operator.scylladb.com/outage-test"
+
+func setOperatorDeploymentReplicas(ctx context.Context, deploymentsClient appsv1client.DeploymentInterface, replicas int32) {
```
rzetelskik: nit: s/setOperatorDeploymentReplicas/setOperatorDeploymentReplicasAndWaitForRollout

test/e2e/set/scyllacluster/scyllacluster_outage.go
```diff
@@ -0,0 +1,128 @@
...
+
+	sc, err = scyllaClient.Patch(ctx, sc.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(sc.Spec.PodMetadata).NotTo(o.BeNil())
+	o.Expect(sc.Spec.PodMetadata.Annotations).To(o.HaveKeyWithValue(key, value))
+
+	return sc
+}
+
+var _ = g.Describe("ScyllaCluster spec update during operator outage", framework.SuiteSerial, func() {
```
rzetelskik: the test name reads as "ScyllaCluster spec update during operator outage should reconcile spec changes made while the operator was not running", which doesn't make much sense, how about something closer to "ScyllaCluster" + "should reconcile spec after an operator outage" (or whatever, as long as the sentence makes sense)

test/e2e/set/scyllacluster/scyllacluster_outage.go
```diff
@@ -0,0 +1,128 @@
...
+		sc, err = controllerhelpers.WaitForScyllaClusterState(initialRolloutCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		scyllaclusterverification.Verify(ctx, f.KubeClient(), f.ScyllaClient(), sc)
+		scyllaclusterverification.WaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+
+		framework.By("Saving the original operator replica count")
+		operatorDeploy, err := f.KubeAdminClient().AppsV1().Deployments(operatorNamespace).Get(ctx, operatorDeploymentName, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		originalOperatorReplicas := *operatorDeploy.Spec.Replicas
```
rzetelskik, line 95: assert it's >0?

test/e2e/set/scyllacluster/scyllacluster_outage.go
```diff
@@ -0,0 +1,128 @@
...
+			setOperatorDeploymentReplicas(ctx, operatorDeployments, originalOperatorReplicas)
+		})
+
+		setOperatorDeploymentReplicas(ctx, operatorDeployments, 0)
+
+		framework.By("Updating ScyllaCluster spec.podMetadata.annotations while operator is down")
+		outageAnnotationValue := sc.ResourceVersion
+		sc = updateClusterPodAnnotations(ctx, f.ScyllaClient().ScyllaV1().ScyllaClusters(f.Namespace()), sc, outageAnnotationKey, outageAnnotationValue)
+
+		setOperatorDeploymentReplicas(ctx, operatorDeployments, originalOperatorReplicas)
```
rzetelskik: nit: framework.By("Restoring the original operator replica count")

test/e2e/set/scyllacluster/scyllacluster_outage.go
```diff
@@ -0,0 +1,128 @@
...
+
+		framework.By("Waiting for the ScyllaCluster to reconcile the spec change")
+		reconcileCtx, reconcileCtxCancel := utils.ContextForRollout(ctx, sc)
+		defer reconcileCtxCancel()
+		sc, err = controllerhelpers.WaitForScyllaClusterState(reconcileCtx, f.ScyllaClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Verifying the annotation was propagated to ScyllaCluster pods")
+		pods, err := f.KubeClient().CoreV1().Pods(f.Namespace()).List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", naming.ClusterNameLabel, sc.Name),
```
rzetelskik: podSelector := labels.Set(naming.ScyllaDBNodePodsSelectorLabelsForScyllaCluster(sc)).AsSelector()

otherwise you'll pick up other pods, e.g. cleanup job

test/e2e/set/scyllacluster/scyllacluster_outage.go
abramche: > "ScyllaCluster" + "should reconcile spec after an operator outage"

simple and works 👍

test/e2e/set/scyllacluster/scyllacluster_outage.go
abramche: addressed all of the above & packaged all commits into 897f6cefe6564b762a33fc5cdf25ce51e3c6998b

rzetelskik (approved):
Thanks for the updates!
/lgtm

PR #3500, opened 2026-07-13
test: collect logs from the replaced node in node replacement E2E test

rzetelskik:
Node replacement E2E does not capture the logs of the replaced Pod. That leaves the replacement path hard to debug because the post-failure dump sees only the new Pod.

This PR extends the object collection machinery to allow for following the logs of currently running containers of a Pod.
It also extends the machinery to allow for transforming the artifact's filesystem path.

It extends the replacement E2Es to capture the pre-replacement Pod before replacement starts and make its artifacts distinct not to have it overwritten by the collection of the replacing Pod. It keeps the current container logs open while replacement runs, so the logs are collected until the Pod is deleted.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-204

/cc

### DIFF
--- a/test/e2e/framework/dump.go
+++ b/test/e2e/framework/dump.go
@@ -2,7 +2,9 @@ package framework
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/scylladb/scylla-operator/pkg/gather/collect"
 	"github.com/scylladb/scylla-operator/pkg/naming"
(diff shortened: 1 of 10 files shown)


test/e2e/framework/dump.go
```diff
@@ -69,3 +73,94 @@ func DumpNamespace(ctx context.Context, restConfig *rest.Config, discoveryClient
 		name,
 	)
 }
+
+// DumpPodAndFollowLogs starts dumping the pod manifest and following its current container logs.
+// The first returned function waits until all current log streams are open or collection fails,
+// and the second one cancels log following and waits for collection to finish.
+func DumpPodAndFollowLogs(ctx context.Context, restConfig *rest.Config, dynamicClient dynamic.Interface, corev1Client corev1client.CoreV1Interface, artifactsDir string, pod *corev1.Pod, options collect.CollectObjectOptions) (func(context.Context) error, func(context.Context) error, error) {
```
czeslavo: As discussed, let's consider passing name/namespace instead of a Pod snapshot to avoid using stale state.

rzetelskik: done

PR #3513, opened 2026-07-23
fix: prevent ScyllaDBDatacenter rollout deadlock when rack is prepended during a progressing rollout

rzetelskik:

Currently, ScyllaDBDatacenter rollout can deadlock when a a rack is prepended during a progressing rollout with bootstrap sync enabled.

This is due to an [incorrect implementation of ScyllaDBDatacenter controller](https://github.com/scylladb/scylla-operator/blob/412f2606a6ea7ccbbd5b2c655ea3681afb71a75f/pkg/controller/scylladbdatacenter/sync_statefulsets.go#L397-L471) .
`createMissingStatefulSets` convoluted logic mixes waiting for the rollout of existing StatefulSets with creation of new StatefulSets, assuming it should only wait for the rollout of StatefulSets earlier in the spec’s slice. Prepending a rack during an active rollout breaks it.

This PR fixes it by separating these actions: the controller now waits for all existing StatefulSets to roll out before a new StatefulSet is created. It also splits the logic into separate functions.

To test this, the PR introduces ScyllaDBDatacenter controller envtests ensuring sequential creation of StatefulSets in different scenarios.

Envtest failure before the fix: https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3513/pull-scylla-operator-master-envtest/2080348833983762432.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-247

/cc

### DIFF
--- a/pkg/controller/scylladbdatacenter/controller.go
+++ b/pkg/controller/scylladbdatacenter/controller.go
@@ -50,7 +50,9 @@ import (
 const (
 	ControllerName = "ScyllaDBDatacenterController"
 
-	artificialDelayForCachesToCatchUp = 10 * time.Second
+	// defaultStatefulSetCachePropagationDelay is the default value for the delay after applying StatefulSet changes
+	// to let informer caches observe the update.
+	defaultStatefulSetCachePropagationDelay = 10 * time.Second
 )
 
 var (
@@ -87,6 +89,18 @@ type Controller struct {
 	handlers *controllerhelpers.Handlers[*scyllav1alpha1.ScyllaDBDatacenter]
 
 	keyGetter crypto.KeyGenerator
+
+	statefulSetCachePropagationDelay time.Duration
+}
+
+type ControllerOption func(ctrl *Controller)
+
+// WithStatefulSetCachePropagationDelay overrides the delay after applying StatefulSet changes to let informer caches
+// observe the update.
+func WithStatefulSetCachePropagationDelay(delay time.Duration) ControllerOption {
+	return func(c *Controller) {
+		c.statefulSetCachePropagationDelay = delay
+	}
 }
 
 func NewController(
@@ -108,6 +122,7 @@ func NewController(
 	operatorImage string,
 	cqlsIngressPort int,
 	keyGetter crypto.KeyGenerator,
+	options ...ControllerOption,
 ) (*Controller, error) {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartStructuredLogging(0)
@@ -160,6 +175,12 @@ func NewController(
 		),
 
 		keyGetter: keyGetter,
+
+		statefulSetCachePropagationDelay: defaultStatefulSetCachePropagationDelay,
+	}
+
+	for _, option := range options {
+		option(sdcc)
 	}
 
 	var err error
--- a/pkg/controller/scylladbdatacenter/sync_statefulsets.go
+++ b/pkg/controller/scylladbdatacenter/sync_statefulsets.go
@@ -561,7 +561,7 @@ func (sdcc *Controller) syncStatefulSets(
 	if len(createProgressingConditions) > 0 {
 		// Wait for the informers to catch up.
 		// TODO: Add expectations, not to reconcile sooner then we see this new StatefulSet in our caches. (#682)
-		time.Sleep(artificialDelayForCachesToCatchUp)
+		time.Sleep(sdcc.statefulSetCachePropagationDelay)
 		return progressingConditions, nil
 	}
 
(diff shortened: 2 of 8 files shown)


test/envtest/controllers/scylladbdatacenter_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+)
+
+const scyllaDBDatacenterControllerResyncPeriod = 12 * time.Hour
+
+// scyllaDBDatacenterControllerEventuallyTimeout has to account for a hardcoded 10s sleep in StatefulSet reconciliation loop.
+const scyllaDBDatacenterControllerEventuallyTimeout = 30 * time.Second
+
+// scyllaDBDatacenterControllerConsistentlyTimeout has to cover at least one hardcoded 10s sleep in StatefulSet reconciliation loop.
+const scyllaDBDatacenterControllerConsistentlyTimeout = 15 * time.Second
```
czeslavo: Can we make the sleep configurable so that we override it in tests to some reasonably low value? Should allow reducing the timeouts.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -460,16 +426,83 @@ func (sdcc *Controller) createMissingStatefulSets(
...
-			return progressingConditions, nil
 		}
 	}
 
 	return progressingConditions, apimachineryutilerrors.NewAggregate(errs)
 }
 
+// createMissingStatefulSets creates missing StatefulSets.
+// At most one StatefulSet is created to ensure racks bootstrap sequentially.
+func (sdcc *Controller) createMissingStatefulSets(
```
czeslavo: Let's consider adding unit tests if possible.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -460,16 +426,83 @@ func (sdcc *Controller) createMissingStatefulSets(
...
+
+		// StatefulSets must be created sequentially. Return early.
+		return []*appsv1.StatefulSet{sts}, progressingConditions, nil
+	}
+
+	return []*appsv1.StatefulSet{}, []metav1.Condition{}, nil
+}
+
+// ensureRackNamesInRackStatuses records statuses for newly created racks before informer caches catch up.
+func (sdcc *Controller) ensureRackNamesInRackStatuses(
```
czeslavo: Ditto - unit tests.

czeslavo (commented):
Great catch, great changes. 🎖️

test/envtest/controllers/scylladbdatacenter_controller_test.go
rzetelskik: done

pkg/controller/scylladbdatacenter/sync_statefulsets.go
rzetelskik: Added unit tests. The nested functions need fixing asap, e.g.
```
scyllaDBImageVersion, err := naming.ImageToVersion(sdc.Spec.ScyllaDB.Image)
if err != nil {
	klog.ErrorS(err, "can't get version of image", "Image", sdc.Spec.ScyllaDB.Image)
}

status.UpdatedVersion = scyllaDBImageVersion
```
in `calculateRackStatus`. I'll try to raise tracking issues, but I won't be fixing it here as the fixes seem endless.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
rzetelskik: added

pkg/controller/scylladbdatacenter/controller.go
```diff
@@ -108,6 +119,7 @@ func NewController(
 	operatorImage string,
 	cqlsIngressPort int,
 	keyGetter crypto.KeyGenerator,
+	options ControllerOptions,
```
czeslavo: Nit: to avoid forcing this option struct to be passed down in every call, we could implement it with [functional options](https://golang.cafe/blog/golang-functional-options-pattern) like so, allowing it to be passed only in tests:

```go
type ControllerOption func(ctrl *Controller)

func WithStatefulSetCachePropagationDelay(delay time.Duration) func(*Controller) {
  return func(ctrl *Controller) {
    ctrl.statefulSetCachePropagationDelay = delay
  }
}


```

rzetelskik: done

czeslavo (approved):
🚀 
/lgtm

PR #3514, opened 2026-07-24
test: backup-restore from minio

abramche:
- add new test for backup-restore to minio storage
- the test structure is very similar to existing test, extract common code to helpers

Resolves https://scylladb.atlassian.net/browse/OPERATOR-166

### DIFF
--- a/hack/.ci/lib/e2e.sh
+++ b/hack/.ci/lib/e2e.sh
@@ -320,6 +320,14 @@ function run-e2e {
     kubectl create -n=e2e secret generic s3-credentials --dry-run=client -o=yaml | kubectl_create -f=-
   fi
 
+  if [[ -n "${SO_S3_AGENT_CONFIG_PATH+x}" ]]; then
+    agent_config_in_container_path=/var/run/secrets/agent-config/agent-config.yaml
+    kubectl create -n=e2e secret generic agent-config --from-file=agent-config.yaml="${SO_S3_AGENT_CONFIG_PATH}" --dry-run=client -o=yaml | kubectl_create -f=-
+  else
+    kubectl create -n=e2e secret generic agent-config --dry-run=client -o=yaml | kubectl_create -f=-
+    agent_config_in_container_path=""
+  fi
+
   # Build a comma-separated string following a `<cluster_identifier>=<bucket_name>` format expected by `--worker-object-storage-buckets` flag.
   worker_object_storage_buckets=$(
     res=()
@@ -383,6 +391,7 @@ function run-e2e {
     "--object-storage-bucket=${SO_BUCKET_NAME}"
     "--gcs-service-account-key-path=${gcs_sa_in_container_path}"
     "--s3-credentials-file-path=${s3_credentials_in_container_path}"
+    "--s3-agent-config-path=${agent_config_in_container_path}"
     "--scylladb-version=${SCYLLADB_VERSION}"
     "--scylladb-manager-version=${SCYLLADB_MANAGER_VERSION}"
     "--scylladb-manager-agent-version=${SCYLLADB_MANAGER_AGENT_VERSION}"
@@ -454,6 +463,8 @@ $(printf '    - "%s"\n' "${e2e_command_args[@]}")
       mountPath: /var/run/secrets/gcs-service-account-credentials
     - name: s3-credentials
       mountPath: /var/run/secrets/s3-credentials
+    - name: agent-config
+      mountPath: /var/run/secrets/agent-config
     - name: kubeconfig
       mountPath: /var/run/secrets/kubeconfig
       subPath: kubeconfig
@@ -476,6 +487,9 @@ $(printf '    - "%s"\n' "${e2e_command_args[@]}")
   - name: s3-credentials
     secret:
       secretName: s3-credentials
+  - name: agent-config
+    secret:
+      secretName: agent-config
   - name: kubeconfig
     secret:
       secretName: kubeconfig
--- a/hack/.ci/manifests/namespaces/minio/00_namespace.yaml
+++ b/hack/.ci/manifests/namespaces/minio/00_namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: minio
--- a/hack/.ci/manifests/namespaces/minio/50_deployment.yaml
+++ b/hack/.ci/manifests/namespaces/minio/50_deployment.yaml
@@ -1,0 +1,45 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: minio
+  namespace: minio
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: minio
+  template:
+    metadata:
+      labels:
+        app: minio
+    spec:
+      containers:
+      - name: minio
+        # Final MinIO community release with a published container image; subsequent releases are source-only.
+        image: minio/minio:RELEASE.2025-09-07T16-13-09Z@sha256:14cea493d9a34af32f524e538b8346cf79f3321eff8e708c1e2960462bd8936e
+        command:
+        - minio
+        - server
+        - /data
+        env:
+        - name: MINIO_ROOT_USER
+          value: minioadmin
+        - name: MINIO_ROOT_PASSWORD
+          value: minioadmin
+        ports:
+        - name: api
+          containerPort: 9000
+          protocol: TCP
+        readinessProbe:
+          httpGet:
+            path: /minio/health/ready
+            port: 9000
+          initialDelaySeconds: 5
+          periodSeconds: 5
+        volumeMounts:
+        - name: data
+          mountPath: /data
+      volumes:
+      - name: data
+        persistentVolumeClaim:
+          claimName: minio-data
--- a/hack/kind/cluster-setup.sh
+++ b/hack/kind/cluster-setup.sh
@@ -119,4 +119,13 @@ else
 
   # Deploy the full operator stack (cert-manager, prometheus, haproxy-ingress, operator, scylla-manager).
   "${repo_root}/hack/ci-deploy.sh" "${SO_IMAGE}"
+
+  # Deploy MinIO as the object storage for backup/restore e2e tests.
+  minio_deploy_dir="${ARTIFACTS}/deploy/minio"
+  mkdir -p "${minio_deploy_dir}"
+  cp "${repo_root}"/hack/.ci/manifests/namespaces/minio/*.yaml "${minio_deploy_dir}"
+  yq e --inplace '.spec.storageClassName = env(SO_SCYLLACLUSTER_STORAGECLASS_NAME)' "${minio_deploy_dir}/10_pvc.yaml"
+
+  kubectl apply --server-side=true -f "${minio_deploy_dir}"
+  kubectl -n minio rollout status --timeout=5m deployment.apps/minio
 fi
--- a/test/e2e/set/scyllacluster/scyllamanager_object_storage.go
+++ b/test/e2e/set/scyllacluster/scyllamanager_object_storage.go
@@ -27,14 +27,17 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
-// Not part of SuiteKindFast: requires external object storage configured on the cluster.
-var _ = g.Describe("Scylla Manager integration", framework.SuiteParallel, framework.SuiteParallelOpenShift, func() {
+var _ = g.Describe("Scylla Manager integration", framework.SuiteParallel, framework.SuiteParallelOpenShift, framework.SuiteKindFast, func() {
 	var f *framework.Framework
 
 	g.BeforeEach(func(ctx context.Context) {
 		f = framework.NewFramework(ctx, "scyllacluster")
 	})
 
+	// The object storage backend is determined by the settings passed to the test framework: the credentials
+	// type selects how the agent is set up (GCS service account key or S3 credentials, optionally with a custom
+	// agent config, e.g. an S3 endpoint override - for kind minio) and, together with the bucket name, the backup
+	// location ("gcs:<bucket>" or "s3:<bucket>"). The test itself is agnostic to the specific storage provider.
 	g.It("should register cluster, sync backup tasks and support manual restore procedure", func(ctx g.SpecContext) {
 		sourceSC := f.GetDefaultScyllaCluster()
 		sourceSC.Spec.Datacenter.Racks[0].Members = 1
@@ -740,7 +743,7 @@ func setUpObjectStorageCredentials(ctx context.Context, ns string, nsClient fram
 		s3CredentialsFile := objectStorageSettings.S3CredentialsFile()
 		o.Expect(s3CredentialsFile).NotTo(o.BeEmpty())
 
-		setUpS3Credentials(ctx, nsClient.KubeClient().CoreV1(), sc, ns, s3CredentialsFile)
+		setUpS3Credentials(ctx, nsClient.KubeClient().CoreV1(), sc, ns, s3CredentialsFile, objectStorageSettings.S3AgentConfig())
 
 	}
 }
@@ -784,7 +787,7 @@ func setUpGCSCredentials(ctx context.Context, coreClient corev1client.CoreV1Inte
 	}
 }
 
-func setUpS3Credentials(ctx context.Context, coreClient corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster, namespace string, s3CredentialsFile []byte) {
+func setUpS3Credentials(ctx context.Context, coreClient corev1client.CoreV1Interface, sc *scyllav1.ScyllaCluster, namespace string, s3CredentialsFile []byte, s3AgentConfig []byte) {
 	g.GinkgoHelper()
 
 	secret := &corev1.Secret{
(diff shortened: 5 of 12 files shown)


hack/kind/cluster-setup.sh
```diff
@@ -119,4 +119,15 @@ else
...
+  # Deploy MinIO as the object storage for backup/restore e2e tests.
+  minio_deploy_dir="${ARTIFACTS}/deploy/minio"
+  mkdir -p "${minio_deploy_dir}"
+  cp "${repo_root}"/deploy/minio/*.yaml "${minio_deploy_dir}"
+  yq e --inplace '.spec.storageClassName = env(SO_SCYLLACLUSTER_STORAGECLASS_NAME)' "${minio_deploy_dir}/10_pvc.yaml"
+
+  kubectl apply --server-side=true -f "${minio_deploy_dir}"
+  kubectl -n minio rollout status --timeout=5m deployment.apps/minio
+
+  kubectl -n minio exec deployment/minio -- mkdir -p "/data/${SO_BUCKET_NAME}"
```
mflendrich: Since this MinIO deployment is local to the single backup&restore test (and not reused across all the tests), it would result in less clutter if the initialization of minio happened within the test case implementation, and not here on the global level.

suggestion: move MinIO initialization from here (global to the entire kind suite) to just the test case, and clean up appropriately

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -27,8 +27,8 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
-// Not part of SuiteKindFast: requires external object storage configured on the cluster.
-var _ = g.Describe("Scylla Manager integration", framework.SuiteParallel, framework.SuiteParallelOpenShift, func() {
```
mflendrich, line 31: If I understand this right -- that this test case runs either with cloud S3 or with MinIO depending on the suite (paralllel/parallel ocp vs kind fast) - and requires the s3 agent config path only for the "kind flavor", then I find it awkward that the test behavior changes on the suite it runs in.

IMO it would be more flexible and maintainable if we had a separate case for "Scylla Manager can back up and restore to S3" and "Scylla Manager can back up and restore to MinIO". This would also allow us to enable/disable those cases per suite in the future as we please, without relying on environmental factors like presence of a config.

hack/kind/cluster-setup.sh
rzetelskik: I disagree, this is more aligned with how we provision storage in other platforms, the existing wiring for storage buckets,  and comes with the benefit of making backup/restore easily testable in a local env (outside of tests)

hack/.ci/manifests/namespaces/minio/00_namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rzetelskik, line 1: if this is intended for internal consumption it shouldn't live in `deploy/` (meant for user consumption), but along other internal manifests (hack/.ci/manifests)

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
rzetelskik: The behaviour depends on the flags passed to the test runner, the test itself is platform-agnostic

abramche: The minio test is identical to other storage tests (the test does not even know it is backing up to minio - to the test, it is S3 storage) - it would introduce 100% code duplication for no reason - the test is already polymorphic even without minio (same code runs for S3 and GCS). 

imo current approach follows the pattern and is a better choice for maintaining the codebase

hack/.ci/manifests/namespaces/minio/00_namespace.yaml
abramche: moved the crds (4a28569b942a204d51f1ae30bb49161f9e19f615) and updated the cluster setup script (34aedaa2d8fe891c5970305647d2b9412d4c5859)

rzetelskik: Thanks. On a side note, those are not CRDs but manifests (and there are no CRDs or CR manifests among them), so the commit message subject is misleading

hack/.ci/manifests/namespaces/minio/50_deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+    matchLabels:
+      app: minio
+  template:
+    metadata:
+      labels:
+        app: minio
+    spec:
+      containers:
+      - name: minio
+        image: minio/minio:latest
```
rzetelskik: we should pin this

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
```diff
@@ -27,8 +27,8 @@ import (
...
+// Runs with MinIO object storage in SuiteKindFast
```
rzetelskik: not a good place for this comment - the test is agnostic to this, I'd remove the comment altogether

hack/.ci/manifests/namespaces/minio/50_deployment.yaml
abramche: minio is not under active development any more (the latest patch was an year ago, no changes since), is it worth pinning it?

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
abramche: why not? minio adds the entire possibility to run in Kind - it explains why external storage support is allowed in a kind test (which possibly runs locally)

hack/.ci/manifests/namespaces/minio/50_deployment.yaml
rzetelskik: it's always a good practice to ensure reproducibility and  prevent supply chain tampering - not that important here but still a good practice

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
rzetelskik: Because the test is agnostic to this, the wiring happens in a different place. You could wire the kind suite to use AWS S3 and you wouldn't have to change anything in this test.

hack/.ci/manifests/namespaces/minio/00_namespace.yaml
abramche: renamed in d98d679867e1466662cf82db898381fd1086e9e0

hack/.ci/manifests/namespaces/minio/50_deployment.yaml
abramche: pinned to the latest docker image release 
d98d679867e1466662cf82db898381fd1086e9e0

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
abramche: removed in 871a9bb5c03301676378f68f526e6cbd63dd3ecd

hack/.ci/run-e2e-shared.env.sh
```diff
@@ -22,3 +22,8 @@ export SCYLLA_OPERATOR_FEATURE_GATES
 SO_SCYLLACLUSTER_STORAGECLASS_NAME="${SO_SCYLLACLUSTER_STORAGECLASS_NAME=scylladb-local-xfs}"
 export SO_SCYLLACLUSTER_STORAGECLASS_NAME
 
+# Object storage bucket for backup/restore e2e tests. Cloud providers back up to a real bucket of this name;
+# the kind suite creates a bucket of this name in MinIO (see hack/kind/cluster-setup.sh).
+SO_BUCKET_NAME="${SO_BUCKET_NAME:-manager-backup-tests}"
```
rzetelskik: why not set it in hack/kind/run-e2e-tests.sh then?

abramche: it's a bit of a dependency mess

`hack/kind/cluster-setup.sh` and `hack/kind/run-e2e-tests.sh` both separately use this value (first creates the bucket, the second passes it on to the runner pod). `cluster-setup.sh` runs first, so we cannot set it in the (later) `hack/kind/run-e2e-tests.sh` invocation. 

Both of them source `hack/.ci/run-e2e-shared.env.sh` so it works (both in CI and locally). The alternative is to move the `SO_BUCKET_NAME` init to a kind-fast Prow action, but it would require that local executions take care of this...

rzetelskik: I see. I don't really like how it leaks kind-specific env to other jobs, maybe we could source something like `hack/kind/shared-env.sh`? @czeslavo do you have better ideas on how a shared env var could be shared between kind scripts without leaking to other scripts? The only example of this I see is the CLUSTER_NAME set in Makefile

czeslavo: I can see the cloud tests set the variable explicitly in their ProwJob definitions (e.g., [here](https://github.com/scylladb/scylla-operator-release/blob/80f56450e2b4896833809521c891967d559a909b/config/prow/jobs/scylladb/scylla-operator/scylla-operator-master-presubmits.yaml#L2091), creating it dynamically). 

I think that what Eugene proposed is fine - we can have a default that's set for every test flavor and can be overridden. It happens that for kind it is not, as its requirements are satisfied with just that. I'd leave it as proposed, especially since it's not a new variable that's only-kind specific. It's been moved from `hack/.ci/lib/e2e.sh`.

rzetelskik: It's been moved, but now it has a default, kind-specific value spilling into all e2e runner scripts. Any new script would have to explicitly unset it not to fail validation https://github.com/scylladb/scylla-operator/blob/76dd8ef467b99e0dae7d63a9519bfcf861f2a9f0/pkg/cmd/tests/object_storage_options.go#L55-L57

hack/kind/cluster-setup.sh
mflendrich: IMO there is no single good answer - it's an architectural design problem we're facing here with conflicting interests to consider.

> this is more aligned with how we provision storage in other platforms, the existing wiring for storage buckets,

MinIO is not part of external test infrastructure though, so it does not share the need to be provisioned outside of  the test runner.

Furthermore, making this an unconditional part of cluster initialization cements the notion of having a single test suite (where I'd envision multiple smaller localized/targeted test suites in the future).

Therefore, this approach complicates the shared test environment for the purpose of satisfying a single isolated test case, which is an anti-pattern (at least from the automated testing perspective).
And I'm not buying the notion of aligning new code with an existing anti-pattern where it's not necessary, just for the sake of staying consistent.

> making backup/restore easily testable in a local env (outside of tests)

This is a good point, though. The local env use case convinces me. I'm happy to resolve my comment based on this.

hack/.ci/lib/e2e.sh
```diff
@@ -320,6 +319,14 @@ function run-e2e {
     kubectl create -n=e2e secret generic s3-credentials --dry-run=client -o=yaml | kubectl_create -f=-
   fi
 
+  agent_config_in_container_path=""
+  if [[ -n "${SO_S3_AGENT_CONFIG_PATH+x}" ]]; then
+    agent_config_in_container_path=/var/run/secrets/agent-config/agent-config.yaml
+    kubectl create -n=e2e secret generic agent-config --from-file=agent-config.yaml="${SO_S3_AGENT_CONFIG_PATH}" --dry-run=client -o=yaml | kubectl_create -f=-
+  else
+    kubectl create -n=e2e secret generic agent-config --dry-run=client -o=yaml | kubectl_create -f=-
+  fi
```
mflendrich: nit - possibly clearer flow without overwriting

```suggestion
  if [[ -n "${SO_S3_AGENT_CONFIG_PATH+x}" ]]; then
    agent_config_in_container_path=/var/run/secrets/agent-config/agent-config.yaml
    kubectl create -n=e2e secret generic agent-config --from-file=agent-config.yaml="${SO_S3_AGENT_CONFIG_PATH}" --dry-run=client -o=yaml | kubectl_create -f=-
  else
    kubectl create -n=e2e secret generic agent-config --dry-run=client -o=yaml | kubectl_create -f=-
    agent_config_in_container_path=""
  fi
```

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
mflendrich: Understood. In that case I ask that the polymorphism is expressed explicitly in the "docstring" comment for the test case - defining the contract of the polymorphism that's being expanded here (precisely speaking: what controls which backend the test is going to be run against).

The reason I'm asking for this is because this makes different executions of this test case not equal and I want to manage the [surprise factor](https://i.redd.it/al1sgwz8drha1.png).

The surprise is the fact that the parameterization is hidden in opaque settings hardcoded within the test executor for various platforms.

abramche: Kacper specifically requested the docstring to be removed in https://github.com/scylladb/scylla-operator/pull/3514#discussion_r3765988372

I can reintroduce it with an expanded explanation of the test behaviour but let's reach a consensus

hack/.ci/run-e2e-shared.env.sh
abramche: I can move minio bucket creation to `hack/kind/run-e2e-tests.sh` (and `SO_BUCKET_NAME` assignment), then we do not need it in cluster-setup.sh and elsewhere. `e2e.sh` defaults to empty name -> existing behaviour -> new suites do not need to define it.

The only nit is `run-e2e-tests.sh` will have a bit of a preparation logic; but it ticks all other checkboxes. @rzetelskik WDYT?

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
rzetelskik: I absolutely don't mind a docstring explaining what controls how manager/manager agent is configured with storage in that test, i.e. that it depends on the bucket name, credentials and config passed to the test framework. The comment I requested to remove was leaking storage setup for a specific suite which this test is agnostic to.

hack/.ci/run-e2e-shared.env.sh
rzetelskik: That way we'd lose minio setup in the dev env, which is arguably one of the bigger benefits this PR introduces. Is there a specific reason not to set up a file with env shared just between the kind scripts (like the `hack/kind/shared-env.sh` I proposed)? This is what you'd normally do in code - have a shared const within the scope without leaking it.

abramche: > That way we'd lose minio setup in the dev env

No, why? You run the tests locally as a Makefile target or directly with hack/kind scripts and everything is set up the same way as before. Minio is still deployed with cluster-setup.sh and the bucket is still created within the reach of kind test kickoff point (`make test-e2e-kind` or `run-e2e-tests-sh`).

> Is there a specific reason not to set up a file with env shared just between the kind scripts (like the hack/kind/shared-env.sh I proposed)? 

Yes, I believe this is a worse way: the minio bucket matters only to this particular kind suite, any other one (think: operator upgrade - and it uses its own invocation script) will not use it. So a shared const will not be really "shared" - only one particular test suite is using it.

You could argue that minio setup could also move to `run-e2e-tests.sh` - minio setup is minuscule and takes almost no resources though.

hack/.ci/lib/e2e.sh
abramche: applied in 4a90ad55e3e37f0b502c6f440b14fe678a3668ee

hack/.ci/run-e2e-shared.env.sh
abramche: See 4a90ad55e3e37f0b502c6f440b14fe678a3668ee for reorganization suggestion

test/e2e/set/scyllacluster/scyllamanager_object_storage.go
abramche: Added docstring in d6380f234c8159f2cd52614c9ffa18a71c2fa52f

hack/.ci/run-e2e-shared.env.sh
rzetelskik: > Minio is still deployed with cluster-setup.sh and the bucket is still created within the reach of kind test kickoff point (make test-e2e-kind or run-e2e-tests-sh).

Not if you just run `make kind-setup`, which is the way to set up the local env outside of running e2e tests

rzetelskik: I think the current form is good - we set up minio on cluster setup, but the particular bucket is scoped to the test runner, so if you want to test locally outside of e2es you just create a subdir. Happy to resolve now.

rzetelskik:
lgtm
/assign @mflendrich

hack/.ci/run-e2e-shared.env.sh
abramche: > if you want to test locally outside of e2es you just create a subdir

no need - `make test-e2e-kind` (or `run-e2e-tests.sh` directly) will create it

rzetelskik: You can run a dev env without running e2es - this is what I'm saying in https://github.com/scylladb/scylla-operator/pull/3514#discussion_r3783153998. This is what I do on a daily basis during development/debugging. Either way, I think it's good enough as is. We could use a dev README explaining how to create a bucket and provide credentials/config in such case (when you set up a dev cluster and want to spin a ScyllaDB cluster **without** runing e2es) but let's consider this out of scope - we don't document it for other storage types.

mflendrich (approved):
/lgtm

abramche:
rebased onto master to fix the orphaned PV test failures

abramche:
/test e2e-gke-parallel

cert-manager upstream issue

abramche:
/test images
/retest

upstream fixed in https://github.com/scylladb/scylla-operator/pull/3609

PR #3515, opened 2026-07-25
feat(nodeconfig): support filesystem creation flags

### LINKED ISSUE #2353: Reduce XFS metadata overhead

We should disable two XFS flags, simillar to how ScyllaDB AMI's sets up the storage.

 Minimize metadata overhead, synchronize with how ScyllaDB AMI's are set up.

Original issue: https://github.com/scylladb/scylladb/issues/22028
PR changing the AMI: https://github.com/scylladb/scylladb/pull/22072

vnaruvan:

This PR adds support for generic filesystem creation flags in NodeConfig local disk setup.

The new `spec.localDiskSetup.filesystems[].flags` field is passed through to the `mkfs` command, allowing filesystem-specific options to be configured from the NodeConfig API instead of being hardcoded in the implementation.

The NodeConfig examples now use this field to pass the ScyllaDB-recommended XFS metadata options:

- `-m rmapbt=0`
- `-m reflink=0`

This keeps the implementation filesystem-agnostic while allowing XFS metadata overhead to be reduced through configuration.

Generated API, CRD, docs, and bundle updates are included in separate commits.

Resolves #2353

**Validation:**

- `go test ./pkg/disks ./pkg/controller/nodesetup`
- `make update`
- `make verify`


czeslavo:
Hey @vnaruvan. Thank you for the contribution! It looks good.

Just one thing that I'd ask you for: we have a page dedicated to configuring nodes where we could mention this new config knob and explain why we recommend disabling both `rmapbt` and  `reflink` for xfs. If you could extend that, that'd be great. Thanks!

vnaruvan:
Thanks! I’ll update the NodeConfig docs page to document the new `flags` field and explain the XFS `rmapbt=0` / `reflink=0` recommendation.

vnaruvan:
Updated the NodeConfig docs page with the new `flags` field and the XFS `rmapbt=0` / `reflink=0` recommendation.

PR #3525, opened 2026-07-28
chore(kind): allow local registry port override

abramche:
- replace static kind local registry port with an env var lookup
- dynamically generate the (previously hardcoded) containerd registry structure
- port fallback to 5001 if unset

Resolves https://scylladb.atlassian.net/browse/OPERATOR-252

/priority important-soon
/cc czeslavo

### DIFF
--- a/hack/kind/cluster-setup.sh
+++ b/hack/kind/cluster-setup.sh
@@ -31,6 +31,15 @@ if ! podman network inspect kind >/dev/null 2>&1; then
   podman network create kind
 fi
 
+# Source shared constants and generate containerd registry configuration.
+source "${repo_root}/hack/kind/lib.sh"
+internal_reg_port=5000
+containerd_reg_dir="${repo_root}/hack/kind/containerd-registries/localhost:${KIND_REGISTRY_PORT}"
+mkdir -p "${containerd_reg_dir}"
+cat > "${containerd_reg_dir}/hosts.toml" <<EOF
+[host."http://${KIND_REGISTRY_NAME}:${internal_reg_port}"]
+EOF
+
 # Ensure KinD cluster exists.
 if ! kind get clusters | grep -q "^${CLUSTER_NAME}$"; then
     KIND_CREATE_CMD=(kind create cluster --name="${CLUSTER_NAME}" --config="${repo_root}/hack/kind/cluster-config.yaml" --retain)
@@ -53,17 +62,15 @@ else
 fi
 
 # Set up a local registry for the KinD cluster following https://kind.sigs.k8s.io/docs/user/local-registry/.
-reg_name='kind-registry'
-reg_port='5001'
-if [ "$(podman inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
+if [ "$(podman inspect -f '{{.State.Running}}' "${KIND_REGISTRY_NAME}" 2>/dev/null || true)" != 'true' ]; then
   podman run \
-    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --replace --network bridge --name "${reg_name}" \
+    -d --restart=always -p "127.0.0.1:${KIND_REGISTRY_PORT}:${internal_reg_port}" --replace --network bridge --name "${KIND_REGISTRY_NAME}" \
     registry:2
 fi
 
 # Connect registry to KinD network.
-if [ "$(podman inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = 'null' ]; then
-  podman network connect "kind" "${reg_name}"
+if [ "$(podman inspect -f='{{json .NetworkSettings.Networks.kind}}' "${KIND_REGISTRY_NAME}")" = 'null' ]; then
+  podman network connect "kind" "${KIND_REGISTRY_NAME}"
 fi
 
 # Inform KinD cluster about the local registry.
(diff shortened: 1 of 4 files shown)


abramche:
/test e2e-kind-fast

API server in k8s was unresponsive

hack/kind/lib.sh
```diff
@@ -18,7 +18,9 @@ function build-and-push-operator-image {
 
   # If SO_IMAGE is not set, build the image.
   if [ -z "${SO_IMAGE:-}" ]; then
-    SO_IMAGE="localhost:5001/scylladb/scylla-operator:e2e-$( date +%Y%m%d%H%M%S )"
+    local reg_port
+    reg_port=$( podman port kind-registry 5000 | cut -d':' -f2 )
```
czeslavo: Nit: it would make sense to extract the registry container name to a constant and use it across the board.

czeslavo (commented):
Please update `CONTRIBUTING.md` - there's a reference to the registry with the 5001 port hardcoded.

hack/kind/lib.sh
abramche: addressed both in b524ce552e3ad57cf28e915e569c0405136153b7

mflendrich (changes requested):
I oppose the idea of picking a random port - this is unnecessarily non-deterministic.

Why can't this be a configurable port? The user can defer to picking a random port if they want by using whatever port-picker tool on their end.

abramche:
> Why can't this be a configurable port? 

It can be - I can introduce the same port picking logic to the Jenkins pipeline directly; and then this PR only has overrides (via env var). Yaron though asked that we simplify the pipeline code and move the logic as much as possible into our own machinery, which is why I went this way (also consulting with Grzegorz).

mflendrich:
Understood.

Ports (and their availability) are a feature of the environment. If it's the jenkins environment that does not like 5001, then the choice of a different port belongs there.

Let's go with an exposed param for the port. As I said - port picking can always be done on top of that. And doing it inside increases accidental complexity.

abramche:
> Ports (and their availability) are a feature of the environment. If it's the jenkins environment that does not like 5001, then the choice of a different port belongs there.

Raised https://scylladb.atlassian.net/browse/OPERATOR-252 specifically for port override option (this PR); the original ticket (do not use port 5001) will be done in a scylla-pkg PR.

czeslavo:
lgtm
/assign @mflendrich

hack/kind/cluster-setup.sh
```diff
@@ -53,8 +61,8 @@ else
 fi
 
 # Set up a local registry for the KinD cluster following https://kind.sigs.k8s.io/docs/user/local-registry/.
-reg_name='kind-registry'
-reg_port='5001'
+reg_name="${KIND_REGISTRY_NAME}"
+reg_port="${KIND_REGISTRY_PORT}"
```
mflendrich: Why not use the `KIND_REGISTRY_NAME` and `KIND_REGISTRY_PORT` variables directly?

hack/kind/cluster-setup.sh
```diff
@@ -31,6 +31,14 @@ if ! podman network inspect kind >/dev/null 2>&1; then
   podman network create kind
 fi
 
+# Source shared constants and generate containerd registry configuration.
+source "${repo_root}/hack/kind/lib.sh"
+containerd_reg_dir="${repo_root}/hack/kind/containerd-registries/localhost:${KIND_REGISTRY_PORT}"
+mkdir -p "${containerd_reg_dir}"
+cat > "${containerd_reg_dir}/hosts.toml" <<EOF
+[host."http://${KIND_REGISTRY_NAME}:5000"]
```
mflendrich: If the registry container called `reg_name` exposes the registry on `reg_port`, then why is the override with port 5000 necessary?

And if proven necessary, then this generated file might be better suited for a temp directory as opposed to a gitignored place in the repo.

abramche: 5000 is the port of podman/kind [internal network](https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-host-namespace), which - being internal - does not conflict with anything.

This port is mapped to an **external/host OS port** (ex-5001) which was causing conflicts occasionally - this PR only deals with the external port.

The mounted mapping is needed so that kind can recognize `localhost` repository lookups.

We could use a temp dir, but https://github.com/scylladb/scylla-operator/blob/master/hack/kind/cluster-config.yaml#L19 (and all other mount points) specifically refer to this path. We could employ some `sed`-ing to patch a temp directory... but it feels like it's replacing one hack with another (and not necessarily a better one).

Creating `cluster-config.yaml` inline is an even bigger hack.

hack/kind/cluster-setup.sh
abramche: Good point, these are unnecessary any more

hack/kind/cluster-setup.sh
mflendrich: > Creating `cluster-config.yaml` inline is an even bigger hack.

Fair enough - that convinces me.

> 5000 is the port of podman/kind [internal network](https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-host-namespace), which - being internal - does not conflict with anything.

In that case let's make the 5000 a constant (stored in a bash variable) to avoid throwing the 5000 magic number around

abramche: done in ee24a2080a71b1816ac9497d5ba2e5b6b4c4a818 (squashed changes)

mflendrich (approved):
/lgtm

PR #3529, opened 2026-07-29
chore(renovate): manage local-csi-driver image references

czeslavo:
Adds Renovate coverage for the `local-csi-driver` image pins in `examples/common/local-volume-provisioner/` and `hack/.ci/manifests/`, so that version bumps are proposed automatically, and consolidates the per-file weekday schedule rules for container images into a single datasource-matched rule. `ignorePaths` is overridden to be empty because `config:recommended` ignores `**/examples/**`, which hid the example manifest from Renovate.

Closes https://scylladb.atlassian.net/browse/OPERATOR-263


rzetelskik (approved):
/lgtm

PR #3530, opened 2026-07-29
feat: relax bootstrap barrier to allow for parallel bootstrap

rzetelskik:
The bootstrap barrier requires every node in `ScyllaDBDatacenterNodesStatusReport` to be present and UP before a node proceeds with bootstrap, and membership in that required set was derived from Kubernetes state (`rackStatus.CurrentNodes`). A node that already exists as a Kubernetes object but hasn't joined the ScyllaDB cluster is therefore required, while being structurally unable to satisfy the precondition - it has no host ID and no status to propagate. This moves the criterion to actual ScyllaDB cluster membership: the sidecar checks whether the node owns normal tokens and records it on the member `Service` as `internal.scylla-operator.scylladb.com/node-joined-scylladb-cluster`, and report inclusion is gated on that. Normal tokens are used as signal rather than operation mode because token metadata is restored from disk before gossip starts, so a bootstrapped node never stops owning them across a restart. As a consequence, absence of an entry in the report no longer implies the node is missing or unhealthy.

This doesn't deliver parallel bootstrap on its own - `Pod` and `StatefulSet` creation are still sequential. There's no separate barrier path for sequential and parallel bootstrap, so this is E2E-tested as is.

**WARNING! Important notes**:
The barrier can still theoretically sequentialize joining nodes. A node that is joining is absent from the report's own entries by construction, but it can still reach `ObservedNodes` through its peers' gossip view, and the status report doesn't distinguish a joining node from a member - so a host ID appearing there without a corresponding report entry fails the precondition and holds up other nodes. I'm keeping that conservative: erring the other way lets nodes that have joined but aren't yet reported drop out of the required set. The practical implication is negligible, since raft sequentializes the joins anyway.

The other risk is that membership is now observed asynchronously, so it's inherently racy: a node can join and go down before its sidecar records the annotation, leaving it absent from the report and letting the barrier clear other nodes against a cluster that isn't actually healthy. The same unfiltered `ObservedNodes` channel keeps this narrow - a node that joined is still reported `DOWN` by its peers and re-enters the required set through their reports - so the window only really opens if no peer's `ObservedNodes` were propagated in time. I'm accepting that: the annotation is only ever written from a successful observation and never cleared, so once a node has been observed as a member it stays required permanently, and closing the window means either treating unobservable nodes as required, which is the deadlock this PR fixes, or clearing the annotation when the ScyllaDB API is unreachable, which would drop healthy nodes out of the required set on every restart.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-159

/cc

### DIFF
--- a/pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
+++ b/pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
@@ -37,11 +37,15 @@ type RackNodesStatusReport struct {
 	Name string `json:"name"`
 
 	// Nodes holds the list of node status reports collected from nodes from this rack.
+	// Entries correspond to ScyllaDB nodes, not to the Kubernetes objects representing them: a node is reported only
+	// once its ScyllaDB instance has joined the cluster and owns tokens in it.
+	// The absence of an entry doesn't imply the node is missing or unhealthy.
 	Nodes []NodeStatusReport `json:"nodes"`
 }
 
 // ScyllaDBDatacenterNodesStatusReport serves as an internal interface for propagating and aggregating node statuses observed by nodes in a ScyllaDB datacenter.
 // It is used by ScyllaDB Operator for coordinating operations such as topology changes and is not intended for direct user interaction.
+// The report only covers nodes that have joined the ScyllaDB cluster, which is distinct from the set of Kubernetes objects that represent them.
 // +kubebuilder:object:root=true
 // +kubebuilder:storageversion
 // +genclient
--- a/pkg/controller/scylladbdatacenter/resource.go
+++ b/pkg/controller/scylladbdatacenter/resource.go
@@ -2439,18 +2439,9 @@ func makeRackNodesStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec 
 		return nil, fmt.Errorf("can't get rack %q node count of ScyllaDBDatacenter %q: %w", rackSpec.Name, naming.ObjRef(sdc), err)
 	}
 
-	actualRackNodeCount := int32(0)
-	rackStatus, _, found := oslices.Find(sdc.Status.Racks, func(status scyllav1alpha1.RackStatus) bool {
-		return status.Name == rackSpec.Name
-	})
-	if found && rackStatus.CurrentNodes != nil {
-		actualRackNodeCount = *rackStatus.CurrentNodes
-	}
-
 	var nodeStatusReports []scyllav1alpha1.NodeStatusReport
 	for ord := int32(0); ord < *desiredRackNodeCount; ord++ {
-		isNodeExpectedInK8sState := ord < actualRackNodeCount
-		nodeStatusReport, ok, err := makeNodeStatusReport(sdc, rackSpec, int(ord), services, podLister, isNodeExpectedInK8sState)
+		nodeStatusReport, ok, err := makeNodeStatusReport(sdc, rackSpec, int(ord), services, podLister)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("can't make node status report for node %d of rack %q of ScyllaDBDatacenter %q: %w", ord, rackSpec.Name, naming.ObjRef(sdc), err))
 			continue
@@ -2475,17 +2466,18 @@ func makeRackNodesStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec 
 
 // makeNodeStatusReport creates a NodeStatusReport for a specific node in a rack.
 // It returns an optional NodeStatusReport, a boolean indicating whether the NodeStatusReport is non-nil, and an error.
-func makeNodeStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec *scyllav1alpha1.RackSpec, ordinal int, services map[string]*corev1.Service, podLister corev1listers.PodLister, isNodeExpectedInK8sState bool) (*scyllav1alpha1.NodeStatusReport, bool, error) {
-	var hostID string
+// A node is included in the report only while its Service is annotated with naming.NodeJoinedScyllaDBClusterAnnotation
+// set to true, signaling it is a member of the ScyllaDB cluster.
+func makeNodeStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec *scyllav1alpha1.RackSpec, ordinal int, services map[string]*corev1.Service, podLister corev1listers.PodLister) (*scyllav1alpha1.NodeStatusReport, bool, error) {
 	svcName := naming.MemberServiceName(*rackSpec, sdc, ordinal)
-	svc, svcExists := services[svcName]
-	if svcExists {
-		hostID = svc.Annotations[naming.HostIDAnnotation]
+	svc, ok := services[svcName]
+	if !ok {
+		klog.V(4).InfoS("Member Service is missing, skipping", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KRef(sdc.Namespace, svcName))
+		return nil, false, nil
 	}
 
-	if !isNodeExpectedInK8sState && len(hostID) == 0 {
-		// The node is not expected to be a part of the cluster in K8s state, and it has no known identity in ScyllaDB. Skip it.
-		klog.V(5).InfoS("Node is not expected to be part of the cluster in Kubernetes state and has no known identity in ScyllaDB, skipping", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KRef(sdc.Namespace, svcName))
+	if svc.Annotations[naming.NodeJoinedScyllaDBClusterAnnotation] != naming.LabelValueTrue {
+		klog.V(5).InfoS("Node has not yet been annotated as a member of the ScyllaDB cluster, skipping", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KObj(svc), "AnnotationKey", naming.NodeJoinedScyllaDBClusterAnnotation)
 		return nil, false, nil
 	}
 
@@ -2493,14 +2485,15 @@ func makeNodeStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec *scyl
 		Ordinal: ordinal,
 	}
 
+	hostID := svc.Annotations[naming.HostIDAnnotation]
 	if len(hostID) == 0 {
-		// Host ID hasn't been propagated yet, report an empty status without a hostID.
-		klog.V(4).InfoS("HostID of an expected node has nod been propagated yet, reporting an empty status", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KRef(sdc.Namespace, svcName))
+		// HostID hasn't been propagated yet, report an empty status.
+		// It should never happen unless the annotation has been manually removed.
+		klog.Warningf("Required node %q of ScyllaDBDatacenter %q has no HostID annotation, reporting an empty status", klog.KObj(svc), klog.KObj(sdc))
 		return nodeStatusReport, true, nil
 	}
 	nodeStatusReport.HostID = &hostID
 
-	// HostID is non-empty, we can now safely use the svc object.
 	podName := naming.PodNameFromService(svc)
 	pod, err := podLister.Pods(sdc.Namespace).Get(podName)
 	if err != nil {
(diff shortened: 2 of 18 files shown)


pkg/naming/constants.go
```diff
@@ -49,6 +49,11 @@ const (
 
 	// NodeStatusReportAnnotation reflects the current status report from the ScyllaDB node.
 	NodeStatusReportAnnotation = "internal.scylla.scylladb.com/scylladb-node-status-report"
+
+	// IncludeInStatusReportAnnotation reflects whether the node's operation mode has reached or passed NORMAL at least once.
+	// If set to "true", the node is considered to have joined the cluster and should be included in the status report.
+	// It is sticky: once set to "true", it should never be cleared or downgraded.
+	IncludeInStatusReportAnnotation = "internal.scylla-operator.scylladb.com/include-in-status-report"
```
czeslavo: I'd suggest renaming this to directly express its underlying meaning, e.g., `/node-has-reached-normal-operation-mode`.

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -2475,32 +2466,29 @@ func makeRackNodesStatusReport(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackSpec
...
-	svc, svcExists := services[svcName]
-	if svcExists {
-		hostID = svc.Annotations[naming.HostIDAnnotation]
-	}
-
-	if !isNodeExpectedInK8sState && len(hostID) == 0 {
-		// The node is not expected to be a part of the cluster in K8s state, and it has no known identity in ScyllaDB. Skip it.
-		klog.V(5).InfoS("Node is not expected to be part of the cluster in Kubernetes state and has no known identity in ScyllaDB, skipping", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KRef(sdc.Namespace, svcName))
+	svc, ok := services[svcName]
+	if !ok || svc.Annotations[naming.IncludeInStatusReportAnnotation] != naming.LabelValueTrue {
```
czeslavo: Let's consider adding a dedicated field to the status report that will flag such a node so that downstream we can decide based on that whether we need to wait for it or not. Skipping it completely feels to me like losing information unnecessarily.

pkg/naming/constants.go
rzetelskik: I purposefully made it abstract because it doesn't reflect a current state of the node at all times. E.g. if you replace a node the proposed `node-has-reached-normal-operation-mode` doesn't hold its meaning any longer. I'd keep it as an explicit signal for inclusion in the status report. I can change the wording to something like expected/required.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: We don't really loose information, because at that point the only thing in the report would be the ordinal and optionally a HostID. Perhaps the status report is a bit more confusing though, as it only includes nodes past NORMAL operation mode.
The most intuitive solution would be to include the current state in the report, and let the barrier decide if it should include it, but then we run into issues with losing the stickiness.
We could add a new field, but having an empty entry with "Required/Excluded" field is isomorphic with no entry.
And the field has to be something that defaults to "false" to avoid issues with operator/CRD version skew - so something like "Exclude" - but I'm not a fan as it leaks the barrier implementation into the CR.
Let's discuss on a call tomorrow.

pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
```diff
@@ -37,11 +37,15 @@ type RackNodesStatusReport struct {
 	Name string `json:"name"`
 
 	// Nodes holds the list of node status reports collected from nodes from this rack.
+	// Membership here refers to the ScyllaDB cluster itself, not to the Kubernetes objects representing it: a node is
+	// reported only once the corresponding ScyllaDB node has joined the cluster and owns tokens in it.
+	// The absence of an entry doesn't imply the corresponding node is missing or unhealthy.
```
czeslavo: Nit: wording could be more precise to make it clear what we mean by `corresponding node`.

pkg/controller/sidecar/sync.go
```diff
@@ -123,10 +124,62 @@ func (c *Controller) syncAnnotations(ctx context.Context, svc *corev1.Service) e
...
+	case !isKnown:
+		// The node hasn't shown up in the cluster's token metadata yet. Leave the annotation as is and retry later.
+		klog.V(2).InfoS("Node with the expected HostID hasn't joined the ScyllaDB cluster yet, will retry in a bit", "Service", klog.KObj(svc), "HostID", hostID)
+		c.queue.AddRateLimited(c.key)
+
+	case !isMember:
+		metav1.SetMetaDataAnnotation(&svcCopy.ObjectMeta, naming.NodeJoinedScyllaDBClusterAnnotation, naming.LabelValueFalse)
+
+		klog.V(4).InfoS("Node doesn't own any tokens, looks like it's still bootstrapping, requeuing", "Service", klog.KObj(svc), "HostID", hostID)
+		c.queue.AddAfter(c.key, requeueWaitDuration)
```
czeslavo: Why not `AddAfter` in both cases?

pkg/controller/sidecar/sync.go
```diff
@@ -123,10 +124,62 @@ func (c *Controller) syncAnnotations(ctx context.Context, svc *corev1.Service) e
...
+
+	if !equality.Semantic.DeepEqual(svc, svcCopy) {
+		_, err = c.kubeClient.CoreV1().Services(svcCopy.Namespace).Update(ctx, svcCopy, metav1.UpdateOptions{})
+		if err != nil {
+			errs = append(errs, fmt.Errorf("can't update service %q: %w", naming.ObjRef(svc), err))
+		} else {
+			klog.V(2).InfoS("Successfully updated service annotations", "Service", klog.KObj(svc))
+		}
+	}
+
```
czeslavo: Could we extract this to a separate, well-named method so it's clear at first glance what's happening here, and reduce the syncAnnotations's length?

pkg/controller/sidecar/sync.go
```diff
@@ -123,10 +124,62 @@ func (c *Controller) syncAnnotations(ctx context.Context, svc *corev1.Service) e
 		return fmt.Errorf("can't get HostID: %w", err)
 	}
 
-	ipToHostIDMap, err := scyllaClient.GetIPToHostIDMap(ctx, c.localhostAddress)
+	svcCopy := svc.DeepCopy()
+	metav1.SetMetaDataAnnotation(&svcCopy.ObjectMeta, naming.HostIDAnnotation, hostID)
+
+	var errs []error
```
czeslavo: Given that we have the switch below and only one case can be executed, why do we need a slice of errors here? IMO, it'd be more readable if all cases were returning on their own:

```go
    isMember, isKnown, err := nodeIsScyllaDBClusterMember(ctx, scyllaClient, c.localhostAddress, hostID)
    if err != nil {
       return fmt.Errorf("can't determine ScyllaDB cluster membership: %w", err)
    }

	switch {
	case !isKnown:
		// The node hasn't shown up in the cluster's token metadata yet. Leave the annotation as is and retry later.
		klog.V(2).InfoS("Node with the expected HostID hasn't joined the ScyllaDB cluster yet, will retry in a bit", "Service", klog.KObj(svc), "HostID", hostID)
		c.queue.AddRateLimited(c.key)
        return nil

	case !isMember:
		metav1.SetMetaDataAnnotation(&svcCopy.ObjectMeta, naming.NodeJoinedScyllaDBClusterAnnotation, naming.LabelValueFalse)

		klog.V(4).InfoS("Node doesn't own any tokens, looks like it's still bootstrapping, requeuing", "Service", klog.KObj(svc), "HostID", hostID)
		c.queue.AddAfter(c.key, requeueWaitDuration)
        return nil

    default:
       return c.syncTokenRingHashAnnotation(...)
```

czeslavo (commented):
🥇 for the sidecar controller envtest. A few nits left.

pkg/controller/sidecar/sync.go
rzetelskik: Which part?

czeslavo: Ah, I missed the fact that the svc update part is outside of the switch. I still feel we could do something to make this function read better.

How about something like below:

```go
// observeRequiredAnnotations reads the ScyllaDB API and returns the annotations required on the member Service,
// how soon the node should be resynced, and an observation error.
// Annotations are observed on a best-effort basis: only those backed by a successful observation are returned, so
// a non-nil error comes with whatever was observed before it and the caller leaves the rest untouched.
func (c *Controller) observeRequiredAnnotations(ctx context.Context, scyllaClient *scyllaclient.Client) (map[string]string, time.Duration, error) {
      hostID, err := c.getHostID(ctx, scyllaClient, c.localhostAddress)
      if err != nil {
              return nil, 0, fmt.Errorf("can't get HostID: %w", err)
      }

      annotations := map[string]string{
              naming.HostIDAnnotation: hostID,
      }

      isMember, isKnown, err := nodeIsScyllaDBClusterMember(ctx, scyllaClient, c.localhostAddress, hostID)
      if err != nil {
              // Membership can't be determined, so the annotation keeps its last observed value.
              return annotations, 0, fmt.Errorf("can't determine ScyllaDB cluster membership: %w", err)
      }

      if !isKnown {
              // The node hasn't shown up in the cluster's tokenion keeps its last observed value.
              klog.V(2).InfoS("Node with the expected HostID hasn't joined the ScyllaDB cluster yet, will retry in a bit", "HostID", hostID)
              return annotations, requeueWaitDuration, nil
      }

      if !isMember {
              klog.V(4).InfoS("Node doesn't own any tokens, looks like it's still bootstrapping, requeuing", "HostID", hostID)
              annotations[naming.NodeJoinedScyllaDBClusterAnnotation] = naming.LabelValueFalse
              return annotations, requeueWaitDuration, nil
      }

      annotations[naming.NodeJoinedScyllaDBClusterAnnotation] = naming.LabelValueTrue

      currentTokenRingHash, err := getTokenRingHash(ctx, scyllaClient, c.localhostAddress)
      if err != nil {
              return annotations, 0, fmt.Errorf("can't get current token ring hash: %w", err)
      }
      annotations[naming.CurrentTokenRingHashAnnotation] = currentTokenRingHash

      return annotations, 0, nil
}

func (c *Controller) updateServiceAnnotations(ctx context.Context, svc *corev1.Service, annotations map[string]string) error {
      svcCopy := svc.DeepCopy()
      for k, v := range annotations {
              metav1.SetMetaDataAnnotation(&svcCopy.ObjectMeta, k, v)
      }

      if equality.Semantic.DeepEqual(svc, svcCopy) {
              return nil
      }

      _, err := c.kubeClient.CoreV1().Services(svcCopy.Namespace.UpdateOptions{})
      if err != nil {
              return fmt.Errorf("can't update Service %q: %w", n
      }

      klog.V(2).InfoS("Successfully updated Service annotations", "Service", klog.KObj(svc))

      return nil
}

func (c *Controller) syncAnnotations(ctx context.Context, svc *c
      startTime := time.Now()
      klog.V(4).InfoS("Started syncing Service annotations", "Service", klog.KObj(svc), "startTime", startTime)
      defer func() {
              klog.V(4).InfoS("Finished syncing Service annotations", "Service", klog.KObj(svc), "duration", time.Since(startTime))
      }()

      scyllaClient, err := c.newScyllaClient()
      if err != nil {
              return fmt.Errorf("can't create a new ScyllaClient: %w", err)
      }
      defer scyllaClient.Close()

      requiredAnnotations, requeueAfter, observeErr := c.observeRequiredAnnotations(ctx, scyllaClient)
      if requeueAfter > 0 {
              c.queue.AddAfter(c.key, requeueAfter)
      }

      updateErr := c.updateServiceAnnotations(ctx, svc, requiredAnnotations)

      return apimachineryutilerrors.NewAggregate([]error{observeErr, updateErr})
}
```

pkg/controller/sidecar/sync.go
rzetelskik: I just preserved the preexisting code. To me this actually looks like a bug. Returning with no errorr results in the controller calling Forget(key), so the exponential backoff doesn't actually accumulate. I'll change to addAfter. I think it's worth verifying if the duration is right though - 5s seems like a lot in this scenario.

pkg/controller/sidecar/sync.go
czeslavo: It's being discussed in https://github.com/scylladb/scylla-operator/pull/3530#discussion_r3710658882. Resolving as duplicate.

pkg/controller/sidecar/sync.go
rzetelskik: done with slight modifications

pkg/controller/sidecar/sync.go
rzetelskik: done, added a comment on the requeue duration

pkg/api/scylla/v1alpha1/types_scylladbdatacenternodesstatusreport.go
rzetelskik: done, ptal

czeslavo (approved):
/lgtm

PR #3532, opened 2026-07-29
test(kind): cap e2e test parallelism

abramche:
- lower Ginkgo parallelism for parallel test suites

Resolves https://scylladb.atlassian.net/browse/OPERATOR-258

### DIFF
--- a/pkg/cmd/tests/tests.go
+++ b/pkg/cmd/tests/tests.go
@@ -17,7 +17,9 @@ import (
 )
 
 const (
-	EnvVarPrefix = "SCYLLA_OPERATOR_TESTS_"
+	EnvVarPrefix               = "SCYLLA_OPERATOR_TESTS_"
+	defaultParallelParallelism = 8
+	defaultSerialParallelism   = 1
 )
 
 // Suites is the canonical list of scylla-operator test suites.
@@ -28,7 +30,7 @@ var Suites = ginkgotest.TestSuites{
 		Description: templates.LongDesc(`
 		Runs all tests.`,
 		),
-		DefaultParallelism: 100,
+		DefaultParallelism: defaultParallelParallelism,
 	},
 	{
 		Name: "scylla-operator/conformance/parallel",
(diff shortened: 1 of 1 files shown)


pkg/cmd/tests/tests.go
```diff
@@ -76,7 +76,7 @@ var Suites = ginkgotest.TestSuites{
 		Relatively fast tests that can be run on kind clusters.
 		`),
 		LabelFilter:        framework.SuiteKindFastLabelName,
-		DefaultParallelism: 60,
+		DefaultParallelism: 30,
```
rzetelskik: There are stability issues in other suites as well. Given we're supposed to enable parallel bootstrap, which will further worsen the initial spike, I'd set this to 8 across all suites (excluding serial) and go from there with potential tail latency improvements (priorities, splitting into more suites, giving it more infra etc). See https://github.com/scylladb/scylla-operator/commit/c161cb4579213b0abda124856d095fd5b0285473

abramche: I yanked your changes in to this PR (only parallelism changes), let's see how do the tests fare. In the original kind run with =30, the tests [were](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3532/pull-scylla-operator-master-e2e-kind-fast/2082471231235297280) quick and passed (not a sample, I know).

rzetelskik:
I suggested the value so I'd like a second pair of eyes for the review.
/auto-cc

rzetelskik:
A quick explanation for why I suggested a more conservative default: trying to find a perfect default for all possible infra configurations is impossible without a variable input. Using a more conservative value for the hardcoded default makes more sense to me than trying to tune it to the highest value retaining stability, i.e. I'd rather have a reasonable, stable baseline and override if necessary.

rzetelskik (approved):

abramche:
I feel a reasonable comfort spot for current hardware is higher (maybe at 15-ish), but I find Kacper's argument about an extra safe baseline convincing: re-running is expensive if just a single test flakes + I do not have capacity atm for an in-depth parallelism exercise (and CI is biting us already, a lot). 

This change as is will immediately improve the situation.

mflendrich (approved):
/lgtm

This is a provisional step towards mitigation of resource contention during massively parallel test executions - let's see how it performs in reality and decide further steps.

abramche:
/retest

the only suite not affected by these changes... 🥲

PR #3534, opened 2026-07-30
fix(SC API): change SM Agent configmap to a secret

abramche:
- as seen by pkg/controller/scyllacluster/migrate.go:235, SC API spec was not meant to be a configmap (it can contain auth token); this PR amends the API, the documentation and generated files

Resolves https://scylladb.atlassian.net/browse/OPERATOR-248

### DIFF
--- a/docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
+++ b/docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
@@ -405,7 +405,7 @@ object
      - resources the Scylla container will use.
    * - scyllaAgentConfig
      - string
-     - Scylla config map name to customize scylla manager agent
+     - ScyllaAgentConfig specifies a reference to custom ScyllaDB Manager Agent configuration stored as Secret.
    * - scyllaConfig
      - string
      - Scylla config map name to customize scylla.yaml
--- a/pkg/api/scylla/v1/types_cluster.go
+++ b/pkg/api/scylla/v1/types_cluster.go
@@ -544,7 +544,7 @@ type RackSpec struct {
 	// Scylla config map name to customize scylla.yaml
 	ScyllaConfig string `json:"scyllaConfig"`
 
-	// Scylla config map name to customize scylla manager agent
+	// ScyllaAgentConfig specifies a reference to custom ScyllaDB Manager Agent configuration stored as Secret.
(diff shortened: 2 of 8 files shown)


rzetelskik:
Agent config is supposed to be provided as a Secret since https://github.com/scylladb/scylla-operator/pull/62. Wild that no one ever complained about the field description. Anyway, I believe this completely settles the debate that aligning the docs with the actual behaviour is the correct approach.

docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
```diff
@@ -405,7 +405,7 @@ object
      - resources the Scylla container will use.
    * - scyllaAgentConfig
      - string
-     - Scylla config map name to customize scylla manager agent
+     - Secret name containing ScyllaDB Manager Agent custom config
```
rzetelskik: this is an autogenerated change, please move to the right commit

pkg/api/scylla/v1/types_cluster.go
```diff
@@ -544,7 +544,7 @@ type RackSpec struct {
 	// Scylla config map name to customize scylla.yaml
 	ScyllaConfig string `json:"scyllaConfig"`
 
-	// Scylla config map name to customize scylla manager agent
+	// Secret name containing ScyllaDB Manager Agent custom config
```
rzetelskik: I think it deserves a brief changelog entry given it's a user facing change which might raise some eyebrows

abramche: 183c5b047d7844c85359daefc0a44e7660b93b38

docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
abramche: done, missed accidentally

pkg/api/scylla/v1/types_cluster.go
rzetelskik: sorry for not pointing that out in the first pass - the wording could be a bit better, preferrably aligned with SDBDC's `CustomConfigSecretRef` "ScyllaAgentConfig specifies a reference to custom ScyllaDB Manager Agent configuration stored as Secret."

abramche: 0adf36024ae5c7ca791af38338ce7174fe5e9934

PR #3535, opened 2026-07-31
fix(sidecar): check node state before fetching HostID map

### LINKED ISSUE #3020: Sidecar should directly check if node joined the cluster before fetching IP to HostID map
### Context

One of the sidecar controller's responsibilities is syncing Scylla node Service annotations (host ID and current token ring hash). Currently, if the IP to HostID mapping doesn't return the local node among its values, we extrapolate this condition as "the node hasn't joined the cluster yet", and we report it as such in logs.

According to Scylla developers, when the API is available and the node status is normal, this mapping should always return the local node. 

The goal of this issue is to make this logic _fully correct_ by explicitly verifying whether the node is normal state before fetching the mapping. This way, if the node isn't in a normal state, we can log, as we do now, that the node hasn't joined the cluster yet, and we have to wait for it. But, if the node is in a normal state, and then the IP to HostID mapping doesn't include it, we can return an error, as it's an unexpected condition.

### Acceptance criteria
- [ ] Sidecar controller explicitly verifies whether a node joined the cluster by checking its state is normal
- [ ] Sidecar controller logs that the node hasn't joined the cluster yet when the node's state is not normal
- [ ] Sidecar controller returns an error if the node's state is normal and the IP to HostID mapping doesn't include the expected hostID

vnaruvan:

This PR updates the sidecar annotation sync flow to check the local ScyllaDB node operation mode when the node's HostID is missing from the IP-to-HostID map.

Previously, if the local HostID was missing from the IP-to-HostID map, the sidecar assumed the node had not joined the cluster yet and requeued. This change makes that check explicit: the sidecar now waits while the node is not in `NORMAL` operation mode, and returns an error if the node is already `NORMAL` but the expected HostID is still missing from the mapping.

Changes:
- Fetch the IP-to-HostID map before checking the local node operation mode.
- Check the local node operation mode only when the local HostID is missing from the IP-to-HostID map.
- Requeue while the node is not in `NORMAL` operation mode.
- Return an error if the node is already in `NORMAL` mode but its HostID is still missing from the IP-to-HostID map.

Testing:
- `go test ./pkg/controller/sidecar -run TestNodeIsScyllaDBClusterMember -count=1`
- `go test -tags=envtest ./test/envtest/controllers/... -count=1 -timeout=30m -ginkgo.focus='SidecarController'`

Resolves #3020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
   wrapper incorrectly compared against `OperationalModeDecommissioned` instead of `OperationalModeDecommissioning`,
   which resulted in an unhandled error being logged.
   [#3538](https://github.com/scylladb/scylla-operator/pull/3538)
+- Fixed sidecar annotation sync to check the local ScyllaDB node operation mode before fetching the IP-to-HostID map.
+  Nodes that are not yet in `NORMAL` mode are requeued, while a missing HostID for a `NORMAL` node is surfaced as an error.
+  [#3535](https://github.com/scylladb/scylla-operator/pull/3535)
 - Fixed nodes being skipped by cleanup after the token ring changed. The operator seeded a node's last cleaned up token
   ring hash with whatever ring state its sidecar sampled first, which exempted that node from cleanup of the ring it
   first observed. Which nodes ended up exempt depended on the race between node joins and the time for sidecar to propagate the token ring hash.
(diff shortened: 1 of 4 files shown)


vnaruvan:
Thanks, I’ll rebase and resolve the conflict.

vnaruvan:
Updated the envtest fake ScyllaDB API to serve `/storage_service/operation_mode`.
The focused SidecarController envtest suite now passes locally:
`SUCCESS! -- 14 Passed | 0 Failed | 34 Skipped`.

rzetelskik:
@vnaruvan please bear with me, I'll get to reviewing this once we've added support for parallel bootstrap as this change is on the critical path

vnaruvan:
Thanks for the update, no problem. I’ll wait for the parallel bootstrap support and address any review comments once you get to it.

PR #3546, opened 2026-08-03
test: introduce the upgrade operator test in ginkgo

abramche:
- introduces a kind of upgrade operator test, but in Ginkgo (compare: [GHA version](https://github.com/scylladb/scylla-operator/pull/3527) - note that the test, being run outside k8s, has to `kubectl exec` to connect to internal containers. I found no good way of routing traffic otherwise.
- introduce a separate suite and machinery to run the test
- introduce optional runner arguments, config values and wired machinery for `OPERATOR_UPGRADE_FROM_VERSION` (defaults to the latest minor.patch released version, renovate-managed) and `OPERATOR_UPGRADE_TO_VERSION` (defaults to latest unreleased version). Either can be unreleased versions (i.e. `1.20-latest`, `1.21-latest`), in which case relevant images will be built off tip of the branch and the branch's manifests will be used for deployment.
- small refactor (extract common helper)

NB: the -release action (using the machinery) follows in a [separate PR](https://github.com/scylladb/scylla-operator-release/pull/655). This PR is blocking the release PR and should be merged first.

Part of https://scylladb.atlassian.net/browse/OPERATOR-274

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,7 @@ GO_LD_FLAGS ?=-ldflags '$(strip $(call version-ldflags,$(GO_PACKAGE)/pkg/build) 
 RENOVATE_CONFIG ?=./renovate.json
 
 KIND_CLUSTER_NAME=scylla-operator-e2e
+KIND_OPERATOR_UPGRADE_CLUSTER_NAME=kind-operator-upgrade
 
 # TODO: look into how to make these local to the targets
 export DOCKER_BUILDKIT :=1
@@ -877,3 +878,8 @@ kind-teardown:
 test-e2e-kind: kind-setup
 	CLUSTER_NAME=$(KIND_CLUSTER_NAME) ./hack/kind/run-e2e-tests.sh
 .PHONY: test-e2e-kind
+
+# This target runs a modified kind-setup on its own (no auto deployment)
+test-e2e-kind-operator-upgrade:
+	CLUSTER_NAME=$(KIND_OPERATOR_UPGRADE_CLUSTER_NAME) ./hack/kind/run-e2e-operator-upgrade.sh
+.PHONY: test-e2e-kind-operator-upgrade
--- a/assets/config/config.yaml
+++ b/assets/config/config.yaml
@@ -13,6 +13,10 @@ operator:
   grafanaDefaultPlatformDashboard: "scylladb-2026.2/scylla-overview.2026.2.json"
   prometheusVersion: "v3.12.0" # Generated from scylla-monitoring/versions.sh PROMETHEUS_VERSION.
 operatorTests:
+  operatorVersions:
+    # renovate: datasource=github-releases depName=scylla-operator packageName=scylladb/scylla-operator versioning=semver
+    upgradeFrom: "1.21.0" # latest released minor.patch, the version operator-upgrade tests upgrade from
+    upgradeTo: "latest" # the version operator-upgrade tests upgrade to
   scyllaDBVersions:
     updateFrom: "2026.2.1" # One patch lower than .operator.scyllaDBVersion (if no patches yet for this minor, use a release candidate)
     upgradeFrom: "2026.1.9" # One minor lower than .operator.scyllaDBVersion
--- a/hack/kind/run-e2e-operator-upgrade.sh
+++ b/hack/kind/run-e2e-operator-upgrade.sh
@@ -5,7 +5,8 @@
 set -euExo pipefail
 shopt -s inherit_errexit
 
-readonly repo_root="$( dirname "${BASH_SOURCE[0]}" )/../.."
+# Absolute, so it stays valid when the test is run from the build root below.
+readonly repo_root="$( realpath "$( dirname "${BASH_SOURCE[0]}" )/../.." )"
 
 # Ensure all kind calls use podman.
 export KIND_EXPERIMENTAL_PROVIDER=podman
@@ -17,6 +18,9 @@ fi
 
 REENTRANT="${REENTRANT:-true}"
 export REENTRANT
+
+# Release branch worktrees created by resolve-operator-version below; removed by cleanup.
+worktrees=()
 
 # cleanup collects artifacts (best-effort, before teardown as it needs the cluster) and always tears the cluster
 # down, preserving the original exit code. Unlike other kind e2e tests, this one dirties the environment (deploys
@@ -30,6 +34,9 @@ function cleanup {
     must_gather_image="docker.io/scylladb/scylla-operator:${must_gather_image}"
   fi
   ( gather-artifacts-on-exit "${must_gather_image}" ) || true
+  for worktree in "${worktrees[@]}"; do
+    git -C "${repo_root}" worktree remove --force "${worktree}" || true
+  done
   "${repo_root}/hack/kind/cluster-teardown.sh" || true
   rm -f "${KUBECONFIG:-}" || true
   exit "${exit_code}"
@@ -70,16 +77,44 @@ ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}/upgrade-operator"
 mkdir -p "${ARTIFACTS}"
 export ARTIFACTS
 
-# Optional upgrade target override (e.g. parsed from the triggering CI comment). A value other than "latest"
-# (a released version or a full image ref) is used directly as the upgrade target; unset or "latest" builds the
-# current tree and pushes it to the local registry. Env name matches run-e2e in hack/.ci/lib/e2e.sh.
-if [ "${OPERATOR_UPGRADE_TO_VERSION:-latest}" == "latest" ]; then
-  OPERATOR_UPGRADE_TO_VERSION=""
-fi
-build-and-push-operator-image "${repo_root}" OPERATOR_UPGRADE_TO_VERSION
-
-# Version to upgrade from; env-overridable, defaulting from the config assets like run-e2e in hack/.ci/lib/e2e.sh.
+# Versions to upgrade between; env-overridable (e.g. parsed from the triggering CI comment), defaulting from the
+# config assets like run-e2e in hack/.ci/lib/e2e.sh. Each accepts a released version, a full image ref, "latest"
+# (build the current tree) or "<major.minor>-latest" (e.g. "1.21-latest": build the tip of the corresponding
+# release branch, v1.21); the "-latest" forms are kind-runner-only, as they require building an image.
 OPERATOR_UPGRADE_FROM_VERSION="${OPERATOR_UPGRADE_FROM_VERSION:-$( yq '.operatorTests.operatorVersions.upgradeFrom' "${repo_root}/assets/config/config.yaml" )}"
+OPERATOR_UPGRADE_TO_VERSION="${OPERATOR_UPGRADE_TO_VERSION:-latest}"
+
+# resolve-operator-version resolves the "latest"/"<major.minor>-latest" forms in the version variable named by $1:
+# it builds the image from the corresponding tree (the current one, or a temporary release branch worktree) and
+# replaces the value with the SHA-pinned image ref. The tree is stored in the variable named by $2 — the test runs
+# that version's deploy script from it, so the deployed manifests match the tree the image was built from.
+# Released versions and full image refs are left as-is; a released version deploys via hack/ci-deploy-release.sh,
+# which resolves the manifests from the image's OCI source/revision labels — the release's exact git SHA.
+function resolve-operator-version {
+  local -n version="${1:?Missing version variable name}"
+  local -n deploy_dir="${2:?Missing deploy dir variable name}"
+
+  deploy_dir="${repo_root}"
+  if [[ "${version}" =~ ^([0-9]+\.[0-9]+)-latest$ ]]; then
+    local release_branch="v${BASH_REMATCH[1]}"
+    deploy_dir="$( mktemp -d )"
+    worktrees+=( "${deploy_dir}" )
+    # Fetch by URL: CI checkouts (Prow clonerefs) have no "origin" remote, and a local clone's
+    # "origin" may be a fork without the release branches.
+    git -C "${repo_root}" fetch https://github.com/scylladb/scylla-operator.git "${release_branch}"
+    git -C "${repo_root}" worktree add --detach "${deploy_dir}" FETCH_HEAD
+    version=""
+  elif [ "${version}" == "latest" ]; then
+    version=""
+  else
+    # A released version or a full image ref needs no build.
+    return 0
+  fi
+  build-and-push-operator-image "${deploy_dir}" "${1}"
+}
+
+resolve-operator-version OPERATOR_UPGRADE_FROM_VERSION operator_upgrade_from_deploy_dir
+resolve-operator-version OPERATOR_UPGRADE_TO_VERSION operator_upgrade_to_deploy_dir
 
 apply-e2e-workarounds
 
@@ -96,4 +131,6 @@ go run "${repo_root}/cmd/scylla-operator-tests" run kind-operator-upgrade \
   --scyllacluster-storageclass-name=standard \
   --scyllacluster-reactor-backend=io_uring \
   --operator-upgrade-from-version="${OPERATOR_UPGRADE_FROM_VERSION}" \
-  --operator-upgrade-to-version="${OPERATOR_UPGRADE_TO_VERSION}"
+  --operator-upgrade-from-deploy-dir="${operator_upgrade_from_deploy_dir}" \
+  --operator-upgrade-to-version="${OPERATOR_UPGRADE_TO_VERSION}" \
+  --operator-upgrade-to-deploy-dir="${operator_upgrade_to_deploy_dir}"
--- a/hack/kind/run-e2e-operator-upgrade.sh
+++ b/hack/kind/run-e2e-operator-upgrade.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+
+# Copyright (C) 2026 ScyllaDB
+
+set -euExo pipefail
+shopt -s inherit_errexit
+
+readonly repo_root="$( dirname "${BASH_SOURCE[0]}" )/../.."
+
+# Ensure all kind calls use podman.
+export KIND_EXPERIMENTAL_PROVIDER=podman
+
+if [ -z "${CLUSTER_NAME}" ]; then
+  echo "CLUSTER_NAME must be set" > /dev/stderr
+  exit 1
+fi
+
+REENTRANT="${REENTRANT:-true}"
+export REENTRANT
+
+# cleanup collects artifacts (best-effort, before teardown as it needs the cluster) and always tears the cluster
+# down, preserving the original exit code. Unlike other kind e2e tests, this one dirties the environment (deploys
+# and upgrades the operator stack), so it must not leave the cluster around for reuse.
+function cleanup {
+  local exit_code=$?
+  # The must-gather image has to be a full ref; resolve a bare released version against the released repository
+  # (mirroring getOperatorImageRef in the test).
+  local must_gather_image="${OPERATOR_UPGRADE_TO_VERSION:-}"
+  if [ -n "${must_gather_image}" ] && [[ "${must_gather_image}" != */* ]]; then
+    must_gather_image="docker.io/scylladb/scylla-operator:${must_gather_image}"
+  fi
+  ( gather-artifacts-on-exit "${must_gather_image}" ) || true
+  "${repo_root}/hack/kind/cluster-teardown.sh" || true
+  rm -f "${KUBECONFIG:-}" || true
+  exit "${exit_code}"
+}
+
+# The test deploys the operator stack itself (released version, then the upgrade target), so cluster-setup.sh only
+# prepares the cluster and registry. Force a fresh cluster (RECREATE) in case a previous run was killed before its
+# teardown trap fired and left a dirty cluster behind.
+export SO_SKIP_DEPLOYMENT=true
+export RECREATE=true
+"${repo_root}/hack/kind/cluster-setup.sh"
+trap cleanup EXIT
+
+# Set KUBECONFIG to point to the kind cluster.
+KUBECONFIG="$(mktemp --suffix ".kubeconfig")"
+kind get kubeconfig --name="${CLUSTER_NAME}" > "${KUBECONFIG}"
+export KUBECONFIG
+
+source "${repo_root}/hack/kind/lib.sh"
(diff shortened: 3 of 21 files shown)


pkg/cmd/tests/tests.go
```diff
@@ -30,63 +28,72 @@ var Suites = ginkgotest.TestSuites{
...
 		Description: templates.LongDesc(`
 		Tests that verify ScyllaDBMonitoring functionality, possible to be run on kind clusters.
 		`),
 		LabelFilter:        framework.SuiteKindScyllaDBMonitoringLabelName,
-		DefaultParallelism: defaultParallelParallelism,
+		DefaultParallelism: 8,
+	},
+	{
+		Name: "kind-operator-upgrade",
+		Description: templates.LongDesc(`
```
czeslavo, line 93: We should document that this suite is expected to be run in the repository root's context.

assets/config/config.yaml
```diff
@@ -1,21 +1,23 @@
...
   # renovate: datasource=docker depName=scylla-manager-agent packageName=docker.io/scylladb/scylla-manager-agent versioning=semver
   scyllaDBManagerAgentVersion: "3.11.2@sha256:9fd3ba869f1d610b5f3f635a958ecfd3083c7abb8c9cdb12d0d26c6add0b67d9"
   bashToolsImage: "registry.access.redhat.com/ubi9/ubi:9.5-1745854298@sha256:f4ebd46d3ba96feb016d798009e1cc2404c3a4ebdac8b2479a2ac053e59f41b4"
-  grafanaImage: "docker.io/grafana/grafana:12.4.6" # Generated from scylla-monitoring/versions.sh GRAFANA_VERSION.
+  grafanaImage: "docker.io/grafana/grafana:12.4.5" # Generated from scylla-monitoring/versions.sh GRAFANA_VERSION.
   grafanaDefaultPlatformDashboard: "scylladb-2026.2/scylla-overview.2026.2.json"
   prometheusVersion: "v3.12.0" # Generated from scylla-monitoring/versions.sh PROMETHEUS_VERSION.
 operatorTests:
+  # renovate: datasource=github-releases depName=scylla-operator packageName=scylladb/scylla-operator versioning=semver
+  operatorVersion: "1.21.0" # latest minor.patch release
```
czeslavo: I suggest being more explicit about this name, as its usage is well-defined (e.g., `operatorUpgradeFromVersion`).

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -0,0 +1,75 @@
...
+apply-e2e-workarounds
+
+# Run the test directly on the host (not in a pod) because it needs access to the repo
+# to call hack/ci-deploy.sh for the operator upgrade.
+go run "${repo_root}/cmd/scylla-operator-tests" run kind-operator-upgrade \
+  --kubeconfig="${KUBECONFIG}" \
+  --loglevel=2 \
+  --color=false \
+  --artifacts-dir="${ARTIFACTS}" \
+  --timeout=20m \
```
czeslavo: Let's make this timeout a bit longer, as 20m is already set at the test case level. We should leave some time to test setup/artifacts gathering, etc.

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -0,0 +1,75 @@
...
+# Build and push the upgrade target image to the local registry if SO_UPGRADE_IMAGE is not set.
+if [ -z "${SO_UPGRADE_IMAGE:-}" ]; then
+  SO_UPGRADE_IMAGE="localhost:5001/scylladb/scylla-operator:e2e-upgrade-$( date +%Y%m%d%H%M%S )"
+  echo "Building upgrade operator image: ${SO_UPGRADE_IMAGE}"
+  podman build --format docker -t "${SO_UPGRADE_IMAGE}" -f "${repo_root}/Dockerfile" "${repo_root}"
+  echo "Pushing upgrade operator image to local registry: ${SO_UPGRADE_IMAGE}"
+  podman push --tls-verify=false "${SO_UPGRADE_IMAGE}"
+else
+  echo "Using existing upgrade operator image: ${SO_UPGRADE_IMAGE}"
+fi
```
czeslavo: Please consider adapting `build-and-push-operator-image` so it can be reused here to not duplicate these commands.

test/e2e/set/operatorupgrade/operator_upgrade.go
```diff
@@ -0,0 +1,81 @@
...
+
+	g.BeforeEach(func(ctx context.Context) {
+		f = framework.NewFramework(ctx, "operator-upgrade")
+	})
+
+	g.It("should upgrade the operator from the latest release to the current version", g.SpecTimeout(testTimeout), func(ctx g.SpecContext) {
+		targetImageRef := framework.TestContext.TargetOperatorImageRef
+		o.Expect(targetImageRef).NotTo(o.BeEmpty(), "--target-operator-image-ref must be set for operator upgrade tests")
+
+		// The released operator stack is already deployed by hack/kind/cluster-setup.sh before this test runs.
```
czeslavo: To be sure we test what we intended, let's add an assertion that will verify that the operator version that's deployed matches `Project.OperatorTests.OperatorVersion`.

BTW, if I'm not mistaken, `Project.OperatorTests.OperatorVersion` isn't used anywhere right now. To make that work for us, I suggest we refactor `cluster-setup.sh`  a bit by adding a `--install-scylla-operator` (default on) switch. 

That way we could call it with `--install-scylla-operator=false` from `run-e2e-operator-upgrade.sh` and make the initial version `ci-deploy.sh` part of the test itself (allowing us to pass `{from}` version from the Go struct)

test/e2e/set/operatorupgrade/operator_upgrade.go
```diff
@@ -0,0 +1,81 @@
...
+		sc, err = controllerhelpers.WaitForScyllaClusterState(initialRolloutCtx, f.ScyllaAdminClient().ScyllaV1().ScyllaClusters(sc.Namespace), sc.Name, controllerhelpers.WaitForStateOptions{}, utils.IsScyllaClusterRolledOut)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Upgrading operator stack to image %s via ci-deploy.sh", targetImageRef)
+		artifactsDir, err := os.MkdirTemp("", "operator-upgrade-artifacts-")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		defer func() {
+			err = os.RemoveAll(artifactsDir)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}()
```
czeslavo: This will remove all the artifacts despite the test result. Was that intended? If we want to preserve the artifacts, we should rather make a dedicated directory under `--artifacts-dir` that the runner gets.

abramche:
@czeslavo all of the comments have been addressed in 4cf060ed1e818e8d9743caf454f4c0fba1ccc9f1 - however I needed to rebase onto master and there was a big conflicting change concerning lib.sh and other machinery (due to kind registry port override) so I had to squash the changes

Makefile
```diff
@@ -862,3 +862,7 @@ kind-teardown:
 test-e2e-kind: kind-setup
 	CLUSTER_NAME=$(KIND_CLUSTER_NAME) ./hack/kind/run-e2e-tests.sh
 .PHONY: test-e2e-kind
+
+test-e2e-kind-operator-upgrade: kind-setup
```
czeslavo: `SO_SKIP_DEPLOYMENT` should be set to true, as now we're first deploying the current operator build, then downgrading it in the test (via ci-deploy.sh) and upgrading again to the current build.

Maybe it would make sense to make a call to `./hack/kind/cluster-setup.sh` part of the `./hack/kind/run-e2e-operator-upgrade.sh` script? We could also make it tear down the kind cluster (`./hack/kind/cluster-teardown.sh` in a trap) after the test concludes so we never run the test on a cluster with the operator installed.

Also, we could use a dedicated cluster name for upgrade tests so it doesn't interfere with other e2es that are ok to reuse a cluster.

test/e2e/set/operatorupgrade/operator_upgrade.go
```diff
@@ -0,0 +1,99 @@
...
+		f = framework.NewFramework(ctx, "operator-upgrade")
+	})
+
+	g.It("should upgrade the operator from the latest release to the current version", g.SpecTimeout(testTimeout), func(ctx g.SpecContext) {
+		targetImageRef := framework.TestContext.TargetOperatorImageRef
+		o.Expect(targetImageRef).NotTo(o.BeEmpty(), "--target-operator-image-ref must be set for operator upgrade tests")
+		releasedImageRef := fmt.Sprintf("docker.io/scylladb/scylla-operator:%s", configassets.Project.OperatorTests.OperatorUpgradeFromVersion)
+
+		// Not cleaned up on purpose: in CI the pod is discarded anyway, and locally the artifacts are worth keeping.
+		artifactsDir, err := os.MkdirTemp("", "operator-upgrade-artifacts-")
```
czeslavo: Please create a dedicated upgrade artifacts directory under `--artifacts-dir` to make sure that we have them preserved in CI. This way, they will be lost, I'm afraid.

test/e2e/set/operatorupgrade/operator_upgrade.go
```diff
@@ -0,0 +1,99 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		framework.By("Deploying the released operator stack (%s) via ci-deploy.sh", releasedImageRef)
```
czeslavo: Let's consider verifying the test runs against a clean cluster before deploying anything (the operator is not installed). We should have a guarantee that we run against a clean cluster - this is not satisfied and goes unnoticed (as pointed in https://github.com/scylladb/scylla-operator/pull/3546/changes#r3718608176).

Makefile
abramche: the variable is set in https://github.com/scylladb/scylla-operator-release/pull/655/changes#diff-fe8b79a8339c28fc5d00b5a845cec34eb093e8bbbcb46bb96a7f4e0a6d1ddc55R3127, but you are right, this glosses over local testing

abramche: set SO_SKIP_DEPLOYMENT in the run upgrade script (also invokes kind-setup), use a unique cluster name in b9d65f78fff9dd66f573326ad56778f79320ba89

Also forced cluster recreation since it must start with a clean slate each time

test/e2e/set/operatorupgrade/operator_upgrade.go
abramche: addressed in b9d65f78fff9dd66f573326ad56778f79320ba89

test/e2e/set/operatorupgrade/operator_upgrade.go
abramche: it was handled for CI, but now also enforced for local executions in b9d65f78fff9dd66f573326ad56778f79320ba89

test/e2e/set/operatorupgrade/operator_upgrade.go
czeslavo: Can't see it addressed; it's still using `MkdirTemp`.

test/e2e/set/operatorupgrade/operator_upgrade.go
czeslavo: I meant that we could have a check at the test level that ensures that all its prerequisites are satisfied - i.e., there's no operator installed yet. This way, we make it more robust and less prone to changes in the outside scripting.

test/e2e/set/operatorupgrade/operator_upgrade.go
abramche: Ah, I thought the shell changes were enough, but the test was not reading from them

8f09719e9684911184c8f2c4581c9b692e12ee2e

test/e2e/set/operatorupgrade/operator_upgrade.go
abramche: added an assertion
8f09719e9684911184c8f2c4581c9b692e12ee2e

test/e2e/set/operatorupgrade/operator_upgrade.go
```diff
@@ -32,9 +33,12 @@ var _ = g.Describe("Operator Upgrade", framework.SuiteOperatorUpgrade, func() {
...
 
-		// Not cleaned up on purpose: in CI the pod is discarded anyway, and locally the artifacts are worth keeping.
-		artifactsDir, err := os.MkdirTemp("", "operator-upgrade-artifacts-")
-		o.Expect(err).NotTo(o.HaveOccurred())
+		framework.By("Verifying the operator is not deployed yet")
+		_, err := f.KubeAdminClient().AppsV1().Deployments(operatorName).Get(ctx, operatorName, metav1.GetOptions{})
+		o.Expect(apierrors.IsNotFound(err)).To(o.BeTrue(), "expected no operator deployment before the test deploys it, got err=%v", err)
+
+		// ci-deploy.sh writes the rendered deploy manifests here; reuse the suite's artifacts dir (--artifacts-dir).
+		artifactsDir := framework.TestContext.ArtifactsDir
```
czeslavo: Have you verified that if we run `ciDeploy` twice with the same artifacts directory, the second run won't overwrite the first one's manifests in there?

If it happens, let's consider creating well-named dedicated directories under `framework.TestContext.ArtifactsDir`, e.g., `before-upgrade`, `after-upgrade`.

abramche: it will overwrite; my logic was it's not important to preserve the release manifests. it's cheap though, and for sake of completeness', I do not mind

c0429334b586215303e40ab94514e49a0eeccb1c (squashed)

test/e2e/set/operatorupgrade/operator_upgrade.go
```diff
@@ -0,0 +1,105 @@
...
+})
+
+// ciDeploy deploys the operator stack with the given operator image via hack/ci-deploy.sh (run from the repo root).
+// ci-deploy.sh sets the operator image from the ref and blocks on `kubectl rollout status`, so on return the operator
+// is running the target image and available.
+func ciDeploy(ctx context.Context, operatorImageRef, artifactsDir string) {
+	cmd := exec.CommandContext(ctx, "hack/ci-deploy.sh", operatorImageRef)
+	cmd.Env = append(os.Environ(), "REENTRANT=true", fmt.Sprintf("ARTIFACTS=%s", artifactsDir))
+	output, err := cmd.CombinedOutput()
+	o.Expect(err).NotTo(o.HaveOccurred(), "ci-deploy.sh failed: %s", string(output))
```
czeslavo: I haven't spotted that before, but we should use `hack/ci-deploy-release.sh` for deploying the initial version. With `hack/ci-deploy.sh`, we're using manifests from the current branch (CRDs, dependencies, etc.), so we do not really test the complete upgrade path; we only change the operator image version.

abramche: f5c6d6db88d135f549f8bf1191bae15f34d3b894 (also refactored the test a bit to reduce duplicated parts)

notably, CSI driver had to be specifically excluded for the test to run in kind (no XFS storage)

czeslavo:
The test is failing - it reproduces locally for me:
> [FAILED] hack/ci-deploy.sh failed: + shopt -s inherit_errexit
> error: timed out waiting for the condition on scyllaclusters/scylla-manager-cluster

abramche:
> The test is failing - it reproduces locally for me:
> 
> > [FAILED] hack/ci-deploy.sh failed: + shopt -s inherit_errexit
> > error: timed out waiting for the condition on scyllaclusters/scylla-manager-cluster

Yes, there appears to be a long-lived network policy defect which only floats up on Operator's upgrade hook (when SM cluster upgrade config map is present). I've posted details to https://scylladb.atlassian.net/browse/OPERATOR-283 and will test a fix momentarily.

czeslavo:
One thing that I discussed with @mflendrich is that we should not only test an upgrade path from the latest patch to the current master, but also from the previous minor latest patch (relative to the release branch) to the tip of the release branch.

That would mean we'd need to add a job to the feature branch as well and have the version to upgrade from set in the `config.yaml` accordingly (i.e., upgrade from `v1.20` on the v1.21 release branch). Let's see if we can backport it to v1.20 to cover that.

abramche:
> One thing that I discussed with @mflendrich is that we should not only test an upgrade path from the latest patch to the current master, but also from the previous minor latest patch (relative to the release branch) to the tip of the release branch.

Previous, as in 1.20.x -> latest? But we [only](https://operator.docs.scylladb.com/stable/upgrade/upgrade-operator.html#upgrading-scylladb-operator) support N+1 upgrades.

czeslavo:
> > One thing that I discussed with @mflendrich is that we should not only test an upgrade path from the latest patch to the current master, but also from the previous minor latest patch (relative to the release branch) to the tip of the release branch.
> 
> Previous, as in 1.20.x -> latest? But we [only](https://operator.docs.scylladb.com/stable/upgrade/upgrade-operator.html#upgrading-scylladb-operator) support N+1 upgrades.

Previous like v1.20.x -> v1.21 release branch tip.

Edit: the reason behind this requirement is that we support 2 minor versions; therefore, we should validate an upgrade path between the two.

abramche:
> Previous like v1.20.x -> v1.21 release branch tip.

doable via `/test e2e-kind-upgrade-operator 1.20.2 1.21.0` (once [linked](https://github.com/scylladb/scylla-operator-release/pull/655) -release PR gets merged) or, with addition of https://github.com/scylladb/scylla-operator-release/pull/657, with i.e.
```
$ ./hack/run-prowjob.sh --job=pull-scylla-operator-master-e2e-kind-upgrade-operator --pull-number=3546 --env-vars=OPERATOR_UPGRADE_FROM_VERSION=1.20.2,OPERATOR_UPGRADE_TO_VERSION=1.21.0
```

[1.20.2 - 1.21.0 test job](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3546/pull-scylla-operator-master-e2e-kind-upgrade-operator/2085412220400832512)

[(default) 1.21.0 - latest test job](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3546/pull-scylla-operator-master-e2e-kind-upgrade-operator/2085406003888656384)

abramche:
updated for the proposed solution of https://scylladb.atlassian.net/browse/OPERATOR-5 - accept unreleased versions for _FROM and _TO (i.e. support flow for `1.21-latest` to `latest`, `1.20-latest` to `1.21.0`...)

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -0,0 +1,133 @@
...
+# (build the current tree) or "<major.minor>-latest" (e.g. "1.21-latest": build the tip of the corresponding
+# release branch, v1.21); the "-latest" forms are kind-runner-only, as they require building an image.
+OPERATOR_UPGRADE_FROM_VERSION="${OPERATOR_UPGRADE_FROM_VERSION:-$( yq '.operatorTests.operatorVersions.upgradeFrom' "${repo_root}/assets/config/config.yaml" )}"
+OPERATOR_UPGRADE_TO_VERSION="${OPERATOR_UPGRADE_TO_VERSION:-latest}"
+
+# resolve-operator-version resolves the "latest"/"<major.minor>-latest" forms in the version variable named by $1:
+# it builds the image from the corresponding tree (the current one, or a temporary release branch worktree) and
+# replaces the value with the SHA-pinned image ref. The tree is stored in the variable named by $2 — the test runs
+# that version's deploy script from it, so the deployed manifests match the tree the image was built from.
+# Released versions and full image refs are left as-is, deploying from the current tree.
```
czeslavo: That doesn't seem correct to me. Released versions should deploy from the git SHA they were released from.

czeslavo:
> > Previous like v1.20.x -> v1.21 release branch tip.
> 
> doable via `/test e2e-kind-upgrade-operator 1.20.2 1.21.0` (once [linked](https://github.com/scylladb/scylla-operator-release/pull/655) -release PR gets merged) or, with addition of [scylladb/scylla-operator-release#657](https://github.com/scylladb/scylla-operator-release/pull/657), with i.e.
> 
> ```
> $ ./hack/run-prowjob.sh --job=pull-scylla-operator-master-e2e-kind-upgrade-operator --pull-number=3546 --env-vars=OPERATOR_UPGRADE_FROM_VERSION=1.20.2,OPERATOR_UPGRADE_TO_VERSION=1.21.0
> ```
> 
> [1.20.2 - 1.21.0 test job](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3546/pull-scylla-operator-master-e2e-kind-upgrade-operator/2085412220400832512)
> 
> [(default) 1.21.0 - latest test job](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3546/pull-scylla-operator-master-e2e-kind-upgrade-operator/2085406003888656384)

It's nice that it's doable on request; it may be useful in some cases. Anyway, we could port the test to v1.21 release branch and set up a presubmit job there. I'll leave it to you to decide what's better - we'll get to that point anyway after releasing v1.22, as we'll transform master jobs to v1.22.

abramche:
> we could port the test to v1.21 release branch and set up a presubmit job there. 

I'll back port to both 1.20 and 1.21 once this PR is merged 👍

hack/kind/run-e2e-operator-upgrade.sh
abramche: This is a poorly worded comment - what was meant, that the deploy script runs from the current tree (that's the only gotcha); the released versions deploy using `ci-deploy-release.sh` and the script resolves the image to the same git SHA they were released from

reworded in cc1d2a20692dd7b42bb696014e5c71c9647b34ad

czeslavo: Alright, makes sense now, thanks.

abramche:
`1.20-latest` (tip of the branch) to `1.21.0` - https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3546/pull-scylla-operator-master-e2e-kind-upgrade-operator/2085706883250786304

`1.21-latest` (tip of the branch) to `latest` (tip of master) - https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3546/pull-scylla-operator-master-e2e-kind-upgrade-operator/2085711598936133632

czeslavo:
:rocket: 
/lgtm

PR #3555, opened 2026-08-04
feat: modify seed selection algorithm to allow for parallel bootstrap

rzetelskik:
This PR modifies the seed selection algorithm to:
- Align with siren's implementation as decided by @mflendrich in parallel bootstrap discovery. The algorithm is not 1-1 as coordinating a single node bootstrap before the other Pods are created would complicate things much more and it's not really necessary - ScyllaDB can wait until a selected, initially non-serving seed responds. Node's operational mode as input is also replaced by Pod's readiness for simplicity.
- Allow for parallel bootstrap.

Here're the pointers to siren's implementation for reference:
- https://github.com/scylladb/siren/blob/ad5411c44a8591d3474f78d8db88912775768925/cluster/init.go#L86-L158
- https://github.com/scylladb/siren/blob/ad5411c44a8591d3474f78d8db88912775768925/cluster/cluster.go#L1678-L1732

Resolves https://scylladb.atlassian.net/browse/OPERATOR-150

/cc

### DIFF
--- a/pkg/sidecar/identity/member.go
+++ b/pkg/sidecar/identity/member.go
@@ -1,10 +1,13 @@
 package identity
 
 import (
+	"cmp"
 	"context"
 	"fmt"
+	"maps"
 	"slices"
-	"strconv"
+	"strings"
+	"time"
 
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/controllerhelpers"
@@ -12,8 +15,21 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	apimachineryutilerrors "k8s.io/apimachinery/pkg/util/errors"
+	apimachineryutilwait "k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
 )
+
+// seedResolutionPollInterval is how often an unready fallback seed candidate's broadcast address is
+// re-checked while waiting for it to be assigned. It's a fixed, short interval because resolving a
+// broadcast address is quick and there's no benefit in waiting longer between attempts.
+const seedResolutionPollInterval = 500 * time.Millisecond
+
+// seedResolutionPollTimeout bounds how long we wait for an unready fallback seed candidate's broadcast address
+// to become resolvable. In practice the container's startup and liveness probes should fail and restart the
+// process long before this fires, so the timeout is only a defensive backstop against waiting forever.
+const seedResolutionPollTimeout = 15 * time.Minute
 
 // Member encapsulates the identity for a single member
 // of a Scylla Cluster.
@@ -23,7 +39,6 @@ type Member struct {
 	// Namespace of the Pod
 	Namespace     string
 	Rack          string
-	RackOrdinal   int
 	Datacenter    string
 	Cluster       string
 	ServiceLabels map[string]string
@@ -39,20 +54,12 @@ type Member struct {
 }
 
 func NewMember(service *corev1.Service, pod *corev1.Pod, nodesAddressType, clientAddressType scyllav1alpha1.BroadcastAddressType, ipFamily corev1.IPFamily, additionalScyllaDBArguments []string) (*Member, error) {
-	rackOrdinalString, ok := pod.Labels[naming.RackOrdinalLabel]
-	if !ok {
-		return nil, fmt.Errorf("pod %q is missing %q label", naming.ObjRef(pod), naming.RackOrdinalLabel)
-	}
-	rackOrdinal, err := strconv.Atoi(rackOrdinalString)
-	if err != nil {
-		return nil, fmt.Errorf("can't get rack ordinal from label %q: %w", rackOrdinalString, err)
-	}
+	var err error
 
 	m := &Member{
 		Namespace:                   service.Namespace,
 		Name:                        service.Name,
 		Rack:                        pod.Labels[naming.RackNameLabel],
-		RackOrdinal:                 rackOrdinal,
 		Datacenter:                  pod.Labels[naming.DatacenterNameLabel],
 		Cluster:                     pod.Labels[naming.ClusterNameLabel],
 		ServiceLabels:               service.Labels,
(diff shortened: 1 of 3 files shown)


rzetelskik:
/cc @czeslavo 
As you pointed out that you'd like to see `GetSeeds` func body shortened, let's discuss what you have in mind - I suppose we can split the two resolution/polling stages into helpers.

One more thing to consider is that in the current implementation waiting on the fallback to respond can stall indefinitely - this is how ScyllaDB handles the unresponsive seed. It's difficult to see what's happening from the nodes waiting for the seed to respond while using service IPs as broadcast addresses without looking at the entire cluster - this can be observed in  the failing orphaned PV test here: https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3521/pull-scylla-operator-master-e2e-gke-parallel-clusterip/2085425389412290560).

rzetelskik:
Tests with naive parallel bootstrap implementation can be observed here: https://github.com/scylladb/scylla-operator/pull/3521.
They are failing on the discovered issues:
- https://scylladb.atlassian.net/browse/OPERATOR-295
- https://scylladb.atlassian.net/browse/OPERATOR-285
- https://scylladb.atlassian.net/browse/OPERATOR-291

pkg/sidecar/identity/member.go
```diff
@@ -76,78 +78,230 @@ func NewMember(service *corev1.Service, pod *corev1.Pod, nodesAddressType, clien
...
+
+		return true, nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("can't resolve fallback seed broadcast address of %q: %w", selectedFallback.pod.Name, err)
+	}
+
+	seeds = append(seeds, resolved)
+	return seeds, nil
+}
```
czeslavo, line 180: I suggest we split it the following way for readability. Flipped `if !fallback` to make the default path non-indented.

```suggestion
// GetSeeds returns the seeds for this member. External seeds, if provided, are always included.
// If no ready seed candidate exists, it blocks polling until the selected fallback candidate's
// broadcast address becomes resolvable or ctx is canceled.
func (m *Member) GetSeeds(ctx context.Context, coreClient corev1client.CoreV1Interface, externalSeeds []string) ([]string, error) {
	candidates, err := m.getDCSeedCandidates(ctx, coreClient)
	if err != nil {
		return nil, fmt.Errorf("can't get seed candidates: %w", err)
	}
	klog.V(4).InfoS("Found DC seed candidates", "Candidates", seedCandidateNames(candidates))

	// Self takes part in selection like any other node.
	selected, fallback := m.selectDCSeedCandidates(candidates)
	klog.V(2).InfoS("Selected DC seed candidates", "Candidates", seedCandidateNames(selected), "Fallback", fallback)

	if fallback {
		// No ready seed candidates exist, so the first-ranked candidate is selected on its own.
		return m.fallbackToDistiguishedUnreadyCandidate(ctx, coreClient, externalSeeds, selected)
	}

	// All selected candidates are ready, so their broadcast addresses must be resolvable.
	if err := m.ensureAllHaveBroadcastAddress(selected); err != nil {
		return nil, fmt.Errorf("can't resolve seed broadcast addresse: %w", err)
	}

	seeds := make([]string, 0, len(externalSeeds)+len(selected))
	seeds = append(seeds, externalSeeds...)
	return seeds, nil
}
```

pkg/sidecar/identity/member.go
```diff
@@ -76,78 +78,230 @@ func NewMember(service *corev1.Service, pod *corev1.Pod, nodesAddressType, clien
...
+		// Self is the fallback node, and it isn't joining an existing cluster - seed of itself.
+		return []string{m.BroadcastAddress}, nil
+	}
+
+	seeds := make([]string, 0, len(externalSeeds)+1)
+	seeds = append(seeds, externalSeeds...)
+
+	// The fallback candidate isn't self and isn't ready, so its broadcast address may be unresolvable.
+	// Wait for it rather than fail.
+	klog.V(2).InfoS("Waiting for the DC seed candidate's broadcast address to become resolvable", "Candidate", selectedFallback.pod.Name)
```
czeslavo:
```suggestion
	klog.V(2).InfoS("Waiting for the DC fallback seed candidate's broadcast address to become resolvable", "Candidate", selectedFallback.pod.Name)
```

pkg/sidecar/identity/member.go
```diff
@@ -76,78 +78,230 @@ func NewMember(service *corev1.Service, pod *corev1.Pod, nodesAddressType, clien
...
+// seedCandidate is a node in this member's DC, which may end up being selected as its seed.
+type seedCandidate struct {
+	pod *corev1.Pod
+	svc *corev1.Service
+}
+
+// GetSeeds returns the seeds for this member. External seeds, if provided, are always included.
+// If no ready seed candidate exists, it blocks polling until the selected fallback candidate's
+// broadcast address becomes resolvable or ctx is canceled.
+func (m *Member) GetSeeds(ctx context.Context, coreClient corev1client.CoreV1Interface, externalSeeds []string) ([]string, error) {
```
czeslavo, line 95: As you suggested in a private conversation, I agree it would make sense to log the final seeds that are picked after `GetSeeds` returns. The message can be added on the `GetSeeds` caller side.

pkg/sidecar/identity/member.go
```diff
@@ -76,78 +78,230 @@ func NewMember(service *corev1.Service, pod *corev1.Pod, nodesAddressType, clien
...
+	}
+
+	seeds := make([]string, 0, len(externalSeeds)+1)
+	seeds = append(seeds, externalSeeds...)
+
+	// The fallback candidate isn't self and isn't ready, so its broadcast address may be unresolvable.
+	// Wait for it rather than fail.
+	klog.V(2).InfoS("Waiting for the DC seed candidate's broadcast address to become resolvable", "Candidate", selectedFallback.pod.Name)
+	var resolved string
+	err = apimachineryutilwait.PollUntilContextCancel(ctx, seedResolutionPollInterval, true, func(ctx context.Context) (bool, error) {
```
czeslavo: I think it would make sense to run it with a reasonable timeout (15min?) to have a recovery mechanism in case the node never gets an address for some reason. The pod will restart and re-evaluate seeds, maybe picking other nodes that became ready in the meantime.

czeslavo (commented):
> As you pointed out that you'd like to see GetSeeds func body shortened, let's discuss what you have in mind - I suppose we can split the two resolution/polling stages into helpers.

Proposed `GetSeeds` shape in a comment. :+1: 

> One more thing to consider is that in the current implementation waiting on the fallback to respond can stall indefinitely - this is how ScyllaDB handles the unresponsive seed. It's difficult to see what's happening from the nodes waiting for the seed to respond while using service IPs as broadcast addresses without looking at the entire cluster - this can be observed in the failing orphaned PV test here: https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3521/pull-scylla-operator-master-e2e-gke-parallel-clusterip/2085425389412290560).

As I understand it, it is an already existing problem that just exhibits a bit differently with parallel bootstrap, as instead of a single node stuck for serial rollout, we have 3, where two others are stuck waiting for the first one designated as their seed. Perhaps we could think about improving logging so such a case is easier to diagnose?

pkg/sidecar/identity/member.go
rzetelskik: The context should be cancelled now by a startup/liveness probe - do you think it deserves a timeout as well?

pkg/sidecar/identity/member.go
rzetelskik: :+1: done

pkg/sidecar/identity/member.go

pkg/sidecar/identity/member.go
rzetelskik: done

rzetelskik: also added some additional ones in special cases (external seeds only, self only)

pkg/sidecar/identity/member.go
rzetelskik: added a defensive backstop with a comment highlighting the probes are likely to kill it earlier

rzetelskik:
/hold for the optional multi-dc test and the remaining comment

> As I understand it, it is an already existing problem that just exhibits a bit differently with parallel bootstrap, as instead of a single node stuck for serial rollout, we have 3, where two others are stuck waiting for the first one designated as their seed. Perhaps we could think about improving logging so such a case is easier to diagnose?

The thing is we can't really do much, as this happens after we've started the ScyllaDB process. When the resolved broadcast address comes from the Service, from the sidecar's perspective it looks pretty much regular - we start ScyllaDB with seeds. In PodIP case the seed resolution will keep failing occasionally as the Pod may not exist on subsequent live calls.
A possible approach is to terminate resolution early when the Pod enters an Unschedulable state - because at that point it won't get up without manual intervention. But I don't really like this as it breaks the separation of concerns a bit - Pods 1..N check the state of Pod 0 to determine it's own state and fail because of it.

czeslavo:
> The thing is we can't really do much, as this happens after we've started the ScyllaDB process. When the resolved broadcast address comes from the Service, from the sidecar's perspective it looks pretty much regular - we start ScyllaDB with seeds. In PodIP case the seed resolution will keep failing occasionally as the Pod may not exist on subsequent live calls.
A possible approach is to terminate resolution early when the Pod enters an Unschedulable state - because at that point it won't get up without manual intervention. But I don't really like this as it breaks the separation of concerns a bit - Pods 1..N check the state of Pod 0 to determine it's own state and fail because of it.

As we discussed, let's assume the logging and the Pod 0 crashlooping should be enough of a signal to look at the reason for its failure.

PR #3568, opened 2026-08-07
feat(api): add bootstrapPolicy API field

czeslavo:

Extends `ScyllaDBDatacenter` and `ScyllaCluster` CRDs with an optional `bootstrapPolicy` enum field. Setting it to `Parallel` requires Scylla version >= 2026.2.

Managed `ScyllaDBDatanceter`: if its parent doesn't set `bootstrapPolicy` explicitly, we always set `bootstrapPolicy` to `Sequential` to ensure preexisting clusters do not use `Parallel` without opting in when we implement https://scylladb.atlassian.net/browse/OPERATOR-290.

For `ScyllaDBCluster`, we hardcode the value to `Sequential` in child `ScyllaDBDatacenter`s as we won't support parallel bootstrap in automated multi-DC.

Closes https://scylladb.atlassian.net/browse/OPERATOR-148

### DIFF
--- a/pkg/semver/version.go
+++ b/pkg/semver/version.go
@@ -26,7 +26,8 @@ var (
 	// ScyllaDBVersionWithoutNodeExporter is "2026.3.0-0", not "2026.3.0": per semver precedence a pre-release
 	// version is always lower than the same version without one, so a plain "2026.3.0" threshold would reject
 	// pre-GA nightly builds that already ship without node-exporter.
-	ScyllaDBVersionWithoutNodeExporter = semver.MustParse("2026.3.0-0")
+	ScyllaDBVersionWithoutNodeExporter          = semver.MustParse("2026.3.0-0")
+	ScyllaDBVersionRequiredForParallelBootstrap = semver.MustParse("2026.2.0")
 )
 
 // ScyllaVersion contains the version of a cluster with unkown version support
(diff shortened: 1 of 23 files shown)


pkg/semver/version.go
```diff
@@ -27,6 +27,10 @@ var (
 	// version is always lower than the same version without one, so a plain "2026.3.0" threshold would reject
 	// pre-GA nightly builds that already ship without node-exporter.
 	ScyllaDBVersionWithoutNodeExporter = semver.MustParse("2026.3.0-0")
+	// ScyllaDBVersionRequiredForParallelBootstrap is "2026.2.0-0", not "2026.2.0": per semver precedence a pre-release
+	// version is always lower than the same version without one, so a plain "2026.2.0" threshold would reject
+	// pre-GA builds that already support bootstrapping nodes in parallel.
+	ScyllaDBVersionRequiredForParallelBootstrap = semver.MustParse("2026.2.0-0")
```
rzetelskik: I don't think we should care about pre-releases as this is already released. `ScyllaDBVersionWithoutNodeExporter` had to include prereleases as we implemented sidecar before the version was released.
Changing this to 2026.2.0 would simplify validation and be aligned with how we document this.

czeslavo: Changed to 2026.2 and aligned tests.

rzetelskik (approved):
/lgtm

czeslavo:
@rzetelskik, I had to reformat to satisfy `verify` check. PTAL

rzetelskik (approved):
/lgtm

PR #3571, opened 2026-08-07
[Backport to 1.21] test: introduce the upgrade operator test in ginkgo

abramche:
- backports https://github.com/scylladb/scylla-operator/pull/3546

Resolves https://scylladb.atlassian.net/browse/OPERATOR-298

### DIFF
--- a/assets/config/config.yaml
+++ b/assets/config/config.yaml
@@ -11,6 +11,10 @@ operator:
   grafanaDefaultPlatformDashboard: "scylladb-2026.1/scylla-overview.2026.1.json"
   prometheusVersion: "v3.9.1" # Generated from scylla-monitoring/versions.sh PROMETHEUS_VERSION.
 operatorTests:
+  operatorVersions:
+    # renovate: datasource=github-releases depName=scylla-operator packageName=scylladb/scylla-operator versioning=regex:^(?<major>1)\.(?<minor>20)\.(?<patch>\d+)$
+    upgradeFrom: "1.20.2" # latest released 1.20 patch, the version operator-upgrade tests upgrade from
+    upgradeTo: "1.21-latest" # the version operator-upgrade tests upgrade to (tip of the v1.21 release branch)
   scyllaDBVersions:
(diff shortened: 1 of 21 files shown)


assets/config/config.yaml
```diff
@@ -11,6 +11,10 @@ operator:
   grafanaDefaultPlatformDashboard: "scylladb-2026.1/scylla-overview.2026.1.json"
   prometheusVersion: "v3.9.1" # Generated from scylla-monitoring/versions.sh PROMETHEUS_VERSION.
 operatorTests:
+  operatorVersions:
+    # renovate: datasource=github-releases depName=scylla-operator packageName=scylladb/scylla-operator versioning=semver
+    upgradeFrom: "1.21.0" # latest released minor.patch, the version operator-upgrade tests upgrade from
+    upgradeTo: "latest" # the version operator-upgrade tests upgrade to
```
rzetelskik: isn't this effectively causing branch 1.21 to test master? Shouldn't this be 1.20->1.21 on branch 1.21?
I think it could easily be generated btw, as upgradeTo should probably match the IMAGE_TAG in Makefile (out of scope of backporting these specific changes though)

abramche: yes, good point, also restricted renovate not to bump above 1.20.x (in the 1.20 backport PR as well)

now that you mention it, I discovered a loophole that I somehow overlooked and there is a fallback to the latest upgradeTo in `hack/kind/run-e2e-operator-upgrade.sh:85` - fixed in both places to force read from config.yaml

it's not a logic gap for master, but still worth fixing for explicitly defined behaviour, I'll create a [tiny targeted PR for master](https://github.com/scylladb/scylla-operator/pull/3573)

rzetelskik: Renovate is only configured for the default branch. It's not going to send updates to release branches unless we configure it to do so, which I'm not sure we'd like to do.

The fact that this requires changes on every release branch cutoff needs an entry in the release procedure, at least before it's automated somehow.

Let's discuss this on Monday team sync.

PR #3573, opened 2026-08-07
fix: remove hardcoded operator upgradeTo fallback

abramche:
- read from config.yaml for consistency

Related to https://scylladb.atlassian.net/browse/OPERATOR-274

not a logic gap for master, but to be on the safe side, treat config.yaml as the only source of truth


abramche:
superseded by https://github.com/scylladb/scylla-operator/pull/3584

Issue #2744, opened 2025-07-04
ScyllaCluster can get stuck on a cleanup job progressing condition


czeslavo:
### Link to the job that flaked.

https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2743/pull-scylla-operator-master-e2e-gke-parallel-clusterip/1941076808367607808

### Snippet of what failed.

```

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000440080>: 
      waiting has been interupted ([false]): timed out waiting for the condition
      {
          msg: "waiting has been interupted ([false]): timed out waiting for the condition",
          err: <wait.errInterrupted>{
              cause: <*errors.errorString | 0xc000328740>{
                  s: "timed out waiting for the condition",
              },
          },
      }
  occurred
  In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scyllacluster/scyllacluster_upgrades.go:62 @ 07/04/25 11:12:24.48 
```

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2833/pull-scylla-operator-master-e2e-gke-parallel/1949834581385416704#1:test-build-log.txt%3A3489

czeslavo:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2845/pull-scylla-operator-v1.18-e2e-gke-parallel-clusterip/1950463100754857984

rzetelskik:
https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/logs/ci-scylla-operator-1.17-e2e-gke-parallel/1952248247976202240#1:test-build-log.txt%3A2763

The test is stuck on waiting for ScyllaCluster's readiness - the job controller is progressing.
```
  - lastTransitionTime: "2025-08-04T06:14:07Z"
    message: 'JobControllerProgressing: Waiting for Job "e2e-test-scyllacluster-xdcrw-0-6hrbx/cleanup-basic-x55tr-us-east-1-rack-0-0"
      to complete.'
    observedGeneration: 1
    reason: WaitingForJobCompletion
    status: "True"
    type: JobControllerProgressing
```

[...]

PR #3574, opened 2026-08-07
feat: seed last cleaned up token ring hash with a never-cleaned-up value

rzetelskik:
Cleanup of a node is triggered by its member Service's current token ring hash differing from the last cleaned up one. The latter is seeded with the current hash the first time the Service is seen carrying one, which exempts the node from cleanup of the ring it first observed. The seeded value is whatever ring state the node's sidecar sampled first, so which nodes end up exempt depends on the race between how far apart the nodes join and the sidecar's 5s poll, not on the nodes themselves. Ordered pod management spaced the joins out enough that the exempt node was always (usually?) only the last one, but parallel pod management removes that spacing and the number of created cleanup jobs varies wildly.

This PR changes this to seed a `NeverCleanedUp` value instead, meaning the node has never been cleaned up. It never matches a current hash, so every node is cleaned up once the ring changes and the count follows the member count.

PR also removes the wait for the annotation before scaling. It only guarded against the ring hash changing before the cleanup logic first observed a node, which could freeze a post-scale hash and exempt the node from cleanup. An absent annotation can now only ever become `NeverCleanedUp`, so scaling early can at most delay a cleanup, never lose it.

The E2E assertions pinned which ordinals get a cleanup job. They now assert that a cleanup job is created for every node.

This fix makes the existing cleanup burst worse by one job per rack: an N node cluster previously got N-1 jobs and now gets N (including the edge-case of a one-node cluster). What's worse, however, is that with parallel bootstrap we can create multiple StatefulSets at once, so a potential burst can grow to RACKS x NODES.

Nothing limits how many cleanup jobs are created at the same time. We can consider capping the number of in-flight jobs (or any other improvement proposed in https://scylladb.atlassian.net/browse/OPERATOR-48).

Resolves https://scylladb.atlassian.net/browse/OPERATOR-291

### DIFF
--- a/pkg/controller/scylladbdatacenter/resource.go
+++ b/pkg/controller/scylladbdatacenter/resource.go
@@ -156,16 +156,19 @@ func MemberService(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName, name string
 		maps.Copy(annotations, sdcAnnotations)
 	}
 
+	// Seed the annotation with a value meaning the node has never been cleaned up. Seeding it with the currently
+	// observed token ring hash would exempt the node from a cleanup of the ring state it happens to observe first,
+	// which depends on when its sidecar samples the ring.
+	// Only a completed cleanup Job may set it to a real hash.
+	lastCleanedUpTokenRingHash := naming.LastCleanedUpTokenRingHashAnnotationValueNeverCleanedUp
 	if oldService != nil {
-		_, hasLastCleanedUpRingHash := oldService.Annotations[naming.LastCleanedUpTokenRingHashAnnotation]
-		currentTokenRingHash, hasCurrentRingHash := oldService.Annotations[naming.CurrentTokenRingHashAnnotation]
-		if !hasLastCleanedUpRingHash && hasCurrentRingHash {
-			annotations[naming.LastCleanedUpTokenRingHashAnnotation] = currentTokenRingHash
+		if v, ok := oldService.Annotations[naming.LastCleanedUpTokenRingHashAnnotation]; ok {
+			lastCleanedUpTokenRingHash = v
 		}
 	}
+	annotations[naming.LastCleanedUpTokenRingHashAnnotation] = lastCleanedUpTokenRingHash
 
-	cleanupJob, ok := jobs[naming.CleanupJobForService(name)]
-	if ok {
+	if cleanupJob, ok := jobs[naming.CleanupJobForService(name)]; ok {
 		if len(cleanupJob.Annotations[naming.CleanupJobTokenRingHashAnnotation]) != 0 && cleanupJob.Status.CompletionTime != nil {
 			annotations[naming.LastCleanedUpTokenRingHashAnnotation] = cleanupJob.Annotations[naming.CleanupJobTokenRingHashAnnotation]
 		}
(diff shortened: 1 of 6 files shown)


rzetelskik:
Tested against naive parallel bootstrap implementation in https://github.com/scylladb/scylla-operator/pull/3521

pkg/controller/scylladbdatacenter/resource.go
```diff
@@ -157,10 +157,12 @@ func MemberService(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName, name string
...
-		if !hasLastCleanedUpRingHash && hasCurrentRingHash {
-			annotations[naming.LastCleanedUpTokenRingHashAnnotation] = currentTokenRingHash
+		// Seed the annotation with an empty value, meaning the node has never been cleaned up. Seeding it with the
+		// currently observed token ring hash would exempt the node from a cleanup of the ring state it happens to
+		// observe first, which depends on when its sidecar samples the ring.
+		// Only a completed cleanup Job may set it to a real hash.
+		if _, hasLastCleanedUpRingHash := oldService.Annotations[naming.LastCleanedUpTokenRingHashAnnotation]; !hasLastCleanedUpRingHash {
+			annotations[naming.LastCleanedUpTokenRingHashAnnotation] = ""
 		}
 	}
```
czeslavo, line 168: Wouldn't it make sense to fill the annotation with an empty string at the time of service creation already?

pkg/naming/constants.go
```diff
@@ -42,6 +42,8 @@ const (
 	CurrentTokenRingHashAnnotation = "internal.scylla-operator.scylladb.com/current-token-ring-hash"
 
 	// LastCleanedUpTokenRingHashAnnotation reflects the last cleaned up hash of token ring of the scylla node.
+	// An empty value means the node has never been cleaned up and is distinct from the annotation being absent,
```
czeslavo: Please consider using a more explicit value for this state instead of an empty string, e.g., `NeverCleanedUp`. This way, when looking at the object annotation, it will be clear what it means. Also, having a dedicated constant in the code would improve readability.

pkg/controller/scylladbdatacenter/resource.go
rzetelskik: done

pkg/naming/constants.go
rzetelskik: done

rzetelskik:
The tests against naive parallel-bootstrap implementation seem to have flaked on https://github.com/scylladb/scylla-operator/issues/2744.
@czeslavo should we merge this or debug?

czeslavo:
Two more nits: 
- commit message still mentions "an empty value"
- there's a changelog entry missing

> The tests against naive parallel-bootstrap implementation seem to have flaked on https://github.com/scylladb/scylla-operator/issues/2744.
@czeslavo should we merge this or debug?

Would be nice to know if this change amplifies this flake. I wouldn't block this PR on it, though. Maybe we could run some script overnight that would trigger jobs running this single test case and see how it performs?

rzetelskik:
> Would be nice to know if this change amplifies this flake. I wouldn't block this PR on it, though. Maybe we could run some script overnight that would trigger jobs running this single test case and see how it performs?

Nevermind, it actually failed on https://github.com/scylladb/scylla-operator/pull/3577. Unrelated then.

rzetelskik:
> Two more nits:
> 
> - commit message still mentions "an empty value"
> - there's a changelog entry missing

Done and done. PTAL @czeslavo.

czeslavo:
:rocket: 
/lgtm

PR #3576, opened 2026-08-10
docs: mention manager upgrade during operator upgrade flow

abramche:
- mention that manager also needs to be upgraded on operator upgrade (missed user path with gitops)

Resolves https://scylladb.atlassian.net/browse/OPERATOR-300

/cc czeslavo

### DIFF
--- a/docs/source/upgrade/upgrade-operator.md
+++ b/docs/source/upgrade/upgrade-operator.md
@@ -13,6 +13,10 @@ If any additional steps are required for a specific version upgrade, they will b
 ## Upgrade via GitOps (kubectl)
 
 A typical upgrade flow using GitOps (kubectl) requires re-applying the manifests using ones from the release you want to upgrade to.
+
+If ScyllaDB Manager is installed, you need to upgrade it manually by re-applying its manifests using ones from the release you want to upgrade to.
+Please refer to the [ScyllaDB Manager installation instructions](./../deploy-scylladb/install-scylladb-manager.md) for details.
+
 Note that ScyllaDB Operator's dependencies also need to be updated to the versions compatible with the target ScyllaDB Operator version.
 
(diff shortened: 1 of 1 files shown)


docs/source/upgrade/upgrade-operator.md
```diff
@@ -13,6 +13,10 @@ If any additional steps are required for a specific version upgrade, they will b
 ## Upgrade via GitOps (kubectl)
 
 A typical upgrade flow using GitOps (kubectl) requires re-applying the manifests using ones from the release you want to upgrade to.
+
+If ScyllaDB Manager is installed, you need to upgrade it manually by re-applying its manifests using ones from the release you want to upgrade to.
+Please refer to the [ScyllaDB Manager installation instructions](./../deploy-scylladb/install-scylladb-manager.md) for details.
```
rzetelskik, line 18: this doesn't only apply to manager but all dependencies, which it says below - I wouldn't make a special case for manager. Do you think the existing wording should be highlighted or made more noticeable?

czeslavo: I think it's easy to miss what exact dependencies a user is expected to update, especially if you go the GitOps path. We're linking to https://operator.docs.scylladb.com/stable/install-operator/install-with-gitops.html here, which doesn't have a Manager step, as we treat it as a separate step and have a dedicated page for it: https://operator.docs.scylladb.com/stable/deploy-scylladb/install-scylladb-manager.html.

rzetelskik: Good point, I forgot it's no longer a required dependency in our docs

rzetelskik:
lgtm
/assign @czeslavo

abramche:
do we need a changelog entry for this, btw?

rzetelskik:
> do we need a changelog entry for this, btw? 

I think we don't need one
/label no-changelog

PR #3577, opened 2026-08-10
test: reserve orphaned-PV clone PVs for their intended PVC

rzetelskik:
The fake clone provisioner binds one-way: it creates the clone PV with an empty claimRef and then sets the real PVC's spec.volumeName. volumeName is only a request, claimRef is what reserves, so the PV controller can match the unclaimed clone PV against a different unbound PVC of the same class and win, leaving the intended PVC permanently Pending.

The race needs two unclaimed clone PVs alive at once. Today, with ordered pod management policy, they are spread out in time enough for it to never trigger. With parallel bootstrap all three land simultanously and the test fails.

Set spec.claimRef to the intended PVC at creation. The UID is left unset because this test replaces the PVC, and the UID is only enforced when it's set.

Retain the clone PV. Pre-binding makes it Bound, which newly activates reclaim, and it shares a backing volume with the clone PVC's PV - on CSI-backed CI a Delete reclaim would delete storage still in use.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-295

/hold to test with naive parallel-bootstrap implementation

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_pv.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_pv.go
@@ -247,18 +247,31 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", framework.SuiteParall
 						pv, err := f.KubeAdminClient().CoreV1().PersistentVolumes().Get(ctx, pvc.Spec.VolumeName, metav1.GetOptions{})
 						o.Expect(err).NotTo(o.HaveOccurred())
 
+						realPVCName := strings.TrimPrefix(pvc.Name, "clone-")
+
 						pvClone := &corev1.PersistentVolume{
 							ObjectMeta: metav1.ObjectMeta{
 								Name:   fmt.Sprintf("clone-%s", pv.Name),
 								Labels: f.CommonLabels(),
 							},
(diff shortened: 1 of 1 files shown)


rzetelskik:
/hold cancel
Tests with naive parallel bootstrap implementation passed.

rzetelskik:
> 
> ci/prow/e2e-gke-parallel	[5df0a49](https://github.com/scylladb/scylla-operator/commit/5df0a49127985f34049c49f86b2acd7027d33bf1)	[link](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3577/pull-scylla-operator-master-e2e-gke-parallel/2086749484586373120)	true	`/test e2e-gke-parallel`
> [Full PR test history](https://prow.scylla-operator.scylladb.com/pr-history?org=scylladb&repo=scylla-operator&pr=3577). [Your PR dashboard](https://prow.scylla-operator.scylladb.com/pr?query=is:pr+state:open+author:rzetelskik).
> 
> Details

Unrelated, flaked on https://github.com/scylladb/scylla-operator/issues/2744.

PR #3578, opened 2026-08-10
feat: support parallel bootstrap of ScyllaDB nodes

rzetelskik:
This PR wires parallel bootstrap policy through to the rack StatefulSets. Parallel maps to parallel pod management, so a rack's nodes bootstrap without waiting for the previous ordinal, and all missing rack StatefulSets are created in a single sync instead of one per requeue.

Changing the policy on an existing cluster recreates the StatefulSets with DeletePropagationOrphan, so the Pods survive and are adopted back by the new one without any disruptions. 
This PR also clears resourceVersion on the recreate path in resourceapply, which otherwise makes the create fail, and only self-heal on a subsequent reconciliation.

Covered by unit tests, envtests, and an E2E toggling the policy both ways asserting the StatefulSets are replaced without any disruptions. All existing E2Es use the default parallel bootstrap policy for newly created clusters.

One caveat: StatefulSets are still only created (even when in parallel) after all existing StatefulSets have rolled out, since the gate also fronts scaling, updates and upgrades.

**Prerequisites**:
- [x] https://scylladb.atlassian.net/browse/OPERATOR-290

Resolves https://scylladb.atlassian.net/browse/OPERATOR-288
Resolves https://scylladb.atlassian.net/browse/OPERATOR-152

/cc

### DIFF
--- a/pkg/controller/scylladbdatacenter/sync_statefulsets.go
+++ b/pkg/controller/scylladbdatacenter/sync_statefulsets.go
@@ -437,8 +437,10 @@ func (sdcc *Controller) checkExistingStatefulSetsRolloutStatus(
 }
 
 // createMissingStatefulSets creates the missing StatefulSets from requiredStatefulSets.
-// Existing StatefulSets are skipped and at most one missing StatefulSet is created so racks bootstrap sequentially.
-// It returns the created StatefulSet and its progressing conditions, or an error if creation failed.
+// Existing StatefulSets are skipped. With the Sequential bootstrap policy at most one missing StatefulSet is created so
+// that racks bootstrap one by one, while with the Parallel one all of them are created at once.
+// It returns the StatefulSets created so far and their progressing conditions, together with an error if a creation
+// failed.
 func createMissingStatefulSets(
 	ctx context.Context,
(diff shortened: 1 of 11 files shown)


pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Shouldn't we return created STSes and respective progressing conditions in case of an error on a single STS apply so that we still surface them in status if some managed to be created/updated?

czeslavo (commented):
Looks good, just one comment.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
rzetelskik: Good point, changed. I also modified the caller to always run the status update and aggregate the errors. It's a bit ugly, bit trying to untangle it turned into a real scope creep. Ideally we should remove the whole status updating logic from this func IMO and delegate all of it to `calculateStatus`, but I'm afraid to touch it along these changes.

rzetelskik:
@czeslavo I also added a commit taking degraded status condition of ScyllaDBDatacenterNodeStatusReport controller into account. I'll also add a changelog entry as I believe this will be the core PR for parallel bootstrap support.

PR #3579, opened 2026-08-10
feat: add mutating webhook infrastructure

czeslavo:

Wires mutating webhook infrastructure by exposing a `/mutate` webhook server endpoint, adding a `MutatingWebhookConfiguration` manifest, and adding noop defaulters for `ScyllaCluster` and `ScyllaDBDatacenter` CRDs. Uses controller-runtime's `admission.Webhook` handler to avoid rolling our own.

Actual `ScyllaDBDatacenter` and `ScyllaCluster` defaulters implementation will happen in a separate PR under https://scylladb.atlassian.net/browse/OPERATOR-290.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-287


test/envtest/controllers/mutating_webhook_test.go
```diff
@@ -0,0 +1,132 @@
...
+	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
+	operatorcmd "github.com/scylladb/scylla-operator/pkg/cmd/operator"
+	"github.com/scylladb/scylla-operator/test/envtest"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+var _ = g.Describe("Mutating admission webhook", func() {
```
rzetelskik, line 22: is this intended to replace e2es for admission webhooks (or the counterpart that we would add for the mutating webhook)? Or will this only arrive with the actual defaulter implementations?

test/envtest/controllers/mutating_webhook_test.go
```diff
@@ -0,0 +1,132 @@
...
+			labels = map[string]string{}
+		}
+		labels["update-trigger"] = "true"
+		obj.SetLabels(labels)
+		err = env.KubeClient().Update(ctx, obj)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Expect(resourceInvocations.Load()).To(o.Equal(resourceInvocationsBeforeUpdate), "the shipped rules should be CREATE-only")
+	},
+		g.Entry("admits a ScyllaCluster unchanged", &entry{
```
rzetelskik, line 97: should we add cases where it would mutate the object?

test/envtest/setup.go
```diff
@@ -221,6 +222,29 @@ func SetupMockValidatingWebhook(ctx context.Context, e *Environment, handleFunc
...
 
+	setupMockWebhook(ctx, e, webhookOpts, webhookPath, admissionreview.NewHandler(handleFunc))
+}
+
+// SetupOperatorMutatingWebhook installs the MutatingWebhookConfiguration shipped in deploy/operator/ and dispatches
+// the intercepted admission requests to handler, so the shipped rules and path are exercised.
+// The webhook server is started automatically and cleaned up when the test ends.
+// NOTE: envtest rewrites the manifest's client config to point at a test-local webhook server, not the webhook
+// server shipped with the Operator.
+func SetupOperatorMutatingWebhook(ctx context.Context, e *Environment, handler admission.Handler) {
```
rzetelskik, line 253: out of scope of this PR, but what's the plan for implementing the default behaviour in envtests? I imagine we'd want the objects created in controller envtests to mimick the actual defaulting behaviour - will every controller's envtests have to set up the webhooks itself? Or would that only be covered by E2Es?

rzetelskik (commented):
overall lgtm, left some questions

test/envtest/controllers/mutating_webhook_test.go
czeslavo: I wanted to cover the real manifest in envtests this way, but I'd like to introduce an E2E in the defaulters implementation PR so that we're sure it's properly wired.

test/envtest/controllers/mutating_webhook_test.go
czeslavo: We will add such cases once we add actual defaulters in the next PR.

test/envtest/setup.go
czeslavo: Wired this by default. Opting out is possible with `WithoutMutatingWebhook` option.

rzetelskik (approved):
/lgtm
🚀

PR #3584, opened 2026-08-11
test: restructure operator upgrade test

abramche:
- remove config values for operator upgrade from/to versions, infer them dynamically: `OPERATOR_UPGRADE_FROM_VERSION` = <previous minor.patch release>, `OPERATOR_UPGRADE_TO_VERSION` = checkout/HEAD
- add a GHA workflow running the test as a presubmit (`pull_request`, docs-only changes skipped) and as on-demand manual `workflow_dispatch`
- add manual ref overrides for manual dispatch
- drop the dead operator-upgrade version plumbing from the generic e2e runner (`hack/.ci/lib/e2e.sh`).

Part of https://scylladb.atlassian.net/browse/OPERATOR-164 (changes to the test / action)


.github/workflows/e2e-kind-operator-upgrade.yaml
```diff
@@ -0,0 +1,45 @@
+name: "E2E: KinD (operator upgrade)"
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'enhancements/**'
+      - '**.md'
+      - 'renovate.json'
+  workflow_dispatch:
```
mflendrich, line 10: as we discussed - workflow_dispatch should allow specifying the version to upgrade from

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-of-onworkflow_dispatchinputs

.github/workflows/e2e-kind-operator-upgrade.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
mflendrich, line 30: paired with my previous comment about inputs to workflow_dispatch, this will no longer work (in the scenario where the user triggers more than one run with different "from" versions)

pkg/cmd/tests/options.go
```diff
@@ -149,8 +147,8 @@ func (o *TestFrameworkOptions) AddFlags(cmd *cobra.Command) {
...
-	cmd.PersistentFlags().StringVarP(&o.OperatorUpgradeFrom, "operator-upgrade-from-version", "", o.OperatorUpgradeFrom, "Version of the operator to upgrade from during operator upgrade tests. Either a released version or a full image reference.")
-	cmd.PersistentFlags().StringVarP(&o.OperatorUpgradeTo, "operator-upgrade-to-version", "", o.OperatorUpgradeTo, "Version of the operator to upgrade to during operator upgrade tests. Either a released version or a full image reference, e.g. of a locally built image.")
+	cmd.PersistentFlags().StringVarP(&o.OperatorUpgradeFrom, "operator-upgrade-from-version", "", o.OperatorUpgradeFrom, "Version of the operator to upgrade from during operator upgrade tests. Either a released version or a full image reference. Required by the operator upgrade suite.")
+	cmd.PersistentFlags().StringVarP(&o.OperatorUpgradeTo, "operator-upgrade-to-version", "", o.OperatorUpgradeTo, "Version of the operator to upgrade to during operator upgrade tests. Either a released version or a full image reference, e.g. of a locally built image. Required by the operator upgrade suite.")
```
mflendrich, line 160: Where does the test take the images from?

If the test builds them themselves (I don't think that's the case), then this should take a Git ref (that could be not only a tag, but also a commit SHA, a branch head, etc.)
If the test takes an existing image that's expected to be built and pushed to a registry, then it makes sense to allow specifying the entire image reference, not just the version.

I just don't see a scenario in which a "version" is a usable abstraction.
Another thing that (as I understand) this test is designed to assume that the "upgrade to" version is typically the version the tests are at.

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -77,12 +77,54 @@ ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}/upgrade-operator"
...
+function get-upgrade-from-version {
+  local branch="${1-}"
+  local tags
+
+  # Prefer local tags (present with full checkouts); fall back to the canonical repository for shallow
+  # clones and forks without the release tags.
+  tags="$( git -C "${repo_root}" tag --list | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' || true )"
+  if [ -z "${tags}" ]; then
+    tags="$( git ls-remote --tags --refs https://github.com/scylladb/scylla-operator.git | sed -e 's|.*refs/tags/||' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' )"
+  fi
```
mflendrich: This logic looks incorrect - it will fetch the remote tags only if the local list of tags is empty.

And there's a case in which the local list of tags is not empty but is outdated - e.g. if this function is run on a clone that is older than some released tag.

mflendrich (commented):
First round focused on correctness

.github/workflows/e2e-kind-operator-upgrade.yaml
mflendrich: And while "upgrade to" is also a parameter, I don't see why it wouldn't be exposed as well. I assume that the default should be the SHA the workflow is run on.

abramche: > as we discussed - workflow_dispatch should allow specifying the version to upgrade from

On a call, @rzetelskik specifically requested that we provide minimal input and the versions to be calculated dynamically -- which makes sense, since regardless whether you release a minor or a patch, we upgrade from the latest available minor.patch, following our N+1 upgrade policy.

> And while "upgrade to" is also a parameter, I don't see why it wouldn't be exposed as well. I assume that the default should be the SHA the workflow is run on.

It's not exposed for simplicity again and yes, it's essentially the SHA the workflow is dispatched from:

-> new minor release `1.22`, branch `v1.22` is out, the workflow is dispatched with `--ref=v1.22`, checks out the 1.22 tree, CI machinery builds a SHA-pinned tip of 1.22 (with an explicit tree checkout) and uses it as a TO ref. FROM is calculated as the latest released one minor junior -> 1.21.1, it does not get built, it gets pulled from docker hub.

-> new patch release `1.21.2`, the workflow is dispatched with `--ref=v1.21`, checks out tip of `v1.21`, CI machinery builds a SHA-pinned tip of 1.21 and uses it as a TO ref. FROM is calculated as the latest released one minor junior -> 1.20.2, it does not get built, it gets pulled from docker hub.

The entire flow above was designed with an idea of testing a minor version upgrade. I can only see manual inputs as a valuable option if we also want to cover patch upgrades (1.21.0 -> 1.21.1). Otherwise it's dismantling guard rails for potentially unsupported upgrade paths.

.github/workflows/e2e-kind-operator-upgrade.yaml
abramche: This is deliberate to avoid CI spam; but if you see value in running several actions at a time with a single ref, it's a small fix.

pkg/cmd/tests/options.go
abramche: The test always accepts built images.

The images are already published to a docker hub registry (latest released minor.patch) - and TO images are built and published to kind-local registry with CI by SHA-pinned tip in the work branch.

Versions are human-readable abstraction sugar in the function name - both arguments already accept a fully qualified image ref.

> "upgrade to" version is typically the version the tests are at.

yes

.github/workflows/e2e-kind-operator-upgrade.yaml
abramche: modified in ce099211ff1d041f21ed5e0139809095bdc0d425

PR updates cancel existing run (no value)
manual dispatches do not

hack/kind/run-e2e-operator-upgrade.sh
abramche: only fetch remote tags (ce099211ff1d041f21ed5e0139809095bdc0d425)

pkg/cmd/tests/options.go
abramche: renamed functions to `-to-ref` and `-from-ref` for clarity in 9fc9a55e2bede1a720d3f19de720371141c96ecf

.github/workflows/e2e-kind-operator-upgrade.yaml
mflendrich: > The entire flow above was designed with an idea of testing a minor version upgrade. I can only see manual inputs as a valuable option if we also want to cover patch upgrades (1.21.0 -> 1.21.1). Otherwise it's dismantling guard rails for potentially unsupported upgrade paths.

Only supporting minor version upgrades is an unnecessary limitation of this tool (especially because of the use case of testing patch version upgrades as you mention).

Instead: I expect the test to be not just a part of a rigid procedure, but a robust tool capable of verifying that, for given version X and Y, the operator at version X survives an upgrade to version Y. There's nothing wrong in asking it to test an upgrade between two versions many releases apart, or two (perhaps almost identical) versions one commit apart.
This possibility makes the tool robust. I don't see what "guardrails" would be useful here for. It's perfectly okay to try an upgrade from operator 1.6 to 1.20 and see it fail. (in fact I'd be worried if it didn't fail).

---

> we provide minimal input and the versions to be calculated dynamically

Two answers here:
Firstly: It is possible for the test to achieve both at the same time:
    * be "batteries included" if run with no arguments (which resolves to "test the diff between the last patch of the previous minor and HEAD")
    * allow specifying arbitrary X and Y with an explicit override


Secondly: Your own changes (in this PR and https://github.com/scylladb/scylla-operator/pull/3546) seem okay with this arbitrary "from" and "to" specification, in contradiction with the sentiment of your comment here.

> `# Versions to upgrade between; env-overridable, defaulting branch-relative via the functions below.`

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -77,12 +77,51 @@ ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}/upgrade-operator"
...
+# operator-upgrade-to-ref prints the checked-out tree ("latest") on pull requests (the PR is what's under test;
+# GITHUB_BASE_REF is only set there), the release-branch tip ("X.Y-latest", what a release would promote) when
+# the given branch is a release branch, the checked-out tree otherwise.
+function operator-upgrade-to-ref {
+  local branch="${1-}"
+
+  if [ -z "${GITHUB_BASE_REF:-}" ] && [[ "${branch}" =~ ^v([0-9]+\.[0-9]+)$ ]]; then
+    echo "${BASH_REMATCH[1]}-latest"
+  else
+    echo "latest"
```
mflendrich: `latest` as a container image tag does not mean "the checked out tree" - it means "the tip of `master`" in our project, and "the latest released stable version" everywhere else.

I don't think that using `latest` to mean "the checked out tree" is acceptable, because (1) it's very surprising, (2) it prevents using the real `latest` as an input in a correct way.

.github/workflows/e2e-kind-operator-upgrade.yaml
mflendrich: I still don't see how this would work (i.e. not cancel the other runs) if the user runs more than one run with different "from" versions.

abramche: I do not think changes are needed. Workflow is only cancelled with a new one in the PR (on the logic that old PR content is immaterial). Manual dispatch creates an own group (`|| github.run_id`) so they should not cancel each other. Do you see something else here?

.github/workflows/e2e-kind-operator-upgrade.yaml
abramche: Introduced optional overrides in c1dfab981536ab1f5539eed44852315c6ee2fa1c

hack/kind/run-e2e-operator-upgrade.sh
abramche: Removed altogether in 5da8a091cbdcb822e18119cc42b9ab257fadacd4

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -77,44 +77,79 @@ ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}/upgrade-operator"
...
+
+  if [ -z "${tags}" ]; then
+    echo "Can't determine the released version to upgrade from: no release tag matching branch \"${branch}\"" >&2
+    return 1
+  fi
+
+  sort -V <<< "${tags}" | tail -n 1 | sed -e 's/^v//'
+}
+
+current_branch="${GITHUB_BASE_REF:-${GITHUB_REF_NAME:-$( git -C "${repo_root}" rev-parse --abbrev-ref HEAD )}}"
```
mflendrich: Why is `GITHUB_BASE_REF` and `GITHUB_REF_NAME` necessary?

Won't the `git rev-parse` work in all the cases?

mflendrich:
@abramche to summarize our conversation today:

- the mechanism introduced in https://github.com/scylladb/scylla-operator-release/blame/017017e0362f6f158d516efcc844ce7399c34606/.github/ISSUE_TEMPLATE/release_minor.md#L235 tests a ref that's not necessarily the same ref that is being promoted - please fix by testing the exact commit that's being released
- the resolution function, ref parsing etc looks like overkill for the relatively simple task. The list of tags available can be pulled with a tool call similar to `skopeo list-tags docker://docker.io/scylladb/scylla-operator` - maybe the git fetch is not necessary at all?

abramche:
tested in the field for v1.22

```
$ gh workflow run e2e-kind-operator-upgrade.yaml --repo scylladb/scylla-operator --ref v1.22-upgrade-operator
✓ Created workflow_dispatch event for e2e-kind-operator-upgrade.yaml at v1.22-upgrade-operator
https://github.com/scylladb/scylla-operator/actions/runs/32957989265

To see the created workflow run, try: gh run view 32957989265
To see runs for this workflow, try: gh run list --workflow="e2e-kind-operator-upgrade.yaml"

$ gh run view --job=98143841471

✓ v1.22-upgrade-operator .github/workflows/e2e-kind-operator-upgrade.yaml · 32957989265
Triggered via workflow_dispatch about 24 minutes ago

✓ Upgrade the operator on KinD in 16m28s (ID 98143841471)
  ✓ Set up job
  ✓ Checkout
  ✓ Run the operator upgrade test
  - Upload artifacts
  ✓ Post Checkout
  ✓ Complete job
```

.github/workflows/e2e-kind-operator-upgrade.yaml
```diff
@@ -0,0 +1,62 @@
...
+  workflow_dispatch:
+    inputs:
+      upgrade-from:
+        description: >-
+          Ref to upgrade from: a tag of docker.io/scylladb/scylla-operator (e.g. a released version "1.20.2" or
+          a CI build "1.21-<git-commit-hash>") or a full image ref. Empty: the highest released patch of the
+          minor preceding the dispatched ref (on master, the highest released version).
+        required: false
+        type: string
+      upgrade-to:
```
mflendrich, line 19: As we agreed that the destination upgrade version will be the version that the test (or workflow) is run on - why is this still an input?

hack/kind/run-e2e-operator-upgrade.sh
```diff
@@ -77,44 +73,44 @@ ARTIFACTS="${ARTIFACTS:-$( mktemp -d )}/upgrade-operator"
...
+    return 1
   fi
-  build-and-push-operator-image "${deploy_dir}" "${1}"
+
+  sort -V <<< "${tags}" | tail -n 1 | sed -e 's/^v//'
 }
 
-resolve-operator-version OPERATOR_UPGRADE_FROM_VERSION operator_upgrade_from_deploy_dir
-resolve-operator-version OPERATOR_UPGRADE_TO_VERSION operator_upgrade_to_deploy_dir
+current_ref="${GITHUB_BASE_REF:-${GITHUB_REF_NAME:-$( git -C "${repo_root}" rev-parse --abbrev-ref HEAD )}}"
```
mflendrich, line 108: Why is `GITHUB_BASE_REF` / `GITHUB_REF_NAME` necessary here? Is there any case in which `git rev-parse` is not sufficient?

.github/workflows/e2e-kind-operator-upgrade.yaml
abramche: They are not required. The workflow already uses the ref it runs on as a default -to.

You specifically requested ref overrides in https://github.com/scylladb/scylla-operator/pull/3584#discussion_r3786844393 (`allow specifying arbitrary X and Y with an explicit override`).

hack/kind/run-e2e-operator-upgrade.sh
abramche: Because it's not enough when you run from a CI context (PR), since `action/checkout` fetches a shallow clone. This is a PR safeguard specifically.

Suppose you make a PR against the `v1.20` branch. Without this change, the merge ref is simply `HEAD` - the resolution logic would have dumb-defaulted to the latest 1.21 release, causing illegitimate downgrade test.

PR #3585, opened 2026-08-11
ci: wait for cert-manager webhook to be callable before installing operator

abramche:
- Probe cert-manager apiserver-to-webhook path on webhook deploy: retry a server-side dry-run create of a self-signed Issuer until it succeeds. This mutates nothing and covers both the caBundle-lag and CA-rotation variants of the race.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-303

Solves racy deployments like https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/3584/pull-scylla-operator-master-e2e-gke-parallel/2087169673371062272


czeslavo (approved):
/lgtm

PR #3588, opened 2026-08-12
feat: set bootstrapPolicy to parallel for new clusters >= 2026.2

czeslavo:

Implements mutating webhook defaulters for `ScyllaDBDatacenter` and `ScyllaCluster` that set their `bootstrapPolicy` to `Parallel` on creation, when their version is a valid semver >= 2026.2. For other cases, leaves the policy set to `nil`. It will allow us to change the default deduced in controllers to `Parallel` at some point, effectively migrating users that didn't specify the policy explicitly and had clusters created before Operator v1.22.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-290

/hold for https://github.com/scylladb/scylla-operator/pull/3578 so the changelog entry can be updated with this PR link

### DIFF
--- a/docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
+++ b/docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
@@ -92,7 +92,7 @@ object
      - backups specifies backup tasks in Scylla Manager. When Scylla Manager is not installed, these will be ignored.
    * - bootstrapPolicy
      - string
-     - bootstrapPolicy controls whether ScyllaDB nodes are bootstrapped one at a time (Sequential) or started in parallel (Parallel). Parallel requires ScyllaDB 2026.2 or later. If not provided, it's treated as Sequential.
+     - bootstrapPolicy controls whether ScyllaDB nodes are bootstrapped one at a time (Sequential) or started in parallel (Parallel). Parallel requires ScyllaDB 2026.2 or later. If not provided, it's treated as Sequential. On creation, it's set to Parallel when spec.version is a semver-parseable version supporting parallel bootstrap. The set value is persisted, so it stays in effect across ScyllaDB version changes until it's explicitly changed. It's otherwise left unset, and never set on an already existing object.
    * - cpuset
      - boolean
      - cpuset determines if the cluster will use cpu-pinning. Deprecated: `cpuset` is deprecated. It is now treated as if it is always set to true regardless of its value.
--- a/pkg/api/scylla/defaulting/scylladbdatacenter_defaulting_test.go
+++ b/pkg/api/scylla/defaulting/scylladbdatacenter_defaulting_test.go
@@ -1,0 +1,120 @@
+// Copyright (c) 2026 ScyllaDB.
+
+package defaulting_test
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/scylladb/scylla-operator/pkg/api/scylla/defaulting"
+	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
+	"github.com/scylladb/scylla-operator/pkg/test/unit"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSetDefaultsScyllaDBDatacenter(t *testing.T) {
+	t.Parallel()
+
+	newScyllaDBDatacenter := func(image string, bootstrapPolicy *scyllav1alpha1.BootstrapPolicy) *scyllav1alpha1.ScyllaDBDatacenter {
+		return &scyllav1alpha1.ScyllaDBDatacenter{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "basic",
+				Namespace: "test",
+			},
+			Spec: scyllav1alpha1.ScyllaDBDatacenterSpec{
+				ClusterName: "basic",
+				ScyllaDB: scyllav1alpha1.ScyllaDB{
+					Image: image,
+				},
+				Racks: []scyllav1alpha1.RackSpec{
+					{
+						Name: "rack",
+					},
+				},
+				BootstrapPolicy: bootstrapPolicy,
+			},
+		}
+	}
+
+	tt := []struct {
+		name       string
+		datacenter *scyllav1alpha1.ScyllaDBDatacenter
+		expected   *scyllav1alpha1.ScyllaDBDatacenter
+	}{
+		{
+			name:       "unset bootstrapPolicy with the minimal version supporting parallel bootstrap is defaulted to Parallel",
+			datacenter: newScyllaDBDatacenter(unit.ScyllaDBImageAtParallelBootstrapThreshold, nil),
+			expected:   newScyllaDBDatacenter(unit.ScyllaDBImageAtParallelBootstrapThreshold, new(scyllav1alpha1.BootstrapPolicyParallel)),
+		},
+		{
+			name:       "unset bootstrapPolicy with a version newer than the minimal one supporting parallel bootstrap is defaulted to Parallel",
+			datacenter: newScyllaDBDatacenter(unit.ScyllaDBImageAboveParallelBootstrapThreshold, nil),
+			expected:   newScyllaDBDatacenter(unit.ScyllaDBImageAboveParallelBootstrapThreshold, new(scyllav1alpha1.BootstrapPolicyParallel)),
+		},
+		{
+			name:       "unset bootstrapPolicy with an older version is left unset",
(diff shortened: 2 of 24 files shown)


rzetelskik:
/hold cancel
To break the circular dependency. I'll add this PR to the changelog entry in https://github.com/scylladb/scylla-operator/pull/3578.

pkg/api/scylla/defaulting/scylladbdatacenter_defaulting_test.go
```diff
@@ -0,0 +1,125 @@
...
+		{
+			// Mutating admission runs before the object is validated against the CRD schema, so the defaulter has to
+			// tolerate a spec missing the fields it reads.
+			name:       "unset bootstrapPolicy with an empty image is left unset",
+			datacenter: newScyllaDBDatacenter("", nil),
+			expected:   newScyllaDBDatacenter("", nil),
+		},
+		{
+			name:       "explicit Sequential bootstrapPolicy is preserved",
+			datacenter: newScyllaDBDatacenter("docker.io/scylladb/scylla:2026.2.0", pointer.Ptr(scyllav1alpha1.BootstrapPolicySequential)),
```
rzetelskik: should we use consts? Something like https://github.com/rzetelskik/scylla-operator/blob/2a7f67fe44d2919f09db60d6537acbfd940ba3ed/pkg/test/unit/const.go

test/envtest/controllers/mutating_webhook_test.go
```diff
@@ -77,38 +83,105 @@ var _ = g.Describe("Mutating admission webhook", func() {
...
 			},
+			expectedDefaultedSpecFields: nil,
 		}),
-		g.Entry("admits a ScyllaDBDatacenter unchanged", &entry{
+		g.Entry("stamps a Parallel bootstrapPolicy on a ScyllaCluster with a version supporting parallel bootstrap", &entry{
+			resource:            "scyllaclusters",
+			expectedIntercepted: true,
+			newObject: func(name, namespace string) client.Object {
+				sc := newBasicScyllaCluster(name, namespace)
+				sc.Spec.Version = "2026.2.0"
```
rzetelskik: ditto about consts

test/envtest/controllers/mutating_webhook_test.go
```diff
@@ -160,3 +233,13 @@ func getUnstructuredSpec(obj client.Object) any {
...
 }
+
+// setUpdateTriggerLabel labels obj so that updating it is a no-op for everything but the resource version.
+func setUpdateTriggerLabel(obj client.Object) {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels["update-trigger"] = "true"
+	obj.SetLabels(labels)
```
rzetelskik, line 244: I believe this could be simplified with metav1.SetMetaDataLabel?

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
```diff
@@ -51,6 +53,13 @@ var _ = g.Describe("ScyllaCluster upgrades", framework.SuiteParallel, framework.
 			sc.Spec.Repository = initialRepo
 			sc.Spec.Version = initialVersion
 
+			// The ScyllaDB version the cluster is created with may not support bootstrapping nodes in parallel, in
+			// which case Parallel is rejected on admission. Pin the policy explicitly so that the one the cluster is
+			// created under is a property of the version under test, and doesn't depend on the fixture defaults.
+			if !semver.SupportsParallelBootstrap(initialVersion) {
+				sc.Spec.BootstrapPolicy = pointer.Ptr(scyllav1.BootstrapPolicySequential)
```
rzetelskik: nit: use `new`

rzetelskik (commented):
first pass, mostly cosmetic - I'm yet to review the test logic

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
```diff
@@ -51,6 +53,13 @@ var _ = g.Describe("ScyllaCluster upgrades", framework.SuiteParallel, framework.
 			sc.Spec.Repository = initialRepo
 			sc.Spec.Version = initialVersion
 
+			// The ScyllaDB version the cluster is created with may not support bootstrapping nodes in parallel, in
+			// which case Parallel is rejected on admission. Pin the policy explicitly so that the one the cluster is
+			// created under is a property of the version under test, and doesn't depend on the fixture defaults.
+			if !semver.SupportsParallelBootstrap(initialVersion) {
```
rzetelskik: I also wonder if we should change the policy for the targetVersion on the update. In theory we claim it should be possible to switch at any point, and that way we'd at least partially test with the desired bootstrap policy, but it would tamper with the test semantics. Wdyt?

czeslavo: I'd keep the upgrade test to verify only the upgrade path, without exercising features like this one. I think it's better to focus this test on this single responsibility.

In the end, what we'd test this way is that once we upgrade ~operator~ ScyllaDB version, we can change the cluster's bootstrap policy to `Parallel` and it will work as expected. Isn't this simply covered in the E2Es added in https://github.com/scylladb/scylla-operator/pull/3578?

test/envtest/controllers/mutating_webhook_test.go
czeslavo: Won't work here as `SetMetaDataLabel` expects an `ObjectMeta` pointer, while in this case we have only a `client.Object` interface at hand that cannot give us such a pointer.

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
rzetelskik: Fair enough. What I don't like about this is that we'll effectively test with a different policy than requested, and if someone requests it explicitly the more correct path would be to error. Given this should default to sequential on <2026.2 (which I didn't think about when writing the comment on the ticket), should we just remove this override? Sorry for the confusion

docs/source/reference/api/groups/scylla.scylladb.com/scyllaclusters.rst
```diff
@@ -92,7 +92,7 @@ object
      - backups specifies backup tasks in Scylla Manager. When Scylla Manager is not installed, these will be ignored.
    * - bootstrapPolicy
      - string
-     - bootstrapPolicy controls whether ScyllaDB nodes are bootstrapped one at a time (Sequential) or started in parallel (Parallel). Parallel requires ScyllaDB 2026.2 or later. If not provided, it's treated as Sequential.
+     - bootstrapPolicy controls whether ScyllaDB nodes are bootstrapped one at a time (Sequential) or started in parallel (Parallel). Parallel requires ScyllaDB 2026.2 or later. If not provided, it's treated as Sequential. On creation, it's set to Parallel when spec.version is a semver-parseable version supporting parallel bootstrap. The set value is persisted, so it stays in effect across ScyllaDB version changes until it's explicitly changed. It's otherwise left unset, and never set on an already existing object.
```
rzetelskik, line 95: should go into a separate commit with generated changes

test/e2e/set/scylladbdatacenter/scylladbdatacenter_defaulting.go
```diff
@@ -0,0 +1,54 @@
...
+
+	g.It("should default bootstrapPolicy on creation", func(ctx g.SpecContext) {
+		sdc := f.GetDefaultScyllaDBDatacenter()
+		// The fixture carries an explicit bootstrapPolicy when the suite is invoked with one.
+		sdc.Spec.BootstrapPolicy = nil
+
+		// Parallel is only stamped for ScyllaDB versions known to support bootstrapping nodes in parallel, and
+		// Sequential is never stamped, so the expected value depends on the image under test. It's deliberately not
+		// hardcoded: the image can be pinned by a digest, leaving no parseable version to determine the support from.
+		scyllaDBVersion, _ := naming.ImageToVersion(sdc.Spec.ScyllaDB.Image)
```
rzetelskik: why ignore the error?

czeslavo: Added an assertion.

test/e2e/set/scyllacluster/scyllacluster_upgrades.go
czeslavo: Dropped the override.

rzetelskik (approved):
/lgtm
💪

rzetelskik:
/test e2e-openshift-aws-parallel
(since this is modifying the bundle)
/hold for test result

rzetelskik:
openshift failed on image pull back off, let's test it on the implementation PR
/hold cancel

PR #3598, opened 2026-08-18
fix(test/e2e): pair ScyllaDB Pods with their own PVC clone consumer Pods

rzetelskik:
`ScyllaCluster Orphaned PV controller > should replace a node with orphaned PV` times out on rollout after the replacement it tests has already succeeded in k8s clusters with >1 node dedicated to run ScyllaDB.

The clone PVs have no `NodeAffinity` - it's immutable, and the test needs to set it later to fake the orphan - so `PodAffinity` is the only thing keeping a ScyllaDB Pod on the Node holding its data. The term from `a79360d6` matched *any* Pod labelled `cloneLabelKey`, and all three consumer Pods share that value. In a failed run the consumer Pod of the replaced ordinal was recreated on another Node, which made that Node satisfy every ScyllaDB Pod's term, and Pod 0 was rescheduled onto it - coming up with empty storage while keeping its old identity, never flagged as orphaned, so bootstrap deadlocked..

`MatchLabelKeys` with `apps.kubernetes.io/pod-index` resolves the term per ordinal, which rack placement can't do on its own. Pod 0 now selects `cloneLabelKey = <namespace> AND pod-index in (0)`. The field is silently ignored when its feature gate is off, so the test asserts each Pod shares a Node with its own consumer Pod.

Still lacking:

Consumer Pods are only a stand-in for where the data is. If one is recreated outside the test's control - eviction, preemption - the paired ScyllaDB Pod follows it onto a Node without storage. This fixes *which* consumer Pod is followed, not the consumer Pod moving.

I dropped a prototype remembering the Node per ordinal - a PVC can bind before its consumer Pod is scheduled, and we only observe PVCs in the goroutine, so it's best effort only. 
Real node pinning on the clone PV is the fix, but it conflicts with the immutability workaround.

So this only fixes the scenario in which the consumer Pods don't involuntarily move between nodes.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-285

/cc

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_pv.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_pv.go
@@ -336,6 +336,9 @@ var _ = g.Describe("ScyllaCluster Orphaned PV controller", framework.SuiteParall
 					o.Expect(err).To(framework.NotHaveOccurredExceptNotFound())
 
 					err = framework.WaitForObjectDeletion(ctx, f.DynamicAdminClient(), corev1.SchemeGroupVersion.WithResource("persistentvolumeclaims"), f.Namespace(), clonePVC.Name, &clonePVC.UID)
+					if err != nil {
+						return false, fmt.Errorf("couldn't wait for clone PVC %q to be deleted: %v", naming.ManualRef(f.Namespace(), clonePVCName), err)
+					}
 				}
 				return false, nil
 			})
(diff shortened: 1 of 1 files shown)


rzetelskik:
@czeslavo ptal at this approach. As per the description, this only fixes the issue on a happy path where consumer Pods are not involuntarily rescheduled. Making them stick to an initially selected node would only be best effort without significantly complicating an already over-engineered test (unless you have an idea about how to enforce the stickiness of the consumer Pods without the NodeAffinity on their PVs).

An alternative to this would be to create a predefined ordinal-host mapping, but this doesn't really reproduce a real life scenario where the orphaned node would be scheduled on a different node.

Yet another alternative would be to simply remove this test - it's costing us a lot of maintenance. We have envtest coverage of the orphaned PV controller, as well as an E2E covering node replacement. This test covers the entire path, but the cost is high.

rzetelskik:
/hold
I'll experiment with enabling MutablePVNodeAffinity in our kind setup and simplify the test, leaving it in kind suite only.

rzetelskik:
superseded by https://github.com/scylladb/scylla-operator/pull/3599

PR #3599, opened 2026-08-18
fix(test/e2e): simulate a lost node by mutating PV node affinity

rzetelskik:
To simulate PV orphaning, the orphaned PV test works around the immutability of `PersistentVolume.spec.nodeAffinity` with a heavy, hard to maintain aparatus: a fake provisioner, a dedicated StorageClass, PV cleanup, and pod affinity pinning each ScyllaDB Pod to a stand-in consumer Pod.

The test in its current form is flaky, as described in https://scylladb.atlassian.net/browse/OPERATOR-285.
I tried to fix it in #3598 patching that apparatus to bind ScyllaDB Pods and their corresponding consumer Pods. However, it wouldn't protect consumer Pods from moving between nodes on involuntary rescheduling, and the only fix for that I could come up with was to create an arbitrary mapping of consumer Pods to k8s nodes, which wouldn't really simulate an orphaned node being rescheduled to a different node any longer.

As an alternative, I propose to make use of `MutablePVNodeAffinity` feature gate which make the entire aparatus unnecessary. The real PV's `nodeAffinity` is now patched to point at a node that doesn't exist, which triggers the orphaned PV controller to initiate node replacement.
This also drastically reduces the maintenance cost of this test.

The gate is alpha in 1.35 and is only enablable on KinD, so the test is dropped from other suites. We can restore it once the remaining environments get the feature gate enabled by default.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-285

/cc


czeslavo (approved):
> Lines changed: 95 additions & 794 deletions

:heart:

/lgtm

PR #3601, opened 2026-08-19
chore: configure Renovate base branch patterns

mflendrich:
Please review in tandem with https://github.com/scylladb/scylla-operator-release/pull/660.
Part of: OPERATOR-338

## Description

By default Renovate only processes the repository's default branch, so dependency updates are never proposed against release branches. This configures `baseBranchPatterns` with `master` and the currently supported release branches `v1.20` and `v1.21`.

```json
"baseBranchPatterns": [
  "master",
  "v1.20",
  "v1.21"
]
```

Notes:

- (Claude claims that) Renovate reads its config from the default branch only, so this `renovate.json` will govern all three streams. The separate (and until now inert) `renovate.json` files on `v1.20`/`v1.21` remain ignored unless `useBaseBranchConfig: "merge"` is added.

## Verification

- `make verify-renovate-config` passes (the change is stable across the `jq` round-trip performed by `hack/sync-renovate-config-with-go-mod-replace.sh`).
- `baseBranchPatterns` validated as a top-level `array` of `string` against Renovate's published schema (v44.34.0).
- `master`, `v1.20` and `v1.21` all confirmed to exist on the remote.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,6 +4,11 @@
     "config:recommended",
     ":semanticCommitTypeAll(chore)"
   ],
+  "baseBranchPatterns": [
+    "master",
+    "v1.20",
+    "v1.21"
+  ],
   "enabledManagers": [
     "custom.regex",
     "gomod"


renovate.json
```diff
@@ -4,6 +4,11 @@
     "config:recommended",
     ":semanticCommitTypeAll(chore)"
   ],
+  "baseBranchPatterns": [
+    "master",
+    "v1.20",
+    "v1.21"
+  ],
```
rzetelskik, line 11: I don't think this is sufficient. It will try to bump dependencies out of the supported range for a given branch which we declare on release (ScyllaDB, ScyllaDB Manager and so on). We can just ignore those PRs but it's unneccesary noise. Are there any other dependencies than go deps (and base images?) that you'd want to see updated for released versions?

mflendrich: I started looking into it and would actually love to hear your perspective on this.

I first thought about leaving just base images + gomod on the release branches, but then reached a conclusion that:
- we generally want patch releases to make it to the release branches
- we generally don't want minor/major releases (but that would be on a case by case basis I guess - I think it's not obviously wrong to expand support forward)

I therefore believe that the situations where we need to turn down an upgrade on a release branch:
- might better be "default yes" than the current "default no"
- even if we end up closing bump PRs, there are rules in place in renovate that prevent recreating them (e.g. if you close a major/minor version bump PR, renovate will skip that major/minor version altogether going forward).

Some middle ground might be by restricting the PRs to, for example, only patch version upgrades. But this definitely won't work with gomod - many vulnerability fixes are effectively at least minor version bumps.

In short:
- I think "default yes" might be better than "default no" for supported versions
- I would like to see how real the problem of unwanted dep bump PRs is

rzetelskik: > Some middle ground might be by restricting the PRs to, for example, only patch version upgrades.

If this can be done - sure. Closing PRs instead of defining the intent in the config is a very unpractical way of handling this imo. I wouldn't want to make the decision on every PR if this versions should be merged or not - I'd rather it was already encoded and only the desired PRs were sent.

mflendrich: I believe that this answer misses my point (explained in the latter part of the quoted paragraph).
It is my understanding that (at least for gomod) we'll want almost every gomod update to land, and my last comment explains why restricting to patch only is not enough to meaningfully resolve vulnerabilities.

rzetelskik: I agree, but this doesn't only touch go mod. The point I'm making is about all the other dependencies. The discussion we had on a call was based on the assumption we can restrict this to selected dependencies, which this PR doesn't seem to attempt.

abramche: We could consider grouping renovate PRs for older minor branches and test/merge them in bulk once a week or something (possibly as a part of triage duty). A single update PR per older branch exists then at any point of time -- almost no noise at all.

PR #3603, opened 2026-08-19
fix: complete an ongoing decommission when the rack node count is raised back

czeslavo:
**Description of your changes:**

Implements https://github.com/scylladb/scylla-operator/pull/3611.

**Which issue is resolved by this Pull Request:**
Closes https://scylladb.atlassian.net/browse/OPERATOR-343

PR #3611, opened 2026-08-20
docs(enhancement): handling node count changes during an ongoing decommission

czeslavo:

Adds a proposal for handling rack node count changes made during an ongoing decommission (OPERATOR-327).

Posting it as a PR (not a gist, not a Confluence page) to make Mermaid rendering work and make commenting on specific lines easier.

Closes https://scylladb.atlassian.net/browse/OPERATOR-327.

### DIFF
--- a/enhancements/proposals/OPERATOR-327-proposal.md
+++ b/enhancements/proposals/OPERATOR-327-proposal.md
@@ -1,0 +1,134 @@
+# Node count changes during an ongoing decommission (OPERATOR-327)
+
+## TLDR
+
+When a user scales a rack down, the operator decommissions nodes. Today, if the user changes the node count again before the decommission finishes, the rack gets stuck forever ([OPERATOR-343](https://scylladb.atlassian.net/browse/OPERATOR-343)).
+
+One rule is proposed, defined per node: **once a node's decommission has started, that node must finish leaving**. It is fully removed, together with its service and PVC, and its identity must not be used again before that. Only then the new node count is applied. Capacity that comes back is added as new, empty nodes.
+
+The proposed design records the leaving nodes in a per-rack status field before any decommission starts, and uses that record as a barrier until they are fully removed.
+
+The barrier blocks the whole rack only because we use StatefulSets today: a StatefulSet can only remove its highest-numbered node, so leaving nodes and new nodes cannot coexist in a rack. The rule itself is per node, and the design is ready for a future move away from StatefulSets, where the barrier shrinks to single nodes and a rack can scale out while other nodes are still leaving.
+
+Decisions to make:
+
+1. A decommission that ScyllaDB keeps rejecting stays trapped behind the barrier. Do we build an escape hatch now, knowing it becomes unnecessary once node management moves away from StatefulSets — or do we accept the risk until then? See Risks and mitigations.
+
+## Problem
+
+The operator asks a node to decommission by setting the `scylla/decommissioned=false` label on its member service. The sidecar runs `nodetool decommission` and sets the label to `true` when done.
+
+**The contract: once a service is stamped with the decommission label, we must assume the node was (or will be) decommissioned. The operator cannot revoke this operation.** ScyllaDB cannot cancel a running decommission, and the operator can never be sure the sidecar has not started it yet.
+
+Today, if the user raises the node count back during an ongoing decommission, the rack gets stuck forever:
+
+```mermaid
+sequenceDiagram
+    autonumber
+    participant User
+    participant Operator
+    participant Svc as Service rack-1
+    participant Sidecar
+
+    User->>Operator: scale rack down, nodes 2 to 1
+    Operator->>Svc: set label scylla/decommissioned="false"
+    Note over Operator: waits for the decommission to finish
+    User->>Operator: change of mind, scale back up, nodes 1 to 2
+    Note over Operator: still waits, the new node count changes nothing
+    Sidecar->>Svc: decommission done, set label to "true"
+    Note over Operator: replicas already match the spec,<br/>so there is nothing to scale down
+    Note over Operator: the service is needed again for node count 2,<br/>so it is never removed
+    Note over User,Sidecar: STUCK FOREVER: the pod runs a decommissioned ScyllaDB<br/>and is never ready, the "true" label stays on the service
+```
+
+The pod keeps running a decommissioned ScyllaDB and never becomes ready. Nothing ever removes the label. Even worse: the leftover `"true"` label makes the next scale-down of this node skip the decommission completely (see Risks and mitigations).
+
+## The core invariant, and why it becomes a rack-level barrier
+
+The durable part of this design is a **node-level** invariant:
+
+> A node whose decommission intent is recorded is *leaving*. It must complete its decommission and be fully cleaned up (service and PVC deleted) before its identity may exist again. Nothing may re-include a leaving node in the desired set.
+
+With the current StatefulSet backend, this invariant forces a **rack-level** barrier. A StatefulSet identifies nodes by ordinals `0..N-1` and can only remove the highest ordinal. So the leaving nodes must stay at the top of the range until they are removed. If we allowed a scale-out while a node is leaving, the new node would be created *above* the leaving one — and the leaving one could never be removed without also removing the newer, healthy node. This is why, with StatefulSets, a rack cannot scale out while a scale-in is in progress.
+
+The rack barrier is therefore a consequence of the StatefulSet backend. It does not prevent a future refactor from moving the barrier to the node level once we move away from StatefulSets. See "Readiness for heterogeneous nodes in a rack" below.
+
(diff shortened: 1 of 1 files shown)


enhancements/proposals/OPERATOR-327-proposal.md
```diff
@@ -0,0 +1,118 @@
...
+- **Operator upgrade during a decommission.** The in-flight operation looks like a stamped service with no record, so it is picked up and finished the same way.
+
+## Risks
+
+**Carried-over `decommissioned=true` labels.** [Issue #3539](https://github.com/scylladb/scylla-operator/issues/3539) shows that clusters in the wild can have a *healthy* node whose service carries a stale `"true"` label (left there by the bug above, plus manual recovery). After the fix, the operator removes such nodes and deletes their PVCs **right after the upgrade, without any user action**. For a real decommission leftover this is correct self-healing. For a healthy node that someone recovered by hand, it destroys its data.
+
+Decision to make: do we guard against this with a sanity check — before deleting the PVC of a `"true"`-labeled node, check that the node is no longer a member of the cluster (for example, using the nodes status reports)? If the check cannot confirm the node has left, refuse to prune and report `Degraded`.
```
rzetelskik: It might be difficult to assert from the operator controllers. The right place to verify this would be the sidecar imo and in theory it already does that, i.e. it only sets the Decommissioned label to true once the node reaches "Decommissioned" operational mode. Perhaps it would suffice to ensure that Decommissioned=True does not stick after the scaledown, i.e. ensure that the service is deleted before we scale back up.

enhancements/proposals/OPERATOR-327-proposal.md
```diff
@@ -0,0 +1,118 @@
...
+    Note over Operator: waits for the decommission to finish
+    User->>Operator: change of mind, scale back up, nodes 1 to 2
+    Note over Operator: still waits, the new node count changes nothing
+    Sidecar->>Svc: decommission done, set label to "true"
+    Note over Operator: replicas already match the spec,<br/>so there is nothing to scale down
+    Note over Operator: the service is needed again for node count 2,<br/>so it is never removed
+    Note over User,Sidecar: STUCK FOREVER: the pod runs a decommissioned ScyllaDB<br/>and is never ready, the "true" label stays on the service
+```
+
+The pod keeps running a decommissioned ScyllaDB and never becomes ready. Nothing ever removes the label. Even worse: the leftover `"true"` label makes the next scale-down of this node skip the decommission completely (see Risks).
```
rzetelskik: > Even worse: the leftover `"true"` label makes the next scale-down of this node skip the decommission completely (see Risks).

But the node is already decommissioned at that point, right? So that node shouldn't ever reach readiness?
This makes me think the race might actually be worse, i.e. that the STS is actually scaled down, PVC removed, and then both are brought up without the service being pruned? That could be how https://github.com/scylladb/scylla-operator/issues/353 happened

enhancements/proposals/OPERATOR-327-proposal.md
```diff
@@ -0,0 +1,118 @@
...
+    decommission:       # present only while a scale-down is in progress
+      desiredNodes: 1   # the committed target, frozen until the operation finishes
+```
+
+What we get:
+
+- **Single commit point.** One status write defines the target. We do not depend on the stamping order or on repair rules after a crash.
+- **Single release point.** The new node count applies only after all stamped services are pruned. With labels only, if pruning fails halfway, the node count can go back up in small steps, with one extra node bootstrap per step.
+- **Visible progress.** Users and support can see the ongoing operation and its target in the status.
+- **Webhook warnings.** On a node count change during an operation, the webhook can warn: "this change will apply after the ongoing scale-down finishes".
```
rzetelskik: We could explicitly deny scaling up when decommissioning is in progress? Like we deny removing a rack when it has any members. We'd still need to implement the logic you proposed in the controller as a safety measure, but this would communicate the intent better imo.

enhancements/proposals/OPERATOR-327-proposal.md
rzetelskik: Perhaps  splitting the label/annotation declaring the intent to decommission from actual decommissioning status could be a way to make it more robust?
I.e. "DecommissionedLabel" could always just report if the node is decommissioned or not, and a separate label/annotation would serve as the trigger? Just thinking out loud here.

rzetelskik (commented):
Overall this seems very sound. I marginally like the status approach better, especially if we could enforce scale-down completion with validation.
I'm not really keen on the safety check in the controller - either way it's sidecar that has to check the operational mode or the node being a part of the token ring (it already does that for the status report, but the stickiness of the annotation would have to be verified). I believe we should be able to model this that way.

enhancements/proposals/OPERATOR-327-proposal.md
```diff
@@ -0,0 +1,118 @@
...
+## Alternative (improved): dedicated status field
+
+Store the operation in the rack status **before** stamping anything:
+
+```yaml
+status:
+  racks:
+  - name: a
+    decommission:       # present only while a scale-down is in progress
+      desiredNodes: 1   # the committed target, frozen until the operation finishes
```
rzetelskik: I wonder if we'd need a new field for this? Seems like this is already the intent behind `rackStatus.Nodes`:
https://github.com/scylladb/scylla-operator/blob/9ee0ed1e2bdec54a41865137b068673804301587/pkg/api/scylla/validation/scylladbdatacenter_validation.go#L499-L502.

enhancements/proposals/OPERATOR-327-proposal.md
mflendrich: I see value in this approach, but I'm afraid that it's against the principle of declarative configuration where you have "desired state" and "actual state" and the job of a controller is to bring the world to the desired state.

And as a consequence thereof, it should be acceptable to control the desired state regardless of the actual state. Think e.g. of the implications the deny behavior you're proposing on future autoscaling.

czeslavo: Yes, we could also make it a hard requirement and reject such changes in the validating webhook.

The reason why I suggested making it a warning in the first place was that it's a more graceful way to signal the intended change won't take effect _immediately_ (decommissioning in-flight has to conclude first), but _eventually_. With the fix in place, there would be no harm in accepting it.

enhancements/proposals/OPERATOR-327-proposal.md
czeslavo: Yes, the race you mention is technically possible now - if the PVC is deleted, but the service deletion fails, there's a window for a scale-out to happen before the service deletion retries. The design solves that by ensuring that the node cannot be re-created with the same identity until its service is gone (in the STS implementation: the rack cannot be scaled until all decommissioning nodes' services are gone).

enhancements/proposals/OPERATOR-327-proposal.md
czeslavo: `Status.Nodes` maps directly to `sts.Spec.Replicas`, so it wouldn't work (we would have to change the field's semantics). Anyway, with the updated approach (`decommissioningNodes` list), it's no longer relevant.

enhancements/proposals/OPERATOR-327-proposal.md
czeslavo: With the proposed design (a barrier preventing STS scaling before a decommissioned node's service is gone), there will be no way to hit this issue. 

I suggest that instead of hardening this, we add the operator upgrade instructions to ensure there are no stale `decommissioned=true`-labeled services in users' clusters and treat the issue as solved. 

Updated the "Carried-over decommissioned=true labels" subsection to propose that.

czeslavo:
@rzetelskik @mflendrich I've updated the doc. I made sure that the potential future refactor of the way we manage nodes (allowing manipulation of individual nodes in a rack instead of relying on the STS backend) will be compatible with the proposed invariant.

Instead of using the status field to specify the desired replica count (which was driven by StatefulSet semantics), I propose listing nodes to be decommissioned in a list (`status.rack[].decommissioningNodes`). This makes it compatible with both models - either it is StatefulSet-based or our custom backend; we only care about the invariant: the node must complete its decommissioning and be fully cleaned up (PVC and Service gone) before another node using the same identity (in the STS implementation: ordinal) can be bootstrapped.

This way, with the STS backend, we implement a barrier at the rack level, but our API is ready for a change that will move the barrier to the node level.

https://github.com/scylladb/scylla-operator/pull/3603 holds a PoC implementation.

mflendrich (approved):
As discussed in our meetings - the updated design that is decoupled from the statefulset LGTM 🚢. This way we're better positioned for a future where the rack is decoupled from the sts.

Thanks!

/lgtm

czeslavo:
As agreed on the call, I'm merging this as-is. API surface changes will be updated in the doc together with an implementation.

/hold cancel

PR #3590, opened 2026-08-13
chore: switch BootstrapSynchronisation feature gate to opt-out

rzetelskik:
**Description of your changes:** This PR switches the BootstrapSynchronisation feature gate to opt-out.

**Which issue is resolved by this Pull Request:**
Resolves https://scylladb.atlassian.net/browse/OPERATOR-310

/cc

PR #3615, opened 2026-08-21
fix: stop blocking bootstrapping new nodes for the duration of the decommission

rzetelskik:
Currently, bootstrapping a new node is blocked by the bootstrap barrier while another node is decommissioning, which in the worst case scenario can block unstucking a decommission by adding new nodes.

The bootstrap barrier requires every host ID observed by the peers to have an entry with a status of its own in the nodes status report. A decommissioning node stays a token owning member the whole time, so its peers keep observing it, but makeRackNodesStatusReport enumerated the rack's nodes from the desired node count and dropped it as soon as the count shrank. Once the decommission completes the node flips to DECOMMISSIONED and its gossiper stops, so its sidecar can no longer answer the host ID map request it determines the membership with, nor report a status. The membership annotation is only written from a successful observation, so it kept "true", and the node was still entitled to an entry, now one without a status. Either way the barrier rejected the report and the joining node waited for the decommission, or the decommissioned node's Service, to go away. That Service is pruned only after the StatefulSet is scaled down, and the scaling loop isn't reached while any rack has a pending rollout, which is exactly the state a node stuck in the barrier leaves its rack in. Adding nodes to unstick a decommission was blocked by the same mechanism.

The report is now enumerated from the rack's member Services. The sidecar consults the operation mode before the token metadata: a DECOMMISSIONED node has left the ring for good and is annotated as not joined, dropping out of the report right away, while a STARTING, DRAINED or MAINTENANCE node, whose gossiper is down as well but which still owns its tokens, keeps its last observed membership instead of erroring on every sync. To do that the OperationalMode set is aligned with ScyllaDB's storage service mode enum, identical on every release we support: it carried CLIENT and DECOMMISSIONING, which ScyllaDB never reports, and lacked STARTING, BOOTSTRAP, MOVING and MAINTENANCE. As a side effect, decommissioning a node which hasn't finished bootstrapping now waits instead of failing on an UNKNOWN mode.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-353

/cc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,19 @@
   could be removed after the last node's `StatefulSet` was scaled down but before its member `Service` and `PVC` were
   removed, after which the operator could no longer resolve their rack and never cleaned them up.
   [#3633](https://github.com/scylladb/scylla-operator/pull/3633)
+- Fixed nodes being decommissioned dropping out of `ScyllaDBDatacenterNodesStatusReport` and blocking bootstrapping new nodes for the duration of the decommission.
+  The report enumerated a rack's nodes from the desired node count, so a decommissioning node
+  disappeared from it as soon as the count shrank, while its peers kept reporting it until the decommission
+  completed. The bootstrap barrier requires every observed host ID to have a report entry of its own, so a node joining
+  during a scale down waited for the decommission to finish.
+  This could additionally prevent unblocking decommissioning by adding new nodes.
+  The report is now built from the existing member Services, and a node which has finished decommissioning is dropped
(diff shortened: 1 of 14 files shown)


rzetelskik:
/cc @mflendrich 
Let's decide if we want to merge this or revert https://github.com/scylladb/scylla-operator/pull/3590.

rzetelskik:
This can still deadlock. Reverting the opt-out bootstrap sync for now, will come back to this later.

czeslavo (approved):
/lgtm

PR #3618, opened 2026-08-24
Revert "chore: switch BootstrapSynchronisation feature gate to opt-out"

rzetelskik:
Reverts scylladb/scylla-operator#3590 due to https://scylladb.atlassian.net/browse/OPERATOR-353. We don't want to send a last-minute fix and would rather be on the safe side. I'll send a separate PR adding a note to the docs about the known issue.

/cc @mflendrich


mflendrich (approved):
/lgtm

PR #3623, opened 2026-08-26
feat: record and drain the set of decommissioning nodes in the rack status

czeslavo:
Populates `status.racks[].decommissioningNodes` (`decommissioningMembers` on `ScyllaCluster`) with the nodes leaving a rack: a node is recorded before its decommission is requested and dropped once its PVC and Service have been pruned. This is a status-only change with no change to scale-down behaviour; the barrier that holds the rack node count while nodes are leaving is to be implemented separately under OPERATOR-343.

Closes https://scylladb.atlassian.net/browse/OPERATOR-363.

### DIFF
--- a/pkg/controller/scylladbdatacenter/status.go
+++ b/pkg/controller/scylladbdatacenter/status.go
@@ -3,6 +3,8 @@ package scylladbdatacenter
 import (
 	"context"
 	"fmt"
+	"slices"
+	"strings"
 
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/naming"
@@ -58,18 +60,20 @@ func getScyllaVersion(podLister corev1listers.PodLister, sts *appsv1.StatefulSet
 	return version, nil
 }
 
-// calculateRackStatus builds rack status from its StatefulSet.
+// calculateRackStatus builds rack status from its StatefulSet and member Services.
 // If sts is nil, it returns a stale zero status so the rack still appears in status.
 // Empty racks report the target image version; non-empty racks report the version from ordinal-0.
-func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, sts *appsv1.StatefulSet) *scyllav1alpha1.RackStatus {
+// The decommissioning nodes are derived from the decommissioned labels of the rack's member Services in services.
+func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, sts *appsv1.StatefulSet, services map[string]*corev1.Service) *scyllav1alpha1.RackStatus {
 	status := &scyllav1alpha1.RackStatus{
-		Name:           rackName,
-		Nodes:          new(int32(0)),
-		CurrentNodes:   new(int32(0)),
-		UpdatedNodes:   new(int32(0)),
-		ReadyNodes:     new(int32(0)),
-		AvailableNodes: new(int32(0)),
-		Stale:          new(true),
+		Name:                 rackName,
+		Nodes:                new(int32(0)),
+		CurrentNodes:         new(int32(0)),
+		UpdatedNodes:         new(int32(0)),
+		ReadyNodes:           new(int32(0)),
+		AvailableNodes:       new(int32(0)),
+		DecommissioningNodes: calculateDecommissioningNodes(rackName, services),
+		Stale:                new(true),
 	}
 
 	if sts == nil {
@@ -105,6 +109,33 @@ func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.
 	return status
 }
 
+// calculateDecommissioningNodes returns the nodes of the rack whose member Service carries the decommissioned label,
+// with either value: a node is leaving from the moment its decommission is requested until its Service is pruned.
+// The labels are the ground truth, so the list is derived from them on every sync.
+// Entries are sorted by name so that the list is stable across syncs. It is nil when no node is leaving.
+func calculateDecommissioningNodes(rackName string, services map[string]*corev1.Service) []scyllav1alpha1.DecommissioningNodeStatus {
+	var nodes []scyllav1alpha1.DecommissioningNodeStatus
+	for _, svc := range services {
+		if svc.Labels[naming.RackNameLabel] != rackName {
+			continue
+		}
+
+		if _, ok := svc.Labels[naming.DecommissionedLabel]; !ok {
+			continue
+		}
+
+		nodes = append(nodes, scyllav1alpha1.DecommissioningNodeStatus{
+			Name: svc.Name,
+		})
+	}
+
+	slices.SortFunc(nodes, func(a, b scyllav1alpha1.DecommissioningNodeStatus) int {
+		return strings.Compare(a.Name, b.Name)
+	})
+
+	return nodes
+}
+
 func updateAggregatedStatusFields(status *scyllav1alpha1.ScyllaDBDatacenterStatus) {
 	status.Nodes = new(int32(0))
 	status.ReadyNodes = new(int32(0))
@@ -132,7 +163,7 @@ func (sdcc *Controller) calculateStatus(sdc *scyllav1alpha1.ScyllaDBDatacenter, 
 	// Calculate the status for racks.
 	for _, rack := range sdc.Spec.Racks {
 		stsName := naming.StatefulSetNameForRack(rack, sdc)
-		status.Racks = append(status.Racks, *calculateRackStatus(sdcc.podLister, sdc, rack.Name, statefulSetMap[stsName]))
+		status.Racks = append(status.Racks, *calculateRackStatus(sdcc.podLister, sdc, rack.Name, statefulSetMap[stsName], serviceMap))
 	}
 
 	updateAggregatedStatusFields(status)
--- a/pkg/controller/scylladbdatacenter/status.go
+++ b/pkg/controller/scylladbdatacenter/status.go
@@ -7,6 +7,7 @@ import (
 	scyllav1alpha1 "github.com/scylladb/scylla-operator/pkg/api/scylla/v1alpha1"
 	"github.com/scylladb/scylla-operator/pkg/naming"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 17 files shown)


pkg/controller/scylladbdatacenter/status.go
```diff
@@ -105,6 +121,44 @@ func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.
 	return status
 }
 
+// getRackStatus returns the status of the named rack from racks, or nil if there is none.
+func getRackStatus(racks []scyllav1alpha1.RackStatus, rackName string) *scyllav1alpha1.RackStatus {
+	rackStatus, _, ok := oslices.Find(racks, func(rackStatus scyllav1alpha1.RackStatus) bool {
```
rzetelskik: nit: let's use std slices (slices.IndexFunc would be a corresponding func)

pkg/controller/scylladbdatacenter/status.go
```diff
@@ -58,10 +60,12 @@ func getScyllaVersion(podLister corev1listers.PodLister, sts *appsv1.StatefulSet
 	return version, nil
 }
 
-// calculateRackStatus builds rack status from its StatefulSet.
+// calculateRackStatus builds rack status from its StatefulSet and member Services.
 // If sts is nil, it returns a stale zero status so the rack still appears in status.
 // Empty racks report the target image version; non-empty racks report the version from ordinal-0.
-func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, sts *appsv1.StatefulSet) *scyllav1alpha1.RackStatus {
+// The decommissioning nodes record is carried over from previous, if any, and merged with the decommissioned labels
+// of the rack's member Services in services.
```
rzetelskik: I'd explicitly add that previous can be nil

pkg/controller/scylladbdatacenter/status.go
```diff
@@ -105,6 +121,44 @@ func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.
...
+		if _, ok := svc.Labels[naming.DecommissionedLabel]; ok {
+			names.Insert(svc.Name)
+		}
+	}
+
+	if names.Len() == 0 {
+		return nil
+	}
+
+	return oslices.ConvertSlice(apimachineryutilsets.List(names), func(name string) scyllav1alpha1.DecommissioningNodeStatus {
```
rzetelskik: nit: again let's use std slices

pkg/controller/scylladbdatacenter/status.go
```diff
@@ -72,6 +76,18 @@ func calculateRackStatus(podLister corev1listers.PodLister, sdc *scyllav1alpha1.
 		Stale:          new(true),
 	}
 
+	var recorded []scyllav1alpha1.DecommissioningNodeStatus
+	if previous != nil {
+		recorded = previous.DecommissioningNodes
```
rzetelskik: when are the entries deleted?

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -711,7 +711,32 @@ func (sdcc *Controller) syncStatefulSets(
 				return progressingConditions, nil
 			}
 
+			rackName := sts.Labels[naming.RackNameLabel]
+			_, rackStatusIdx, ok := oslices.Find(status.Racks, func(rackStatus scyllav1alpha1.RackStatus) bool {
```
rzetelskik: nit: std slices

pkg/controller/scylladbdatacenter/sync_services.go
```diff
@@ -194,6 +195,14 @@ func (sdcc *Controller) pruneServices(
...
 			continue
 		}
+
+		// The node is fully pruned, drop it from the record of decommissioning nodes.
+		_, rackStatusIdx, ok := oslices.Find(status.Racks, func(rackStatus scyllav1alpha1.RackStatus) bool {
+			return rackStatus.Name == rackName
+		})
+		if ok {
+			removeDecommissioningNode(&status.Racks[rackStatusIdx], svc.Name)
+		}
```
rzetelskik: On principle I'd avoid writing to status here to not move us further away from having a single place computing the status. I think this unnecessarily complicates things - this should be picked up by calculateStatus on service deletion.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -720,30 +720,41 @@ func (sdcc *Controller) syncStatefulSets(
...
-					klog.V(2).InfoS("Recording decommissioning node", "ScyllaDBDatacenter", klog.KObj(sdc), "Rack", rackName, "Service", klog.KObj(lastSvc))
-					progressingConditions = append(progressingConditions, metav1.Condition{
-						Type:               statefulSetControllerProgressingCondition,
-						Status:             metav1.ConditionTrue,
-						Reason:             "RecordingDecommissioningNode",
-						Message:            fmt.Sprintf("Recording node %q as decommissioning in the status of rack %q.", naming.ObjRef(lastSvc), rackName),
-						ObservedGeneration: sdc.Generation,
-					})
-					return progressingConditions, nil
+				recordDecommissioningNode(rackStatus, lastSvc.Name)
```
rzetelskik: on principle I wouldn't extend writing to status here - could we just wait for the label to propagate to status? That way we are closer to a single place syncing the status, and this only reacts to it

rzetelskik (commented):
first pass

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: I agree; it would be better to stick to the principle where the status can be rebuilt from the current state of the k8s cluster.

I gave it a go, ensuring we do not break any edge cases. I realized that this way we may even improve behavior because we will reconcile to the desired state quicker rather than sticking to the commited set of nodes to be decommissioned.

For example, let's consider the following scenario:

_Without the status write-ahead_ 
- we start parallel decommissioning of a 5-node cluster by scaling it to 2
- the controller approaches stamping nodes 4, 3, 2; succeeds in stamping 4 and 3, but crashes after that
- user in the meantime changes desired replicas back to 5
- the controller restarts, sees that 4 and 3 are decommissioning so it doesn't pick up the desired replicas count yet; 
- 4 and 3 decommission successfully; the controller prunes their PVCs and services; no more nodes are decommissioning, so the desired replica set is used for scaling; node 2 is never decommissioned as it wasn't necessary

_With the status write-ahead_
- we start parallel decommissioning of a 5-node cluster by scaling it to 2
- the controller writes into the status a set of nodes to be decommissioned: 4, 3, 2
- the controller, based on the status, approaches stamping 4, 3, 2; succeeds in stamping 4 and 3, but crashes after that
- user in the meantime changes desired replicas back to 5
- the controller restarts, sees that the status states we should decommission 4, 3, 2, but 2 has no label - stamps it **even though it is no longer required - we already know that we're going to eventually scale out**

What effectively changes is that `decommissioningNodes` means "currently decommissioning" rather than "planned to leave", and a crash between individual services stamping turns one parallel batch into two (if there was no change of the desired replicas in the above scenario, the controller would first wait for nodes 4 and 3 to successfully decommission, scale the STS in, and then stamp node 2 and decommission it in a separate batch). What we gain is quicker reconciliation in the case of a "change of mind" scenario, as we skip unnecessary decommissions.

rzetelskik: So what's the current idea for gating the scale-up? What happens if we label a service and then hit a stale cache when assessing if we can scale back up (assuming a scenario where a scale up immediately follow a scale down)? Isn't it the same issue we have with statefulsets, and we'd need a local cache (expectations) to prevent this?

czeslavo: Hm, yes, the problem with stale cache is something that I didn't take into account here. It could hit us despite the approach (both with the status write-ahead and with it being derived from the services' labels).

There's a chance that if SDC rack scaling up triggers another reconciliation just after the service gets stamped, we'd scale up when the service is not stamped in cache. As you said, that could be solved either with a local expectation cache or a live call to list services before scaling up a rack.

I suggest we tackle this in the next PR that will introduce the barrier. We can discuss whether the live call or the local cache is preferred.

rzetelskik: Let's sync on this - I'm no longer sure how the new status fields resolve the issue documented in the enhancement proposal with these changes

test/envtest/controllers/scylladbdatacenter_controller_test.go
```diff
@@ -106,6 +106,146 @@ var _ = g.Describe("ScyllaDBDatacenter controller", func() {
...
+			waitForStatefulSetReplicas(ctx, env, rackStatefulSetName, initialNodes-1)
+
+			g.By("Waiting for the leaving node's Service to be pruned and the record to drain")
+			waitForServiceToBePrunedAndRecordToDrain(ctx, env, sdc.Name, rackName, leavingServiceName)
+		})
+
+		// This is a reproducer of OPERATOR-343: raising the node count while a node is being decommissioned wedges the
+		// rack, because both the StatefulSet scale and the Service pruning are driven by the spec node count, which no
+		// longer calls for a scale-in. The record keeps the leaving node throughout, which is exactly what the barrier
+		// implemented under OPERATOR-343 needs to finish the decommission first and only then apply the raised count.
```
rzetelskik: > which is exactly what the barrier implemented under OPERATOR-343 needs to finish the decommission first and only then apply the raised count

This no longer holds with status only being there for observability, right? The controller will look at service labels

rzetelskik (commented):
Thanks for the updates, looks much cleaner now. Left one comment. I'd also make sure we reflect the changes in the role of the status entries in the enhancement proposal, but it doesn't have to go in this PR.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
rzetelskik: Discussed - the status will not be used for gating and will only reflect the state. Stale cache issue will be handled in subsequent PRs.

test/envtest/controllers/scylladbdatacenter_controller_test.go
czeslavo: Reworded to make it clear.

rzetelskik (approved):
/lgtm

PR #3626, opened 2026-08-27
fix: report the Services waiting for a hostID in a deterministic order

czeslavo:

The cert sync built the progressing messages and the hostIDs by iterating a map, so the CertControllerProgressing condition listed the waiting Services in a random order. Every sync produced a different status, every status update triggered another sync, and the controller spun in a hot loop for as long as any member Service lacked a hostID.

Spotted when writing envtests for https://scylladb.atlassian.net/browse/OPERATOR-343.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


mflendrich (approved):
Nice one.
/lgtm

PR #3629, opened 2026-08-28
fix: introduce decommissioning scaling barrier

czeslavo:

Fixes a rack getting stuck forever when its node count is changed while one of its nodes is being decommissioned: while a rack has decommissioning nodes, it now reconciles to the lowest leaving ordinal instead of the spec count, so spec changes are deferred (reported as `DeferringRackNodeCountChange`) until the leaving nodes are fully removed, and returning capacity bootstraps as fresh nodes.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-343.

### DIFF
--- a/pkg/controller/scylladbdatacenter/sync_statefulsets.go
+++ b/pkg/controller/scylladbdatacenter/sync_statefulsets.go
@@ -548,6 +548,18 @@ func getDecommissioningNodeNames(decommissioningNodes []scyllav1alpha1.Decommiss
 	return names
 }
 
+// makeDeferredNodeCountChangeCondition makes the progressing condition reporting that a change of the rack's spec node
+// count waits for the decommissioning nodes.
+func makeDeferredNodeCountChangeCondition(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, specNodeCount int32, decommissioningNodes []scyllav1alpha1.DecommissioningNodeStatus) metav1.Condition {
+	return metav1.Condition{
+		Type:               statefulSetControllerProgressingCondition,
+		Status:             metav1.ConditionTrue,
+		Reason:             "DeferringRackNodeCountChange",
+		Message:            fmt.Sprintf("Deferring node count change of rack %q to %d until the Services of its decommissioning nodes %q are pruned.", rackName, specNodeCount, getDecommissioningNodeNames(decommissioningNodes)),
+		ObservedGeneration: sdc.Generation,
+	}
+}
+
 // makeWaitingForRackServicePruningCondition makes the progressing condition reporting that the rack's decommissioned
 // nodes wait for their Services to be pruned.
 func makeWaitingForRackServicePruningCondition(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, decommissioningNodes []scyllav1alpha1.DecommissioningNodeStatus) metav1.Condition {
@@ -570,7 +582,8 @@ func makeWaitingForRackServicePruningCondition(sdc *scyllav1alpha1.ScyllaDBDatac
 //  4. Prune: pruneServices removes the node's Service and PVC (syncServices).
 //
 // The branches below are dispatched by the rack's current state, so they appear in guard order, not lifecycle order.
-// While any node is leaving, changes to the spec node count, up or down, are deferred.
+// While any node is leaving, changes to the spec node count, up or down, are deferred and reported with a progressing
+// condition.
 // It returns the progressing conditions to report: the rack still has leaving nodes and is held for as long as there
 // are any.
 func (sdcc *Controller) syncRackDecommission(ctx context.Context, sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, sts *appsv1.StatefulSet, rackServices map[string]*corev1.Service) ([]metav1.Condition, error) {
@@ -593,6 +606,12 @@ func (sdcc *Controller) syncRackDecommission(ctx context.Context, sdc *scyllav1a
 			return progressingConditions, fmt.Errorf("can't get decommission target node count of rack %q of ScyllaDBDatacenter %q: %w", rackName, naming.ObjRef(sdc), err)
 		}
 
+		// A plain scale-down has the spec already at the target count, so a difference means the spec changed since
+		// and the change is deferred until the leaving nodes' Services are pruned.
+		if *specNodeCount != targetNodeCount {
+			klog.V(4).InfoS("Deferring rack node count change until the Services of the decommissioning nodes are pruned", "ScyllaDBDatacenter", klog.KObj(sdc), "Rack", rackName, "NodeCount", *specNodeCount, "TargetNodeCount", targetNodeCount, "DecommissioningNodes", getDecommissioningNodeNames(decommissioningNodes))
+			progressingConditions = append(progressingConditions, makeDeferredNodeCountChangeCondition(sdc, rackName, *specNodeCount, decommissioningNodes))
+		}
 	}
 
 	// Wait if any decommissioning is in progress: a node with the label set to false was requested to decommission
--- a/pkg/controller/scylladbdatacenter/sync_statefulsets.go
+++ b/pkg/controller/scylladbdatacenter/sync_statefulsets.go
@@ -518,6 +518,163 @@ func ensureRackNamesInRackStatuses(
 	return apimachineryutilerrors.NewAggregate(errs)
 }
 
+// getRackDecommissionTargetNodeCount returns the node count the rack reconciles to while nodes are leaving:
+// the leaving nodes are at the top of the StatefulSet, which can only remove its highest ordinals, so the count is the
+// lowest leaving ordinal, capped at the StatefulSet replicas.
+func getRackDecommissionTargetNodeCount(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, sts *appsv1.StatefulSet, decommissioningNodes []scyllav1alpha1.DecommissioningNodeStatus) (int32, error) {
+	if len(decommissioningNodes) == 0 {
+		return 0, fmt.Errorf("rack %q of ScyllaDBDatacenter %q has no decommissioning nodes", rackName, naming.ObjRef(sdc))
+	}
+
+	nodeCount := *sts.Spec.Replicas
+	for _, node := range decommissioningNodes {
+		ordinal, err := naming.IndexFromName(node.Name)
+		if err != nil {
+			return 0, fmt.Errorf("can't get ordinal of decommissioning node %q of rack %q of ScyllaDBDatacenter %q: %w", node.Name, rackName, naming.ObjRef(sdc), err)
+		}
+
+		nodeCount = min(nodeCount, ordinal)
+	}
+
+	return nodeCount, nil
+}
+
+// getDecommissioningNodeNames returns the names of the given decommissioning nodes.
+func getDecommissioningNodeNames(decommissioningNodes []scyllav1alpha1.DecommissioningNodeStatus) []string {
+	names := make([]string, 0, len(decommissioningNodes))
+	for _, node := range decommissioningNodes {
+		names = append(names, node.Name)
+	}
+	return names
+}
+
+// makeWaitingForRackServicePruningCondition makes the progressing condition reporting that the rack's decommissioned
+// nodes wait for their Services to be pruned.
+func makeWaitingForRackServicePruningCondition(sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, decommissioningNodes []scyllav1alpha1.DecommissioningNodeStatus) metav1.Condition {
+	return metav1.Condition{
+		Type:               statefulSetControllerProgressingCondition,
+		Status:             metav1.ConditionTrue,
+		Reason:             "WaitingForRackServicePruning",
+		Message:            fmt.Sprintf("Waiting for decommissioned service(s) %q of rack %q to be pruned.", getDecommissioningNodeNames(decommissioningNodes), rackName),
+		ObservedGeneration: sdc.Generation,
+	}
+}
+
+// syncRackDecommission drives the decommission of the rack's leaving nodes and holds the rack while any is leaving.
+// A node is leaving from the moment its member Service carries the scylla/decommissioned label until the Service is
+// pruned, and a scale-down begins here by labelling the highest node. One node at a time, from the highest ordinal,
+// every leaving node goes through the following lifecycle:
+//  1. Request: the highest node is stamped with the scylla/decommissioned=false label.
+//  2. Wait: the node's sidecar runs the decommission and flips the label to true.
+//  3. Conclude: the StatefulSet is scaled below the node.
+//  4. Prune: pruneServices removes the node's Service and PVC (syncServices).
+//
+// The branches below are dispatched by the rack's current state, so they appear in guard order, not lifecycle order.
+// While any node is leaving, changes to the spec node count, up or down, are deferred.
+// It returns the progressing conditions to report: the rack still has leaving nodes and is held for as long as there
+// are any.
+func (sdcc *Controller) syncRackDecommission(ctx context.Context, sdc *scyllav1alpha1.ScyllaDBDatacenter, rackName string, sts *appsv1.StatefulSet, rackServices map[string]*corev1.Service) ([]metav1.Condition, error) {
+	var progressingConditions []metav1.Condition
+
+	specNodeCount, err := controllerhelpers.GetRackNodeCount(sdc, rackName)
+	if err != nil {
+		return progressingConditions, fmt.Errorf("can't get rack %q node count of ScyllaDBDatacenter %q: %w", rackName, naming.ObjRef(sdc), err)
+	}
+
+	decommissioningNodes := calculateDecommissioningNodes(rackName, rackServices)
+	if len(decommissioningNodes) == 0 && *specNodeCount >= *sts.Spec.Replicas {
+		return progressingConditions, nil
+	}
+
+	targetNodeCount := *specNodeCount
+	if len(decommissioningNodes) != 0 {
+		targetNodeCount, err = getRackDecommissionTargetNodeCount(sdc, rackName, sts, decommissioningNodes)
+		if err != nil {
+			return progressingConditions, fmt.Errorf("can't get decommission target node count of rack %q of ScyllaDBDatacenter %q: %w", rackName, naming.ObjRef(sdc), err)
+		}
+
+	}
+
+	// Wait if any decommissioning is in progress: a node with the label set to false was requested to decommission
+	// and hasn't finished. It is not necessarily the highest ordinal, e.g. a label that outlived its scale-down, so
+	// this can't be folded into the scale-down below, and nothing may be stamped or scaled while it is in flight.
+	for _, svc := range rackServices {
+		if svc.Labels[naming.DecommissionedLabel] == naming.LabelValueFalse {
+			klog.V(4).InfoS("Waiting for the node to finish decommissioning, signalled by its Service being labelled decommissioned=true", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KObj(svc))
+			progressingConditions = append(progressingConditions, metav1.Condition{
+				Type:               statefulSetControllerProgressingCondition,
+				Status:             metav1.ConditionTrue,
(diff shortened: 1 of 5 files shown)


pkg/controller/scylladbdatacenter/status.go
```diff
@@ -136,6 +138,57 @@ func calculateDecommissioningNodes(rackName string, services map[string]*corev1.
...
+		return 0, fmt.Errorf("can't parse ordinal from service name %q", name)
+	}
+
+	ordinal, err := strconv.ParseInt(ordinalStrings[1], 10, 32)
+	if err != nil {
+		return 0, fmt.Errorf("can't parse ordinal from service name %q: %w", name, err)
+	}
+
+	return int32(ordinal), nil
+}
```
rzetelskik: We have naming.IndexFromName already

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -663,10 +663,11 @@ func (sdcc *Controller) syncStatefulSets(
 			},
 		}
 
+		rackName := sts.Labels[naming.RackNameLabel]
```
rzetelskik: should we assert it's not empty?

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -672,6 +672,30 @@ func (sdcc *Controller) syncStatefulSets(
...
+		if len(decommissioningNodes) > 0 {
+			// The node count is held at the effective count while nodes are leaving. A plain scale-down has the spec
+			// already at the effective count, so a difference means the spec changed since and the change is deferred.
+			specNodeCount, err := controllerhelpers.GetRackNodeCount(sdc, rackName)
+			if err != nil {
+				return progressingConditions, fmt.Errorf("can't get rack %q node count of ScyllaDBDatacenter %q: %w", rackName, naming.ObjRef(sdc), err)
+			}
+
+			if *specNodeCount != *req.Spec.Replicas {
+				decommissioningNodeNames := oslices.ConvertSlice(decommissioningNodes, func(node scyllav1alpha1.DecommissioningNodeStatus) string {
```
rzetelskik: nit: switch to std slices

test/envtest/controllers/scylladbdatacenter_controller_test.go
```diff
@@ -172,14 +191,59 @@ var _ = g.Describe("ScyllaDBDatacenter controller", func() {
...
 
-		// This is a reproducer of OPERATOR-343: raising the node count while a node is being decommissioned wedges the
-		// rack, because both the StatefulSet scale and the Service pruning are driven by the spec node count, which no
-		// longer calls for a scale-in. The leaving node keeps its decommissioned label throughout, so it stays listed in
-		// the status. The barrier implemented under OPERATOR-343 will source the leaving nodes from the same labels to
-		// finish the decommission first and only then apply the raised count. Once that lands, the trailing assertions
-		// flip: the StatefulSet has to scale down, the Service has to be pruned and the list has to drain before the
-		// rack grows back.
-		g.It("should keep listing a decommissioning node when the node count is raised back mid-decommission", func(ctx g.SpecContext) {
+		g.It("should decommission a multi-node scale-down one node at a time from the highest ordinal", func(ctx g.SpecContext) {
```
rzetelskik, line 194: should we have such test for cross-rack serialization as well? I.e. at least two racks, decommission a node in each and verify they are decommissioned sequentially?

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+var _ = g.Describe("ScyllaDBDatacenter", framework.SuiteParallel, framework.SuiteParallelOpenShift, framework.SuiteKindFast, func() {
+	var f *framework.Framework
+
+	g.BeforeEach(func(ctx context.Context) {
+		f = framework.NewFramework(ctx, "scylladbdatacenter")
+	})
+
+	g.It("should finish an ongoing decommission before applying a node count raised back mid-decommission", func(ctx g.SpecContext) {
```
rzetelskik: nit: the test name describes implementation more than intent, how about something closer to "should finish an ongoing decommission before scaling up"?

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
```diff
@@ -0,0 +1,152 @@
...
+		for ordinal := int32(0); ordinal < leavingOrdinal; ordinal++ {
+			svc, err := nsClient.KubeClient().CoreV1().Services(ns.Name).Get(ctx, naming.MemberServiceName(rack, sdc, int(ordinal)), metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			host, err := utilsv1alpha1.GetBroadcastRPCAddress(ctx, nsClient.KubeClient().CoreV1(), sdc, svc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			stayingHosts = append(stayingHosts, host)
+		}
+		di := verification.InsertAndVerifyCQLData(ctx, stayingHosts)
+		defer di.Close()
```
rzetelskik: I think it would be better to modify the data inserter to allow for a custom RF - this approach will have a side effect of setting up the client with only the staying hosts, which doesn't reproduce reality that well. We should set up the client for all hosts but configure RF to allow for decommissioning.

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
```diff
@@ -0,0 +1,152 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(freshService.UID).NotTo(o.Equal(leavingService.UID))
+		o.Expect(freshService.Labels).NotTo(o.HaveKey(naming.DecommissionedLabel))
+
+		hostIDs, err := utilsv1alpha1.GetHostIDs(ctx, nsClient.KubeClient().CoreV1(), sdc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(hostIDs).To(o.HaveLen(int(initialNodes)))
+		// The staying nodes keep their identity, while the leaving node's identity is gone for good.
+		stayingHostIDs := oslices.Filter(hostIDs, func(hostID string) bool {
+			return oslices.Contains(initialHostIDs, func(initialHostID string) bool {
```
rzetelskik: nit: std slices

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Asserting now.

pkg/controller/scylladbdatacenter/status.go
czeslavo: Good point, reused.

test/envtest/controllers/scylladbdatacenter_controller_test.go
czeslavo: Added.

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
czeslavo: Reworded as suggested.

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
czeslavo: As agreed, we won't add a specific E2E test case for the barrier itself. It's an edge case covered well in envtests.

czeslavo:
@rzetelskik As we discussed, I tried to rework this so the barrier is clearly visible, not just implicitly from the "effective" node count. This brings the required STS calculation back to its original state (with no dependency on services). Instead, there's a dedicated `syncRackDecommission` step in the reconciliation loop that will handle scaling down as a whole, including the barrier. PTAL. I think it reads better this way indeed.

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
rzetelskik: envtests + existing e2es after splitting into more granular cases should cover this sufficiently - the e2e part will be covered in https://github.com/scylladb/scylla-operator/pull/3635

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
rzetelskik: no longer applies

test/e2e/set/scylladbdatacenter/scylladbdatacenter_scaling.go
rzetelskik: no longer applies

pkg/controller/scylladbdatacenter/sync_statefulsets.go
rzetelskik: I don't see the assertion

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,155 @@ func ensureRackNamesInRackStatuses(
...
+			// Services are managed in the other loop.
+			// When informers see the new service, will get re-queued.
+			return true, progressingConditions, nil
+		}
+
+		if len(lastSvc.Labels[naming.DecommissionedLabel]) == 0 {
+			lastSvcCopy := lastSvc.DeepCopy()
+			// Record the intent to decommission the member.
+			// TODO: Move this into syncServices so it reconciles properly. This is edge triggered
+			//  and nothing will reconcile the label if something goes wrong or the flow changes.
```
rzetelskik: could we handle this TODO with this change? Not necessarily doing what it says

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -652,78 +801,40 @@ func (sdcc *Controller) syncStatefulSets(
...
-				return progressingConditions, nil
-			}
+		// The decommission flow: while any node of the rack is leaving, it owns the StatefulSet scale and holds the
+		// rack, and any scale-down starts there.
+		isDecommissioning, decommissionProgressingConditions, err := sdcc.syncRackDecommission(ctx, sdc, sts, rackServices)
+		progressingConditions = append(progressingConditions, decommissionProgressingConditions...)
+		if err != nil {
+			return progressingConditions, fmt.Errorf("can't sync decommission of rack %q of ScyllaDBDatacenter %q: %w", rackName, naming.ObjRef(sdc), err)
+		}
+		if isDecommissioning {
```
rzetelskik: should we also return if len(progressingConditions) > 0 ?

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,155 @@ func ensureRackNamesInRackStatuses(
...
+		if len(lastSvc.Labels[naming.DecommissionedLabel]) == 0 {
+			lastSvcCopy := lastSvc.DeepCopy()
+			// Record the intent to decommission the member.
+			// TODO: Move this into syncServices so it reconciles properly. This is edge triggered
+			//  and nothing will reconcile the label if something goes wrong or the flow changes.
+			lastSvcCopy.Labels[naming.DecommissionedLabel] = naming.LabelValueFalse
+			controllerhelpers.AddGenericProgressingStatusCondition(&progressingConditions, statefulSetControllerProgressingCondition, lastSvcCopy, "update", sdc.Generation)
+			_, err = sdcc.kubeClient.CoreV1().Services(lastSvcCopy.Namespace).Update(ctx, lastSvcCopy, metav1.UpdateOptions{})
+			if err != nil {
+				return true, progressingConditions, err
```
rzetelskik: wrap err

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,155 @@ func ensureRackNamesInRackStatuses(
...
+				Reason:             "WaitingForRackServiceDecommission",
+				Message:            fmt.Sprintf("Waiting for rack service %q to decommission.", naming.ObjRef(svc)),
+				ObservedGeneration: sdc.Generation,
+			})
+
+			return true, progressingConditions, nil
+		}
+	}
+
+	if *sts.Spec.Replicas > targetNodeCount {
```
rzetelskik: nit: reduce indentation with *sts.Spec.Replicas <= targetNodeCount { return ... } ?

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,155 @@ func ensureRackNamesInRackStatuses(
...
+		if err != nil {
+			return true, progressingConditions, err
+		}
+
+		if *specNodeCount != targetNodeCount {
+			decommissioningNodeNames := make([]string, 0, len(decommissioningNodes))
+			for _, node := range decommissioningNodes {
+				decommissioningNodeNames = append(decommissioningNodeNames, node.Name)
+			}
+			klog.V(4).InfoS("Deferring rack node count change until the decommissioning nodes are removed", "ScyllaDBDatacenter", klog.KObj(sdc), "Rack", rackName, "NodeCount", *specNodeCount, "TargetNodeCount", targetNodeCount, "DecommissioningNodes", decommissioningNodeNames)
```
rzetelskik: the "removed" is unclear to me - the replicas are only removed later in this func. Is this referring to services being pruned? We should either say that or "leave the cluster"

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,155 @@ func ensureRackNamesInRackStatuses(
...
+				Type:               statefulSetControllerProgressingCondition,
+				Status:             metav1.ConditionTrue,
+				Reason:             "WaitingForRackServiceDecommission",
+				Message:            fmt.Sprintf("Waiting for rack service %q to decommission.", naming.ObjRef(svc)),
+				ObservedGeneration: sdc.Generation,
+			})
+
+			return true, progressingConditions, nil
+		}
+	}
```
rzetelskik: isn't this duplicating the logic above? These services will be among `decommissioning nodes`, no?

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: PTAL now.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: It's also no longer edge-triggered; we will keep stamping until we succeed, so yeah, we can get rid of it, as that's not really actionable.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Simplified this so there's no separate boolean needed to steer the logic. `len(progressingConditions) > 0` means we have to wait.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Reworded with explicit intention (waiting for node services to be pruned).

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Not really, `decommissioningNodes` includes all nodes with either decommissioning label value (true/false). Here we're explicitly waiting for a node that's in the process of decommissioning (false). This guard won't be needed in the parallel decommission case.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,188 @@ func ensureRackNamesInRackStatuses(
...
+	decommissioningNodes := calculateDecommissioningNodes(rackName, rackServices)
+	if len(decommissioningNodes) == 0 && *specNodeCount >= *sts.Spec.Replicas {
+		return progressingConditions, nil
+	}
+
+	targetNodeCount := *specNodeCount
+	if len(decommissioningNodes) != 0 {
+		targetNodeCount, err = getRackDecommissionTargetNodeCount(sdc, rackName, sts, decommissioningNodes)
+		if err != nil {
+			return progressingConditions, err
```
rzetelskik: nit: wrap err

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,188 @@ func ensureRackNamesInRackStatuses(
...
+		}
+
+		if nodeCount == nil || ordinal < *nodeCount {
+			nodeCount = pointer.Ptr(ordinal)
+		}
+	}
+
+	if *sts.Spec.Replicas < *nodeCount {
+		nodeCount = pointer.Ptr(*sts.Spec.Replicas)
+	}
```
rzetelskik: nit: I believe you could simplify it and avoid the pointer if you started with `nodeCount := *sts.Spec.Replicas` and then did `nodeCount = min(nodeCount, ordinal)` in the loop

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,188 @@ func ensureRackNamesInRackStatuses(
...
+	targetNodeCount := *specNodeCount
+	if len(decommissioningNodes) != 0 {
+		targetNodeCount, err = getRackDecommissionTargetNodeCount(sdc, rackName, sts, decommissioningNodes)
+		if err != nil {
+			return progressingConditions, err
+		}
+
+		// A plain scale-down has the spec already at the target count, so a difference means the spec changed since
+		// and the change is deferred until the leaving nodes' Services are pruned.
+		if *specNodeCount != targetNodeCount {
```
rzetelskik, line 611: won't you block forever if spec node count is lowered during an active decommission?

rzetelskik: nevermind, there is no return

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -518,6 +518,188 @@ func ensureRackNamesInRackStatuses(
...
+			progressingConditions = append(progressingConditions, makeDeferredNodeCountChangeCondition(sdc, rackName, *specNodeCount, decommissioningNodes))
+		}
+	}
+
+	// Wait if any decommissioning is in progress: a node with the label set to false was requested to decommission
+	// and hasn't finished. It is not necessarily the highest ordinal, e.g. a label that outlived its scale-down, so
+	// this can't be folded into the scale-down below, and nothing may be stamped or scaled while it is in flight.
+	for _, svc := range rackServices {
+		if svc.Labels[naming.DecommissionedLabel] == naming.LabelValueFalse {
+			klog.V(4).InfoS("Waiting for service to be decommissioned", "ScyllaDBDatacenter", klog.KObj(sdc), "Service", klog.KObj(svc))
```
rzetelskik: nit: wording is kind of weird, i.e. it's the scylla node being decommissioned, not the service

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: > This guard won't be needed in the parallel decommission case.

I meant that we won't have to return on every single service, but just if there are any still `decommissioned=false`.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Reworded to be more explicit about expectations. We're waiting for node decommissioning, which is signalled by the node service label turning true.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Done.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Makes sense. Done. Also dropped `pointer.Ptr` use.

rzetelskik (approved):
Thanks for the updates and well done on making this a bit more edible
/lgtm

PR #3633, opened 2026-08-31
feat: reject and warn on node count changes conflicting with an ongoing decommission

czeslavo:
Guards a rack's node count against an ongoing decommission on admission, on both `ScyllaDBDatacenter` and `ScyllaCluster`, using the record added in OPERATOR-362 and OPERATOR-363.

**Removing a rack that still has nodes leaving is rejected**, naming the nodes that hold it up. The existing rule only blocked the removal while the rack's `StatefulSet` had replicas: the node count in the rack status mirrors those replicas and reaches zero before the leaving nodes are pruned, while the member `Service` and `PVC` live until `pruneServices` succeeds. Once the rack is gone from the spec, the controller can no longer resolve the rack of those objects, so it never prunes them and its sync stays degraded. The new check is placed before the staleness check, so a stale status doesn't mask it with a less actionable error. The error is retryable: the record is derived from a `Service` cache snapshot, so it can still list a node for about one reconcile after its `Service` is gone.

**Changing the number of nodes of a rack that has nodes leaving produces a warning.** The change is accepted, but the rack reconciles as if it hadn't been made until the leaving nodes have finished decommissioning and been removed, which is otherwise invisible to the user.

Covered by unit tests for both APIs and by an envtest that holds the last leaving node's `Service` with a finalizer to freeze the window in which the removal used to be admitted.

Closes https://scylladb.atlassian.net/browse/OPERATOR-364.

### DIFF
--- a/docs/source/operate/scale-add-remove-racks.md
+++ b/docs/source/operate/scale-add-remove-racks.md
@@ -165,11 +165,13 @@ You must scale the rack to zero members first, wait for decommissioning to finis
 Update the ScyllaCluster spec to set `members: 0` for the rack being removed:
 
 ```bash
-kubectl -n scylla patch scyllacluster scylla --type=json \
-  -p='[{"op":"replace","path":"/spec/datacenter/racks/<index>/members","value":0}]'
-```
+# Zero-based index of the rack being removed in spec.datacenter.racks. Change it for your cluster.
+RACK_INDEX=0
 
-Replace `<index>` with the zero-based index of the rack in the `racks` array.
+kubectl -n scylla patch scyllacluster scylla --type=json --patch-file=/dev/stdin <<EOF
+[{"op": "replace", "path": "/spec/datacenter/racks/${RACK_INDEX}/members", "value": 0}]
+EOF
+```
 
 Wait for the Operator to decommission all nodes in the rack:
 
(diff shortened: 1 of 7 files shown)


docs/source/operate/scale-add-remove-racks.md
```diff
@@ -186,6 +186,23 @@ kubectl -n scylla get pods -l scylla/rack=<rack-name>
 
 Expected output: no pods listed.
 
+Verify no member of the rack is still leaving the cluster:
+
+```bash
```
mflendrich, line 193: This jsonpath looks heavy, especially with the burden of hand-patching the rack name and a poor failure mode _If `members` prints nothing either, the rack name is wrong._

suggestion: ask to just `describe scyllacluster` and eyeball the rack status for the relevant fields.
This makes the command less intimidating and error-prone (especially for the misspelled rack name case)

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -196,6 +213,9 @@ kubectl -n scylla edit scyllacluster scylla
 
 Delete the entire rack entry. Save and apply.
 
+The removal is rejected while the rack still has members leaving the cluster, with the members that hold it up named in the error.
+Wait for them to be removed and retry.
```
mflendrich: 1. passive voice "the removal is rejected" (by whom?); suggestion: clarify the subject (`The admission webhook will reject a removal while the rack still has members leaving the cluster`)
2. I suggest showing an example of the rejection error

docs/source/operate/scale-add-remove-racks.md
czeslavo: Agree. Changed it to `kubectl describe` with an explicit expected output.

Also adjusted commands for rack member updates to 0 and rack removal to use a proper env variable instead of hand-waving a problem with `kubectl edit` or telling the user to replace `<index>` manually in the patch.

docs/source/operate/scale-add-remove-racks.md
czeslavo: 1. Dropped the passive voice.
2. Added an explicit example of the rejection error.

mflendrich (approved):
/lgtm

PR #3634, opened 2026-08-31
test(e2e): let ScyllaDB default the test keyspace replication

rzetelskik:
The e2e `DataInserter` couples two unrelated things: the hosts it gets are both the driver's contact points and the source of the keyspace's replication factor. Tests that need a specific RF have to contort the host list to get it - e.g. in https://github.com/scylladb/scylla-operator/pull/3629#discussion_r3893900930 a test would have to pass only the nodes remaining after a scale-down operation to keep RF below the scaled-down node count, which as a side effect limited the client's contact points to those nodes and stopped reproducing reality (although this could be further simplified and improved by passing just the identity service address as an endpoint).
The derived RF is also RF-rack-invalid on our single-rack fixtures - RF=3 on a 1-rack DC.

This decouples the two by dropping RF derivation entirely: the keyspace now omits the replication factor, and with optionless `NetworkTopologyStrategy` ScyllaDB places a replica on every rack of every datacenter, which is RF-rack-valid by construction and independent of the host list. The tests whose data must survive a node's storage loss (`scyllacluster_pv.go`, `scyllacluster_replace.go`) get their replication from topology instead - the zonal fixture with 3 racks of 1 member.

Risks and mitigations:

* Optionless `NetworkTopologyStrategy` requires ScyllaDB 2026.1+. All versions covered by the suites, including the upgrade path from 2026.1.10, support it.
* Tests on single-rack fixtures silently drop from RF = node count to RF=1. For `scyllacluster_rebootstrap.go` this is irrelevant - the PVCs survive, so the data never depends on replication. For `scyllacluster_scaling.go` it's a stricter check: at RF=1, `CL=ALL` reads verify that decommission actually streamed every range.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-372

/cc

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_replace.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_replace.go
@@ -179,5 +179,8 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
 		o.Expect(postReplacementHostIDs).To(o.HaveLen(int(utils.GetMemberCount(sc))))
 		o.Expect(postReplacementHostIDs).NotTo(o.ContainElement(preReplacementHostID))
 		o.Expect(postReplacementHostIDs).To(o.ContainElement(postReplacementHostID))
+
+		framework.By("Verifying the data survived the replacement")
+		verification.VerifyCQLData(ctx, di)
 	})
 })
--- a/test/e2e/set/scyllacluster/scyllacluster_replace.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_replace.go
@@ -32,8 +32,9 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
 		ns, nsClient, ok := f.DefaultNamespaceIfAny()
 		o.Expect(ok).To(o.BeTrue(), "Can't get default namespace")
 
-		sc := f.GetDefaultScyllaCluster()
-		sc.Spec.Datacenter.Racks[0].Members = 3
+		// Use 3 racks of 1 member each instead of a single rack of 3 members, so that the default keyspace
+		// replication of a replica per rack replicates the test data across 3 nodes.
+		sc := f.GetDefaultZonalScyllaClusterWithThreeRacks()
 
 		framework.By("Creating a ScyllaCluster")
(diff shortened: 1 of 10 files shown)


test/e2e/set/scyllacluster/scyllacluster_replace.go
```diff
@@ -51,6 +52,8 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
 		hosts, err := utils.GetBroadcastRPCAddresses(ctx, nsClient.KubeClient().CoreV1(), sc)
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(hosts).To(o.HaveLen(int(utils.GetMemberCount(sc))))
+		// The node's storage is discarded and the replacement streams from the other replicas, so the data only
```
czeslavo, line 55: I noticed this test doesn't really verify if data survives replacement - there's no `VerifyCQLData` afterwards. Should we add it?

test/e2e/utils/verification/cql.go
```diff
@@ -19,12 +21,8 @@ func InsertAndVerifyCQLData(ctx context.Context, hosts []string, options ...util
 	return di
 }
 
-func InsertAndVerifyCQLDataByDC(ctx context.Context, hosts map[string][]string) *utils.DataInserter {
-	di, err := utils.NewMultiDCDataInserter(hosts)
-	o.Expect(err).NotTo(o.HaveOccurred())
-
-	InsertAndVerifyCQLDataUsingDataInserter(ctx, di)
-	return di
```
czeslavo, line 27: Nit: let's consider dropping this function altogether, as it is now only responsible for flattening the map.

test/e2e/set/scyllacluster/scyllacluster_replace.go
rzetelskik: done

test/e2e/utils/verification/cql.go
rzetelskik: done, although ideally we wouldn't pass the hosts directly but let the client discover them using the identity service, following our client connection guide: https://operator.docs.scylladb.com/stable/connect-your-app/connect-via-cql.html#driver-configuration-tips. I'll add a separate tracker for this improvement.

czeslavo (approved):
/lgtm

PR #3635, opened 2026-08-31
test(e2e): split the ScyllaCluster scaling e2e test into granular scenarios

rzetelskik:
Currently, the `ScyllaCluster should support scaling` E2E test walks a single cluster through 3 -> 5 -> 4 -> 3 -> 5 in one spec, so a single test case conflates different functionalities (scale-up, decommission, regression test covering decommission after manual drain etc). This PR splits it into separate test cases.

Prerequisites:
- [x] https://github.com/scylladb/scylla-operator/pull/3634

Resolves https://scylladb.atlassian.net/browse/OPERATOR-371

/cc

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_scaling.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_scaling.go
@@ -5,6 +5,7 @@ package scyllacluster
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -29,202 +30,176 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
-		o.Expect(err).NotTo(o.HaveOccurred())
+			step.verifyFunc(previousHostIDs, hostIDs, leftHostIDs)
 
-		scyllaclusterverification.Verify(ctx, f.KubeClient(), f.ScyllaClient(), sc)
-		scyllaclusterverification.WaitForFullQuorum(ctx, f.KubeClient().CoreV1(), sc)
+			for _, previousHostID := range previousHostIDs {
+				if !slices.Contains(hostIDs, previousHostID) {
+					leftHostIDs = append(leftHostIDs, previousHostID)
+				}
+			}
```
czeslavo, line 82: Shouldn't this be moved before the `step.verifyFunc` call?

rzetelskik: It doesn't change anything practically, but I agree it's cleaner - changed the position and the docstring on leftHostIDs

czeslavo: 🤔 I think it made a difference, as otherwise `step.verifyFunc` call would always get an empty slice, no? Anyways, looks correct now.

czeslavo (approved):
/lgtm

test/e2e/set/scyllacluster/scyllacluster_scaling.go
rzetelskik: it iterates over steps, so leftHostIDs would contain nodes that left in previous steps - now it also includes nodes that left in *this* step

PR #3637, opened 2026-09-02
docs: cover parallel decommission in the scaling docs

czeslavo:
Documents parallel decommission to the same extent as parallel bootstrap: the scaling how-to's parallel node operations section now covers scale-in, with the tablets vs vnodes difference, the deferral of node count changes while nodes are leaving and a short section on a scale-down that doesn't finish, and the StatefulSets and PDB explanations no longer state that nodes always leave one at a time.

Prerequisites:
- [x] https://github.com/scylladb/scylla-operator/pull/3639

Note for clarity: the parallel decommission feature itself hasn't been merged yet (https://github.com/scylladb/scylla-operator/pull/3639). We hold this PR until it is.

Closes https://scylladb.atlassian.net/browse/OPERATOR-332

### DIFF
--- a/docs/source/operate/scale-add-remove-racks.md
+++ b/docs/source/operate/scale-add-remove-racks.md
@@ -9,23 +9,31 @@ Scaling changes the replica count of that StatefulSet:
 
 - **Scale up** — new pods are appended at the end of the ordinal sequence (highest index).
   After the new node joins the token ring, the Operator automatically triggers a [data cleanup](../understand/automatic-data-cleanup.md) on affected nodes.
-- **Scale down** — the Operator decommissions the highest-ordinal pod first, streams its data to the remaining nodes, reduces the replica count, and then deletes the PVC and Service.
-  Only one node is decommissioned at a time.
+- **Scale down** — the Operator decommissions the nodes above the new count, highest ordinal first.
+  It waits for them to stream their data to the remaining nodes, then lowers the replica count and deletes their PVCs and Services.
+  Whether the Operator decommissions the leaving nodes one at a time or all at once depends on [parallel node operations](#sequential-and-parallel-node-operations).
 
 Because StatefulSets maintain contiguous pod ordinals and scale down from the highest ordinal, you cannot remove an arbitrary node from the middle of a rack.
 If a specific node is unhealthy, use [node replacement](replace-nodes.md) instead.
 
 For background on the StatefulSet-per-rack architecture, see [StatefulSets and racks](../understand/statefulsets-and-racks.md).
 
-## Sequential and parallel node provisioning
+## Sequential and parallel node operations
 
-The Operator controls how it provisions new nodes with parallel node operations. They determine whether Operator starts ScyllaDB nodes one at a time or all at once: when you create a cluster, add, or scale-out a rack.
+Parallel node operations control whether the Operator acts on nodes one at a time or all at once.
+This covers starting nodes when you create a cluster, add a rack or scale a rack up, and decommissioning nodes when you scale a rack down.
 
-With parallel node operations disabled, Operator starts ScyllaDB nodes one at a time. Within a rack, each Pod must become ready before Operator starts the next one. Operator creates racks one at a time. Bringing up a cluster takes as long as the sum of every node's startup time.
+With parallel node operations disabled, the Operator starts ScyllaDB nodes one at a time. Within a rack, each Pod must become ready before the Operator starts the next one. The Operator creates racks one at a time. Bringing up a cluster takes as long as the sum of every node's startup time.
+Likewise, scaling down removes one node at a time. The next decommission starts only after the previous node's Pod is gone, and no rack in the datacenter scales up or down in the meantime.
 
-With parallel node operations enabled, Operator starts all ScyllaDB nodes at once. Operator starts Pods of a rack without waiting for the previous ones to become ready. Operator creates all racks at the same time. Bringing up a cluster is faster, and the difference grows with the number of nodes.
+With parallel node operations enabled, the Operator starts all ScyllaDB nodes at once. The Operator starts Pods of a rack without waiting for the previous ones to become ready. The Operator creates all racks at the same time. Bringing up a cluster is faster, and the difference grows with the number of nodes.
+Likewise, scaling down decommissions all the nodes above the new count at once. The other racks can still scale up or down while it does.
+Configuration updates and version upgrades still wait until no node in the datacenter is leaving, in both modes.
 
 Parallel node operations provide better performance when your keyspaces are backed by tablets (the default since ScyllaDB 2025.2), as opposed to vnodes. Therefore, we strongly recommend that you only use tablets for your data keyspaces. This is because with vnode-based keyspaces, a joining node streams its data before it finishes joining, which slows down bringing up new nodes. With tablets, the data is moved in the background after the node joins, so the time to bring up new nodes is not affected by data streaming.
+The same holds for scaling down.
+ScyllaDB migrates tablets off all the leaving nodes at the same time.
+It streams vnode data off one leaving node at a time, no matter how many nodes are leaving, so a parallel scale-down of vnode data takes about as long as a sequential one.
 Consider setting [`tablets_mode_for_new_keyspaces`](https://docs.scylladb.com/manual/stable/architecture/tablets.html#enabling-tablets) to `enforced` in your [ScyllaDB configuration](../deploy-scylladb/deploy-your-first-cluster.md#create-a-scylladb-configuration) to prevent individual keyspaces from opting out of tablets.
 
 :::{note}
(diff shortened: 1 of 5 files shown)


czeslavo:
/hold 
for https://github.com/scylladb/scylla-operator/pull/3639

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -9,23 +9,31 @@ Scaling changes the replica count of that StatefulSet:
...
+Parallel node operations control whether the Operator acts on nodes one at a time or all at once.
+This covers starting nodes when you create a cluster, add a rack or scale a rack up, and decommissioning nodes when you scale a rack down.
 
-With parallel node operations disabled, Operator starts ScyllaDB nodes one at a time. Within a rack, each Pod must become ready before Operator starts the next one. Operator creates racks one at a time. Bringing up a cluster takes as long as the sum of every node's startup time.
+With parallel node operations disabled, the Operator starts ScyllaDB nodes one at a time. Within a rack, each Pod must become ready before the Operator starts the next one. The Operator creates racks one at a time. Bringing up a cluster takes as long as the sum of every node's startup time.
+Likewise, scaling down removes one node at a time. The next decommission starts only after the previous node has left and been removed, and no rack in the datacenter scales up or down in the meantime.
```
rzetelskik: has left and been removed?

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -9,23 +9,31 @@ Scaling changes the replica count of that StatefulSet:
...
 
-With parallel node operations enabled, Operator starts all ScyllaDB nodes at once. Operator starts Pods of a rack without waiting for the previous ones to become ready. Operator creates all racks at the same time. Bringing up a cluster is faster, and the difference grows with the number of nodes.
+With parallel node operations enabled, the Operator starts all ScyllaDB nodes at once. The Operator starts Pods of a rack without waiting for the previous ones to become ready. The Operator creates all racks at the same time. Bringing up a cluster is faster, and the difference grows with the number of nodes.
+Likewise, scaling down decommissions all the leaving nodes of a rack at once, and the Operator removes them as they finish, from the highest ordinal down. The other racks can still scale up or down while it does.
```
rzetelskik: I understand this description tries to differentiate the operations on DB's side and k8s side but this makes it unnecessarily complicated imo: "[...] decommissions all the leaving nodes [...] and the Operator removes them as they finish". We have decommission, leave, remove, finish in one sentence.

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -51,19 +59,16 @@ spec:
...
 :::
 
 If you don't specify the field, the Operator defaults it to `true` on creation, provided the ScyllaDB version is higher or equal to 2026.2.
 
-Operator keeps bootstrapping the nodes of clusters that already existed before the addition of this feature sequentially. 
-It is recommended that you set the field to `true` explicitly to bootstrap new nodes in parallel.
+For clusters created before this feature existed, the Operator keeps starting and decommissioning nodes one at a time.
+Set the field to `true` explicitly to start and decommission their nodes in parallel.
+
+Parallel node operations aren't available for datacenters managed by `ScyllaDBCluster`. The Operator always starts and decommissions their nodes one at a time.
```
rzetelskik: do we need this? We don't really expose this API

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -74,6 +79,19 @@ It is recommended that you enable it whether or not parallel node operations are
 
 ## Scale a ScyllaCluster
 
+:::{warning}
+Before scaling down, make sure the resulting topology is valid for ScyllaDB.
+A scale-down is subject to the same [prerequisites as `nodetool decommission`](https://docs.scylladb.com/manual/stable/operating-scylla/nodetool-commands/decommission.html):
+
+- every node of the cluster is up,
+- at least as many nodes as the highest replication factor of any keyspace stay in the datacenter,
```
rzetelskik: this feels outdated for enforce_rack_list? As in it doesn't paint a complete picture

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -74,6 +79,19 @@ It is recommended that you enable it whether or not parallel node operations are
...
+- the remaining nodes have room for the data of the leaving nodes,
+- rack-aware tablet replication remains valid.
```
rzetelskik: Difficult to decode what that means - we should at least link to a specific document here. Although honestly I'd drop listing the requirements on our side and just point to ScyllaDB docs in general, this  may easily go stale.

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -119,12 +137,45 @@ Wait for the operation to complete:
...
+```bash
+kubectl -n scylla get scyllacluster scylla -o jsonpath='{range .status.conditions[?(@.status=="True")]}{.type}: {.reason}: {.message}{"\n"}{end}'
+```
+
+- `WaitingForRackServiceDecommission` — ScyllaDB hasn't finished decommissioning the named node, or keeps rejecting the request.
+  Run `nodetool status` on another node: `UL` means the data is still moving, `UN` means ScyllaDB hasn't accepted the decommission.
+  The usual cause is a violated [decommission prerequisite](#scale-a-scyllacluster): a node is down, too few nodes would stay for a keyspace's replication factor, or the nodes that stay have no room for the data.
+  The sidecar retries every few seconds. Its attempts and ScyllaDB's reason for a rejection both appear in the log of the leaving Pod's `scylla` container.
+  Once you fix the cause, the next retry succeeds and the decommission proceeds on its own.
+- `WaitingForStatefulSetRollout` — a scale-down only starts when every rack is fully ready. A node in [maintenance mode](use-maintenance-mode.md) is a common cause.
```
rzetelskik: > A node in [maintenance mode](use-maintenance-mode.md) is a common cause.

is this halucinated? Why would it be a common cause?

docs/source/operate/scale-add-remove-racks.md
```diff
@@ -119,12 +137,45 @@ Wait for the operation to complete:
...
+
+The Operator never times out a decommission; ScyllaDB decides how long moving the data takes.
+If the `ScyllaCluster` stays at `Progressing=True` longer than you expect, read the reason:
+
+```bash
+kubectl -n scylla get scyllacluster scylla -o jsonpath='{range .status.conditions[?(@.status=="True")]}{.type}: {.reason}: {.message}{"\n"}{end}'
+```
+
+- `WaitingForRackServiceDecommission` — ScyllaDB hasn't finished decommissioning the named node, or keeps rejecting the request.
+  Run `nodetool status` on another node: `UL` means the data is still moving, `UN` means ScyllaDB hasn't accepted the decommission.
```
rzetelskik: isn't looking at the logs of the decommissioned node a more obvious first step? If it's blocked it will say it explicitly

docs/source/understand/statefulsets-and-racks.md
```diff
@@ -51,9 +51,15 @@ When you decrease the `members` count, the Operator removes nodes from the **end
...
+
+With parallel node operations disabled, the Operator scales down **one node at a time**, regardless of how large the requested scale-down is, so that only one decommission is in progress in the datacenter at any moment.
+With parallel node operations enabled, the Operator requests the decommission of all the nodes above the new count at once, highest ordinal first, and scales the StatefulSet below the decommissioned nodes at the top of the rack as they finish.
+It doesn't wait for the whole batch, because a decommissioned pod is not ready and counts against the [PodDisruptionBudget](pod-disruption-budgets.md) until it is removed.
+A decommission can't be cancelled, so the Operator holds the rack's `members` count against raises while any node is leaving.
+A lowered count extends the scale-down right away.
+A raised count waits until the leaving nodes and their PVCs and Services are removed; nodes whose decommission hasn't been requested yet stay, and the Operator bootstraps the rest of the returning nodes as new, empty nodes.
+The `ScyllaCluster` lists the leaving nodes in `status.racks[].decommissioningMembers` until they are removed.
```
rzetelskik: isn't this a bit over the top for this doc? Shouldn't we just link to the scaling docs for this level of detail?

docs/source/upgrade/upgrade-operator.md
```diff
@@ -64,6 +64,40 @@ The below instructions are **supplementary** to the standard upgrade procedure a
...
+If the node was in fact decommissioned, meaning its Pod isn't ready and its host ID is no longer in `nodetool status`, leave the label in place. The upgraded Operator removes the node.
+
+:::{note}
+**Why is this necessary?**
+Starting with v1.23, the Operator treats a member Service labelled `scylla/decommissioned=true` as a node that has left the cluster.
+It removes the node together with its PVC before the rack reconciles to the requested node count.
+Previous versions only acted on the label during a scale-down, so a stale label could sit on a healthy node without consequences.
+On a rack whose node at ordinal N carries a stale label, the upgraded Operator decommissions every node above N, one at a time or all at once with `enableParallelNodeOperations` enabled.
+It then removes node N together with its PVC without decommissioning it, and bootstraps the rack back to the requested count with new, empty nodes.
+A healthy node N loses its data this way, the Operator rebuilds every node above it, and you have to remove node N from the ScyllaDB cluster by hand with `nodetool removenode`.
```
rzetelskik: This briefly touches on the recovery procedure - I wouldn't include it or link to actual recovery steps.

docs/source/operate/scale-add-remove-racks.md
czeslavo: Not sure what you're asking for here - is this statement wrong? Should this be more precise in technical terms? 

I assume you're pointing out that there's no explanation of what "leaving and being removed" mean? I simplified it to just "the previous node's Pod is gone".

docs/source/operate/scale-add-remove-racks.md
czeslavo: Reduced to just "decommissions all the nodes above the new count at once".

docs/source/operate/scale-add-remove-racks.md
czeslavo: Fair point, dropped.

docs/source/operate/scale-add-remove-racks.md
czeslavo: Agreed, dropped the exact list and left just a link to the decommission command documentation.

docs/source/operate/scale-add-remove-racks.md
czeslavo: Maybe "common" is not the most suitable word here. This is just an example of a reason why a node can be unready.

docs/source/operate/scale-add-remove-racks.md
czeslavo: Fair point, suggested looking into the node's logs first.

docs/source/understand/statefulsets-and-racks.md
czeslavo: Agreed, let's not go into details here.

docs/source/upgrade/upgrade-operator.md
czeslavo: Dropped the part suggesting a recovery: "and you have to remove node N from...".

docs/source/operate/scale-add-remove-racks.md
rzetelskik: Sorry, my point was that it basically means the same thing. I think the differentiation only matters from implementation perspective, but for the end user these states shouldn't differ unless something goes wrong, i.e. we should never let a node leave and then not remove its pod.

rzetelskik (approved):
/lgtm

PR #3639, opened 2026-09-02
feat: decommission nodes in parallel with parallel node operations enabled

czeslavo:

`enableParallelNodeOperations` now also controls how nodes are decommissioned. With it enabled, all the nodes a rack is scaled down by are requested to decommission at once, from the highest ordinal down so that a node count reverted mid-way saves the nodes not requested yet, and the StatefulSet is scaled below the decommissioned nodes at the top of the rack as they finish; a decommission in one rack no longer holds the scaling of the other racks. Along the way, the rollout wait before the scaling loop skips racks with leaving nodes, since a leaving node's Pod is not ready and waiting for it never ended when the node count was raised back to the StatefulSet replicas mid-decommission, and `DeferringRackNodeCountChange` is only reported when the spec actually changed, instead of for every sequential multi-node scale-down.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-329


pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -590,13 +592,17 @@ func makeWaitingForRackServicePruningCondition(sdc *scyllav1alpha1.ScyllaDBDatac
...
+// pruned, and a scale-down begins here by labelling the nodes above the target node count. Every leaving node goes
+// through the following lifecycle:
+//  1. Request: the node is stamped with the scylla/decommissioned=false label.
 //  2. Wait: the node's sidecar runs the decommission and flips the label to true.
 //  3. Conclude: the StatefulSet is scaled below the node.
 //  4. Prune: pruneServices removes the node's Service and PVC (syncServices).
 //
+// With parallel node operations disabled, one node at a time goes through it, from the highest ordinal. With parallel
+// node operations enabled, all the nodes above the target count are requested at once, from the highest ordinal down,
+// and the StatefulSet is scaled below all of them in one step once every one of them is decommissioned.
```
rzetelskik: is there a good reason to do this, instead of scaling down greedily? I.e. scale down as many as possible from the highest ordinal down that have already reached decommissioned=true?

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -657,48 +680,75 @@ func (sdcc *Controller) syncRackDecommission(ctx context.Context, sdc *scyllav1a
...
-			ObservedGeneration: sdc.Generation,
-		})
-		// Services are managed in the other loop.
-		// When informers see the new service, will get re-queued.
-		return progressingConditions, nil
+	// Every leaving node is requested through its member Service, so check that all of them are present before
+	// requesting any: the first request fixes the batch (see below), and a Service missing from the caches, e.g. one
+	// being recreated, would otherwise split it.
+	for ordinal := lowestLeavingOrdinal; ordinal < *sts.Spec.Replicas; ordinal++ {
+		svcName := fmt.Sprintf("%s-%d", sts.Name, ordinal)
```
rzetelskik: nit: we could have a helper that could also be folded into naming.MemberServiceName

pkg/controller/scylladbdatacenter/sync_statefulsets.go
```diff
@@ -657,48 +680,75 @@ func (sdcc *Controller) syncRackDecommission(ctx context.Context, sdc *scyllav1a
...
+	// Request the leaving nodes from the highest ordinal down. The target node count is the lowest leaving ordinal,
+	// so the first request fixes the batch and every further one extends it downwards. A node count reverted after
+	// a pass cut short, e.g. by a restart, thus saves the nodes that weren't requested yet: the batch stays truncated
+	// to the requested ones, which finish leaving and are pruned, and the rest of the scale-down resumes as another
+	// round. This gives up the atomicity of the batch on purpose. A decommission can't be taken back while a
+	// scale-down can be re-issued, so the order errs towards fewer nodes leaving, and a node that isn't requested yet
+	// is never lost to a change of mind in either mode.
+	decommissionRequested := false
+	for ordinal := *sts.Spec.Replicas - 1; ordinal >= lowestLeavingOrdinal; ordinal-- {
+		svc := rackServices[fmt.Sprintf("%s-%d", sts.Name, ordinal)]
```
rzetelskik: ditto

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
```diff
@@ -698,13 +728,22 @@ func Test_syncRackDecommission(t *testing.T) {
...
 		},
 		{
-			name:                                     "a fresh scale-down stamps the highest node",
-			sdc:                                      newSDCWithNodes(1),
-			rackName:                                 rackName,
-			sts:                                      newSts(2),
-			rackServices:                             newRackServices(newMemberService(0, nil), newMemberService(1, nil)),
-			expectedConditions:                       []metav1.Condition{makeStampCondition(newMemberService(1, nil))},
-			expectedDecommissionRequestedServiceName: stsName + "-1",
+			name:               "a fresh scale-down stamps the highest node",
```
rzetelskik: nit: with parallel node operations disabled?

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
```diff
@@ -698,13 +728,22 @@ func Test_syncRackDecommission(t *testing.T) {
...
+			sdc:                newSDCWithNodes(1),
+			rackName:           rackName,
+			sts:                newSts(2),
+			rackServices:       newRackServices(newMemberService(0, nil), newMemberService(1, nil)),
+			expectedConditions: []metav1.Condition{makeStampCondition(newMemberService(1, nil))},
+			expectedDecommissionRequestedServiceNames: []string{stsName + "-1"},
+		},
+		{
+			name:               "a fresh multi-node scale-down stamps the highest node only",
```
rzetelskik: ditto

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
```diff
@@ -809,6 +848,138 @@ func Test_syncRackDecommission(t *testing.T) {
...
+			name:                   "a fully decommissioned multi-node scale-down concludes by scaling the StatefulSet to the target in one step with parallel node operations enabled",
+			sdc:                    newParallelSDCWithNodes(1),
+			rackName:               rackName,
+			sts:                    newSts(3),
+			rackServices:           newRackServices(newMemberService(0, nil), newMemberService(1, new(naming.LabelValueTrue)), newMemberService(2, new(naming.LabelValueTrue))),
+			expectedConditions:     []metav1.Condition{makeScaleCondition(newSts(3), 1)},
+			expectedScaledReplicas: new(int32(1)),
+		},
+		{
+			name:         "a node count lowered further mid-decommission is deferred with parallel node operations enabled",
```
rzetelskik: is there a reason not to let it scale down further? Also doesn't it have potential to sequentialise node decommissions? How do we discern between two consecutive scales vs a single batch that failed on a service update?

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
```diff
@@ -809,6 +848,138 @@ func Test_syncRackDecommission(t *testing.T) {
...
+					Reason:             "DeferringRackNodeCountChange",
+					Message:            fmt.Sprintf(`Deferring node count change of rack %q to 0 until the Services of its decommissioning nodes ["%[2]s-1" "%[2]s-2"] are pruned.`, rackName, stsName),
+					ObservedGeneration: sdc.Generation,
+				},
+				makeWaitingForDecommissionCondition(1),
+				makeWaitingForDecommissionCondition(2),
+			},
+		},
+		{
+			name:         "a multi-node scale-down with a leaving node's Service missing waits for it before requesting any with parallel node operations enabled",
```
rzetelskik, line 1150: don't we have a race here? IIUC two things happen when service receives the decommissioned=true label:
- the service is pruned
- the STS is scaled down
Isn't it possible that we run the entire loop again if a service is pruned before the STS is scaled down?

rzetelskik (commented):
first batch

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
rzetelskik: Discussed offline - a pruned service does not make it into nodes being considered decommissioning and thus does not block scaledown

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: The motivation behind it was mainly being deterministic about how many scale operations are going to happen throughout the whole decommissioning. But, as we spoke, I agree that there are more benefits from doing it greedily (we free up the resources sooner, we do not affect PDB, etc.). Updated the implementation to scale down as soon as N >= 1 of the STS suffix nodes are `decommissioned=true`.

test/envtest/controllers/scylladbdatacenter_controller_test.go
```diff
@@ -481,7 +410,270 @@ var _ = g.Describe("ScyllaDBDatacenter controller", func() {
 				}))
 			}).WithContext(ctx).WithTimeout(scyllaDBDatacenterControllerDefaultEventuallyTimeout).WithPolling(100 * time.Millisecond).Should(o.Succeed())
 		})
-	})
+
+		if enableParallelNodeOperations {
```
rzetelskik: I'd move those to separate describe block with so many separate cases, the table can contain the shared ones

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
czeslavo: No, we shouldn't race here. `pruneServices` only prunes a service with an ordinal >= STS spec replicas. By then, the ordinal is outside of the check that ensures services exist before stamping them.

This case specifically covers a case when a service is absent while the STS still counts it, e.g., after a scale up.

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
```diff
@@ -973,6 +973,24 @@ func Test_syncRackDecommission(t *testing.T) {
 				makeWaitingForDecommissionCondition(2),
 			},
 		},
+		{
+			name:               "a stale decommissioned label on the lowest node drains the rack above it one node at a time",
```
rzetelskik: nit: with parallel node operations disabled?

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
```diff
@@ -973,6 +973,24 @@ func Test_syncRackDecommission(t *testing.T) {
...
+			sdc:                newSDCWithNodes(3),
+			rackName:           rackName,
+			sts:                newSts(3),
+			rackServices:       newRackServices(newMemberService(0, new(naming.LabelValueTrue)), newMemberService(1, nil), newMemberService(2, nil)),
+			expectedConditions: []metav1.Condition{makeDeferringCondition(3, 0), makeStampCondition(newMemberService(2, nil))},
+			expectedDecommissionRequestedServiceNames: []string{stsName + "-2"},
+		},
+		{
```
rzetelskik, line 1018: I think this could use a comment - the fact that this is a unit test makes it seem like that's the desired behavior, but iiuc stale decommission labels shouldn't happen with the barrier?

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
czeslavo: Yes, as we discussed, it was also motivated by the deterministic behavior - once a decommission is started, we wait until all nodes are decommissioned before we stamp more. From a practical standpoint, as you noted, this unnecessarily splits the entire decommissioning into two batches. Agreed that it will be better not to wait here, as it's completely safe to extend the decommission range. It will even simplify the code. Updated the implementation.

pkg/controller/scylladbdatacenter/sync_statefulsets.go
czeslavo: Added a helper :+1:

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
czeslavo: Added the suffix to all test cases.

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
czeslavo: Already added a suffix.

pkg/controller/scylladbdatacenter/sync_statefulsets_test.go
czeslavo: This should not happen now, but it's to document what we'd do if someone had a stale `decommissioned=true` service in the cluster (i.e., didn't follow the upgrade guide). Agreed, added comments on all stale label cases for clarity.

test/envtest/controllers/scylladbdatacenter_controller_test.go
czeslavo: Makes sense. Split.

rzetelskik (approved):
/lgtm

PR #3642, opened 2026-09-03
test(e2e): scale by multiple nodes and across racks

rzetelskik:
The scaling entries each moved a single node, which is indistinguishable between parallel and sequential node operations, and never changed a node count on a multi-rack cluster.

This PR changes them to move multiple nodes and adds additional cases moving nodes across racks.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-331

/cc @czeslavo

### DIFF
--- a/test/e2e/set/scyllacluster/scyllacluster_scaling.go
+++ b/test/e2e/set/scyllacluster/scyllacluster_scaling.go
@@ -31,7 +31,7 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
 	})
 
 	type scalingStep struct {
-		members int32
+		rackLayout
 		// beforeFunc, when set, runs against the cluster in the state preceding the scaling operation.
 		beforeFunc func(ctx context.Context, f *framework.Framework, sc *scyllav1.ScyllaCluster)
 		// verifyFunc asserts how the cluster's members changed over the scaling operation.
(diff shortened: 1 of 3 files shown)


test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -31,7 +31,8 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
 	})
 
 	type scalingStep struct {
-		members int32
+		rackCount      int32
+		membersPerRack int32
```
czeslavo: Nit: let's consider extracting a `rackTopology` struct with `rackCount` and `membersPerRack` fields to avoid passing two naked int32s in helpers.

test/e2e/set/scyllacluster/scyllacluster_scaling.go
```diff
@@ -86,40 +88,69 @@ var _ = g.Describe("ScyllaCluster", framework.SuiteParallel, framework.SuitePara
...
+					beforeFunc:     markHighestOrdinalForMaintenanceAndDrain,
+					verifyFunc:     verifyScaledIn,
 				},
 			},
 		}),
 		// Scaling back out verifies that a decommissioned node's storage isn't left in place and reused.
 		g.Entry("out, with new storage after decommissioning", &horizontalScalingEntry{
-			initialMembers: 3,
+			initialRackCount:      1,
+			initialMembersPerRack: 2,
```
czeslavo: Shouldn't this be 3 so that we verify this for parallel operations as well?

rzetelskik: It's a separate functionality but we can cover 1->3->1 here as well, I don't see why not

test/e2e/set/scyllacluster/scyllacluster_scaling.go
rzetelskik: Not a fan of this but done

czeslavo (approved):
/lgtm

PR #3643, opened 2026-09-03
fix(webhook): set controller-runtime logger once per process

czeslavo:
`WebhookOptions.run` called controller-runtime's `SetLogger`, which isn't safe for concurrent calls, so `TestWebhookOptionsRun` raced when its parallel subtests ran it at once. The call moves to `Run`, which runs once per process, and the tests set it once in `TestMain`.

Closes https://scylladb.atlassian.net/browse/OPERATOR-375


rzetelskik (approved):
/lgtm

PR #3627, opened 2026-08-28
docs: add a note about EnableParallelNodeOperations gating future enablement of parallel decommission

rzetelskik:
**Description of your changes:**
This PR adds a docs note about EnableParallelNodeOperations gating future enablement of parallel decommission.
It also strengthens the wording in the changelog. 

**Which issue is resolved by this Pull Request:**
Resolves https://scylladb.atlassian.net/browse/OPERATOR-365

PR #3646, opened 2026-09-04
[1.22] docs: add a note about EnableParallelNodeOperations gating future enablement of parallel decommission

rzetelskik:
Backport of https://github.com/scylladb/scylla-operator/pull/3627.
Not picking up the change in changelog as it only lives in master and should be dropped from release branch.

Resolves #

/cc @czeslavo


czeslavo (approved):
/lgtm

PR #3647, opened 2026-09-04
fix(hack/kind): export KUBECONFIG before applying the registry ConfigMap

rzetelskik:
cluster-setup.sh set KUBECONFIG as a command prefix on cat (KUBECONFIG=... cat <<EOF | kubectl apply -f -), which scopes the assignment to cat and never reaches kubectl on the other side of the pipe. kubectl then fell back to the default localhost:8080 and `make kind-setup` failed with
`dial tcp [::1]:8080: connect: connection refused` on any host without a pre-existing ~/.kube/config pointing at the kind cluster.

The script already wrote and exported a KUBECONFIG for the deployment step further down, so this moves that block above the ConfigMap apply and drops the duplicated mktemp/`kind get kubeconfig` pair. The EXIT trap now covers the temporary kubeconfig for the whole run.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-373

### DIFF
--- a/hack/kind/cluster-setup.sh
+++ b/hack/kind/cluster-setup.sh
@@ -77,10 +77,14 @@ if [ "$(podman inspect -f='{{json .NetworkSettings.Networks.kind}}' "${KIND_REGI
   podman network connect "kind" "${KIND_REGISTRY_NAME}"
 fi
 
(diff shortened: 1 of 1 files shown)


czeslavo (approved):
/lgtm

PR #3650, opened 2026-09-09
[PoC] feat: read your own writes in ScyllaDBDatacenter controller with client-go informers

czeslavo:

That's a PoC of implementing read-your-own-writes semantics in the ScyllaDBDatacenter controller. It relies on client-go's `LastStoreSyncResourceVersion`. `ConsistencyStore` is a type to be used by controllers/client to signal that they made a write/delete and later wait for a given action to be observed before taking any actions based on a read from cache.

It unifies how the controller handles pruning so that it's easier to track deletes. Also, where direct client calls were used for writing/deleting, tracking is covered by decorated clients (`NewRecordingKubeClient`, `NewRecordingScyllaV1alpha1Client`) so that any write/delete is covered in the future without making it an explicit call in the controller code.

I tried making the decorated clients code-generated, but it wasn't working well - the codegen script had to be aware of the package structure, as not everything was possible to infer using reflection etc., so I gave up. Ideas welcome, but I guess that it's not a big deal to add some methods if a controller will need them retroactively.
 
Part of https://scylladb.atlassian.net/browse/OPERATOR-393.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -831,7 +831,7 @@ test-binary: build
 .PHONY: test-binary
 
 test-envtest:
-	ginkgo --tags=envtest -p './test/envtest/controllers/...'
+	ginkgo --tags=envtest -p './test/envtest/...'
 
 help:
 	$(info The following make targets are available:)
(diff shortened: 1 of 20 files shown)


czeslavo:
Closing, the result of discovery and comparison against the controller-runtime approach is to be found in an enhancement proposal: https://github.com/scylladb/scylla-operator/pull/3659

PR #3651, opened 2026-09-09
ci: add a check-required job aggregating GitHub Actions checks

rzetelskik:
We're in the middle of moving CI to GHA, and until GHA gates merges itself tide still does - but tide only knows about contexts it derives from the Prow presubmit config, so with `skip-unknown-contexts: true` every GHA check is optional to it and a PR with red GHA checks merges. A stopgap for the overlap, not the end state.

Rather than list each GHA check in the tide config in another repo, this adds one aggregate, `check-required`. Tide requires that one context; making a GHA job required from here on means adding it to this job's `needs`.

It ships with an empty `needs`, so the merge gate is unchanged for now - `needs.*` expands to nothing and the guard step is skipped. Wiring in `check-changelog` and `e2e-kind-helm` is a follow-up. No `name:` on the job on purpose, so the context is the bare `check-required`.

Merges before the tide change in scylladb/scylla-operator-release, or every open PR here blocks on a context that doesn't exist.

Refs OPERATOR-394.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,0 +1,17 @@
+name: PR
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  check-required:
+    runs-on: ubuntu-latest
+    if: always()
+    steps:
+    - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
+      run: |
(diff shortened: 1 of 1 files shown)


czeslavo (approved):
Note to remember: when we migrate parts of presubmits, we will not include them directly in the `PR` workflow, but will create a separate reusable ' CI ' checks workflow and call it from here.
/lgtm

PR #3652, opened 2026-09-09
[PoC] feat: migrate operator to controller-runtime

czeslavo:

Proof of concept for migrating the operator to controller-runtime, in the order it could be landed. The first part puts the ScyllaDBDatacenter controller on controller-runtime's read-your-writes consistent client while every other controller keeps its client-go code, with no API, RBAC, flag or endpoint changes. The rest moves every other controller of every binary onto controller-runtime, so the whole road is visible.

Part 1, ScyllaDBDatacenter on the consistent client:

- Bumps `k8s.io/*` to v0.37.0 and `sigs.k8s.io/controller-runtime` to v0.25.0 (the first release with the consistent client), with regenerated files.
- Adds `pkg/controllermanager`, which owns a controller-runtime manager and registers every controller of the operator binary with it. Controllers not migrated yet get their typed informers and listers from the manager's cache through a small generic bridge, so there is one informer per kind in the process regardless of which world a controller lives in. Leader election, health probes and the lease name are unchanged; controller-runtime metrics can be enabled with the new `--metrics-bind-address` flag and are off by default.
- Adds `pkg/ctrlclient`, generic adapters from a controller-runtime client to the `resourceapply`/`controllerhelpers` interfaces, and client-backed typed listers for the pure functions that take listers. Refactors `kubecrypto.CertificateManager` onto per-kind object controls; the typed-client constructor stays.
- Ports the ScyllaDBDatacenter controller to a controller-runtime reconciler reading and writing through the manager's read-your-writes client. The cache-propagation sleeps and their option are removed; the upgrade hooks' polling becomes a `RequeueAfter`.
- envtest: the ScyllaDBDatacenter specs run the controller under a manager with informers lagging 500ms behind the API server, a whole-manager spec covers the shared cache, and two specs lag one kind further to cover the decisions previously made from a cache that had not observed the controller's own writes.

Part 2, the rest of the operator:

- Ports the side binaries (bootstrap barrier, ignition, sidecar with status report, node setup daemon) onto managers with namespaced, selector-filtered caches, one commit each.
- Ports the nine remaining single-cluster controllers of the operator binary, one commit each.
- Adds `pkg/remotecluster`, one controller-runtime cluster per RemoteKubernetesCluster driven through the existing dynamic-handler interface, and ports the RemoteKubernetesCluster and ScyllaDBCluster controllers onto it. With that, the informer bridge, the remote informer factories and every legacy runnable leave the controller manager.
- Makes `Reconcile` the entry point of every controller and leaves error handling to controller-runtime; the wrappers that classified errors are gone.

Part of https://scylladb.atlassian.net/browse/OPERATOR-397.


mflendrich:
/test foobarbazqux

(this is a test to verify that the qodo message does not reappear)

czeslavo:
Closed, will be implemented properly following the enhancement proposal in https://github.com/scylladb/scylla-operator/pull/3659.

PR #3654, opened 2026-09-09
test(envtest): cover the ScyllaDBDatacenter controller ahead of its refactor

czeslavo:

Extends the envtest coverage of the ScyllaDBDatacenter controller so that its upcoming refactor has a safety net. The specs pin the current behavior of the paths that had no envtest coverage so far:

- **ScyllaDB upgrade through the hook phases.** The controller runs against a fake ScyllaDB API and the spec follows the whole state machine: the upgrade context ConfigMap, the schema agreement check and the system keyspace snapshots of the pre-upgrade hooks, every StatefulSet partitioned at its node count, the nodes upgraded one at a time from the highest ordinal of the first rack down (drained and snapshotted under maintenance, Pod deleted, partition moved below it), the data snapshot removed once the new Pod is ready, the post-upgrade cleanup of the system snapshots and the upgrade context, and an unknown phase reset to the first one without the snapshots being taken again. Reaching this path required making the ScyllaDB client factory injectable.
- **Node replacement and status aggregation with member Pods.** Envtest runs no StatefulSet controller, so a helper creates the member Pods in its place. The specs cover the replacement by host ID from the wait for the host ID annotation, through the PVC-before-Pod deletion order and the replacing host ID label, to the labels being cleared once the new node reports its host ID, and the aggregation of node counts, versions, readiness and staleness of the racks into the status and conditions.
- **Object ownership and the waiting and degraded paths of the StatefulSet sync.** Excess owned PodDisruptionBudgets, RoleBindings, ServiceAccounts and Ingresses are pruned while the required ones are kept, orphans with matching labels are adopted, a rack StatefulSet or a managed config ConfigMap owned by another controller is left untouched and reported as degraded, a rack StatefulSet without the data volume claim template degrades the StatefulSet sync, a removed rack loses its StatefulSet and rack status, and the sync reports why it waits for the node exporter image, the managed config or another rack's rollout.
- **Status invariants held throughout a change.** The observed generation of the status and its conditions never decreases nor exceeds the generation, also with the ScyllaDBDatacenter informer deliberately kept two seconds behind the API server, and a node is never listed as decommissioning while its Service exists without the decommissioned label. These catch decisions made from a stale informer cache.
- **Config plumbing.** A managed config change rolls the rack StatefulSet out through a new inputs hash, the agent auth token is generated once, kept across syncs, and replaced by an override Secret and then by a custom agent config Secret, following the precedence of its sources, and the TLS Secrets, CA bundle ConfigMaps and CQL connection configs are created.

Spec-side updates of the ScyllaDBDatacenter now retry against the controller's status writes for longer than `retry.DefaultRetry`, which fixed a load-dependent flake of the parallel decommission spec.

**Notes for reviewers**

- The suite is large. Reviewing commit by commit is easier: each commit covers one theme, and only two commits touch non-test code: `refactor(naming): name the labels the ScyllaDBDatacenter controller selects its objects by` and `refactor(scylladbdatacenter): make the ScyllaDB client factory injectable`.
- Three harness devices are deliberate and opt-in per spec: the informer transform that lags one kind behind the API server, the fake ScyllaDB client transport that redirects every request to the fake server while keeping the targeted node in the `Host` header, and the member Pods created in place of the StatefulSet controller, left unscheduled so that the API server deletes and evicts them without a kubelet or a maintained PodDisruptionBudget status.
- Locally, the default `ginkgo -p` may start more envtest control planes than the machine handles and fail in `BeforeEach` with etcd request timeouts. `--procs=4` runs the suite fine.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-399


test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
```diff
@@ -0,0 +1,353 @@
...
+		sdc := setupRolledOutRacks(ctx, env, false, []string{rackName}, 1)
+
+		// The extra objects carry the controller's labels and a controllerRef to the datacenter, as leftovers of
+		// a previous shape of the datacenter would.
+		g.By("Creating excess owned PodDisruptionBudget, RoleBinding, ServiceAccount and Ingress")
+		excessObjectMeta := func(name string) metav1.ObjectMeta {
+			return metav1.ObjectMeta{
+				Name:            name,
+				Namespace:       env.Namespace(),
+				Labels:          naming.ClusterLabels(sdc),
```
rzetelskik: nit: the controller uses only `naming.ClusterNameLabel` for its selector. Maybe we could clean up the mess around labels/selectors in this controller and helpers' naming a bit - see e.g. how it's handled with ScyllaDBClusterDatacenterSelectorLabels

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
```diff
@@ -0,0 +1,353 @@
...
+// nothing owned by another controller. These specs pin what it does with each of them.
+var _ = g.Describe("ScyllaDBDatacenter controller object ownership", func() {
+	const rackName = "rack-a"
+
+	var env *envtest.Environment
+	g.BeforeEach(func(ctx g.SpecContext) {
+		env = envtest.Setup(ctx)
+	})
+
+	g.It("should prune owned objects that aren't required and keep the required ones", func(ctx g.SpecContext) {
```
rzetelskik: nit: perhaps we could turn it into a table test, instead of aggregating all the object types into one?

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
```diff
@@ -0,0 +1,353 @@
...
+		g.By("Creating a ScyllaDBDatacenter with a single rack")
+		sdc, err = env.ScyllaClient().ScyllaV1alpha1().ScyllaDBDatacenters(env.Namespace()).Create(ctx, sdc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Waiting for the StatefulSet sync to be reported as degraded")
+		o.Eventually(func(eo o.Gomega, ctx context.Context) {
+			degradedCondition := getScyllaDBDatacenterCondition(ctx, env, sdc.Name, internalapi.MakeKindControllerCondition("StatefulSet", scyllav1alpha1.DegradedCondition))
+			eo.Expect(degradedCondition).NotTo(o.BeNil())
+			eo.Expect(degradedCondition.Status).To(o.Equal(metav1.ConditionTrue))
+			eo.Expect(degradedCondition.Message).To(o.ContainSubstring("isn't controlled by us"))
```
rzetelskik: nit: seems like a change detector, could we match a reason instead?

test/envtest/controllers/scylladbdatacenter_controller_progressing_test.go
```diff
@@ -0,0 +1,262 @@
...
+		g.By("Creating a ScyllaDBDatacenter with a single rack")
+		sdc, err = env.ScyllaClient().ScyllaV1alpha1().ScyllaDBDatacenters(env.Namespace()).Create(ctx, sdc, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Waiting for the config sync to be reported as degraded and the StatefulSet sync to wait for the managed config")
+		o.Eventually(func(eo o.Gomega, ctx context.Context) {
+			configDegradedCondition := getScyllaDBDatacenterCondition(ctx, env, sdc.Name, internalapi.MakeKindControllerCondition("Config", scyllav1alpha1.DegradedCondition))
+			eo.Expect(configDegradedCondition).NotTo(o.BeNil())
+			eo.Expect(configDegradedCondition.Status).To(o.Equal(metav1.ConditionTrue))
+			eo.Expect(configDegradedCondition.Message).To(o.ContainSubstring("isn't controlled by us"))
```
rzetelskik: ditto

test/envtest/controllers/scylladbdatacenter_controller_progressing_test.go
```diff
@@ -0,0 +1,262 @@
...
+			events, err := env.TypedKubeClient().CoreV1().Events(env.Namespace()).List(ctx, metav1.ListOptions{
+				FieldSelector: fields.SelectorFromSet(fields.Set{
+					"involvedObject.name": sdc.Name,
+					"reason":              "InvalidRack",
+				}).String(),
+			})
+			eo.Expect(err).NotTo(o.HaveOccurred())
+			eo.Expect(events.Items).NotTo(o.BeEmpty())
+			eo.Expect(events.Items[0].Type).To(o.Equal(corev1.EventTypeWarning))
+			eo.Expect(events.Items[0].Message).To(o.ContainSubstring("can't find data PVC template"))
```
rzetelskik: do we actually want to test this? Our error surface is the degraded condition, the events seem somewhat inconsequential

rzetelskik (commented):
first batch

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
```diff
@@ -0,0 +1,433 @@
...
+		}).WithContext(ctx).WithTimeout(scyllaDBDatacenterControllerDefaultEventuallyTimeout).WithPolling(100 * time.Millisecond).Should(o.Succeed())
+		pod, err := env.TypedKubeClient().CoreV1().Pods(env.Namespace()).Get(ctx, memberPod.Name, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(pod.UID).To(o.Equal(memberPod.UID))
+		o.Expect(pod.DeletionTimestamp).To(o.BeNil())
+
+		g.By("Waiting for the Pod to be evicted and the member Service to record the host ID being replaced")
+		o.Eventually(func(eo o.Gomega, ctx context.Context) {
+			_, err := env.TypedKubeClient().CoreV1().Pods(env.Namespace()).Get(ctx, memberPod.Name, metav1.GetOptions{})
+			eo.Expect(apierrors.IsNotFound(err)).To(o.BeTrue(), "Pod %q should be evicted: %v", memberPod.Name, err)
```
rzetelskik: isn't this racy? It first asserts the pod is there and then waits for it to be removed without any action in between

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
```diff
@@ -0,0 +1,433 @@
...
+			eo.Expect(svc.Annotations).To(o.HaveKeyWithValue(naming.HostIDAnnotation, envtestReplacingNodeHostID))
+
+			events, err := env.TypedKubeClient().CoreV1().Events(env.Namespace()).List(ctx, metav1.ListOptions{
+				FieldSelector: fields.SelectorFromSet(fields.Set{
+					"involvedObject.name": memberServiceName,
+					"reason":              "FinishedReplacingNode",
+				}).String(),
+			})
+			eo.Expect(err).NotTo(o.HaveOccurred())
+			eo.Expect(events.Items).NotTo(o.BeEmpty())
```
rzetelskik: same remark about asserting events - do we want to commit to it being an observability surface? Otherwise this is just a change detector

rzetelskik (commented):
second batch

test/envtest/controllers/scylladbdatacenter_controller_upgrade_test.go
```diff
@@ -0,0 +1,453 @@
...
+		o.Eventually(func(eo o.Gomega, ctx context.Context) {
+			eo.Expect(fakeAPI.SchemaAgreementChecks()).To(o.BeNumerically(">", schemaAgreementChecks))
+
+			upgradeContext = getUpgradeContext(ctx, env, upgradeContextConfigMapName)
+			eo.Expect(upgradeContext).NotTo(o.BeNil())
+			eo.Expect(upgradeContext.State).To(o.Equal(internalapi.RolloutRunUpgradePhase))
+		}).WithContext(ctx).WithTimeout(scyllaDBDatacenterControllerDefaultEventuallyTimeout).WithPolling(100 * time.Millisecond).Should(o.Succeed())
+		for _, host := range nodeHosts {
+			o.Expect(fakeAPI.SnapshotsTaken(host, upgradeContext.SystemSnapshotTag)).To(o.Equal(len(systemKeyspaces)), "host %q", host)
+		}
```
rzetelskik: q: the step description claims that this verifies the snapshots were not taken again, how does this verify it? I think I'm not following what this is asserting

test/envtest/controllers/scylladbdatacenter_controller_config_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		g.By("Pointing the ScyllaDBDatacenter at the override Secret")
+		updateScyllaDBDatacenter(ctx, env, sdc.Name, func(sdc *scyllav1alpha1.ScyllaDBDatacenter) {
+			if sdc.Annotations == nil {
+				sdc.Annotations = map[string]string{}
+			}
+			sdc.Annotations[naming.ScyllaDBManagerAgentAuthTokenOverrideSecretRefAnnotation] = overrideSecret.Name
+		})
+
+		g.By("Waiting for the agent auth token Secret to carry the override token")
```
rzetelskik, line 152: nit: this is only one of the paths, the precedence is: custom config secret > secret from override annotation (used internally with automated mutli-dc) > existing auth token > new generated auth token

test/envtest/controllers/scylladbdatacenter_controller_config_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		g.By("Verifying the generated token is kept across syncs")
+		o.Consistently(func(co o.Gomega, ctx context.Context) {
+			secret, err := env.TypedKubeClient().CoreV1().Secrets(env.Namespace()).Get(ctx, agentAuthTokenSecretName, metav1.GetOptions{})
+			co.Expect(err).NotTo(o.HaveOccurred())
+
+			token, err := helpers.GetAgentAuthTokenFromSecret(secret)
+			co.Expect(err).NotTo(o.HaveOccurred())
+			co.Expect(token).To(o.Equal(generatedToken))
+		}).WithContext(ctx).WithTimeout(scyllaDBDatacenterControllerDefaultConsistentlyTimeout).WithPolling(100 * time.Millisecond).Should(o.Succeed())
```
rzetelskik: if there is no forced sync, does this assert anything really?

rzetelskik (commented):
no further comments, good job adding all of this coverage

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
czeslavo: :+1: Added `naming.ScyllaDBDatacenterSelectorLabels(sdc)` returning exactly what the controller selects on and used it across the board.

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
czeslavo: Done.

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
czeslavo: Dropped the message match, asserting the reason instead.

test/envtest/controllers/scylladbdatacenter_controller_progressing_test.go
czeslavo: Done.

test/envtest/controllers/scylladbdatacenter_controller_progressing_test.go
czeslavo: Fair, we do not really document events as user-facing, so let's not test against them.

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
czeslavo: It's not in the current circumstances due to a hard-coded [sleep](https://github.com/scylladb/scylla-operator/blob/d4961312c39c66ed3aac68dd0d63950cd594b94c/pkg/controller/scylladbdatacenter/sync_services.go#L460) we have.

I'm on the fence if covering this in this test is the right way to go, but in the end, it's better to have this covered, as that's the current expected behavior. We will be able to drop the sleep when we only support k8s versions with cache consistency in the STS controller, but until then, it's expected that we won't remove the PVC and the Pod right after. 

Added comments explaining that and changed `Get` to `Consistently` (for 2s after the PVC deletion) to ensure we do not delete the Pod right away.

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
czeslavo: Dropped.

test/envtest/controllers/scylladbdatacenter_controller_config_test.go
czeslavo: Good catch. Replaced it with a spec change that bumps the generation (an extra ScyllaDB argument), a wait for the observed generation to catch up, and then the token comparison.

test/envtest/controllers/scylladbdatacenter_controller_config_test.go
czeslavo: Extended the spec: after the override token lands, it creates a custom agent config Secret with its own token and references it from `rackTemplate.scyllaDBManagerAgent.customConfigSecretRef` while the override annotation stays, and expects the custom config token to win.

So the spec now walks all four levels in order: generated, kept, override, custom config.

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
rzetelskik: If it's just ErrorReason we have no guarantee that it's not caused by something else. Improving this is probably a bigger effort though, so I'll leave it to your decision what to do with it, I'm not expecting an improvement in this PR

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
rzetelskik: Even with the sleep it's still technically racy, i.e. we have no guarantee that the write will be picked up. Also, I don't see the test overriding `scyllaDBDatacenterControllerDisabledStatefulSetCachePropagationDelay`, which is set to 0 seconds. I wouldn't rely on this behaviour (the sleep) in the test either way.

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
czeslavo: Yeah, probably we could change the reason so it's narrow enough to pin this specific case, but I'd rather avoid doing this right now. How about asserting both the reason and the message, but asserting only the message includes the object name?

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
czeslavo: It's not the sleep we control with `scyllaDBDatacenterControllerDisabledStatefulSetCachePropagationDelay`. https://github.com/scylladb/scylla-operator/blob/d4961312c39c66ed3aac68dd0d63950cd594b94c/pkg/controller/scylladbdatacenter/sync_services.go#L460 is hardcoded to 10s between deleting a PVC and evicting a Pod.

What this test tries to cover is the fact that the Pod is not deleted right after the PVC (at least for 2s long `nodeReplacementPodRetentionWindow` that's used to check that the Pod stays consistently not deleted).

I think it is a behavior worth covering; without that, we can remove the sleep and have no regression detected. 

An alternative could be watching events for the Pod and PVC and asserting that their deletion is at least ' nodeReplacementPodRetentionWindow ' apart, but it would be effectively the same check.

rzetelskik: My bad, I didn't understand what sleep this was referring to. I maintain my position that the test is racy by design, but let's reassess when we try to get rid of the sleep.

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
rzetelskik: I'm fine with adding the message assertion but it can also stay as is imo

rzetelskik (approved):
/lgtm
/hold optional, for https://github.com/scylladb/scylla-operator/pull/3654#discussion_r4027444968, feel free to cancel

test/envtest/controllers/scylladbdatacenter_controller_ownership_test.go
czeslavo: Added already :+1:

test/envtest/controllers/scylladbdatacenter_controller_nodes_test.go
czeslavo: Created a tracker for discovery work to assess whether the sleep is still needed: https://scylladb.atlassian.net/browse/OPERATOR-429

PR #3656, opened 2026-09-10
ci: publish multiversion docs from GitHub Actions

rzetelskik:
Ports the post-scylla-operator-multiversion-docs Prow postsubmit. GHA resolves workflow files from the ref of the push event, so master's copy never runs for a push to a release branch: docs-pages-trigger.yaml is a stub that delegates back to master via workflow_call.

**Notes:** 
- this will require switching the pages source to GHA.
- backporting to the supported release branches is required. I believe we can skip the unsupported ones.

**To be considered by the reviewer**: currently the workflow continues using our quay.io/scylladb/scylla-operator-images:uv-0.10 image. We should make an informed decision if we keep using it or switch to setting up python and uv as part of the workflow (see sphinx-scylladb-theme repo: https://github.com/scylladb/sphinx-scylladb-theme/blob/03292f61ed74e10f75334e1d1b0078471e606686/.github/workflows/docs-pages.yaml#L25-L31). In case of a switch, assuming we'd stop building the image, we'd also need to update development instructions.

Related to https://scylladb.atlassian.net/browse/OPERATOR-395

### DIFF
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -1,0 +1,98 @@
+name: "Docs / Publish"
+
+# The push trigger is expected to stay limited to master. Release branches don't belong there:
+# GitHub resolves workflow files from the ref of the push event, so this file's copy on master never
+# runs for a push to a release branch. Those are picked up by docs-pages-trigger.yaml,
+# which calls back into this workflow.
+on:
+  push:
+    branches:
+      - master
+    # The docs build reads more than docs/: literalinclude pulls from examples/ (e.g.
+    # docs/source/deploy-scylladb/set-up-monitoring/setup.md), and the version_context_substitutions
+    # extension reads assets/config/config.yaml, assets/metadata/metadata.yaml and .gitmodules.
+    paths:
+      - "docs/**"
+      - "examples/**"
+      - "assets/**"
+      - ".gitmodules"
+  workflow_call:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    name: Build multiversion docs
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/scylladb/scylla-operator-images:uv-0.10
+    timeout-minutes: 60
+
+    steps:
+      # actions/checkout clones as the runner user, but this container image declares no USER, so the job runs as root
+      # and every git invocation outside the action trips safe.directory. sphinx-multiversion shells out to git to
+      # enumerate remote branches, so this is needed for the build itself, not just for the verification steps.
+      - name: Mark workspace safe for git
+        run: git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: master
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-pages.yaml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          ref: master
+          persist-credentials: false
+          fetch-depth: 0
+
+      # A release branch's conf.py pins an older BRANCHES and smv_latest_version, so building its tree would publish a regressed site.
+      - name: Verify checkout is master
+        run: |
+          [[ "$( git rev-parse --abbrev-ref HEAD )" == 'master' ]]
+          [[ "$( git rev-parse HEAD )" == "$( git rev-parse refs/remotes/origin/master )" ]]
```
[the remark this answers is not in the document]
rzetelskik: a no-op with our uv image, but added as insurance against image change

.github/workflows/docs-pages.yaml
```diff
@@ -0,0 +1,78 @@
+name: "Docs / Publish"
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "docs/**"
```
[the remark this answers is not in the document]
rzetelskik: good point, added

czeslavo:
> To be considered by the reviewer: currently the workflow continues using our quay.io/scylladb/scylla-operator-images:uv-0.10 image. We should make an informed decision if we keep using it or switch to setting up python and uv as part of the workflow (see sphinx-scylladb-theme repo: https://github.com/scylladb/sphinx-scylladb-theme/blob/03292f61ed74e10f75334e1d1b0078471e606686/.github/workflows/docs-pages.yaml#L25-L31). In case of a switch, assuming we'd stop building the image, we'd also need to update development instructions.

I think that it will be better to keep using the image for now. In this state, we at least have a way for reproduction locally using our image.

I can see it as a separate effort that we could make: start using something like https://mise.jdx.dev/ for dev dependency management/pinning so that we can install deps in CI using the same toolset (with https://github.com/jdx/mise-action) that we do locally.

.github/workflows/docs-pages.yaml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          ref: master
+          persist-credentials: false
+          fetch-depth: 0
+
+      # A release branch's conf.py pins an older BRANCHES and smv_latest_version, so building its tree would publish a regressed site.
+      - name: Verify checkout is master
+        run: |
+          [[ "$( git rev-parse --abbrev-ref HEAD )" == 'master' ]]
+          [[ "$( git rev-parse HEAD )" == "$( git rev-parse refs/remotes/origin/master )" ]]
```
czeslavo: nit: I think it should be enough to leave a comment on the `on.push.branches` to explain that this should never be changed to any other branch.

.github/workflows/docs-pages.yaml
```diff
@@ -0,0 +1,88 @@
+name: "Docs / Publish"
+
+on:
+  push:
+    branches:
+      - master
```
czeslavo, line 10: nit: we could add an explicit comment that this is expected to always stay on master.

czeslavo (commented):
Just one nit.

.github/workflows/docs-pages.yaml
rzetelskik: done

rzetelskik:
/hold cancel
https://github.com/rzetelskik/scylla-operator/actions/runs/34574954904/job/103185192808 (after a fix)

PR #3657, opened 2026-09-10
[v1.21] ci: trigger multiversion docs publishing from release branches

rzetelskik:
Backport of #3656.

Only b666f3946c19c70690acdf97485492ba2ae65658 is being backported. The `docs-pages.yaml` workflow it delegates to is only required on master: GitHub resolves workflow files from the ref of the push event, so master's copy never runs for a push to a release branch, which is why the branch-resident stub exists in the first place. The build itself always checks out master and rebuilds every version listed in master's `docs/source/conf.py`, so a release branch has no use for its own copy of the implementation.

Related to https://scylladb.atlassian.net/browse/OPERATOR-395.

/cc @czeslavo


czeslavo (approved):
/lgtm

PR #3659, opened 2026-09-11
docs(enhancements): compare the read-your-writes PoCs for OPERATOR-392

czeslavo:

Adds `enhancements/proposals/jira-392-read-your-writes/README.md`, the outcome of the two discovery tickets under OPERATOR-392. It describes the two proofs of concept for giving the operator's controllers read-your-writes semantics, an in-house consistency layer over client-go informers (#3650) and a migration to controller-runtime (#3652), compares their properties, lays out the evidence gathered for each, and ends with the three options the team has to choose between.

Part of https://scylladb.atlassian.net/browse/OPERATOR-392
Resolves https://scylladb.atlassian.net/browse/OPERATOR-393
Resolves https://scylladb.atlassian.net/browse/OPERATOR-397


enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,771 @@
...
+| scyllacluster | 631 | 136 | 1 |
+| nodetune | 539 | 249 | 0 |
+| nodeconfig | 527 | 143 | 1 |
+| scylladbmanagerclusterregistration | 440 | 216 | 4 |
+| globalscylladbmanager | 385 | 77 | 0 |
+
+Two grow slightly, statusreport (160 to 194) and ignition (248 to 259), because their cache options and trigger wiring
+are more lines than the two informers they replaced.
+
+### 4.4 The multi-DC question
```
mflendrich, line 430: Let's simplify by assuming that ScyllaDBCluster & RKC will be deleted. There is no path forward for them to graduate to GA.

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,771 @@
...
+
+Moving onto someone else's runtime means its defaults arrive with it, and several are not ours. The PoC set or restored
+the following deliberately, so that swapping the binary changes nothing an operator can observe:
+
+- **Leader election stays ours.** The manager's own election is off and `mgr.Start` runs inside the existing
+  `leaderelection.Run`, so the lease keeps its name (`scylla-operator-lock`) and identity format, and standby replicas
+  still do not start caches
+  ([`manager.go:97`](https://github.com/czeslavo/scylla-operator/blob/f334d7813/pkg/controllermanager/manager.go#L97)).
+- **No new listeners.** The manager would serve metrics and health probes by default; both are bound to `"0"`, and
+  metrics are opt-in behind a new `--metrics-bind-address` flag.
```
mflendrich, line 474: this is expressed backwards in writing but what it really means is that controller-runtime makes metrics native (and it can be a good enabler for business logic metric like in progress node replace duration, and similar pageable events).

that would be great to have

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,771 @@
...
+- **One logger.** controller-runtime's cache and reflectors log through a package-level logger, not the manager's.
```
mflendrich, line 475: can you explain in human-speak what happens to logging? will the logs still be forwarded to our existing klog, or something different will happen?

I'd like to a void a situation where we have 2 separate root loggers with their streams interleaved - I assume that's not it?

czeslavo: Yes, we keep using the same klog package-level logger. We could migrate to a logger passed down the call stack in the future for better logger key/context propagation, but it's not a must, so I avoided this in the PoC.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Yes, in the PoC, I intentionally didn't want to enable the metrics as that's worth a separate feature announcement (perhaps, adding some custom metrics on top of that, allowing setting alarms to detect potential misbehavior of the operator).

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,771 @@
...
+| Who owns the concurrency-critical code | us: 1030 lines of watermarks, pending deletes, locking and wake-ups | upstream, marked experimental |
+| Code we maintain (gross added / net) | 1151 / **+999** | milestone 1: 1614 / **+186**; whole operator: 4026 / **−3940** |
+| What the PoC delivers | the ScyllaDBDatacenter controller | milestone 1: the same one controller; milestone 2: 18 controllers across five binaries |
+| Work to give another controller the guarantee | a consistency store, a registration table, client wrapping and a `WaitReady`, plus new recording clients for about 15 kinds and the monitoring client; nothing exists for remote clusters | none: the client every reconciler is handed is already consistent |
+| Effect on the rest of the codebase | none; the hand-written controller machinery stays as it is | workqueues, rate limiting, watch plumbing, metrics and the wiring idiom become upstream's; so do our log-level and error-reporting conventions |
+| Boilerplate | unchanged | 7966 lines of controller boilerplate removed across the operator; the ScyllaDBDatacenter controller's 39 event handlers become one `SetupWithManager` |
```
rzetelskik, line 526: "unchanged" for client-go is not the whole truth imo - we could simply reduce the boilerplate by abstracting away a generic controller implementation instead of repeating it over and over again

mflendrich: I guess controller runtime is exactly that, but the controller is maintained upstream for us?

rzetelskik: Ok? But reducing the boilerplate is possible with both approaches.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: As @rzetelskik pointed out, we may agree that the problem is simplified if we drop ScyllaDBCluster and RKC now, but we should still have in mind that automated multi-DC may come back at us at some point, and it may be worth considering whether controller-runtime will be able to handle that for us.

Kacper mentioned an alternative approach would be to run multi-DC on federated clusters, so we may look into that to see if controller-runtime would make that harder due to its opinionated approach.

mflendrich: At any rate, I have very low confidence in the RKC approach to multi-DC.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: To make it clear, the comparison here is about the specific PoCs that are completed on respective branches. 

And yeah, your point about the potential for reducing boilerplate is valid; it hasn't been assessed in this initiative. My feeling is that reducing it would be very similar to coming up with abstractions that controller-runtime already provides. We could try to pursue this path to see how far we can go, if you think that would be worth it to have a complete picture.

enhancements/proposals/jira-392-read-your-writes/README.md
rzetelskik: Yes, I don't think the RKC approach is what we'd want. But the point is not to completely close the path to managed multi-DC with the chosen approach. If any other approach also required watches for other clusters, based on this document we'd have an issue with CR

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,771 @@
...
+deleted. Of the 1353, about 300 is per-kind lister glue that grows by 30 lines for each further kind a lister-taking
+function needs; the rest is fixed. Newly upstream's, and no longer ours: the workqueue and its rate limiting, the watch
+and event plumbing, and the consistency implementation itself.
+
+## 5. Comparison
+
+| | A: the client-go layer | B: controller-runtime |
+|---|---|---|
+| Where the guarantee is enforced | one barrier at the top of each sync, covering every registered kind and every write still in flight | at each read: a `Get` waits for that one object, a `List` waits for everything this client wrote of that kind, and write barriers stop a read slipping between a write and its recording |
+| Granularity, and what it couples | global per controller: a write in flight for one ScyllaDBDatacenter delays the sync of every other one, on every worker | per key and per kind; unrelated objects do not wait for each other |
```
rzetelskik: I don't think this is correct for List.

Get is per key, but List waits for every pending write of that kind through the client, and at least in the PoC there's one client for the whole binary. So e.g. when the SDC sync lists Pods it waits for Pod writes from node config, other SDC workers etc. So it's per kind across the whole process and happens at each of the lister reads in the sync.

So iiuc A couples everything one controller wrote, once per sync, but B couples everything every controller wrote of one kind, per list call.

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,771 @@
...
+left as they are: cheap controllers take 2 to 5 times longer per reconcile (still hundreds of microseconds; heavy
+controllers unchanged), and the single-key global-manager controller retries faster during namespace termination, so it
+prints roughly twice the error lines for the same underlying condition. Otherwise: no panics, no cache-sync timeouts, no
+throttling, no leader-election anomalies, startup to all workers under 0.4 s, and a total error rate slightly *below*
+master. Conflict lines dropped from 1003 to 15, which is the read-your-writes client doing its job.
+
+Not measured for the client-go PoC. Reasoning: it does not change the controller machinery, the workqueues or the log
+sites outside the ScyllaDBDatacenter controller, so a comparable log delta would be small by construction.
+
+### 6.4 CPU and memory
```
rzetelskik, line 673: I believe that API request rate would be a more interesting metric, as this is what the CR to client-go switch was backed by: https://github.com/scylladb/scylla-operator/pull/534

enhancements/proposals/jira-392-read-your-writes/README.md
mflendrich: > If any other approach also required watches for other clusters, based on this document we'd have an issue with CR

To me "watches for other clusters" is synonymous with RKC - if there's important nuance let's sort it out.

The approach in my mind for a future hypothetical fully managed multi-DC would be using our business logic talking over some kind of high-level trans-cluster RPC. This way watches remain a local thing.

enhancements/proposals/jira-392-read-your-writes/README.md
rzetelskik: I don't think it's worth the time investment for now, but let's just keep the comparison honest - surely it's possible to reduce the boiler plate and yes, we'd just be building our own framework. But this is what the comparison comes down to at the end of the day - build our own or use the upstream, both coming with their pros and cos. Let's focus on understanding the pros and cons on the assumption we have a complete framework in both cases and choose accordingly.

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+### 7.1 Approach A
+
+1. **We own a concurrency-critical component with no upstream.** 1030 lines whose correctness rests on
+   resource-version monotonicity, informer event ordering and lock discipline. It is tested, including against a real
+   API server, but every future change to it is ours to get right.
+2. **Correctness depends on discipline, not on the type system.** Every write must go through a recording client, every
+   kind read from a cache must be registered, and every registered informer must watch every object written through
+   that client, in namespace and selector terms. All three hold today; `Register` checks none of them, and the side
+   binaries with narrow informers are exactly where the third would bite.
```
rzetelskik: This third condition holds for B as well, it's just enforced by the cache selector instead of a Register table. The consistent client records a watermark on every write without checking whether the cache's selector covers the object, and the observed RV only moves on events for objects the cache does watch. So a sidecar binary on a filtered cache (e.g. sidecar's Service cache) that writes an object outside its filter makes the next read of that kind block until a watched object happens to change. It's not the case today, but "no registration table to forget" isn't true.

So this could cause "a cache that never catches up" mentioned in 7.2 item 3 "Consistent reads have no deadline of their own". It seem like without a timeout, B has a worse failure mode.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Simplified the doc assuming we drop automated multi-DC for now; made it clear that controller-runtime doesn't block us from reconsidering it in the future in any form.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Reworded to make it clear that it would be possible to extract common parts of controllers in the client-go approach, but it would be a significant amount of work, and it would be reinventing abstractions that controller-runtime provides.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Added 6.6 covering API calls based on measurements made on the cluster topology e2e suite on kind with audit logging enabled.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Yes, you're right. Updated the doc to make it surfaced for B. A mitigation we could apply in this case would be to set the reconciliation timeout as originally suggested and limit the use of the read-your-writes client only where the cache is unfiltered (operator binary), while leaving the side binaries on a plain cache client.

rzetelskik: Sounds good

czeslavo:
Updated the doc addressing all the comments. PTAL

rzetelskik:
I'm convinced CR is the better path forward. Let's update the doc with the decision (option 1 with opportunistic milestone 2 in this sprint or a part of observability epic) and we should be good to go :rocket:

czeslavo:
> I'm convinced CR is the better path forward. Let's update the doc with the decision (option 1 with opportunistic milestone 2 in this sprint or a part of observability epic) and we should be good to go 🚀

Recorded the decision. Thank you guys for the quick review.

rzetelskik (approved):
lgtm
/assign @mflendrich

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,872 @@
...
+multi-DC controllers are removed, not migrated (4.4). At the end `Reconcile` is the only entry point and there is no
+error classification of our own (`controllertools.NonRetriable` becomes `reconcile.TerminalError`).
+
+Milestone 2 adds three pieces of shared glue and deletes one.
+
+**The bridge is deleted** (−80), and `controllers.go` drops from 546 to 184 lines: with every controller a reconciler,
+registration is a constructor call and a `SetupWithManager` each.
+
+**Reconciler helpers**
+([`pkg/controllertools/reconciler.go`](https://github.com/czeslavo/scylla-operator/blob/f334d7813/pkg/controllertools/reconciler.go),
```
alvaroaleman, line 364: It does have that: https://github.com/kubernetes-sigs/controller-runtime/blob/6ab2188a1fb14a8c4837b418bc0a20049682cde4/examples/typed/main.go#L52

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,872 @@
...
+
+Two examples are in 4.3, for the bootstrap barrier and node tune. Filtering itself is not where controller-runtime is
+rigid.
+
+**One informer per kind per cache.** This is the real constraint. A cache holds one informer per kind and
+representation, so two different selectors on the same kind in one process cannot both exist. The choice is to widen
+the selector and filter client-side, or to run a second cache. The PoC hit it twice:
+
+- Node setup had two Pod factories with disjoint selectors, ScyllaDB Pods on the node across all namespaces
+  ([`nodesetupdaemon.go:197`](https://github.com/scylladb/scylla-operator/blob/5d041a933/pkg/cmd/operator/nodesetupdaemon.go#L197)) and the daemon's own Pod by namespace and
```
alvaroaleman, line 495: Did you try setting both field and label selector for the same type and did it not work? It definitely should

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,872 @@
...
+  name ([`:203`](https://github.com/scylladb/scylla-operator/blob/5d041a933/pkg/cmd/operator/nodesetupdaemon.go#L203)). They became one informer on `spec.nodeName` alone
```
alvaroaleman, line 496: This should work like this:
```
       ByObject: map[client.Object]cache.ByObject{
           &corev1.Pod{}: {
               Namespaces: map[string]cache.Config{
                   o.Namespace: {FieldSelector: fields.OneTermEqualSelector("metadata.name", o.PodName)},
               },
           },
       },
```

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,872 @@
...
+multi-datacenter cluster is the manual one, ScyllaClusters with `externalSeeds`, which does not involve them.
+
+Removing them takes out about 4300 lines of controller code, `pkg/remoteclient` (595 lines), the ScyllaDBCluster,
+RemoteKubernetesCluster and RemoteOwner CRDs, and their e2e specs. For the migration it means 16 controllers instead of
+18, no per-remote-cluster caches, and 541 fewer lines of glue (4.3).
+
+For the record, the PoC did port them, through one `cluster.Cluster` per RemoteKubernetesCluster with watches attached
+as clusters appear, and the multi-DC e2e jobs passed. It was the least comfortable part of the work: controller-runtime
+cannot remove a watch once attached, so a cluster replaced while a watch was still syncing keeps its stopped cache
+alive; a cluster is reported ready before its cache has synced; the upstream answer, `sigs.k8s.io/multicluster-runtime`,
```
alvaroaleman, line 433: > a cluster is reported ready before its cache has synced

Could you expand on that? FWIW, its possible to stop watches through `cache.RemoveInformer`, but yeah, the cache itself will not be gced

enhancements/proposals/jira-392-read-your-writes/README.md
```diff
@@ -0,0 +1,872 @@
...
+Section 8 spells out what each option commits us to; section 9 lists the decisions.
+
+## 3. Approach A: the client-go layer
+
+PoC: [PR #3650](https://github.com/scylladb/scylla-operator/pull/3650), seven commits. No dependency bump, no vendored
+change.
+
+### 3.1 Implementation
+
+A new package, [`pkg/cacheconsistency`](https://github.com/czeslavo/scylla-operator/blob/02390041f/pkg/cacheconsistency),
```
alvaroaleman, line 73: You may also find this interesting: https://kubernetes.io/blog/2026/04/28/kubernetes-v1-36-staleness-mitigation-for-controllers/

It is a bit of a different approach that requires business logic changes rather than doing it in the client, but its more granular

alvaroaleman (commented):
Don't mind me, I was just curious - Left a few random comments in case you are interested :)

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Missed that; thanks for pointing it out. :bow: Updated the doc with a recommendation to use the typed-request form.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Both on one informer would be an AND, and this case is a union of two disjoint sets: ScyllaDB Pods on the node in any namespace, and the daemon's own Pod, which carries none of the ScyllaDB labels. So a single field+label pair can't express it. Your per-namespace suggestion below does.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Our remote cluster set publishes a cluster right after `go c.Start(ctx)` without waiting for the cache to sync, so the first reconciles against it either get `ErrCacheNotStarted` or block inside the read until sync. It's not really a CR issue per se, and is fixable on our side by waiting on the attached sources' `WaitForSync` before publishing. I've rewritten the paragraph to say so.

On `RemoveInformer` - agreed it exists, but here the per-cluster cache is already stopped as a whole with its context. What survives is the `source.Kind` start goroutine, which `controller.Watch` runs under the controller's context and which waits for the cache and then its handler to sync. Neither returns on a stopped cache, so it holds the cache until the controller stops. A source wrapper that derives its context from the cluster's is a feasible fix.

Either way, this is no longer that relevant on our side as we've decided to drop those experimental multi-DC controllers.

enhancements/proposals/jira-392-read-your-writes/README.md
czeslavo: Yeah, this PoC was inspired by this writeup. Made that clear in the doc.

mflendrich (approved):
/lgtm

PR #3670, opened 2026-09-16
feat: run the operator's controllers on one controller-runtime cache

czeslavo:

Runs every controller of the operator binary on one controller-runtime manager and cache, without changing what any of them does. This is the plumbing for moving the ScyllaDBDatacenter controller onto a read-your-writes client, so that it decides from a cache that has observed its own writes.

- **`pkg/controllermanager`** owns the manager: the operator's scheme, the existing resync period as the cache sync period, a read-your-writes client, a one-minute default reconciliation timeout for the controllers that will be registered through controller-runtime, controller-runtime's package logger routed to klog, and neither metrics nor health probes served. Leader election stays in `pkg/leaderelection` around the manager, so the lease and the standby behavior are unchanged. The controller wiring moved here from `pkg/cmd/operator/operator.go`; the multi-datacenter part sits in its own file. The manager also decides whether the ScyllaDBMonitoring controller runs, from the presence of the Prometheus Operator CRDs.
- **The `InformerFor` bridge** hands the legacy controllers their client-go informers and listers from the manager's cache, so a controller on controller-runtime and one that is not share the same informer for a kind: one informer per kind in the process.
- **`pkg/ctrlclient`** adapts a controller-runtime client to the function-shaped interfaces `resourceapply` and `controllerhelpers` take (apply, prune and claim controls, object control), provides client-backed Pod and Secret listers, and converts metav1 options field by field, since controller-runtime's option structs overwrite `Raw`.
- **`kubecrypto.CertificateManager`** takes an `ObjectControl` per kind, with the typed-client construction kept for its current callers.
- **controller-runtime v0.25.1**, which fixes the resource version parse error of a subresource create through the read-your-writes client that we hit in the PoC.

Multi-DC controllers are moving away with https://github.com/scylladb/scylla-operator/pull/3671, so depending on the merge order, `pkg/controllermanager/multidatacenter.go` might be dropped if #3671 is merged first.

Resolves https://scylladb.atlassian.net/browse/OPERATOR-417


pkg/controllermanager/controllers.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	m.addRunnable(smtc.Run, o.ConcurrentSyncs)
+
+	return nil
+}
+
+func (m *Manager) addRunnable(run func(ctx context.Context, workers int), workers int) {
+	m.runnables = append(m.runnables, func(ctx context.Context) {
+		run(ctx, workers)
+	})
```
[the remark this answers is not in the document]
czeslavo: It's so by design. Old controllers keep their own timeouts to not change the behavior. The timeout we set is going to affect only controllers that will be gradually migrated to CR reconcilers.

pkg/ctrlclient/ctrlclient_test.go
```diff
@@ -0,0 +1,119 @@
...
+		}).
+		Build()
+
+	ctx := t.Context()
+	control := ApplyControl[appsv1.StatefulSet](ctx, c, "ns")
+
+	deleteOpts := metav1.DeleteOptions{
+		GracePeriodSeconds: ptr.To[int64](5),
+		Preconditions: &metav1.Preconditions{
+			UID:             ptr.To(types.UID("uid")),
```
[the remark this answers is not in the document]
czeslavo: Covered in UT now.

pkg/controllermanager/controllers.go
```diff
@@ -0,0 +1,283 @@
...
+	scyllaDBClusters := informerFor(ctx, c, errs, &scyllav1alpha1.ScyllaDBCluster{}, scyllav1alpha1listers.NewScyllaDBClusterLister)
+	scyllaDBMonitorings := informerFor(ctx, c, errs, &scyllav1alpha1.ScyllaDBMonitoring{}, scyllav1alpha1listers.NewScyllaDBMonitoringLister)
+	scyllaDBManagerClusterRegistrations := informerFor(ctx, c, errs, &scyllav1alpha1.ScyllaDBManagerClusterRegistration{}, scyllav1alpha1listers.NewScyllaDBManagerClusterRegistrationLister)
+	scyllaDBManagerTasks := informerFor(ctx, c, errs, &scyllav1alpha1.ScyllaDBManagerTask{}, scyllav1alpha1listers.NewScyllaDBManagerTaskLister)
+	nodeConfigs := informerFor(ctx, c, errs, &scyllav1alpha1.NodeConfig{}, scyllav1alpha1listers.NewNodeConfigLister)
+	// ScyllaOperatorConfig is a singleton, so the name-filtered informer the operator used to keep next to the
+	// unfiltered one is not needed with a single cache.
+	scyllaOperatorConfigs := informerFor(ctx, c, errs, &scyllav1alpha1.ScyllaOperatorConfig{}, scyllav1alpha1listers.NewScyllaOperatorConfigLister)
```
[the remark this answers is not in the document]
czeslavo: Nothing critical, as people would rather not create other ScyllaOperatorConfigs, but added a cache filter that will make us watch only the singleton we're concerned about.

pkg/controllermanager/manager.go
```diff
@@ -0,0 +1,163 @@
...
+	// route it to the same place, once per process, so those logs aren't dropped.
+	setControllerRuntimeLogger.Do(func() {
+		ctrllog.SetLogger(options.Logger)
+	})
+
+	mgr, err := ctrlmanager.New(options.RestConfig, ctrlmanager.Options{
+		Scheme: scheme.Scheme,
+		Logger: options.Logger,
+		Cache: cache.Options{
+			SyncPeriod: new(options.ResyncPeriod),
```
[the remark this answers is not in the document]
czeslavo: It's no longer true since Go 1.26.

pkg/cmd/operator/operator.go
```diff
@@ -730,133 +284,7 @@ func (o *OperatorOptions) run(ctx context.Context, streams genericclioptions.IOS
...
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		smtc.Run(ctx, o.ConcurrentSyncs)
-	}()
-
-	<-ctx.Done()
-
-	return nil
+	return cm.Run(ctx)
```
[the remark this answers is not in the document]
czeslavo: Adds context cancellation that will run before wait.

pkg/ctrlclient/controls_test.go
```diff
@@ -0,0 +1,224 @@
...
+				Name:            name,
+				OwnerReferences: ownerRefs,
+			},
+		}
+		if matching {
+			svc.Labels = selectorLabels
+		}
+		return svc
+	}
+	c := fake.NewClientBuilder().WithScheme(scheme.Scheme).WithObjects(
```
rzetelskik, line 62: might be more robust to not rely on scheme package but to construct a test scheme instead with runtime.NewScheme()

pkg/ctrlclient/controls_test.go
```diff
@@ -0,0 +1,224 @@
+// Copyright (c) 2026 ScyllaDB.
```
rzetelskik, line 1: nit: shouldn't this file be merged with ctrlclient_test.go?

pkg/ctrlclient/controls_test.go
```diff
@@ -0,0 +1,224 @@
...
+	applied, changed, err := resourceapply.ApplyConfigMapWithControl(ctx, control, recorder, required, resourceapply.ApplyOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !changed || applied.Data["key"] != "v2" {
+		t.Errorf("expected the apply of changed data to update the ConfigMap, changed=%v data=%v", changed, applied.Data)
+	}
+}
+
+func TestListersReadThroughTheClient(t *testing.T) {
```
rzetelskik, line 188: nit: it feels like this should be a table test with expected as test case input instead of hardcoded

pkg/ctrlclient/controls_test.go
```diff
@@ -0,0 +1,224 @@
...
+	required := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:       testNamespace,
+			Name:            "cm",
+			OwnerReferences: []metav1.OwnerReference{newControllerRef(controller)},
+		},
+		Data: map[string]string{"key": "v1"},
+	}
+
+	_, changed, err := resourceapply.ApplyConfigMapWithControl(ctx, control, recorder, required, resourceapply.ApplyOptions{})
```
rzetelskik, line 162: isn't this just testing the resourceapply implementation? It's not verifying what functions were invoked?

pkg/ctrlclient/ctrlclient_test.go
```diff
@@ -0,0 +1,120 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+)
+
+// The metav1 options given to the typed-client-shaped functions have to reach the API server whole: a StatefulSet
+// recreated with a lost Orphan propagation policy takes its Pods down with it.
+func TestOptionsReachTheClient(t *testing.T) {
```
rzetelskik, line 21: nit: seems like the particular funcs could be split into separate tests

pkg/kubecrypto/certmanager_test.go
```diff
@@ -190,3 +196,227 @@ func Test_ManageCertificates_staleCacheDoesNotRegenerateCA(t *testing.T) {
...
+			}
+
+			objects.syncCaches()
+			manage()
+			second := snapshot()
+			if writes := objects.writes(); writes != len(first) {
+				t.Errorf("expected the second run to issue no writes, got %d in total", writes)
+			}
+			if diff := cmp.Diff(first, second); diff != "" {
+				t.Errorf("expected the second run to leave the objects alone (-first +second):\n%s", diff)
```
rzetelskik, line 418: nit: leave the objects unchanged? 😅

pkg/cmd/operator/operator.go
```diff
@@ -279,584 +256,40 @@ func (o *OperatorOptions) run(ctx context.Context, streams genericclioptions.IOS
...
-		smcrc.Run(ctx, o.ConcurrentSyncs)
-	}()
+	// The key generator has to stop with the controller manager, also when the manager fails to start; the deferred
+	// cancel runs before the deferred wait.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
```
rzetelskik, line 288: nit: can be simplified with wg.Go

test/envtest/controllers/controllermanager_test.go
```diff
@@ -0,0 +1,137 @@
...
+		defer wg.Done()
+		defer g.GinkgoRecover()
+		err := cm.Run(ctx)
+		o.Expect(err).NotTo(o.HaveOccurred())
+	}()
+
+	g.DeferCleanup(func() {
+		cancel()
+		wg.Wait()
+	})
```
rzetelskik, line 136: not 100% sure about this, but shouldn't it be set up before the goroutine is started? Otherwise it may not be registered if the assertion fails in the gorutine?

rzetelskik (commented):
Solid work! Left a couple of comments, mostly nits.

pkg/kubecrypto/certmanager_test.go
```diff
@@ -190,3 +196,227 @@ func Test_ManageCertificates_staleCacheDoesNotRegenerateCA(t *testing.T) {
...
+				return res
+			}
+
+			manage()
+			first := snapshot()
+			if writes := objects.writes(); writes != len(first) {
+				t.Errorf("expected the first run to issue %d writes, got %d", len(first), writes)
+			}
+
+			objects.syncCaches()
```
rzetelskik, line 411: should we add another run before syncing the cache to verify that it makes a live call and doesn't attempt to create the object again?

rzetelskik: This is what I mean more or less: https://github.com/rzetelskik/scylla-operator/commit/5b5d71692fae9e591169a173c73e507cbc185401.

pkg/ctrlclient/controls_test.go
czeslavo: Done.

pkg/ctrlclient/ctrlclient_test.go
czeslavo: Done, split into separate tests.

pkg/kubecrypto/certmanager_test.go
czeslavo: :smile: Fixed.

pkg/cmd/operator/operator.go
czeslavo: Done.

test/envtest/controllers/controllermanager_test.go
czeslavo: Reordered.

pkg/kubecrypto/certmanager_test.go
czeslavo: Added this for the typed clients and listers' case as the CR no longer distinguishes live/cached calls - it uses one consistent client (enforced with `ReadYourWritesClient` wrapper type) for all calls.

pkg/ctrlclient/controls_test.go
czeslavo: :+1: Tests now build a scheme holding only the kinds used in them. 

Also noticed that `newList` used a global scheme instead of relying on the client's one - fixed this.

pkg/ctrlclient/objectcontrol.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (oc ObjectControl[T, PT]) GetCached(namespace, name string) (PT, error) {
+	return Get[T, PT](oc.ctx, oc.c, namespace, name)
+}
+
+// Get reads through the client under ctx. With a read-your-writes client that is as fresh as a live read for
+// everything this controller wrote.
+func (oc ObjectControl[T, PT]) Get(ctx context.Context, namespace, name string) (PT, error) {
+	return Get[T, PT](ctx, oc.c, namespace, name)
```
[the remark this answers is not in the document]
czeslavo: With a read-your-writes client, the cache never lags behind the controller's own writes, so `Get` there is the same read as `GetCached`, kept to satisfy the kubecrypto interface. `NewObjectControl` now takes a `ReadYourWritesClient` wrapper type, built in `controllermanager.New`, right after the cache option that makes the client consistent, so a plain cached client can't be passed by accident.

pkg/ctrlclient/controls_test.go
czeslavo: Turned into a test case per scenario - missing, unchanged, updated.

pkg/ctrlclient/controls_test.go
czeslavo: The intention was to run the control through its consumer, as a controller would do, and assert side effects, but I guess that having specific calls asserted won't hurt us. :+1: Updated.

pkg/ctrlclient/ctrlclient_test.go
```diff
@@ -0,0 +1,404 @@
...
+	tt := []struct {
+		name            string
+		existing        *corev1.ConfigMap
+		expectedCalls   []string
+		expectedChanged bool
+	}{
+		{
+			name:            "creates the ConfigMap when it is missing",
+			existing:        nil,
+			expectedCalls:   []string{"Get", "Create"},
```
rzetelskik: nit: we could use some test-only enum for the calls

test/envtest/controllers/readyourwrites_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+type heldBackBody struct {
+	io.ReadCloser
+	delay time.Duration
+}
+
+func (b heldBackBody) Read(p []byte) (int, error) {
+	n, err := b.ReadCloser.Read(p)
+	time.Sleep(b.delay)
```
rzetelskik: why don't we use a synchronisation mechanism instead of a racy sleep?

rzetelskik (commented):
one comment

test/envtest/controllers/readyourwrites_test.go
czeslavo: Fair point, replaced with a gate that we can `Hold` and `Release` from the test body.

rzetelskik (approved):
/lgtm

PR #3672, opened 2026-09-16
ci: run the govulncheck periodic on GHA

rzetelskik:
Move the govulncheck periodic to GHA: `schedule` only fires from the default branch, so master's workflow fans out over master, and supported branches by checking each one out per matrix entry, and it authenticates as the same GitHub App through `actions/create-github-app-token` instead of a mounted key file. 

It needs `GOVULNCHECK_GH_BOT_APP_ID` and `GOVULNCHECK_GH_BOT_APP_PRIVATE_KEY` as repo secrets before the first run can pass: https://scylladb.atlassian.net/browse/OPERATOR-430.
Dropping the Prow jobs is a follow-up in `scylla-operator-release`.

Related to https://scylladb.atlassian.net/browse/OPERATOR-383


.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+      # Issues live in a private repo, which the job's GITHUB_TOKEN can't write to.
+      - name: Create a GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.GOVULNCHECK_GH_BOT_APP_ID }}
+          private-key: ${{ secrets.GOVULNCHECK_GH_BOT_APP_PRIVATE_KEY }}
+          owner: scylladb
+          repositories: scylla-operator-team
```
czeslavo, line 51: IIRC, govulncheck-gh-bot can accept an app ID and app private key and handle auth on its own.

rzetelskik: It can, but the bot only takes the key as a file, so on GHA I'd have to write the raw private key to the runner fs (under Prow it's a mounted secret volume). `actions/create-github-app-token` also scopes to owner: scylladb / repositories: scylla-operator-team and revokes the token in its post step, while the bot's own app auth is unscoped and unrevoked.

czeslavo: Alright. We can consider dropping this dead path from the CLI once we merge this.

czeslavo (approved):
/lgtm

PR #3673, opened 2026-09-16
fix(manager-agent): readiness via `/ping` instead of a TCP socket

mflendrich:
## Description of your changes

The scylla-manager-agent container's readiness probe was a bare `tcpSocket` check on port 10001 — the agent's HTTPS API port. Every kubelet probe opened a TCP connection and closed it without a TLS ClientHello, so the agent logged `http: TLS handshake error from <kubelet>: EOF` every 10 seconds on every node.

This switches the probe to an HTTPS GET of `/ping`:
- `/ping` is registered on the agent's router **before** the auth-token middleware, so it needs no Authorization header, and it's the same endpoint ScyllaDB Manager itself uses to health-check agents (`GET https://<host>:10001/ping`, expects 204).
- Kubelet probes HTTPS with `InsecureSkipVerify` and no client cert, and treats 204 as success; the agent logs nothing for statuses <400 at the default log level → zero steady-state log noise.
- The check is strictly stronger than the TCP probe: it targets the same address and port (so any state passing it would have passed the TCP check) and additionally verifies the TLS handshake and the HTTP router. The agent builds its TLS config and router before opening the listener, so there is no state where TCP accepts but `/ping` can't be served. Failure semantics during startup (waiting for ignition/ScyllaDB API) are unchanged: nothing listens on 10001, the probe fails.

Testing: `test/e2e/set/scyllacluster/scyllacluster_restarts.go` asserts the scylla-manager-agent container's readiness explicitly in both directions using `utils.GetContainerReadinessMap`: ready=true after the ScyllaCluster rolls out and ready=false while ignition is blocked.

Fixes OPERATOR-431


rzetelskik (approved):
lgtm, but please squash the commits adding and dropping the comment

mflendrich:
@rzetelskik done, PTAL

PR #3674, opened 2026-09-17
ci: run the docs verification periodic on GHA

rzetelskik:
verify-docs runs as three prow periodics, one per built branch, because
`make -C docs verify` only linkchecks the checked-out source. That scales
by a job per release branch and needs changes on every release, and doesn't
currently cover the deprecated branches.

sphinx-multiversion forwards unknown args to sphinx-build, so `-b=linkcheck`
gives the same check across the whole multiversion set - the set docs-pages
already publishes. That's the new `verify-multiversion` target, and one
scheduled workflow on master replaces all three periodics.

Note `-b=linkcheck`, not `-b linkcheck` - sphinx-multiversion parses with
`parse_known_args`, so the space form binds `linkcheck` to its own
`filenames` positional and sphinx gets `-b` with no value. Same reason the
existing `autobuild` target writes `-b=dirhtml`.

Removing the prow periodics in scylla-operator-release follows separately.

Related to https://scylladb.atlassian.net/browse/OPERATOR-432.


czeslavo (approved):
/lgtm

